PR #61, opened 2014-10-01
Add bindings to CRYPTO_memcmp

alexcrichton:
This should help other projects use a constant-time memory comparison.


jwilm:
Thanks for this :)

PR #84, opened 2014-10-13
x509: add the api needed to obtain entries by nid

jmesmon:
Also includes a large list of nids (from openssl/object.h). Might be
transcoding errors on my part there.

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -275,7 +275,7 @@ extern "C" {
                                                            n: c_int,
                                                            file: *const c_char,
                                                            line: c_int));
-    pub fn CRYPTO_free(buf: *const c_char);
+    pub fn CRYPTO_free(buf: *mut c_void);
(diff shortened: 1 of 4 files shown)


src/ffi.rs
```diff
@@ -167,6 +168,193 @@ pub const X509_V_ERR_UNSUPPORTED_EXTENSION_FEATURE: c_int = 45;
...
 
+pub mod nid {
+    use std::mem;
+    use std::collections::enum_set::CLike;
+
+    #[allow(dead_code)]
+    #[allow(non_camel_case_types)]
+    #[repr(uint)]
+    pub enum Nid {
+        Undefined,
```
jwilm: I'm not convinced that's the right location for NID values. I ran into an issue trying to get text for `Nid::UID` from the subject, and changing the UID value to the one defined in `include/openssl/obj_mac.h` resolved the problem.

PR #133, opened 2014-12-31
Change to use updated Mutex API in latest Rust master.

kinghajj:
The Mutex API has changed so that `StaticMutex::lock()` returns `MutexGuard<()>` instead of `StaticMutexGuard`. Also, `.lock()` must be unwrapped.


ehsanul:
@sfackler Nightly has been hit now.

sfackler:
Thanks!

PR #140, opened 2015-01-09
Update to rust master

alexcrichton:
I tried to pick new integer sizes as appropriate, but you may want to double check me!


sfackler:
Looks like a doctest needs an `#[allow(unstable)]` added.

sfackler:
But I'll fix that myself, thanks!

PR #145, opened 2015-01-21
Bring ffi definitions closer to the originals

gkoz:
Add missing return types and fix imprecise type translations.
Repair the fallout in the openssl crate.


sfackler:
Thanks!

PR #153, opened 2015-02-03
Don't bail out of openssl-sys build config when targeting Android

mbrubeck:
Fixes servo/servo#4700.  #109 added special support for building on Android without host libraries, but #113 broke this by bailing out early when cross-compiling.  This adds an exception to the cross-compile check.


sfackler:
Thanks!

PR #154, opened 2015-02-04
crypto::symm rewrite

gkoz:
This is a prototype `symm` module rewrite that takes some inspiration from `hash` refactoring.
It's difficult to come up with a clean general API because each mode's API is slightly incompatible with every other one.
This version provides a `Writer` adapter for composability and an `apply` method that writes to `&mut [u8]`.
Because the static check model I tried out with `hash` and `hmac` wasn't settled on, call order requirements are checked dynamically and panic in case of errors.
Padding/alignment and authentication errors are handled with `Result`s, more "impossible" errors are checked with asserts.

### DIFF
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -397,11 +397,14 @@ extern "C" {
     pub fn EVP_aes_256_gcm() -> *const EVP_CIPHER;
     pub fn EVP_rc4() -> *const EVP_CIPHER;
 
+    pub fn EVP_CIPHER_key_length(ctx: *const EVP_CIPHER) -> c_int;
+
     pub fn EVP_CIPHER_CTX_new() -> *mut EVP_CIPHER_CTX;
     pub fn EVP_CIPHER_CTX_set_padding(ctx: *mut EVP_CIPHER_CTX, padding: c_int) -> c_int;
     pub fn EVP_CIPHER_CTX_free(ctx: *mut EVP_CIPHER_CTX);
     pub fn EVP_CIPHER_CTX_ctrl(ctx: *mut EVP_CIPHER_CTX, type_: c_int,
                                arg: c_int, ptr: *mut c_void) -> c_int;
+    pub fn EVP_CIPHER_CTX_iv_length(ctx: *mut EVP_CIPHER_CTX) -> c_int;
 
     pub fn EVP_CipherInit(ctx: *mut EVP_CIPHER_CTX, evp: *const EVP_CIPHER,
                           key: *const u8, iv: *const u8, mode: c_int) -> c_int;
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -383,13 +383,17 @@ extern "C" {
     pub fn EVP_aes_128_ecb() -> *const EVP_CIPHER;
     #[cfg(feature = "aes_xts")]
     pub fn EVP_aes_128_xts() -> *const EVP_CIPHER;
+    #[cfg(feature = "aes_ctr")]
     pub fn EVP_aes_128_ctr() -> *const EVP_CIPHER;
+    #[cfg(feature = "aes_gcm")]
     pub fn EVP_aes_128_gcm() -> *const EVP_CIPHER;
     pub fn EVP_aes_256_cbc() -> *const EVP_CIPHER;
     pub fn EVP_aes_256_ecb() -> *const EVP_CIPHER;
     #[cfg(feature = "aes_xts")]
     pub fn EVP_aes_256_xts() -> *const EVP_CIPHER;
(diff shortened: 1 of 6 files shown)


openssl-sys/src/lib.rs
```diff
@@ -345,25 +380,31 @@ extern "C" {
     pub fn EVP_aes_128_ecb() -> *const EVP_CIPHER;
     #[cfg(feature = "aes_xts")]
     pub fn EVP_aes_128_xts() -> *const EVP_CIPHER;
-    // fn EVP_aes_128_ctr() -> EVP_CIPHER;
-    // fn EVP_aes_128_gcm() -> EVP_CIPHER;
+    pub fn EVP_aes_128_ctr() -> *const EVP_CIPHER;
```
sfackler: I believe these are only defined on OpenSSL 1.0.1+, so they should be put behind a feature.

openssl/src/crypto/symm_new.rs
```diff
@@ -0,0 +1,1835 @@
...
+    );
+}
+
+/// A common cipher interface
+struct Context {
+    ctx: *mut ffi::EVP_CIPHER_CTX,
+    state: State,
+}
+
+/// A trait for ciphers that allow transcoding several chunks of data consequtively.
```
sfackler: "consecutively"

openssl/src/crypto/symm_new.rs
```diff
@@ -0,0 +1,1835 @@
...
+pub trait Apply {
+    /// Transcode the `data` into the `buf`.
+    ///
+    /// The `buf` have enough space to fit `data` (plus a cipher block length
```
sfackler: "must have"

openssl/src/crypto/symm_new.rs
```diff
@@ -0,0 +1,1835 @@
...
+/// A `Writer` adapter that finishes the padded cipher after writing.
+pub struct PaddedFilter<'a, T: 'a> {
+    inner: Filter<'a, T>,
+    closed: bool,
+}
+
+impl <'a, T: PaddedFinish> PaddedFilter<'a, T> {
+    /// Create a `Writer` adapter that finishes the padded cipher after writing.
+    /// The output is written to the `sink`.
+    /// The cipher has to be `start`ed beforehand and is finished explicitly
```
sfackler: Any reason to not have an implementation of `Drop` that flushes the last set of bytes?

gkoz: This was inspired by Rust RFC discussions of having explicit `close` that can return errors. I'm not aware of a final decision there. We need to propagate a malformed padding error somewhere.

So the choice here is between having a `close() -> Result<(), Error>` on the filter and checking for errors in its parent struct with `finish() -> Result<(), Error>` or `get_last_error() -> Option<Error>`. But the `PaddedFilter` can't leave the cipher unfinished because its sink is going away. Checking with `finish` after the filter is dropped would be the only (and extra confusing) case of calling `finish` on a finished cipher -- just to check for an error. `get_last_error` speaks for itself.

openssl-sys/src/lib.rs
gkoz: Should each feature be gated individually or would it make sense for them to be branch based or even just "modern_openssl" for stuff not present in OpenSSL 0.x? There isn't a build-time feature detection phase anyway, is there?
Speaking of versions, would you mind putting an official statement in the Readme as to which OpenSSL versions or branches the bindings are expected to support?

gkoz:
Added the necessary features.

openssl/src/crypto/symm_new.rs
gkoz: Also consider that it's undesirable for destructors to block and writing out the last set of bytes could block. (Shouldn't this reasoning apply to std::io too?)

gkoz: I could borrow this API and implementation, it must be somewhat battle-tested:
https://github.com/hyperium/hyper/blob/b985d85a5d43806cb60ee564116c5faf3f9e4799/src/http.rs#L188-299
Take ownership of the writer, `s/close/end/`, pass through `flush`es.

sfackler: Sorry again for letting this sit.

I believe the reasoning in `std::io` is that it's more important for streams to not lose data rather than not block in a destructor. I'm a fan of a strategy like the one you pointed out in hyper with a consuming method that flushes. We can leave the destructor out for now and see what people think.

sfackler:
I don't want to have separate `symm` and `symm_new` modules. I have a branch called `breaks` that I'm using to collect breaking changes before making an 0.5 release. Could you open a new PR that replaces `symm` with this against that branch instead of `master`?

Thanks!

gkoz:
Continued as #173

Issue #155, opened 2015-02-07
Undefined symbols for architecture x86_64: _HMAC_CTX_copy


s-panferov:
Hello @sfackler. I'm trying to build the compiler plugin [deuterium-orm](https://github.com/deuterium-orm/deuterium-orm). Build fails with this log:

~~~
➜  deuterium-orm git:(master) cargo test
   Compiling deuterium_orm v0.2.0 (file:///Users/panferov/Workspace/my/deuterium-orm)
error: linking with `cc` failed: exit code: 1
[...]

PR #156, opened 2015-02-07
Add info about OS X to the README

s-panferov:
Ref #155


sfackler:
So I'd prefer to avoid advising force linking homebrew's openssl since doing that breaks stuff like sstephenson/rbenv#542. We should probably just be telling pkg-config to look in /usr/local on OSX.

sfackler:
Blargh, looks like there isn't really a better way to do this :(

PR #158, opened 2015-02-11
Fix the target for windows

quentinbaradat:
As a windows user on x86_64 plateform, this fix allowed me to compile. 

I'm not sure that the targets `win32` and `win64` are still relevant. And the gcc should not compile the `libold_openssl_shim.a`  with a  value of `false` as argument.

### DIFF
--- a/openssl-sys/build.rs
+++ b/openssl-sys/build.rs
@@ -15,12 +15,10 @@ fn main() {
                      (instructions in the README) and provide their location through \
                      $OPENSSL_PATH.");
         println!("cargo:rustc-flags=-L native={} -l crypto:static -l ssl:static", path);
-        // going to assume the user built a new version of openssl
-        build_old_openssl_shim(false);
         return;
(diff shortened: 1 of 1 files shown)


openssl-sys/build.rs
```diff
@@ -41,14 +39,15 @@ fn main() {
 }
 
 fn build_old_openssl_shim(is_old: bool) {
-        let mut config: gcc::Config = Default::default();
-        if is_old {
-            config.definitions.push(("OLD_OPENSSL".to_string(), None));
-        }
+    let mut config: gcc::Config = Default::default();
+    if is_old {
+        config.definitions.push(("OLD_OPENSSL".to_string(), None));
```
sfackler: This needs to be built in all cases because it provides the `HMAC_Init_ex_shim`, `HMAC_Update_shim` and `HMAC_Final_shim` functions that the main crate links to.

PR #159, opened 2015-02-12
Fix build on *BSD

wg:
FreeBSD includes OpenSSL in the base system, and I believe OpenBSD includes their fork known as LibreSSL.

### DIFF
--- a/openssl-sys/build.rs
+++ b/openssl-sys/build.rs
@@ -27,6 +27,7 @@ fn main() {
         return;
     }
 
+    // pkg-config doesn't know of OpenSSL on FreeBSD 10.1 and OpenBSD uses LibreSSL
     if target.contains("bsd") {
         println!("cargo:rustc-flags=-l crypto -l ssl");
         // going to assume the base system includes a new version of openssl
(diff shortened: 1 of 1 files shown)


sfackler:
Is pkg-config unable to find the copy of OpenSSL?

wg:
@sfackler yup, I assume because pkg-config itself is a package whereas OpenSSL is part of the base system.

sfackler:
Ok, it looks like this is an issue in general with BSDs: https://bugs.freedesktop.org/show_bug.cgi?id=14247. Can you add a comment on the block briefly explaining why BSD is special cased? Looks good other than that, thanks!

wg:
Great, thanks! Cargo build is currently broken on FreeBSD due to this issue, could you push another revision that includes it?

sfackler:
Published v0.3.5

gkoz:
Note that FreeBSD also has OpenSSL available from ports so unconditionally linking to the base system one might not be the best idea.

sfackler:
@gkoz good point. I moved the BSD special case to after the pkg-config checks which should make it do the right thing more often.

PR #172, opened 2015-02-22
add support for SSL_CTX_set_options and SSL_CTX_get_options

### LINKED ISSUE #168: SslContext should expose some way to set options on the underlying SSL_CTX
This is done via OpenSSL's `SSL_CTX_set_options` function, this is necessary to pass `SSL_OP_NO_SSLv2` and `SSL_OP_NO_SSLv3` (to say nothing of the variety of other options there, but these are the ones I care about at this moment)

reaperhulk:
This is the first rust I've ever written so I'd appreciate any pointers if there are ways to write it better (especially the tests).

fixes #168

### DIFF
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -118,6 +118,7 @@ pub const NID_ext_key_usage: c_int = 126;
 pub const NID_key_usage:     c_int = 83;
 
 pub const SSL_CTRL_OPTIONS: c_int = 32;
+pub const SSL_CTRL_CLEAR_OPTIONS: c_int = 77;
 
 pub const SSL_CTRL_SET_TLSEXT_HOSTNAME: c_int = 55;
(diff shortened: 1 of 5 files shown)


sfackler:
Thanks! Could you add bindings for `SSL_CTX_clear_options` too?

I'm not a huge fan of passing a raw `c_long` around - the `bitflags` crate should be able to give us a slightly nicer API. Could you add `const`s for all of the options to `openssl-sys` and make an `SslCtxOptions` wrapper type with `bitflags!` in `openssl::ssl`? http://doc.rust-lang.org/bitflags/bitflags/macro.bitflags!.html#example

Looks good other than those two things.

reaperhulk:
@sfackler thanks, I'll take a look at revise this PR.

reaperhulk:
(PR has been revised with bitflags and clear_options)

sfackler:
Thanks!

PR #173, opened 2015-02-25
crypto::symm rewrite

gkoz:
Here's the rewrite rebased against `breaks` and updated for the new `std::io`. Redesigning the writer proved more challenging than I expected (remember that each mode has a slightly different API; and I'm trying to avoid having a second writer just for the padded modes), so still chewing on it.


gkoz:
Umm, actually this model from hyper [`W: Writer`]
```rust
pub fn end(mut self) -> IoResult<W> {
```
doesn't really work so well:
a) why should you lose your writer if `cipher.finish()` fails for whatever reason;
b) `finish` returns `Result<Whatever, Error>` so we need to return `Whatever` (cipher's associated type, it can be `()`, `Vec<u8>`, `u64`) alongside with the writer like so
```rust
pub fn end(self) -> Result<(Whatever, W), Error> {
```
or
```rust
pub fn end(self) -> (Result<Whatever, Error>, W) {
```

This isn't pretty. I remember now why borrowing the writer felt better, if we go back to that, we'd have
```rust
pub fn end(self) -> Result<Whatever, Error> {
```
Alternatively, we might make `finish` only return `Result<(), Error>` and add getters for various stuff that it used to return (`get_tag()` for gcm, `next_counter()` for ctr). But the point about losing the writer stands.
Input welcome :)

gkoz:
There is a way to preserve writer similar to http://doc.rust-lang.org/collections/string/struct.FromUtf8Error.html
All of the choices seem kludgy. For now I'd stick to the existing way of borrowing the writer but use a single struct instead of `Filter` and `PaddedFilter`.

gkoz:
So `Filter` is now `Writer`, everything else much the same.
I guess I need to support XTS too before it's merged?

sfackler:
I think we can probably unify some of the stuff in here, but I think we can merge this if you add in XTS support.

PR #176, opened 2015-03-04
Cut down on unstable features in openssl-sys

alexcrichton:
* Move from `old_path` to `path` (leveraging the `fs` feature as well)
* Move from `StaticMutex` to `Mutex<()>` as they're dynamically initialized


sfackler:
Travis seems to not be building this so I'm just going to assume it compiles.

PR #179, opened 2015-03-10
DTLS support

manuels:
These commits implement DTLS support. Feedback very appreciated.


manuels:
Hmm, how can I adjust the settings of the build bot?

manuels:
This travis error doesn't really look like it's coming from my changes to the codebase.
https://travis-ci.org/sfackler/rust-openssl/jobs/53940818: 

    /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/pkg-config-0.3.0/src/lib.rs:218:55: 218:63 error: binary operation `==` cannot be applied to type `unicode::char::ToUppercase`
    /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/pkg-config-0.3.0/src/lib.rs:218 name.chars().map(|c| c.to_uppercase()).map(|c| if c == '-' {'_'} else {c})

Is there a way to rerun this travis job?

sfackler:
I retriggered the build but it's still running into errors. You'll probably want to use the typdefs in the libc crate instead of `u8`, `u16` etc since `sa_family_t` isn't the same size on all platforms.

manuels:
This is interesting: I get an error on the travis machine which I do not get on my linux machine:

https://travis-ci.org/sfackler/rust-openssl/jobs/54166321#L1538

 	thread 'ssl::tests::new_sslstream::dtlsv1' panicked at 'called `Result::unwrap()` on an `Err` value: OpenSslErrors([UnknownError { library: "digital envelope routines", function: "EVP_MD_size", reason: "message digest is null" }])', /home/rustbuild/src/rust-buildbot/slave/nightly-dist-rustc-linux/build/src/libcore/result.rs:744

openssl/src/ssl/connected_socket.rs
```diff
@@ -0,0 +1,343 @@
+use libc::funcs::bsd43::connect;
```
sfackler: This seems like something that should live in a separate library.

openssl/src/ssl/mod.rs
```diff
@@ -105,9 +112,37 @@ impl SslMethod {
...
-            SslMethod::Tlsv1_2 => ffi::TLSv1_2_method()
+            SslMethod::Tlsv1_2 => ffi::TLSv1_2_method(),
+            #[cfg(feature = "dtlsv1")]
+            SslMethod::Dtlsv1 => ffi::DTLSv1_method(),
+            #[cfg(feature = "dtlsv1_2")]
+            SslMethod::Dtlsv1_2 => ffi::DTLSv1_2_method(),
         }
     }
+
+    #[cfg(feature = "dtlsv1")]
```
sfackler: Why do these exist?

openssl/src/ssl/mod.rs
```diff
@@ -282,6 +321,12 @@ impl SslContext {
         }
     }
 
+    pub fn set_read_ahead(&self) {
+        unsafe {
+            ffi::SSL_CTX_ctrl(*self.ctx, ffi::SSL_CTRL_SET_READ_AHEAD, 1, ptr::null_mut());
```
sfackler: This should apparently never be called directly: http://unixhelp.ed.ac.uk/CGI/man-cgi?SSL_CTX_ctrl+3

vhbit: @sfackler it's also states that the reason for not calling directly is that
there are C macros which wrap it.

@manuels As a better approach I'd recommend to create a function with the
same name as OpenSSL macro in openssl-sys/src/lib.rs - I believe there is a
section already for such wrappers just before huge extern block.

manuels:
@sfackler I introduced the `is_dtls()` methods to be able to figure out if we use a DTLS method even if DTLS is not enabled during compilation.
I moved the `connected_socket` code to its own module now.

manuels:
So, this pull requests finally passes all tests. I'm happy for any feedback.

openssl/src/ssl/mod.rs
```diff
@@ -282,6 +320,12 @@ impl SslContext {
         }
     }
 
+    pub fn set_read_ahead(&self, m: c_long) {
+        unsafe {
+            ffi::SSL_CTX_set_read_ahead(*self.ctx, m);
```
vhbit: As a client, I'd prefer to have `m: usize` or `m: u32`, i.e. any rust type which is directly available without importing `libc::c_long`.

manuels: You're right, I changed it to u32. usize semantically implies that it would be a size.

.travis.yml
```diff
@@ -5,12 +5,22 @@ os:
...
+- DIR=`pwd`
+- (test $TRAVIS_OS_NAME == "osx" || (sudo apt-get install gcc make))
+- (test $TRAVIS_OS_NAME == "osx" || (wget http://openssl.org/source/openssl-1.0.2.tar.gz -O /tmp/openssl-1.0.2.tar.gz))
+- (test $TRAVIS_OS_NAME == "osx" || (cd /tmp && tar xzf openssl-1.0.2.tar.gz))
+- (test $TRAVIS_OS_NAME == "osx" || (cd /tmp/openssl-1.0.2/ && ./config --prefix=/usr/ shared))
+- (test $TRAVIS_OS_NAME == "osx" || (cd /tmp/openssl-1.0.2/ && make))
+- (test $TRAVIS_OS_NAME == "osx" || (cd /tmp/openssl-1.0.2/ && sudo make install))
+- cd ${DIR}
+- (test $TRAVIS_OS_NAME == "osx" || (./openssl/test/test.sh &))
+- (test $TRAVIS_OS_NAME == "osx" || (cd openssl && LD_LIBRARY_PATH=/usr/lib:$LD_LIBRARY_PATH cargo test --features "$FEATURES"))
```
sfackler: Why do we need to do this?

openssl/Cargo.toml
```diff
@@ -27,3 +29,7 @@ rustc-serialize = "0.2"
 
 [dependencies]
 bitflags = "0.1.1"
+
+[dependencies.connected_socket]
```
sfackler: This is only needed by tests, right? Could you change it to a `dev-dependency`?

sfackler:
Ping. Also needs a rebase.

manuels:
Rebased and fixed all issues. We need to compile the lastest OpenSSL version in the Travis test because the official version from the repos has a bug in DTLS

sfackler:
Thanks!

openssl/src/ssl/mod.rs
```diff
@@ -339,7 +373,13 @@ impl SslContext {
             return Err(SslError::get());
         }
 
-        Ok(SslContext { ctx: ctx })
+        let ctx = SslContext { ctx: ctx };
+
+        if method.is_dtls() {
+            ctx.set_read_ahead(1);
```
sfackler: The openssl man pages seem to imply that this has no effect if DTLS is being used - am I misinterpreting that? http://wiki.openssl.org/index.php/Manual:SSL_CTX_set_read_ahead%283%29

manuels:
Thanks!

openssl/src/ssl/mod.rs
manuels: You are correct about this manpage, but the manpage is incorrect ;)
I think the scenario in the manpage is when using a UDP socket directly (without a MemBIO). The call to `set_read_ahead()` is definitely required in the context where we use MemBIO.

sfackler: Cool, thanks. I am somehow unsurprised that an OpenSSL man page is incorrect about something :)

PR #182, opened 2015-03-12
Specify minor version of pkg-config (0.3.1)

davbo:
Depending on opensaml in my project seems to pull down 0.3.0 of pkg-config which doesn't compile against the latest rust nightly builds.

Here's a build for example: https://travis-ci.org/davbo/matasano/builds/54021798


sfackler:
Running `cargo update` will cause Cargo to update the dependency version it's using to `0.3.1`.

davbo:
Brilliant! Thanks for that @sfackler

PR #185, opened 2015-03-18
Add TLS Next Protocol Negotiation

mlalic:
This pull request adds FFI bindings to OpenSSL's NPN-related functions to the `openssl-sys` crate. A new crate feature (`npn`) is also added and the functions are exposed only when it is set.

Additionally, the `openssl::ssl` module's API is augmented to provide Next Protocol Negotiation capabilities.

The following public methods were added:

- `SslContext::set_npn_protocols` - allows clients to set a list of protocols that should be used in NPN negotation. This applies to both server, as well as client sockets created using this `SslContext` instance.
- `Ssl::get_selected_npn_protocol` - returns an optional name of the protocol that was selected via NPN.
- `SslStream::get_selected_npn_protocol` - same as the previous method -- it actually delegates to the internal `Arc<Ssl>` instance.

These methods are also predicated on an `npn` feature being set in the `openssl` crate.

This API is very similar to what is found in Python's standard library.

Tests for the feature are included and properly integrated with the automated travis-ci test suite; an additional `openssl` process, with NPN enabled, is spawned before the tests are ran.

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -1,4 +1,4 @@
-use libc::{c_int, c_void, c_long};
+use libc::{c_int, c_void, c_long, c_uint, c_uchar};
 use std::ffi::{CStr, CString};
 use std::fmt;
 use std::io;
@@ -6,6 +6,7 @@ use std::io::prelude::*;
 use std::ffi::AsOsStr;
 use std::mem;
 use std::net;
+use std::slice;
 use std::num::FromPrimitive;
 use std::num::Int;
 use std::path::Path;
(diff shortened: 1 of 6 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -192,6 +221,68 @@ extern fn raw_verify_with_data<T>(preverify_ok: c_int,
...
+extern fn raw_next_proto_select_cb(ssl: *mut ffi::SSL,
+                                   out: *mut *mut c_uchar, outlen: *mut c_uchar,
+                                   inbuf: *const c_uchar, inlen: c_uint,
+                                   _arg: *mut c_void) -> c_int {
+    unsafe {
+        // First, get the list of protocols (that the client should support) saved in the context
+        // extra data.
+        let ssl_ctx = ffi::SSL_get_SSL_CTX(ssl);
+        let protocols = ffi::SSL_CTX_get_ex_data(ssl_ctx, get_npn_protos_idx());
+        let protocols: Box<Vec<u8>> = mem::transmute(protocols);
```
sfackler: You can transmute to a `&Vec<u8>` here instead of `Box` to avoid having to forget it at the end.

openssl/src/ssl/mod.rs
```diff
@@ -192,6 +221,68 @@ extern fn raw_verify_with_data<T>(preverify_ok: c_int,
...
+        // First, get the list of (supported) protocols saved in the context extra data.
+        let ssl_ctx = ffi::SSL_get_SSL_CTX(ssl);
+        let protocols = ffi::SSL_CTX_get_ex_data(ssl_ctx, get_npn_protos_idx());
+        if protocols.is_null() {
+            *out = b"".as_ptr();
+            *outlen = 0;
+        } else {
+            // If the pointer is valid, put the pointer to the actual byte array into the
+            // output parameter `out`, as well as its length into `outlen`.
+            let protocols: Box<Vec<u8>> = mem::transmute(protocols);
```
sfackler: Same here

openssl/src/ssl/mod.rs
```diff
@@ -342,6 +433,36 @@ impl SslContext {
         };
         SslContextOptions::from_bits(ret).unwrap()
     }
+
+    /// Set the protocols to be used during Next Protocol Negotiation (the protocols that
+    /// the client supports).
```
sfackler: Could you add a note to the docs here mentioning that this method needs the npn feature?

openssl/src/ssl/mod.rs
```diff
@@ -470,6 +591,26 @@ impl Ssl {
         }
     }
 
+    /// Returns the protocol selected by performing Next Protocol Negotiation, if any.
+    ///
+    /// The protocol's name is returned is an opaque sequence of bytes. It is up to the client
+    /// to interpret it.
```
sfackler: Same here

openssl/src/ssl/mod.rs
```diff
@@ -623,6 +764,15 @@ impl<S: Read+Write> SslStream<S> {
 
         Some(s)
     }
+
+    /// Returns the protocol selected by performing Next Protocol Negotiation, if any.
+    ///
+    /// The protocol's name is returned is an opaque sequence of bytes. It is up to the client
+    /// to interpret it.
+    #[cfg(feature = "npn")]
```
sfackler: And here

mlalic:
Thanks a lot for your comments! I am away from a computer for the weekend, but I'll make the changes you suggest by Monday morning (CET).

mlalic:
All of your comments have been addressed and the branch rebased onto the current HEAD.

sfackler:
Thanks!

PR #186, opened 2015-03-19
Use raw keys and certs in SslContext 

manuels:
Add `SslContext::set_private_key(&PKey)` and `SslContext::set_certificate(&X509)` to use 'raw' private keys and certificates for SSL connections.
Additionally, this pull requests adds `PKey::private_key_from_pem(...)` which is needed for testing but might come in handy for some people.

### DIFF
--- a/openssl/src/crypto/pkey.rs
+++ b/openssl/src/crypto/pkey.rs
@@ -352,7 +352,6 @@ impl Drop for PKey {
 
 #[cfg(test)]
 mod tests {
-    use std::io;
     use std::path::Path;
     use std::fs::File;
     use crypto::hash::Type::{MD5, SHA1};
--- a/openssl/src/crypto/pkey.rs
+++ b/openssl/src/crypto/pkey.rs
@@ -1,4 +1,5 @@
 use libc::{c_int, c_uint, c_ulong};
+use std::io;
 use std::io::prelude::*;
 use std::iter::repeat;
 use std::mem;
@@ -66,6 +67,22 @@ impl PKey {
                 evp: ffi::EVP_PKEY_new(),
                 parts: Parts::Neither,
             }
+        }
+    }
+
+    /// Reads private key from PEM, takes ownership of handle
+    pub fn private_key_from_pem<R>(reader: &mut R) -> Result<PKey, SslError> where R: Read {
+        let mut mem_bio = try!(MemBio::new());
+        try!(io::copy(reader, &mut mem_bio).map_err(StreamError));
+
+        unsafe {
+            let evp = try_ssl_null!(ffi::PEM_read_bio_PrivateKey(mem_bio.get_handle(),
+                                                                 ptr::null_mut(),
+                                                                 None, ptr::null_mut()));
+            Ok(PKey {
+                evp:   evp,
+                parts: Parts::Both,
+            })
         }
     }
 
@@ -335,6 +352,9 @@ impl Drop for PKey {
 
 #[cfg(test)]
 mod tests {
+    use std::io;
+    use std::path::Path;
+    use std::fs::File;
     use crypto::hash::Type::{MD5, SHA1};
 
     #[test]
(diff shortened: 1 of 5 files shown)


openssl/src/crypto/pkey.rs
```diff
@@ -69,6 +70,22 @@ impl PKey {
         }
     }
 
+    /// Reads certificate from PEM, takes ownership of handle
+    pub fn private_key_from_pem<R>(reader: &mut R) -> Result<PKey, SslError> where R: Read {
```
vhbit: why `private_key_from_pem` instead of plain `from_pem`?

manuels: Because there could also be a method `public_key_from_pem` and I thought it would be good to have am unambiguous name for the function.

vhbit: Ohh, right, my bad.

manuels: But the comment was incorrect. I just corrected it.

openssl/src/ssl/mod.rs
```diff
@@ -312,6 +321,22 @@ impl SslContext {
...
+    /// Specifies the private key
+    pub fn set_private_key(&mut self, key: &PKey) -> Option<SslError> {
+        wrap_ssl_result(
+            unsafe {
+                ffi::SSL_CTX_use_PrivateKey(*self.ctx, key.get_handle())
+            })
+    }
+
+    /// Check consistency of private key and certificate
+    pub fn check_private_key(&mut self) -> Option<SslError> {
```
sfackler: It's a bit nicer to use `Result<(), SslError>` rather than `Option<SslError>` so that things compose more nicely with `try!` etc.

sfackler:
Might need bindings around `SSL_CTX_add_extra_chain_cert` as well.

openssl/src/ssl/mod.rs
manuels: Yes, in deed. All other functions also use `wrap_ssl_result()` which returns `Option<SslError>`.
I think it's best (and consistent) if I modify `wrap_ssl_result()` to return `Result<(), SslError>`.
Do you agree?

sfackler: Yep, that sounds reasonable to me.

openssl/src/crypto/pkey.rs
```diff
@@ -69,6 +70,22 @@ impl PKey {
         }
     }
 
+    /// Reads private key from PEM, takes ownership of handle
```
sfackler: Could this be moved to a method on `PKey`? Something like `PKey::from_pem` maybe.

sfackler:
Also needs a rebase

openssl/src/crypto/pkey.rs
manuels: Not sure what you mean, this function is already a method on `PKey` (`PKey::private_key_from_pem`).
And renaming it to `from_pem` would make it ambiguous because there could also be a function for reading public keys (`public_key_from_pem`)

manuels:
rebased.

sfackler:
Needs one more rebase (sorry!)

manuels:
rebased

openssl/src/crypto/pkey.rs
sfackler: Why would the struct representing a private key have a function to read a public key?

sfackler:
Looks like there are some test failures when building with features turned on: https://travis-ci.org/sfackler/rust-openssl/jobs/55973708

openssl/src/crypto/pkey.rs
manuels: PKey isn't for private keys only, is it?
At least from the codenbase it looks like it can also hold just a public key:

    pub struct PKey {
        evp: *mut ffi::EVP_PKEY,
        parts: Parts,
    }
    enum Parts {
        Neither,
        Public,
        Both
    }

and e.g. `X509_get_pubkey()` also returns a `EVP_PKEY*`

manuels:
Rebased and fixed issues.

PR #187, opened 2015-03-21
Add X509Generator::sign()

manuels:
`X509Generator::generate()` does two things at once: generating a private key and signing a certificate.
However, if you already have a private key and want to use it to sign a certificate, you can't do this right now.
This patch introduces a functions that does just that.


sfackler:
Thanks!

PR #191, opened 2015-03-29
Fix error with current rust nightly

fhartwig:


paulosuzart:
I see this will fix the message: `error: use of unstable library feature 'collections': use &mut s[..] instead` This is preventing a project that uses Hyper to compile. Is it easy to release a new version soon?

filipegoncalves:
@paulosuzart You can clone this repo., apply the patch, and use cargo's override features to get a working, compilable version.

PR #195, opened 2015-04-03
Remove Copy and Clone implementations for ffi structs

gkoz:
Those implementations aren't used in the openssl crate and actually using them would likely lead to memory safety violations or leaks.

### DIFF
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -49,11 +49,6 @@ pub struct EVP_MD_CTX {
     update: *mut c_void
 }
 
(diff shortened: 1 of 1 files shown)


sfackler:
So in my opinion, the FFI layer should be an unopinionated mirror of the C API that doesn't try to add any safety factors on top of it - the fields should be public, the pointers should be raw, there should be no destructors on the structs, etc. Notably, the structs should also be copy/cloneable since move semantics do not necessarily map down to some C API.

gkoz:
Thanks for the explanation. This PR would indeed be incompatible with that objective.

Issue #197, opened 2015-04-11
Work on sockets instead of MemBio


manuels:
I was just investigating if I can implement a SslStream working on a SocketBio instead of MemBio to make SSL_pending work reliably. What is actually the reason why MemBio is used instead of directly a SocketBio? Is it to make the code more versatile.
Would you accept a pull request to add SocketBio (using AsRawFd) in addition to MemBio?

sfackler:
When the ssl module was written, it wasn't possible to extract a file descriptor from Rust's IO types so everything had to go through the extra layer of indirection of MemBio. Since this is now possible in modern Rust, I think it absolutely makes sense to support it, probably even as the default option. Maybe something like this:

```rust
impl<S> SslStream<S> {
    fn new<S: AsRawFd>(stream: S) -> Result<SslStream<S>> { ... }

    // This name is not great...
    fn new_wrapped<S: Read+Write>(stream S) -> Result<SslStream<S>> { ... }
```

sfackler:
I'd also want to make sure we have a good story on windows, though. `AsRawFd` is a unix-only trait, and windows versions of Rust have a version that returns a `HANDLE` instead.

PR #198, opened 2015-04-14
Use RawFd directly

manuels:
This is not completely ready for merging. It's a contribution to the discussion of #197.
Do you think an implementation like this (See https://github.com/manuels/rust-openssl/blob/master/openssl/src/ssl/ssl_stream.rs) would be the right way?

### DIFF
--- a/openssl/src/x509/mod.rs
+++ b/openssl/src/x509/mod.rs
@@ -285,10 +285,13 @@ impl X509Generator {
                                                mem::transmute(&ctx),
                                                nid as c_int,
                                                value.as_ptr() as *mut c_char),
-                None => ffi::X509V3_EXT_conf(ptr::null_mut(),
+                None => {
+                    let name=CString::new(exttype.get_name().unwrap().as_bytes()).unwrap();
+                    ffi::X509V3_EXT_conf(ptr::null_mut(),
                                                mem::transmute(&ctx),
-                                               exttype.get_name().unwrap().as_ptr() as *mut c_char,
-                                               value.as_ptr() as *mut c_char),
+                                               name.as_ptr() as *mut c_char,
+                                               value.as_ptr() as *mut c_char)
+                }
             };
             let mut success = false;
             if ext != ptr::null_mut() {


manuels:
In case you'll have time to build an opinion on this kind of implementation: I'll be on holidays for a week. So please don't expect a response before Thursday next week ;)

sfackler:
Sorry, I'm terrible at looking at PRs quickly :(

So I see the RawFd use as an under the hood optimization - it shouldn't affect the API all that much. Particularly, I'd like to keep the signature of `SslStream` the same (i.e. `SslStream<TcpSocket>` instead of `SslStream<StreamIo<TcpSocket>>` or `SslStream<SocketIo>`). Should be relatively painless by just switching to an enum or something that's stored in the `SslStream`. In both cases, I'd like the `SslStream` to take ownership of the underlying `Stream` - `get_ref` should be available with both a `RawFd` backed implementation and a `MemBio` backed  implementation.

If we had specialization, the ideal API in my opinion would be
```rust
impl SslStream<S: Read+Write> {
    pub fn new(ctx: &SslContext, s: S) -> Result<SslStream<S>, io::Error> {
         // set up with a MemBio
    }
}

impl SslStream<S: Read+Write+AsRawFd> {
    pub fn new(ctx: &SslContext, s: S) -> Result<SslStream<S>, io::Error> {
        // set up with a SocketBio
    }
}
```
so you'd just end up with a more efficient implementation if you wrap the `SslStream` around something that implements `AsRawFd`. As is, we'd probably just want to name the second method `new_fd` or something like that.

There's also somewhat of an issue that Windows deals with `SOCKET` types rather than file descriptors which is a real bummer for an implementation not based off of specialization.

manuels:
That would probably look nicer, but the rust compiler doesn't like it if I want to implement `std::io::Read` twice (for `S:Read+Write` and for `S:AsRawFd`)

    src/ssl/ssl_stream.rs:214:1: 221:2 error: conflicting implementations for trait `std::io::Read` [E0119]
    src/ssl/ssl_stream.rs:214 impl<S:Read+Write> Read for SslStream<S> {
    src/ssl/ssl_stream.rs:215     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
    ...
    src/ssl/ssl_stream.rs:239:1: 246:2 note: note conflicting implementation here
    src/ssl/ssl_stream.rs:239 impl<S:AsRawFd> Read for SslStream<S> {
    src/ssl/ssl_stream.rs:240     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {

PR #199, opened 2015-04-15
Fix documentation URL for openssl-sys.

CarlColglazier:
The [current link](https://sfackler.github.io/rust-openssl/doc/openssl-sys) leads to a 404. The [correct link](https://sfackler.github.io/rust-openssl/doc/openssl_sys) should have an underscore instead of a hyphen.


sfackler:
Thanks!

PR #201, opened 2015-04-16
Add comparison for PKeys

manuels:
Add `==` operation for `PKey`s based on `EVP_PKey_cmp`

### DIFF
--- a/openssl/src/crypto/pkey.rs
+++ b/openssl/src/crypto/pkey.rs
@@ -350,10 +350,8 @@ impl PKey {
     pub unsafe fn get_handle(&self) -> *mut ffi::EVP_PKEY {
         return self.evp
     }
-}
 
-impl PartialEq<PKey> for PKey {
-    fn eq(&self, other: &PKey) -> bool {
+    pub fn public_eq(&self, other: &PKey) -> bool {
         unsafe { ffi::EVP_PKEY_cmp(self.evp, other.evp) == 1 }
     }
 }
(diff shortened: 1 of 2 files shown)


sfackler:
What happens if the PKey is storing a private key instead of a public key?

manuels:
It's included in the tests:

https://github.com/sfackler/rust-openssl/pull/201/files#diff-e96f736ad12f50998e6e83e147873b26R491

The comparison returns false.

sfackler:
I'm a bit concerned that the public key/private key comparison issue will be a bit confusing. Maybe expose this as a `public_eq` method instead?

manuels:
I'm not sure what you mean here. What behavior would you expect?

Currently it's like this:

    Alice's Public Key != Bob's Public Key
    Alice's Public Key == Alice's Public Key
    Alice's Private Key != Bob's Private Key
    Alice's Private Key == Alice's Private Key
    Alice's Private Key != Alice's Public Key

sfackler:
Oh, sorry. I was working off of the OpenSSL man pages which says that it only compares public key components and parameters. If it handles the private key case, then this seems reasonable: https://www.openssl.org/docs/crypto/EVP_PKEY_cmp.html

manuels:
From the manpage:
> Since OpenSSL private keys contain public key components too the function EVP_PKEY_cmp() can also be used to determine if a private key matches a public key.

This seems to be incorrect because it does not correspond to the behavior observed: https://github.com/sfackler/rust-openssl/pull/201/files#diff-e96f736ad12f50998e6e83e147873b26R491

manuels:
Ok, in short: you were right, the manpage was right (no, really ;) ), I was wrong.
I fixed it according to your suggestion

openssl/src/crypto/pkey.rs
```diff
@@ -350,6 +350,10 @@ impl PKey {
     pub unsafe fn get_handle(&self) -> *mut ffi::EVP_PKEY {
         return self.evp
     }
+
+    fn public_eq(&self, other: &PKey) -> bool {
```
sfackler: This should probably be a public function :)

manuels: You're right. That's what unit test won't tell you ;)

Fixed it.

sfackler:
Thanks!

manuels:
Thank you!

PR #202, opened 2015-04-19
Fixes for Native Client.

DiamondLovesYou:


sfackler:
Ideally we'd still build the shim for nacl - is it too awkward to reference the right headers?

DiamondLovesYou:
Yes, because the header includes would have to reference the source
directory of `libressl-pnacl-sys`. Yes, I realize that sort of thing can be
done, via `DEP_libressl_pnacl_sys_$TARGET` env vars, but I find
implementing all that more awkward and quite a bit more work. Frankly, if
`nacl` is the target, that means `libressl` is used, which provides the
wanted function type for each of those. So the shims aren't needed.

Additionally, I may, in the future, make the packaged/published version of
`libressl-pnacl-sys` not have any source/header files at all: for each
arch, a compiled NaCl module is the same for all OSes (ignoring the fact
that my fork can't even generate NaCl modules directly; currently
everything must go through PNaCl, for better or worst), and PNaCl bitcode
is the same regardless of host or arch; so there is no requirement that the
host machine must compile `libressl` and would make `libressl-pnacl-sys`
all that much faster to "build".

If I did the above, the shims would have to be unused.

sfackler:
Cool, sounds good to me.

PR #207, opened 2015-04-28
docs(README.md): remove extra `\` character

Byron:
The original intention might have been to cause a line-break in
the markdown document. However, it was taken literally.

Those who would just copy-paste the `openssl` server line on linux would
be greeted with a `file not found` error.

Without the `\` character, it works as expected.


sfackler:
Thanks!

PR #211, opened 2015-05-07
Add ability to redefine library list via OPENSSL_LIBS environment variable

bozaro:
Add ability to redefine library list via OPENSSL_LIBS environment variable.

It's usefull for compiling with MinGW-w64 installed via MSYS2 (https://wiki.qt.io/MSYS2).

This MinGW-w64 uses **crypto** and **ssl** library names.

### DIFF
--- a/openssl-sys/build.rs
+++ b/openssl-sys/build.rs
@@ -23,7 +23,7 @@ fn main() {
     let libs_env = env::var("OPENSSL_LIBS").ok();
     let libs = match libs_env {
         Some(ref v) => v.split(":").collect(),
-        None => if target.contains("windows") && env::var_os("MSYSTEM").is_none() {
+        None => if target.contains("windows") {
             vec!("eay32", "ssl32")
         } else {
             vec!("crypto", "ssl")
--- a/openssl-sys/build.rs
+++ b/openssl-sys/build.rs
@@ -20,10 +20,14 @@ fn main() {
         }
     }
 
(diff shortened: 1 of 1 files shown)


openssl-sys/build.rs
```diff
@@ -20,10 +20,14 @@ fn main() {
...
     }
 
-    let (libcrypto, libssl) = if target.contains("windows") {
-    	("eay32", "ssl32")
-    } else {
-    	("crypto", "ssl")
+    let libs_env = env::var("OPENSSL_LIBS").ok();
+    let libs = match libs_env {
+        Some(ref v) => v.split(":").collect(),
+        None => if target.contains("windows") && env::var_os("MSYSTEM").is_none() {
```
sfackler: Is the MSYSTEM check the right thing to do here? I've never been able to successfully link against the Msys64-installed libcrypto and libssl.

bozaro: I do not know. I have sucessfully linked with openssl, but work with openssl I did not check (I have it going on according to https://github.com/iron/iron and don't check https).

sfackler: Let's leave the MSYSTEM check out for now. We may be moving to openssl being provided by a separate crate on Windows to avoid the linkage nightmare that it is right now.

bozaro:
I deleted the controversial test environment variable MSYSTEM.
Building anything on Windows I have long been cause pain. I've put up with it :disappointed:

bozaro:
Thanks

PR #213, opened 2015-05-11
Add support for reading X509 subject information

josephglanville:
This is adapted from #84 

Would it make sense to expose Nid from openssl rather than openssl_sys?

I regenerated the test cert/key in order to add a CN for testing purposes.

Guidance most welcome. :smile:


openssl/test/cert.pem
```diff
@@ -1,21 +1,24 @@
...
-Y2o7+1LiwViGONJ1BfRgdFWMjz0BfQIDAQABo1AwTjAdBgNVHQ4EFgQU509vewJT
-k6qdvjT8e52gTE4+V9EwHwYDVR0jBBgwFoAU509vewJTk6qdvjT8e52gTE4+V9Ew
-DAYDVR0TBAUwAwEB/zANBgkqhkiG9w0BAQsFAAOCAQEAnLxMvrhgQrdKUbd+RSl3
-i6tEfexDepO5QOMIL/m7w/OAkXddu7Y/IuLSgAztp7cu8tQeeBTqbO9T9XKRd1ZD
-hMN3QPZwX1ftREPd1mhe3LQgJ/Q3AbrtlW+0XcANBCTiYi19X193sU2a3Z2nzd5v
-vSvrP1W3Yxzvu7Jwvhgl20BQ0fREpPVfZTrJRhg8/jdLA8TmKZA7012LKh59BCEX
-dXgXc3KYb+Fn8usj79P1nP1YhYmX0Lp0IuC0i1z+FoP/NxNMe+ebXPjM9KlkvQK1
-Vy9sIxV9MAlXpGsMvBJAte94JRikYkQTPjMT1DGQQYGpHIYb3a8jcWvmZPEEaqhB
-Ag==
+MIID9DCCAtygAwIBAgIJALuA8gDKi5EzMA0GCSqGSIb3DQEBCwUAMFkxCzAJBgNV
```
sfackler: Why were these replaced?

openssl-sys/src/lib.rs
```diff
@@ -1,7 +1,9 @@
+#![feature(collections)]
```
sfackler: We can't use rustc features without giving up support for stable Rust distributions.

openssl-sys/src/nid.rs
```diff
@@ -0,0 +1,182 @@
...
+    PBES2,
+    PBMAC1,
+    HmacWithSha1,
+    ID_QT_CPS,
+    ID_QT_UNOTICE,
+    RC2_64_CBC,
+    SMIMECaps
+}
+
+impl CLike for Nid {
```
sfackler: This impl doesn't appear necessary.

openssl/test/cert.pem
josephglanville: To add a CN for for testing purposes.

josephglanville:
Dropped need for collections, changed use of CLike inline cast instead.

Thanks for the tips. :)

openssl/test/cert.pem
sfackler: Cool, makes sense.

openssl/src/x509/mod.rs
```diff
@@ -15,11 +16,44 @@ use crypto::pkey::{PKey,Parts};
...
 use ffi;
 use ssl::error::{SslError, StreamError};
+use ffi::nid;
 
 
 #[cfg(test)]
 mod tests;
 
+#[allow(dead_code)]
+pub struct Asn1String<'x> {
```
sfackler: Is there any reason to keep this around if it's not being used?

openssl/src/x509/mod.rs
```diff
@@ -458,6 +492,44 @@ pub struct X509Name<'x> {
...
+            let mut str_from_asn1 : *mut c_char = ptr::null_mut();
+            let len = ffi::ASN1_STRING_to_UTF8(&mut str_from_asn1, asn1_str);
+
+            if len < 0 {
+                return None
+            }
+
+            assert!(!str_from_asn1.is_null());
+
+            Some(SslString::new(str_from_asn1))
```
sfackler: It might make sense to just convert to a `String` here, since OpenSSL is already guaranteeing it to be UTF8.

openssl-sys/src/nid.rs
```diff
@@ -0,0 +1,169 @@
+#[derive(Copy, Clone)]
+#[repr(usize)]
+pub enum Nid {
```
sfackler: I'd move this into the openssl crate so people don't need to add an explicit dependency on openssl-sys to get a hold of it.

sfackler:
Looks good to me other than those last two things.

openssl-sys/src/nid.rs
josephglanville: Will do.

openssl/src/x509/mod.rs
josephglanville: Can I get ASN1_STRING_to_UTF8 to write to an owned String?
Or do I need to get it write to char array first and copy that into the String?

josephglanville: I played with this for a while and I don't think it makes sense to remove SslString.

SslString is likely to be re-usable elsewhere as there are other textual fields we might want to extract from certificates etc. That and it will perform better than buffer copy into an owned String.

openssl/src/x509/mod.rs
josephglanville: Yeah, no longer being used, will get rid of it.

openssl/src/x509/mod.rs
sfackler: The way I'd probably do it is to get the string out of `ASN1_STRING_to_UTF8`, put it into a `CStr`, get a `&str` out of that and then call `to_owned` to get a String.

Keeping `SslString` seems fine though. I would like it to deref directly to `&str` instead of `&CStr` since we know it's UTF8, though.

josephglanville: Changed to deref to `&str` instread.

jethrogb:
Where did this list of NIDs come from? It does not correspond to my OpenSSL installation.

PR #214, opened 2015-05-14
Add accessor for peer_certificate

josephglanville:
This is useful for authenticating peers.

Is there a better way to do this?


sfackler:
This seems reasonable to me, thanks!

sfackler:
Could you add a quick test to make sure it does what we'd expect it to?

josephglanville:
@sfackler test added.

sfackler:
Thanks!

PR #219, opened 2015-05-28
Implement limited X509_REQ functionality

jethrogb:
This PR adds some functionality dealing with certificate signing requests. In particular:
- Reading from PEM format
- Writing to PEM format
- Creating them with X509Generator

### DIFF
--- a/openssl/src/x509/mod.rs
+++ b/openssl/src/x509/mod.rs
@@ -391,6 +391,20 @@ impl X509Generator {
             Ok(x509)
         }
     }
+
+    /// Obtain a certificate signing request (CSR)
+    pub fn request(&self, p_key: &PKey) -> Result<X509Req, SslError> {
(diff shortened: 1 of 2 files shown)


openssl/src/x509/mod.rs
```diff
@@ -538,6 +552,50 @@ impl <'x> X509Name<'x> {
     }
 }
 
+#[allow(dead_code)]
```
sfackler: This dead code annotation should be able to be removed.

sfackler:
Looks good other than one small nit, thanks!

openssl/src/x509/mod.rs
jethrogb: Oops, that got copied from `struct X509`.

sfackler:
Thanks!

PR #221, opened 2015-05-29
Several SSL option fixes

jethrogb:
This PR adds several missing SSL option flags, fixes incorrect option flags, and adds version detection at build time to select the correct option flags.

I don't quite like how the build script works, but I couldn't come up with a better way. The build script links openssl and checks for the version number. It then outputs the correct flag values to a file per flag. This file is then `include!`d.

I considered setting a cfg option with the version number, but `#[cfg()]` only lets you do string equality testing. While it would be possible to implement this correctly, that would be quite cumbersome.

If you think this is too complicated, you should at least merge eba9af6 to get most of the simple fixes.

### DIFF
--- a/openssl-sys/build.rs
+++ b/openssl-sys/build.rs
@@ -2,7 +2,10 @@ extern crate pkg_config;
 extern crate gcc;
 
 use std::env;
+use std::fmt::Write as FmtWrite;
 use std::path::PathBuf;
+use std::fs::File;
+use std::io::Write;
 
 fn main() {
     let target = env::var("TARGET").unwrap();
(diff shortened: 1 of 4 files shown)


jethrogb:
The Travis CI test fails because openssl-sys gets built without any features as a build dependency. It then does not later get recompiled with the selected features. I don't know how to fix this.

openssl/src/ssl/mod.rs
```diff
@@ -67,14 +69,25 @@ bitflags! {
         const SSL_OP_SINGLE_ECDH_USE = 0x00080000,
         const SSL_OP_SINGLE_DH_USE = 0x00100000,
         const SSL_OP_CIPHER_SERVER_PREFERENCE = 0x00400000,
-        const SSL_OP_TLS_ROLLBACK_BUG = 0x00800000,
-        const SSL_OP_NO_SSLV2 = 0x00000000,
+        const SSL_OP_TLS_ROLLBACK_BUG = 0x00800000, // 0x00000400 before 0.9.7
+        const SSL_OP_NO_SSLV2 = 0x01000000,
```
sfackler: In what versions does `SSL_OP_NO_SSLV2` exist? It doesn't seem to be present in either 0.9.8zf or 1.0.2a.

openssl/src/ssl/mod.rs
```diff
@@ -67,14 +69,25 @@ bitflags! {
...
+        const SSL_OP_NO_DTLSV1_2 = 0x08000000,
+        const SSL_OP_NETSCAPE_CA_DN_BUG = 0x20000000,
+        const SSL_OP_NETSCAPE_DEMO_CIPHER_CHANGE_BUG = 0x40000000, // 0x80000000 before 0.9.7
+        const SSL_OP_CRYPTOPRO_TLSEXT_BUG = 0x80000000,
+        const SSL_OP_PKCS1_CHECK_1 = 0x0, // 0x08000000 before 1.0.1
+        const SSL_OP_PKCS1_CHECK_2 = 0x0, // 0x10000000 before 1.0.1
+        const SSL_OP_EPHEMERAL_RSA = 0x0, // used to be 0x00200000; CVE-2015-0204/commit 4b4c1fc
+                                          // Only allow ephemeral RSA keys in export ciphersuites
+        const SSL_OP_MSIE_SSLV2_RSA_PADDING = 0x0, // 0x00000040 before, no effect since 0.9.7h and 0.9.8b
+        const SSL_OP_SSLREF2_REUSE_CERT_TYPE_BUG = 0x0, // 0x00000010 before 1.0.1h / 1.0.2
```
sfackler: Why did these stick around if they've been set to 0 (in openssl, not here)?

sfackler:
There's enough weirdness here with flag values changing/being added in various versions of OpenSSL that we should really divorce the SslContextOptions type from the OpenSSL bitfield. We'd have a set of functions in the C shim to translate between SslContextOptions and OpenSSL's values that can make proper use of the headers.

openssl/src/ssl/mod.rs
jethrogb: It's definitely in both 0.9.8zf and 1.0.2a. Note that the C version uses different capitilization: `SSL_OP_NO_SSLv2`.

openssl/src/ssl/mod.rs
jethrogb: So that legacy code will still compile with newer versions of OpenSSL.

openssl/src/ssl/mod.rs
sfackler: Ah, that'd do it, thanks.

jethrogb:
That could work.

jethrogb:
I have no good story for SSL_OP_ALL. I just define it as all flags that were in range 0x80000bff at the time I wrote the code.

sfackler:
Seems like `SSL_OP_ALL` can just be `!0` since the C code tests for each flag individually, right?

jethrogb:
Aha, I see you think SSL_OP_ALL actually means "every single flag". That, while a rational thought, is in fact incorrect. As per https://www.openssl.org/docs/ssl/SSL_CTX_set_options.html:

> The following bug workaround options are available:
> SSL_OP_...
> ...
> SSL_OP_ALL
> > All of the above bug workarounds.
>
> It is usually safe to use SSL_OP_ALL to enable the bug workaround options if compatibility with somewhat broken implementations is desired.

Now here's the kicker: the value of this flag changes between versions and not all bits mean anything.

Also, I think defining it in terms of actual flags makes more sense, since that way you can do something like: `get_options()&SSL_OP_ALL==SSL_OP_ALL`.

sfackler:
... I hate everything.

The way this deals with `SSL_OP_ALL` seems fine given what we're working with.

sfackler:
I think you should be able to get rid of all of the `#if RUST_FOO != SSL_FOO` guards. If the values are equal, the statement is a trivial noop (if bit x exists, add bit x) that gcc should have no trouble removing.

jethrogb:
You have to consider the case where the flag doesn't exist in C. But you might still be right.

sfackler:
The first part of the shim file will `#define` any of the missing values to 0, won't it?

openssl-sys/build.rs
```diff
@@ -52,7 +55,67 @@ fn main() {
...
+
+    writeln!(shim,"#define COPY_MASK ( \\").unwrap();
+
+    let mut it=options.iter().peekable();
+    while let Some(&(name,_))=it.next()  {
+        let eol=match it.peek() {
+            Some(_) => " | \\",
+            None    => " )"
+        };
+        writeln!(shim,"    ((RUST_{0}==(uint64_t)(uint32_t){0})?RUST_{0}:UINT64_C(0)){1}",name,eol).unwrap();
```
sfackler: Why does this cast to a u32 and then to a u64? To clear the high bits?

jethrogb: I'm not super-familiar with C semantics, and I wanted to be sure it didn't do sign extension.

jethrogb:
I have a commit ready to remove the `#if`s. It does work correctly. The optimization level of the C compiler depends on the optimization level of the current Rust build. This means that the debug build does not eliminate these useless ifs. If you're ok with that, I'll push the commit.

sfackler:
Yeah that seems fine. I don't believe setting ssl options is going to be in the hot path very often :)

jethrogb:
In d0b769c9 you moved the `SSL_CTX_*_options` shims to C. I can either move them back to Rust or change the ABI (we need to pass `u64`), what do you prefer?

sfackler:
Let's go with number 3 and do the translation call followed by the call to the shim in the functions on `SslContext`.

jethrogb:
Done.

Issue #177, opened 2015-03-05
Compile on cygwin ?


ctjhoa:
When I try to compiled openssl-sys on cygwin I've got this error.
```
Compiling openssl-sys v0.5.0 (file:///C:/Users/ctjhoa/rust/rust-openssl/openssl-sys)
failed to run custom build command for `openssl-sys v0.5.0 (file:///C:/Users/ctjhoa/rust/rust-openssl/openssl-sys)`
Process didn't exit successfully: `C:\Users\ctjhoa\rust\rust-openssl\openssl-sys\target\build\openssl-sys-4f8c0d5752fc409a\build-script-build.exe` (exit code: 101)
--- stdout
cargo:rustc-flags=-l dylib=eay32 -l dylib=ssl32
TARGET = Some("x86_64-pc-windows-gnu")
CARGO_MANIFEST_DIR = Some("C:\\Users\\ctjhoa\\rust\\rust-openssl\\openssl-sys")
OUT_DIR = Some("C:\\Users\\ctjhoa\\rust\\rust-openssl\\openssl-sys\\target\\build\\openssl-sys-4f8c0d5752fc409a\\out")
TARGET = Some("x86_64-pc-windows-gnu")
OPT_LEVEL = Some("0")
PROFILE = Some("compile")
compile 0
TARGET = Some("x86_64-pc-windows-gnu")
HOST = Some("x86_64-pc-windows-gnu")
CC_x86_64-pc-windows-gnu = None
CC_x86_64_pc_windows_gnu = None
HOST_CC = None
CC = None
TARGET = Some("x86_64-pc-windows-gnu")
HOST = Some("x86_64-pc-windows-gnu")
CFLAGS_x86_64-pc-windows-gnu = None
CFLAGS_x86_64_pc_windows_gnu = None
HOST_CFLAGS = None
CFLAGS = None
TARGET = Some("x86_64-pc-windows-gnu")
HOST = Some("x86_64-pc-windows-gnu")
CC_x86_64-pc-windows-gnu = None
CC_x86_64_pc_windows_gnu = None
HOST_CC = None
CC = None
[...]

PR #225, opened 2015-06-10
Added support for building on Windows with MinGW

semmaz:
Currently, linking against openssl libraries fails on windows, unless you hint build script with proper [environment variables](https://github.com/sfackler/rust-openssl#windows), or by copying openssl libs in one of the default paths supplied to rustc by cargo. This PR should enable proper linking against openssl while using one of the MinGW distributions (tested with MSYS2).

Somewhat related issue - #177 
I think issue there is same - linking fails because gcc couldn't find libs in paths provided (not because of .dll suffix, gcc bundled with rust and msys is able to link against both .dll & .a)

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,15 @@ rust-openssl to a separate installation.
 
 ### Windows
 
-Install OpenSSL from [here][1]. Cargo will not be able to find OpenSSL if it's
+On Windows, consider building with [mingw-w64](http://mingw-w64.org/).
+Build script will try to find mingw in `PATH` environment variable to provide
+Cargo with location where openssl libs from mingw-w64 package may be found.
+If you followed guide [Building on Windows](https://github.com/rust-lang/rust#building-on-windows)
+from rust repo, then you should have [MSYS2](http://msys2.github.io/) with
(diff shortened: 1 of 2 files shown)


voltagex:
The AppVeyor failure looks unrelated, fwiw.

`Specify a project or solution file. The directory does not contain a project or solution file.`

sfackler:
Yep, I've turned on appveyor support but haven't actually set it up yet.

Just to be clear, does will this link to an openssl installed through pacman in Msys2?

semmaz:
Yep, links to these two to be specific:
```
$ pacman -Ql mingw-w64-x86_64-openssl | grep '^.*bin.*\.dll$'
mingw-w64-x86_64-openssl /mingw64/bin/libeay32.dll
mingw-w64-x86_64-openssl /mingw64/bin/ssleay32.dll
```

sfackler:
Awesome. Could you update the README to mention this new, much less terrible, way of building on Windows? Looks good to me other than that.

semmaz:
Done. Should I also include guide from rust's repo?

sfackler:
Thanks!

PR #226, opened 2015-06-13
Pass symmetric crypto initialization vector as a slice

jethrogb:
There is really no reason not to pass slices here.

Note: This change is backwards-incompatible. Users will need to turn their
parameters into references.


gkoz:
Uh-oh, maybe I need to finish #173 at last. ;)

sfackler:
If we change these signatures to e.g.

```rust
pub fn init<'a, T: As<&'a [u8]>(&self, mode: Mode, key: &[u8], iv: T) {
```

We can avoid breaking back-compat while allowing slices to be passed. It's a bit of a weird thing to use in the long term, so we'd probably switch back to a straight `&[u8]` at the next version bump.

jethrogb:
Done. The syntax is `<T: AsRef<[u8]>>`.

sfackler:
Thanks!

sfackler:
Pushed the revert to the breaks branch: https://github.com/sfackler/rust-openssl/commit/db51bc43bdfd28d27c4826614fffb1576ef4b793

PR #227, opened 2015-06-15
Allow setting of arbitrary X509 names

jethrogb:
_Note added 2015-06-24: GitHub is usually pretty good in keeping obsolete commits in the PR history, but I guess not in this case. The original PR had `names` as a publicly-accessible field on the `X509Generator`._

### DIFF
--- a/openssl/src/x509/mod.rs
+++ b/openssl/src/x509/mod.rs
@@ -209,6 +209,17 @@ impl X509Generator {
         self
     }
 
+    /// Add multiple attributes to the name of the certificate
+    ///
+    /// ```ignore
+    /// generator.add_names(vec![("CN".to_string(),"example.com".to_string())]);
+    /// ```
+    pub fn add_names<I>(mut self, attrs: I) -> X509Generator
+        where I: IntoIterator<Item=(String,String)> {
+        self.names.extend(attrs);
+        self
+    }
+
     /// (deprecated) Sets what for certificate could be used
     ///
     /// This function is deprecated, use `X509Generator.add_extension` instead.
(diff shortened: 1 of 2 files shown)


sfackler:
I'd prefer to keep all the fields private and have an `add_name` method that would add a new name key/value pair.

jethrogb:
I would've preferred to also add append_names or so, but stable Vec does not have a method for that.

sfackler:
What method is `Vec` missing specifically - I would think `push` would be sufficient?

sfackler:
I haven't seen `Into<String>` used much elsewhere, so I think we should avoid it for now. It's backwards compatible to add it later if we feel like.

jethrogb:
With `append_names` I meant something that would add multiple names at the same time, like `Vec.append` or `Vec.push_all`, both unstable.

sfackler:
`Vec::extend` should do that: https://doc.rust-lang.org/stable/collections/vec/struct.Vec.html#method.extend

jethrogb:
I don't know what's up with Travis.

sfackler:
I screwed something up in the Travis setup - should be fixed tonight.

sfackler:
Travis should be working now if you rebase on top of master again.

jethrogb:
Is there anything that's blocking merging this PR?

sfackler:
Nope, looks good - Github doesn't notify me when a PR gets rebased unfortunately :(

sfackler:
Oh, sorry, it also looks like it needs (another) rebase.

jethrogb:
Rebased again. I'll keep that in mind for future PRs.

PR #229, opened 2015-06-25
Add support for ALPN ("successor" to NPN)

jmesmon:
Implementation and interface is very similar, tests I've created are identical.
It's possible we might be able to reduce duplication between the NPN and ALPN code & tests at some point.

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -306,6 +306,24 @@ extern fn raw_next_protos_advertise_cb(ssl: *mut ffi::SSL,
     ffi::SSL_TLSEXT_ERR_OK
 }
 
+/// Convert a set of byte slices into a series of byte strings encoded for SSL. Encoding is a byte
+/// containing the length followed by the string.
+#[cfg(feature = "npn")]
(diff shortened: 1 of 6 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -554,6 +571,29 @@ impl SslContext {
             ffi::SSL_CTX_set_next_protos_advertised_cb(self.ctx, raw_next_protos_advertise_cb, ptr::null_mut());
         }
     }
+
+    /// Set the protocols to be used during ALPN (application layer protocol negotiation).
+    /// If this is a server, these are the protocols we report to the client.
```
sfackler: Can you add a sentence saying that this requires the `alpn` feature?

sfackler:
Needs a rebase and a sentence added to one of the doc blocks, but looks good to me other than that, thanks!

openssl/src/ssl/mod.rs
jmesmon: Done

jmesmon:
I've updated with the comment requested & a rebase

sfackler:
Thanks!

PR #231, opened 2015-06-26
Add ability to determine current cipher for a session.

gcmurphy:


sfackler:
Seems like we might as well wrap all of the relevant SSL_CYPHER functionality:

```rust
impl<S> SslStream<S> {
    pub fn cypher(&self) -> Option<Cypher> { ... }
}

pub struct Cypher { ... }

impl Cypher {
    pub fn name(&self) -> String { ... }
    pub fn version(&self) -> String { ... }
    pub fn description(&self) -> String { ... }
}
```

gcmurphy:
Closing will re-submit after rebase etc.

PR #232, opened 2015-07-01
Fix NID definitions to match OpenSSL.

jethrogb:
The previous numbers were introduced incorrectly in #213.

The numbers are defined in `openssl-src/crypto/objects/obj_mac.num`. This file features occasional deletions and then reassignments (argh), but the last occurence of that was in 2008/OpenSSL 0.9.8.


sfackler:
Could you add a test that reads a couple of values by NID out of a certificate with known values to make sure the variants stay correct?

jethrogb:
Good idea.

sfackler:
Thanks!

PR #233, opened 2015-07-01
Allow setting of arbitrary X509 extensions

jethrogb:
Similar to #227, except for X509 extensions. This is a pretty big overhaul again of the existing X509generator extension interface.

~~Depends on #232~~ _(merged)_


jethrogb:
Currently, a `SubjectAltName` holds a `Vec<(AltNameOption,String)>`. Alternatively, it could hold a `Vec<AltNameOption>`, where each `AltNameOption` variant holds a `String` as data. I'm not sure why to prefer one way over the other.

sfackler:
Can a `SubjectAltName` or `IssuerAltName` have multiple names with the same option? If not, it might make more sense to those to store a `HashMap<AltNameOption, String>`. If duplicates are allowed, then I think the setup with tuples seems fine.

jethrogb:
Definitely multiple allowed. See e.g. the cert on https://steamstore-a.akamaihd.net/public/images/v6/icon_platform_linux.png

sfackler:
Cool, this seems good to me then, thanks!

PR #236, opened 2015-07-06
Unpin the bitflags version.

Ms2ger:
This dependency causes Servo to depend on multiple versions of the bitflags
crate.

### DIFF
--- a/openssl/Cargo.toml
+++ b/openssl/Cargo.toml
@@ -24,7 +24,7 @@ path = "../openssl-sys"
 version = "0.6.3"
 
 [dependencies]
-bitflags = "0.2"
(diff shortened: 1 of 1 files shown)


sfackler:
Can this be changed to a specific version range (e.g. `>= 0.2, < 0.4`)? A `*` dependency is almost always a terrible idea.

Ms2ger:
That only pushes the issue back a little, for probably no practical value, but sure.

sfackler:
The practical value is that I guarantee that this version of openssl will not work with future versions of bitflags.

sfackler:
Published as v0.6.4

PR #240, opened 2015-07-08
Implement certificate extensions for certificate requests

### LINKED ISSUE #238: X509Generator.request doesn't include certificate extensions in the request
Introduced in #233. I'm on it.

jethrogb:
Fixes #238.

I had to implement a shim to access a field of a `struct X509` since that is implemented in Rust as an opaque type.


sfackler:
OpenSSL 1.1 is making X509 an opaque type at the header level - are you aware of the new way of grabbing the extension list?

jethrogb:
Looks like they added X509_get0_extensions for this purpose.

sfackler:
Ah great, thanks

PR #243, opened 2015-07-18
Fix probelms with DTLS when no packets are pending.

manuels:
When using DTLS you might run into the situation where no packets
are pending, so SSL_read returns len=0. On a TLS connection this
means that the connection was closed, but on DTLS it does not
(a DTLS connection cannot be closed in the usual sense).
This commit fixes a bug introduced by c8d23f3.


manuels:
Looks like there is something wrong with the ubuntu repos?

Issue #245, opened 2015-07-22
Clarify alternate Windows build procedure


Zarathustra30:
The alternative procedure for Windows build should be clarified

> You can ... copy the include/openssl directory, libssl32.dll, and libeay32.dll to locations that Cargo can find ...

It would help to list _actual_ locations for where to put the files. The Cargo documentation does not include any information on dynamic linking, so "locations that Cargo can find" is rather opaque, as evidenced by my inability to successfully follow the build procedure.

sfackler:
Did you try the first method of just installing the `mingw-w64-x86_64-toolchain` package in mingw?

Zarathustra30:
I did not. I do not have mingw, and I would prefer to not have to learn another tool while it appears there is a way to build without it.

sfackler:
I'm not totally sure where rustc searches for `dll`s and gcc searches for header files on Windows, unfortunately. The way I've built this library on Windows when not using a mingw-provided openssl is to use the environment variables to explicitly state where the stuff is: `env OPENSSL_LIB_DIR=C:/OpenSSL-Win64 OPENSSL_INCLUDE_DIR=C:/OpenSSL-Win64/include cargo build`.

semmaz:
Even with providing environment variables, rust-openssl still fails to build, mingw works though (after installing toolchain).
This appears to be the issue with [gcc-rs](https://github.com/alexcrichton/gcc-rs/issues/7). I was about to file new issue there, until I discovered it. Should I do that (essentially reopening issue) or just removing alternative build method will be enough?

Zarathustra30:
Removing the alternative build method would be enough to avoid any confusion and close this issue, for the time being. I've updated the report to reflect this.  The build failure [has already been reported](https://github.com/sfackler/rust-openssl/issues/241), so we can keep track of it over there.

However, it would be nice to be able to use SSL without building with mingw, so I think the gcc-rs issue should be reopened.

Zarathustra30:
Okay, now that the alternative to building with MinGW no longer requires MinGW, the original intent of this issue is once again relevant. "Locations that Cargo can find" is too ambiguous, and should be made more specific or removed altogether.

sfackler:
I'm on board with removing that phrase.

luser:
To get this building on Windows with the Win64OpenSSL package I had to do the following:
```
export OPENSSL_LIB_DIR=C:/OpenSSL-Win64/lib
[...]

PR #246, opened 2015-07-23
Removed Windows alternative build procedure

Zarathustra30:
Closes https://github.com/sfackler/rust-openssl/issues/245

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,22 +29,13 @@ rust-openssl to a separate installation.
 
 ### Windows
 
-On Windows, consider building with [mingw-w64](http://mingw-w64.org/).
+On Windows, build with [mingw-w64](http://mingw-w64.org/).
 Build script will try to find mingw in `PATH` environment variable to provide
 Cargo with location where openssl libs from mingw-w64 package may be found.
 If you followed guide [Building on Windows](https://github.com/rust-lang/rust#building-on-windows)
 from rust repo, then you should have [MSYS2](http://msys2.github.io/) with
 `mingw-w64-openssl` installed as part of `mingw-w64-x86_64-toolchain`
 (or `mingw-w64-i686-toolchain`) package.
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,15 +37,6 @@ from rust repo, then you should have [MSYS2](http://msys2.github.io/) with
 `mingw-w64-openssl` installed as part of `mingw-w64-x86_64-toolchain`
 (or `mingw-w64-i686-toolchain`) package.
 
-Alternatively, install OpenSSL from [here][1]. Cargo will not be able to find OpenSSL if it's
-installed to the default location. You can either copy the `include/openssl`
-directory, `libssl32.dll`, and `libeay32.dll` to locations that Cargo can find
-or pass the location to Cargo via environment variables:
-
-```bash
-env OPENSSL_LIB_DIR=C:/OpenSSL-Win64 OPENSSL_INCLUDE_DIR=C:/OpenSSL-Win64/include cargo build
```
sfackler: This method definitely works - why should it be removed?

Zarathustra30: The alternate Windows method only works if MinGW-w64 is already installed, due to the dependency on [gcc-rs](https://github.com/alexcrichton/gcc-rs) (which, in turn, depends on MinGW's version of gcc), so it should not be considered an alternative to using MinGW-w64.

However, when I authored the pull, I was unaware of gcc-rs's dependency on MinGW-w64, and thought the failed compile was an actual bug. The pull would have been a band-aid until the bug was fixed.

This pull may be slightly confused, but the README should reflect that rust-openssl cannot be built without MinGW.

Zarathustra30:
Welp. The gcc from Cygwin [can now be used](https://github.com/alexcrichton/gcc-rs/issues/51), so the alternative build procedure is no longer redundant, rendering this pull obsolete. However, the dependency on gcc should still be more explicit.

PR #247, opened 2015-07-26
Expose ssl::init

panicbit:
Allow other crates (e.g. ones that wrap a C library which requires openssl) to easily initialize openssl


sfackler:
Thanks!

PR #251, opened 2015-08-04
Expose EVP_BytesToKey

ebarnard:
This is based on #66 by pyrho, which looks dead, but copy and pasted as I'm not very used to git.

### DIFF
--- a/openssl/src/crypto/pkcs5.rs
+++ b/openssl/src/crypto/pkcs5.rs
@@ -1,5 +1,68 @@
 use libc::c_int;
+use std::ptr::null;
+
+use crypto::symm_internal::evpc;
+use crypto::hash;
+use crypto::symm;
 use ffi;
+
+#[derive(Clone, Eq, PartialEq, Hash, Debug)]
+pub struct KeyIvPair
+{
+    pub key: Vec<u8>,
+    pub iv: Vec<u8>
+}
+
+/// Derives a key and an IV from various parameters.
+///
+/// If specified `salt` must be 8 bytes in length.
+///
+/// If the total key and IV length is less than 16 bytes and MD5 is used then 
+/// the algorithm is compatible with the key derivation algorithm from PKCS#5 
+/// v1.5 or PBKDF1 from PKCS#5 v2.0.
+///
+/// New applications should not use this and instead use `pbkdf2_hmac_sha1` or 
+/// another more modern key derivation algorithm.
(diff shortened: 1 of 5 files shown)


openssl/src/crypto/pkcs5.rs
```diff
@@ -1,6 +1,42 @@
...
+
+use crypto::symm_internal::evpc;
+use crypto::hash;
+use crypto::symm;
 use ffi;
 
+/// Derives a key and an IV from various parameters.
+/// New applications should use pbkdf2_hmac_sha1 or another more modern key
+/// derevation algorithm
+pub fn evp_bytestokey(typ: symm::Type, message_digest_type: hash::Type,
```
sfackler: This could use a more descriptive name.

openssl/src/crypto/pkcs5.rs
```diff
@@ -1,6 +1,42 @@
...
+                      salt: Option<&[u8]>, data: &[u8],
+                      count: u32) -> (Vec<u8>, Vec<u8>) {
```
sfackler: Could this return a struct with named fields rather than a pair? It seems like it'll be really easy to mix up the key and the IV otherwise.

openssl/src/crypto/pkcs5.rs
ebarnard: Changed it to evp_bytes_to_key_pbkdf1_compatible. It's not great name but it's quite a weird function.

https://www.openssl.org/docs/crypto/EVP_BytesToKey.html

openssl/src/crypto/pkcs5.rs
```diff
@@ -1,6 +1,72 @@
...
+/// v1.5 or PBKDF1 from PKCS#5 v2.0.
+///
+/// New applications should not use this and instead use `pbkdf2_hmac_sha1` or 
+/// another more modern key derivation algorithm.
+pub fn evp_bytes_to_key_pbkdf1_compatible(typ: symm::Type, message_digest_type: hash::Type,
+                      data: &[u8], salt: Option<&[u8]>,
+                      count: u32) -> KeyIvPair {
+
+    // define from include/openssl/evp.h
+    const PKCS5_SALT_LEN: usize = 8;
```
sfackler: This should probably live in openssl-sys.

ebarnard:
@sfackler Addressed nits. Good to go?

sfackler:
Thanks!

PR #253, opened 2015-08-14
Add get_state_string()

manuels:
Pretty useful for debugging ;)

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -6,6 +6,7 @@ use std::fmt;
 use std::io;
 use std::io::prelude::*;
 use std::mem;
+use std::str;
 use std::net;
 use std::path::Path;
 use std::ptr;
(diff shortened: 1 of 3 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -690,6 +691,24 @@ impl Ssl {
         Ok(ssl)
     }
 
+    pub fn get_state_string<'a>(&self) -> &'a str {
+        let state = unsafe {
```
sfackler: What is the lifetime of the strings returned by these methods? If it's static, returning a `&'static str` is a bit more clear than an unbound lifetime.

manuels: The lifetime isn't static, so an unbound lifetiem should be appropriate:

    const char *SSL_state_string(const SSL *ssl);
    const char *SSL_state_string_long(const SSL *ssl);

sfackler: An unbounded lifetime is equivalent to static: http://is.gd/xjEcoa

manuels:
Sorry, I fix the issue with the static lifetime.

Issue #234, opened 2015-07-02
Extend PKey module to better support DH, DSA, EC


jethrogb:
The current PKey module is very RSA-centric. Loading of DSA and EC keys is possible, but the following functions implement RSA-specific functionality only:
- gen
- save_pub
- load_pub
- save_priv
- load_priv
- size
- sign_with_hash
- verify_with_hash

load/save should be implemented using `x2y_PUBKEY` and `x2y_PrivateKey` where x/y are i/d or d/i.

These functions are also RSA-centric, but can't work with any of the other algorithms:
- max_data
- encrypt_with_padding
- decrypt_with_padding

aij:
Is anyone working on this? I'd like to use EC.

jethrogb:
I'm not currently, and also not planning to do this within a couple of weeks.

awelkie:
What's the benefit of having one struct for all of the different types of keys? Would this API be better if there were different key structs (`ECKey`, `RSAKey`, `DSAKey`) instead of one `PKey` struct (even if they were just wrappers around a `PKey` pointer)?

jethrogb:
I can be convinced that they should all be different structs, although this makes loading unknown-type key file harder. They should all exhibit similar interfaces though, since they all support some subset of the same common operations:
* parameter generation
* key generation
* encryption/decryption
* signing/verification
* key derivation
* inputting into an SSL struct

I've been looking into this some more and I also think that most of the EVP_PKEY API is the wrong API to export. We should really be looking more at the `EVP_SealInit` and `EVP_SignInit` families of functions. Unfortunately there is no counterpart for `EVP_PKEY_keygen` and `EVP_PKEY_derive` functions, so those need to be exposed at the lower level.

bbatha:
I'd like to help out with this, at least to add DSA keys, and possibly the overarching traits. Though, I'd need some assistance designing the traits.

sfackler:
I believe this has been resolved in the 0.8 rewrite. If anything's still missing, please open new issues!

PR #256, opened 2015-08-28
Add elliptical curve Diffie Hellman function

awelkie:
This exposes enough of the elliptical curve functionality to generate keys, create shared secrets from another public key, and create shared secrets from public and private keys.

A few things to note (and that may need to change):
* I use the `free` instead of the `clear_free` functions to free memory. Is there any reason to prefer the `clear_free` functions?
* I have an unnecessary copy in `get_public_key`, but I'm not sure how to just safely return the pointer without having a double-free later.
* I made the `BigNum::raw` method public because I needed to use it.
* I exposed the macros from `bn` because I used the `with_ctx` macro.

### DIFF
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -21,6 +21,9 @@ pub type BN_CTX = c_void;
 pub type COMP_METHOD = c_void;
 pub type CRYPTO_EX_DATA = c_void;
 pub type ENGINE = c_void;
+pub type EC_GROUP = c_void;
+pub type EC_KEY = c_void;
+pub type EC_POINT = c_void;
 pub type EVP_CIPHER = c_void;
 pub type EVP_CIPHER_CTX = c_void;
 pub type EVP_MD = c_void;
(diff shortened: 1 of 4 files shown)


jethrogb:
Wouldn't it be better to expose most of this through the PKey API?

awelkie:
I chose to wrap the EC and ECDH APIs because the EVP API seemed a lot more complicated for the same functionality. What's the benefit in using the EVP (and PKey) API? It seems the `PKey` struct is tied to RSA, at least in this library.

jethrogb:
> What's the benefit in using the EVP (and PKey) API?

From a practical perspective, most of OpenSSL uses EVP_PKEY as the main carrier for public/private keys, e.g. in `SSL_CTX_use_PrivateKey`.

From a library design perspective, it would be cleaner to keep all similar functionality (in this: public-key crypto) in the same module.

From a security engineering perspective, using the low-level interfaces to roll your own crypto is strongly discouraged. The EVP API is designed for users of standard crypto, with less potential for your implementation to be a footgun.

> It seems the PKey struct is tied to RSA

Not really

> at least in this library.

Yes, unfortunately, see #234

Issue #235, opened 2015-07-02
Implement Ephemeral DH server-side support


jethrogb:
DHE/ECDHE is currently not supported for servers since there is no way to initialize the server-side DH parameters. This means running servers with DSS or ECDSA keys is also not supported, since all cipher suites with those authentication mechanisms depend on DHE/ECDHE.

The required functions are

- [x] SSL_CTX_set_tmp_dh_callback
- [x] SSL_CTX_set_tmp_dh
- [x] SSL_set_tmp_dh_callback
- [x] SSL_set_tmp_dh

for DHE and 

- [x] SSL_CTX_set_tmp_ecdh_callback
- [x] SSL_CTX_set_tmp_ecdh
- [x] SSL_set_tmp_ecdh_callback
- [x] SSL_set_tmp_ecdh

for ECDHE pre-1.0.2 and

- [x] SSL_CTX_set_ecdh_auto
- [x] SSL_set_ecdh_auto

for ECDHE on 1.0.2 and onwards. It would probably also be a good idea to make available

- [x] get_rfc2409_prime_768 
- [x] get_rfc2409_prime_1024
- [x] get_rfc3526_prime_1536
- [x] get_rfc3526_prime_2048
- [x] get_rfc3526_prime_3072
- [x] get_rfc3526_prime_4096
- [x] get_rfc3526_prime_6144
- [x] get_rfc3526_prime_8192
- [x] EC_KEY_new_by_curve_name

to be able to load the default parameters, although using these 768-bit and 1024-bit values is [not recommended for security reasons](https://weakdh.org)

This depends on #234 to able to pass around EC and DH PKeys.

See for usage of these APIs for example Apache mod_ssl:
https://github.com/apache/httpd/blob/9edfd64/modules/ssl/ssl_engine_kernel.c#L1343-1383 
https://github.com/apache/httpd/blob/9edfd64/modules/ssl/ssl_engine_init.c#L1182-1219

sfackler:
I believe this is now all implemented!

PR #259, opened 2015-08-31
Add support for DHE for forward secrecy

jedisct1:
Add support for `set_tmp_dh()` and RFC5114 DH parameters for forward secrecy.

rust-openssl didn't support forward secrecy at all.

This adds support for DHE, by exposing set_tmp_dh() as well as the RFC5114 parameters, which are conveniently exposed since OpenSSL 1.0.2.

With OpenSSL >= 1.0.2, and the rfc5114 feature gate, enabling DHE is as simple as (here for 2048-bit MODP group with 256-bit prime order subgroup):

        use openssl::dh::DH;
        let dh = DH::get_2048_256().unwrap();
        ctx.set_tmp_dh(dh).unwrap();

With OpenSSL < 1.0.2, DH::from_params() can be used to manually specify the DH parameters (here for 2048-bit MODP group with 256-bit prime order subgroup):

        use openssl::bn::BigNum;
        use openssl::dh::DH;
        let p = BigNum::from_hex_str("87A8E61DB4B6663CFFBBD19C651959998CEEF608660DD0F25D2CEED4435E3B00E00DF8F1D61957D4FAF7DF4561B2AA3016C3D91134096FAA3BF4296D830E9A7C209E0C6497517ABD5A8A9D306BCF67ED91F9E6725B4758C022E0B1EF4275BF7B6C5BFC11D45F9088B941F54EB1E59BB8BC39A0BF12307F5C4FDB70C581B23F76B63ACAE1CAA6B7902D52526735488A0EF13C6D9A51BFA4AB3AD8347796524D8EF6A167B5A41825D967E144E5140564251CCACB83E6B486F6B3CA3F7971506026C0B857F689962856DED4010ABD0BE621C3A3960A54E710C375F26375D7014103A4B54330C198AF126116D2276E11715F693877FAD7EF09CADB094AE91E1A1597").unwrap();
        let g = BigNum::from_hex_str("3FB32C9B73134D0B2E77506660EDBD484CA7B18F21EF205407F4793A1A0BA12510DBC15077BE463FFF4FED4AAC0BB555BE3A6C1B0C6B47B1BC3773BF7E8C6F62901228F8C28CBB18A55AE31341000A650196F931C77A57F2DDF463E5E9EC144B777DE62AAAB8A8628AC376D282D6ED3864E67982428EBC831D14348F6F2F9193B5045AF2767164E1DFC967C1FB3F2E55A4BD1BFFE83B9C80D052B985D182EA0ADB2A3B7313D3FE14C8484B1E052588B9B7D2BBD2DF016199ECD06E1557CD0915B3353BBB64E0EC377FD028370DF92B52C7891428CDC67EB6184B523D1DB246C32F63078490F00EF8D647D148D47954515E2327CFEF98C582664B4C0F6CC41659").unwrap();
        let q = BigNum::from_hex_str("8CF83642A709A097B447997640129DA299B1A47D1EB3750BA308B0FE64F5FBD3").unwrap();
        let dh = DH::from_params(p, g, q).unwrap();
        ctx.set_tmp_dh(dh).unwrap();

### DIFF
--- a/openssl-sys/src/openssl_shim.c
+++ b/openssl-sys/src/openssl_shim.c
@@ -1,5 +1,7 @@
 #include <openssl/hmac.h>
 #include <openssl/ssl.h>
+#include <openssl/dh.h>
+#include <openssl/bn.h>
 
 #if OPENSSL_VERSION_NUMBER < 0x1000000L
 // Copied from openssl crypto/hmac/hmac.c
@@ -77,6 +79,22 @@ long SSL_CTX_add_extra_chain_cert_shim(SSL_CTX *ctx, X509 *x509) {
 
 long SSL_CTX_set_read_ahead_shim(SSL_CTX *ctx, long m) {
     return SSL_CTX_set_read_ahead(ctx, m);
+}
+
+long SSL_CTX_set_tmp_dh_shim(SSL_CTX *ctx, DH *dh) {
+    return SSL_CTX_set_tmp_dh(ctx, dh);
+}
+
+DH *DH_new_from_params(BIGNUM *p, BIGNUM *g, BIGNUM *q) {
+    DH *dh;
+
+    if ((dh = DH_new()) == NULL) {
+        return NULL;
+    }
+    dh->p = p;
+    dh->g = g;
+    dh->q = q;
+    return dh;
 }
 
 long SSL_set_tlsext_host_name_shim(SSL *s, char *name) {
--- a/openssl/src/dh/mod.rs
+++ b/openssl/src/dh/mod.rs
@@ -1,0 +1,97 @@
+use ffi;
+use ssl::error::SslError;
+use bn::BigNum;
+use std::mem;
+use std::ptr;
+
+pub struct DH(*mut ffi::DH);
+
+impl DH {
+    pub fn from_params(p: BigNum, g: BigNum, q: BigNum) -> Result<DH, SslError> {
+        let dh = unsafe { ffi::DH_new_from_params(p.raw(), g.raw(), q.raw()) };
+        if dh == ptr::null_mut() {
+            return Err(SslError::get());
+        }
+        mem::forget(p);
+        mem::forget(g);
+        mem::forget(q);
+        Ok(DH(dh))
+    }
+
+    #[cfg(feature = "rfc5114")]
+    pub fn get_1024_160() -> Result<DH, SslError> {
+        let dh = unsafe { ffi::DH_get_1024_160() };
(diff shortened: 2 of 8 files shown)


openssl-sys/src/openssl_shim.c
```diff
@@ -86,3 +92,295 @@ long SSL_set_tlsext_host_name_shim(SSL *s, char *name) {
 STACK_OF(X509_EXTENSION) *X509_get_extensions_shim(X509 *x) {
     return x->cert_info ? x->cert_info->extensions : NULL;
 }
+
+#if OPENSSL_VERSION_NUMBER < 0x1000200L
```
sfackler: This seems like a bit much to add here - it seems like it would make more sense to put this stuff under a feature flag like we do for other things that require modern versions e.g. DTLS.

jedisct1:
Fair enough. Pull request updated. The standard DH parameters are now behind a `rfc5114` feature gate.
For older OpenSSL versions, `DH::from_params()` has been added, to manually specify the parameters.

sfackler:
Thanks!

jethrogb:
Why not RFC 3526 parameters? See also #235

openssl/src/dh/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+        mem::forget(p);
+        mem::forget(g);
+        mem::forget(q);
+        Ok(DH(dh))
+    }
+
+    #[cfg(feature = "rfc5114")]
+    pub fn get_1024_160() -> Result<DH, SslError> {
+        let dh = unsafe { ffi::DH_get_1024_160() };
+        if dh == ptr::null_mut() {
```
jethrogb: I think you can use `try_ssl_null!` here (and below)

jedisct1: Thanks! There are other places where this macro may be used. A dedicated PR, #261, has been opened.

jedisct1: The  purpose of `from_params` is to allow reimplementing `get_2048_256()` on older OpenSSL versions. For custom parameters, reading the parameters from a file would indeed be a good addition. I'll submit a new PR for this.

jethrogb:
> The purpose of from_params is to allow reimplementing get_2048_256() on older OpenSSL versions.

These could also be read in from their DER version.

openssl/src/dh/mod.rs
jedisct1: Done in #262

jethrogb:
I think this should be under `openssl::crypto`?

PR #261, opened 2015-09-01
Use try_ssl_null!() when relevant

jedisct1:


openssl/src/dh/mod.rs
```diff
@@ -21,27 +19,21 @@ impl DH {
     #[cfg(feature = "rfc5114")]
     pub fn get_1024_160() -> Result<DH, SslError> {
         let dh = unsafe { ffi::DH_get_1024_160() };
-        if dh == ptr::null_mut() {
-            return Err(SslError::get());
-        }
+        try_ssl_null!(dh);
         Ok(DH(dh))
```
jethrogb:
```rust
Ok(DH(try_ssl_null!(unsafe { ffi::DH_get_1024_160() })))
```

jedisct1: Why not, but this pattern is currently not being used anywhere.

jethrogb: I think the 2-line version is pretty common e.g. https://github.com/sfackler/rust-openssl/blob/769b831/openssl/src/x509/mod.rs#L451-L454

PR #263, opened 2015-09-01
Add SSL::set_ecdh_auto()

jedisct1:
This sets automatic curve selection and enables ECDH support.
Requires LibreSSL or OpenSSL >= 1.0.2, so behind a feature gate.

### DIFF
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -675,6 +675,9 @@ extern "C" {
     pub fn SSL_CTX_add_extra_chain_cert(ctx: *mut SSL_CTX, x509: *mut X509) -> c_long;
     #[link_name = "SSL_CTX_set_read_ahead_shim"]
     pub fn SSL_CTX_set_read_ahead(ctx: *mut SSL_CTX, m: c_long) -> c_long;
+    #[cfg(feature = "ecdh_auto")]
+    #[link_name = "SSL_CTX_set_ecdh_auto_shim"]
+    pub fn SSL_CTX_set_ecdh_auto(ssl: *mut SSL_CTX, onoff: c_int) -> c_int;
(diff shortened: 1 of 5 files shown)


openssl-sys/src/lib.rs
```diff
@@ -674,6 +674,9 @@ extern "C" {
     pub fn SSL_CTX_add_extra_chain_cert(ctx: *mut SSL_CTX, x509: *mut X509) -> c_long;
     #[link_name = "SSL_CTX_set_read_ahead_shim"]
     pub fn SSL_CTX_set_read_ahead(ctx: *mut SSL_CTX, m: c_long) -> c_long;
+    #[cfg(feature = "ecdh_auto")]
+    #[link_name = "SSL_CTX_set_ecdh_auto_shim"]
```
sfackler: A shim shouldn't be necessary here. You can just `#[cfg(...)]` the definition directly.

jedisct1: At least on LibreSSL, `SSL_CTX_set_ecdh_auto_shim()` is a macro. Isn't a shim required in that case?

sfackler: Ah, sorry, makes sense.

openssl/src/ssl/mod.rs
```diff
@@ -570,6 +570,14 @@ impl SslContext {
             })
     }
 
+    #[cfg(feature = "ecdh_auto")]
+    pub fn set_ecdh_auto(&mut self, onoff: bool) -> Result<(),SslError> {
```
sfackler: Can you add a quick bit of documentation to this, saying what it does and mentioning that it requires the `ecdh_auto` feature?

jedisct1:
A description has been added.

sfackler:
Thanks!

PR #265, opened 2015-09-01
Swap order of linking ssl/crypto

alexcrichton:
GNU linkers will sometimes aggressively try to strip objects and archives from a
linker command line in a left-to-right fashion. When a linker hits an object
file that doesn't satisfy any unresolved symbols, it will discard the object and
not re-visit it. This means that currently if symbols are depended upon in
libssl then some of the dependencies of libssl (in libcrypto) may have already
been stripped, causing a link error.

By swapping the order of what's linked it reflects the natural flow of
dependencies and the linker should figure everything out for us.

### DIFF
--- a/openssl-sys/build.rs
+++ b/openssl-sys/build.rs
@@ -36,12 +36,12 @@ fn main() {
         Some(ref v) => v.split(":").collect(),
         None => if target.contains("windows") {
             if get_mingw_in_path().is_some() && lib_dir.is_none() && include_dir.is_none() {
-                vec!("eay32", "ssleay32")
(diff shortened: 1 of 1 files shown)


openssl-sys/build.rs
```diff
@@ -41,7 +41,7 @@ fn main() {
                 vec!("eay32", "ssl32")
```
sfackler: I think these may need to be swapped as well? Same with line 39

sfackler:
Linkers are the worst.

alexcrichton:
Swapped the others as well

PR #266, opened 2015-09-01
ssl/npn+alpn: adjust protocol selection to fail if no protocols match

jmesmon:
ssl/npn+alpn: adjust protocol selection to fail if no protocols match
The current behavior causes a server written using rust-openssl to (if
it cannot negotiate a protocol) fallback to the first protocol it has
avaliable.

This makes it impossible to detect protocol mismatches.

This updates our selection to be more similar to how openssl's
s_server behaves: non-matching protocols are not supplied with a
fallback.

Note that some setups may actually want a fallback protocol supplied
via ALPN. To support those cases, we should consider adding a generic
callback that allows protocol selection to be entirely controlled by
the programmer.

For the purposes of having a sane default, however, not supplying a
default (and mimicing s_server's behavior) is the best choice.

### DIFF
--- a/openssl/src/ssl/tests.rs
+++ b/openssl/src/ssl/tests.rs
@@ -628,6 +628,49 @@ fn test_alpn_server_advertise_multiple() {
     assert_eq!(b"spdy/3.1", stream.get_selected_alpn_protocol().unwrap());
 }
 
+/// Test that Servers supporting ALPN don't report a protocol when none of their protocols match
+/// the client's reported protocol.
+#[test]
+#[cfg(feature = "alpn")]
+fn test_alpn_server_select_none() {
+    let localhost = "127.0.0.1:15422";
+    let listener = TcpListener::bind(localhost).unwrap();
+    // We create a different context instance for the server...
+    let listener_ctx = {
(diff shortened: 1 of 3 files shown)


sfackler:
This seems reasonable. Would it be possible to add a test to make sure it behaves properly?

jmesmon:
Updated to add a test.

openssl/src/ssl/tests.rs
```diff
@@ -628,6 +628,48 @@ fn test_alpn_server_advertise_multiple() {
...
+        Ok(_) => {}
+        Err(err) => panic!("Unexpected error {:?}", err)
+    }
+    // Now connect to the socket and make sure the protocol negotiation works...
+    let stream = TcpStream::connect(localhost).unwrap();
+    let stream = match SslStream::new(&ctx, stream) {
+        Ok(stream) => stream,
+        Err(err) => panic!("Expected success, got {:?}", err)
+    };
+    // SPDY is selected since that's the only thing the client supports.
```
sfackler: Comment's not correct here, right?

jmesmon: whoops. commit updated to fix.

PR #267, opened 2015-09-03
Support for introspection of session cipher

gcmurphy:
Adds support for the following OpenSSL native functions:

const char *SSL_CIPHER_get_name(const SSL_CIPHER *cipher);
int SSL_CIPHER_get_bits(const SSL_CIPHER *cipher, int *alg_bits);
char *SSL_CIPHER_get_version(const SSL_CIPHER *cipher);
char *SSL_CIPHER_description(const SSL_CIPHER *cipher, char *buf, int size);

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -668,6 +668,39 @@ impl<'ssl> DerefMut for MemBioRef<'ssl> {
     }
 }
 
+pub struct SslCipher {
+    cipher: *const ffi::SSL_CIPHER
+}
+
+impl SslCipher {
+    pub fn secret_bits(&self) -> u64 {
+        let bits = unsafe { ffi::SSL_CIPHER_get_bits(self.cipher, ptr::null_mut()) };
+        return bits as u64
+    }
+
+    pub fn version(&self) -> String {
+        unsafe {
+            let ver_ptr = ffi::SSL_CIPHER_get_version(self.cipher);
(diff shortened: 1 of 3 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -668,6 +668,54 @@ impl<'ssl> DerefMut for MemBioRef<'ssl> {
     }
 }
 
+pub struct SslCipher {
+    cipher: *const ffi::SSL_CIPHER
+}
+
+impl SslCipher {
+    pub fn secret_bits(&self) -> Option<u64> {
```
sfackler: Why do these methods return options if they are always `Some`?

openssl/src/ssl/mod.rs
```diff
@@ -668,6 +668,54 @@ impl<'ssl> DerefMut for MemBioRef<'ssl> {
...
+
+    pub fn description(&self) -> Option<String> {
+        unsafe {
+            let desc_ptr = ffi::SSL_CIPHER_description(self.cipher, ptr::null_mut(), 0);
+            let desc = String::from_utf8(CStr::from_ptr(desc_ptr).to_bytes().to_vec()).unwrap();
+            Some(desc)
+        }
+    }
+
+    pub unsafe fn from_raw(raw_cipher: *const ffi::SSL_CIPHER) -> Option<SslCipher> {
```
sfackler: I would rather inline this logic in Ssl::get_current_cipher.

PR #270, opened 2015-09-10
Check if public/private RSA key is properly loaded

### LINKED ISSUE #43: Segmentation fault possible on PKey methods
When calling `PKey::encrypt(..)` without a public key set, a Segmentation fault occurs. This also happens when first calling `load_pub` with an invalid key and then calling encrypt.

Unfortunately there is no way to detect if the set keys are valid or not before calling the de/encrypt methods. The `load_` methods should probably return a `Result`

mvdnes:
This fixes #43.

The panics do not break any API since they now just prevent a segmentation fault.

### DIFF
--- a/openssl/src/crypto/pkey.rs
+++ b/openssl/src/crypto/pkey.rs
@@ -110,11 +110,16 @@ impl PKey {
         }
     }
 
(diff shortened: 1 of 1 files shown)


sfackler:
Thanks! This API does need to be overhauled.

Could you add a quick test that a situation like #43 panics instead of segfaulting?

mvdnes:
Done! Even found a bug in the previous fix, so that was definitely a good idea.

PR #272, opened 2015-09-22
Nonblocking Socket Support

jamwt:
Support for nonblocking sockets + tests.

I don't have a windows box to test on, but I'll try to track one down.

cc @alexcrichton

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -22,7 +22,7 @@ use std::slice;
 use bio::{MemBio};
 use ffi;
 use dh::DH;
-use ssl::error::{SslError, SslSessionClosed, StreamError, OpenSslErrors};
+use ssl::error::{NonblockingSslError, SslError, SslSessionClosed, StreamError, OpenSslErrors};
 use x509::{X509StoreContext, X509FileType, X509};
 use crypto::pkey::PKey;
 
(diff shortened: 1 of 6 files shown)


alexcrichton:
Ok, I juggled a few things around after #273 and ended up with [this commit](https://github.com/alexcrichton/rust-openssl/commit/1b800f7b00b8c763f0a95678bb1447812c9352d2) to enable these tests on all platforms. I ended up using `select` which is available across Linux/OSX/Windows as the performance of epoll probably isn't super critical here!

jamwt:
@alexcrichton I actually went looking for a good select somewhere, including in `net2`, and I didn't see it; I take it the path dependency in your modified `Cargo.toml` file means you just added it?

jamwt:
(Implied: agreed epoll is totally unnecessary here.)

jamwt:
Also, might make sense to PR your change after #272 and #273 merge in? Rather than try to combine it somehow.

alexcrichton:
Ah yeah I ended up just writing the `select` bindings by hand, and I'd be totally down for sending a PR after this merges! Whatever's easiest for @sfackler :)

jamwt:
Sgtm

jamwt:
Ping?  Any feedback, @sfackler ?

sfackler:
Sorry, haven't had a chance to look at this :/

It seems somewhat unnecessary to make a full separate copy of `SslStream` that's (I believe?) identical other than the `read` and `write` methods. Maybe it would make more sense to add `read_nonblock` and `write_nonblock` (or whatever names) to `SslStream` that would do what `NonblockingSslStream`'s `read` and `write` do?

jamwt:
Hey, thanks for the feedback.

I actually started out with this approach.  A few reasons why I changed it:

1. The `connect()` and `accept()` methods are different too.  They need to `set_nbio()` on the `BIO`.  They also need to tolerate `WantRead` and `WantWrite` as non-fatal returns from the underlying openssl setup functions.

2. Error wrapping is different, since we have a new error type that includes `WantRead` and `WantWrite` in the ADT.

3. A lot of the "nice" high-level methods on the are on `SslStream`, which wraps either a direct or indirect stream.  Indirect streams don't make any sense (and are actually kind of dangerous, since the retry loop will spin the I/O thread) for nonblocking sockets.

So, once you get down to having your own connect, accept, read, write, make_error, you basically have an entirely different stream type.  And it doesn't really make sense to wrap this stream in the direct/indirect enum.

Ergo, this type becomes a kind of flattened nonblocking stream that's a sibling (conceptually) to `SslStream`, not a lower-level thing like `DirectStream` or `IndirectStream`.

openssl/src/ssl/mod.rs
```diff
@@ -1442,3 +1442,284 @@ impl<S> MaybeSslStream<S> where S: Read+Write {
...
+    pub fn connect<T: IntoSsl>(ssl: T, stream: S) -> Result<NonblockingSslStream<S>, SslError> {
+        let ssl = try!(ssl.into_ssl());
+        let fd = stream.as_raw_socket() as c_int;
+        let ssl = try!(NonblockingSslStream::new_base(ssl, stream, fd));
+        let ret = ssl.ssl.connect();
+        if ret > 0 {
+            Ok(ssl)
+        } else {
+            // WantRead/WantWrite is okay here; we'll finish the handshake in
+            // subsequent send/recv calls.
```
sfackler: This seems somewhat suspicious to me (I'd expect to have to call `connect` again), but if it works, it works I guess. Are you reasonably confident that this case is hit in tests?

jamwt: Yes, in fact, you'll succeed at connect right away.  The negotiation can always be finished in subsequent read/write calls, because to be honest, renegotiation can happen at any time.  So this is already true.

sfackler:
Fair enough. The need to mess with the BIOs in particular seems like it would be hard to work around. 

I'm not totally sold on it being so dangerous to have a nonblocking indirect stream support missing. It seems not totally unreasonable that someone would end up in that situation but be able to appropriately manage the nonblocking part, assuming that the MemBIOs would do the "right thing".

I have been wanting to remove all of the methods on `SslStream` that just forward to `SslContext` in favor of directly doing `my_stream.ctx().whatever()`. Could you leave those off? I'll deprecate the ones on `SslStream` later, but that should cut down on the amount of duplication between the two types.

jamwt:
Okay, so expose a way to get a ref to the underlying `Ssl`, and avoid explicit republishing of some of its info?  Remove all those republishing methods for now?

sfackler:
Yep!

jamwt:
K, will do.

jamwt:
Okay, rebased (including updating the tests to use the new slick `Server` struct) and removed all the methods that call through to the underlying ssl.  Exposed the ssl via a `ssl()` method, so the caller can do whatever they'd like with it.

sfackler:
Thanks!

jamwt:
Thank you!

Ping @alexcrichton .. when you have some time, probably a good idea to submit your PR now with that movement of the tests to `select()` so that we can run on win + osx.

alexcrichton:
Submitted as #294

PR #273, opened 2015-09-22
Enable testing on Windows via AppVeyor

alexcrichton:
This abolishes the test.sh script which spawns a bunch of `openssl` instances to
instead run/manage the binary in-process (providing more isolation to boot). The
tests have been updated accordingly and the `connected_socket` dependency was
also dropped in favor of `net2` as it the former doesn't work on Windows.


alexcrichton:
Wanted to test out https://github.com/sfackler/rust-openssl/pull/272 but required getting Windows testing first! Should probably hold off on merging until the CI actually passes...

openssl/src/ssl/tests.rs
```diff
@@ -344,19 +465,20 @@ run_test!(get_peer_certificate, |method, stream| {
 #[test]
 #[cfg(feature = "dtlsv1")]
 fn test_write_dtlsv1() {
-    let sock = UdpSocket::bind("127.0.0.1:0").unwrap();
-    let stream = sock.connect("127.0.0.1:15410").unwrap();
+    let yes = iter::repeat("y\n").take(1_000_000).collect::<String>();
```
alexcrichton: This is probably the sketchiest part of the tests, it previously was a pipe of `yes` into an `openssl s_server`, but now it's just a whole lot of `y` lines. Looks like it's passing the bots now though but I could switch to a true infinite loop if needed.

sfackler: I'm not sure why `yes` was piped in the first place, since this test doesn't actually read anything. Do things still work if you cut that part out entirely?

alexcrichton: Yeah unfortunately removing it causes the test to fail.

sfackler: I think I'd prefer a true infinite loop. I'm guessing that it's hanging up when it runs out of stuff to send, so the current setup seems like it would be a bit fragile.

alexcrichton:
Man who'd think it'd be more robust to do things correctly? Updated!

sfackler:
Yay!

PR #275, opened 2015-09-23
Require a feature to accept SSLv3 only

jedisct1:
TLSv1 was released in 16 years ago, and LibreSSL permanently removed the SSLv3 code.


sfackler:
Covered by #280. Thanks!

PR #279, opened 2015-10-02
Update documentation about SSLv23

jtdowney:
In OpenSSL world, the SSLv23 option is a poorly name method [that will negotiate](http://stackoverflow.com/questions/23709664/openssl-let-the-server-and-client-negotiate-the-method) what version of TLS or SSL to use. It starts with the best version the library supports (which is TLS 1.2 on OpenSSL >= 1.0.1) and then precedes to keep trying all the way down to SSL 2.0.

The docs here were incorrect because they said it only supported SSLv2, SSLv3, and TLSv1.

It is generally recommended that you set additional options to explicitly disable SSL 2.0 and 3.0, but that seems outside the scope of a thin wrapper library.


sfackler:
Thanks!

PR #280, opened 2015-10-03
Fix build on LibreSSL.

ltratt:
LibreSSL has deprecated SSLv3_method, so this commit makes that a compile-time
feature.

It also removes a test referencing SSL_OP_CISCO_ANYCONNECT, as the LibreSSL
header says it is amongst "Obsolete flags kept for compatibility. No sane code
should use them."


sfackler:
Can you add sslv3 to the default features to maintain backwards compatibility?

ltratt:
Yes, although I wonder if we want to do that? My reading of http://www.libressl.org/patches.html is that v3 isn't something we want to encourage. That said, I don't feel strongly about this, as I have no idea how many people might be using the v3 support.

If we do want to turn v3 on by default, should we do that on a per-platform basis? Otherwise, libressl platforms will still fail to build.

sfackler:
Oh yeah, we definitely don't want to encourage use of v3, but at the same time, I don't want to cause downstream breakage due to the Sslv3 variant disappearing suddenly. I'm going to bump the major version soonish anyways to pull in some other breaking changes and I'll pull this feature flag out of the default features then.

ltratt:
I've pushed a commit which enables v3 by default.

However, I'm at a loss as to how to switch this feature off on some platforms. I want to do something like:

```
[target.x86_64-unknown-openbsd.features]
sslv3 = []
```

but I'm not sure if Cargo supports anything along these lines?

sfackler:
I don't believe it does. Maybe we should just roll out the next major release now with this change to avoid having to deal with annoying default feature stuff.

ltratt:
Ah :/ In which case, I won't object to a major release! If so, shall I do a force push that removes the last commit to this PR? Then you can merge the breaking commit, and prepare for a major release?

sfackler:
Works for me. I'll figure out the release sometime tonight or tomorrow and merge this.

ltratt:
OK, I've removed the "enable v3 by default" commit, so I think this is ready for merging. Thanks!

PR #282, opened 2015-10-05
Set threadid_func on linux/osx (fixes #281)

### LINKED ISSUE #281: Implement threadid_func
According to [this page](https://www.openssl.org/docs/manmaster/crypto/threads.html), `locking_function` and `threadid_func` must be implemented to use multithreaded code safely.

> threadid_func(CRYPTO_THREADID *id) is needed to record the currently-executing thread's identifier into id. The implementation of this callback should not fill in id directly, but should use CRYPTO_THREADID_set_numeric() if thread IDs are numeric, or CRYPTO_THREADID_set_pointer() if they are pointer-based. If the application does not register such a callback using CRYPTO_THREADID_set_callback(), then a default implementation is used - on Windows and BeOS this uses the system's default thread identifying APIs, and on all other platforms it uses the address of errno. The latter is satisfactory for thread-safety if and only if the platform has a thread-local error number facility.

"the address of errno" is afaict an incorrect way of identifying a thread (lldb gives me the same address in both threads -- I don't see _why_ this happens given that errno is in TLS, but ... *shrug*). 

This can cause very scary malloc bugs in openssl.

Code like the following in init would be nice (similar to https://github.com/sfackler/rust-openssl/blob/d9c22137114b61a10105c7c2065e423e33bb456f/openssl-sys/src/lib.rs#L241):

```
extern "C" void my_callback(CRYPTO_THREADID *id) {
   CRYPTO_THREADID_set_numeric(id, (unsigned long)gettid())
}

void init() {
  .....
  CRYPTO_THREADID_set_callback(my_callback)

}
```

I'm not sure if we should be using `gettid()` or the pthread method (the pthread one works on OS X too, which we need since the error is happening on an OS X machine).

cc @larsbergstrom

Manishearth:


larsbergstrom:
fyi, author info is missing "administrator <administrator@administrators-Mac-mini.local>"

Manishearth:
Fixed.

larsbergstrom:
Tested on OSX 10.11.
```
test result: ok. 75 passed; 0 failed; 0 ignored; 0 measured

   Doc-tests openssl

running 12 tests
test crypto::hash::Hasher_0 ... ok
test x509::add_extension_0 ... ok
test crypto::hmac::HMAC_0 ... FAILED
test crypto::hash::Hasher_1 ... ok
test crypto::hmac::HMAC_1 ... FAILED
test x509::add_name_0 ... ok
test x509::extension::Extension::OtherStr_0 ... ok
test x509::extension::Extension::OtherNid_0 ... ok
test x509::add_extensions_0 ... ok
test x509::extension::AltNameOption::Other_0 ... ok
test x509::add_names_0 ... ok
test x509::X509Generator_0 ... ok

failures:

---- crypto::hmac::HMAC_0 stdout ----
	thread 'crypto::hmac::HMAC_0' panicked at 'test executable failed:

thread '<main>' panicked at 'assertion failed: `(left == right)` (left: `-327295773`, right: `1`)', src/crypto/hmac.rs:94
', src/librustdoc/test.rs:259

---- crypto::hmac::HMAC_1 stdout ----
	<anon>:10:5: 10:26 warning: unused result which must be used, #[warn(unused_must_use)] on by default
<anon>:10     h.write_all(data[0]);
              ^~~~~~~~~~~~~~~~~~~~~
<anon>:11:5: 11:26 warning: unused result which must be used, #[warn(unused_must_use)] on by default
<anon>:11     h.write_all(data[1]);
              ^~~~~~~~~~~~~~~~~~~~~
thread 'crypto::hmac::HMAC_1' panicked at 'test executable failed:

thread '<main>' panicked at 'assertion failed: `(left == right)` (left: `-1277369872`, right: `1`)', src/crypto/hmac.rs:94
', src/librustdoc/test.rs:259


failures:
    crypto::hmac::HMAC_0
    crypto::hmac::HMAC_1

test result: FAILED. 10 passed; 2 failed; 0 ignored; 0 measured


[larsberg@lbergstrom openssl]$ 
```

Will check if those pass on master.

larsbergstrom:
Identical test results on master.

Manishearth:
I was getting those failures on the mac builder as well, they happen on master too.

openssl-sys/src/openssl_shim.c
```diff
@@ -3,6 +3,33 @@
...
+void rust_openssl_set_id_callback() {
+  CRYPTO_set_id_callback((unsigned long (*)())thread_id);
+}
+
+#else
+// Openssl already handles Windows directly, so we don't
+// need to explicitly set it
+
+void rust_openssl_set_id_callback() {
+  // We don't know how to set the callback for arbitrary OSes
```
gkoz: Why would the same not work for other POSIX-compatible OSs?

Manishearth: We could. However, `pthread_self()` returns a `pthread_t`, which is opaque (according to the posix spec). On Linux and OSX it's okay to cast it to ulong the way we do, but for a random POSIX OS there's no guarantee.

I don't mind converting this to an `#ifdef __posix`, though. It's *probably* correct for common posix OSes.

openssl-sys/src/openssl_shim.c
```diff
@@ -3,6 +3,33 @@
...
+#include<openssl/crypto.h>
+
+unsigned long thread_id()
+{
+    unsigned long ret = (unsigned long)pthread_self();
+    return ret;
+}
+
+void rust_openssl_set_id_callback() {
+  CRYPTO_set_id_callback((unsigned long (*)())thread_id);
```
sfackler: The explicit cast here isn't necessary, right?

Manishearth: This is what the openssl testing code did, and I'm scared of function pointers, so I did it too :smile:

sfackler:
@larsbergstrom It sounds like you're using modern OpenSSL headers with old OpenSSL binaries. In OpenSSL 1.0, Hmac_init_ex returns an int while in OpenSSL 0.98 it returns void. This crate shims around that by pulling the version out of OpenSSL headers but it looks like in your case the wrong thing happend.

Manishearth:
Could we get a Crates version bump?

sfackler:
Yep, I'll push one when I get home - next release after this one will be a major version, btw.

sfackler:
Published

PR #283, opened 2015-10-07
Add an OPENSSL_EXTRA_DYLIBS var to build.rs

alexcrichton:
When linking OpenSSL statically there are often a few libraries which need to be
pulled in which OpenSSL itself depends on. For example on Linux it normally
depends on `libdl` and on Windows it depends on `libgdi32` and/or `libuser32`.
Typically these sort of dependency relations are expressed via a
`[dependencies]` annotation + an `extern crate` directive, but as these aren't
always required it may not always be the best solution in this case.

For now this adds an extra environment variable which adds a location to throw
in these extra dylibs to the build when linking OpenSSL statically.

### DIFF
--- a/openssl-sys/build.rs
+++ b/openssl-sys/build.rs
@@ -59,6 +59,12 @@ fn main() {
         println!("cargo:rustc-link-lib={}={}", mode, lib);
     }
 
+    if let Ok(s) = env::var("OPENSSL_EXTRA_DYLIBS") {
+        for lib in s.split(":") {
+            println!("cargo:rustc-link-lib={}", lib);
+        }
+    }
+
     let mut include_dirs = vec![];
 
     if let Some(include_dir) = include_dir {


sfackler:
This seems unfortunate. Is there not a reasonable first-class way to have these dependencies properly tracked automatically?

Diggsey:
@sfackler That would be ideal. However, I think even if that is implemented it's still worth having this as a fall-back: without this override it's actually *very difficult* (read impossible) to get additional dylibs to be correctly linked.

Passing them as extra linker arguments doesn't work because they end up too early: `ld` looks at them and ignores the libraries because no symbols from them are imported yet, as the openssl native library is later on in the argument list.

Currently the only workaround I've found is using `cargo rustc` to get direct control over the linker args, and this means I can't run tests using `cargo test`. (This happens because I am statically linking to openssl on windows).

alexcrichton:
The "first class" thing to do here would probably be a build script override instead of communication via environment variables. The problem with overriding this build script, however, would be that the shim is not compiled.

So I suppose the ideal solution might look something like:

* Move the shim from `openssl-sys` to `openssl`
* Move FFI definitions defined by the shim to `openssl` as well
* Have the `openssl` build script read some variables (e.g. `include` and whatnot) from the sys crate to compile the shim.

That should allow overriding `openssl-sys` entirely which allows defining an entire custom set of libraries or whatever is necessary.

alexcrichton:
A downside of what I'm saying though is that it's a breaking change to the sys crate, but an upside would be that "pulling in an OpenSSL dep" would "just work" on OSX 10.11 because the sys crate wouldn't need to compile a shim (which isn't needed unless you're actually linking the `openssl` crate).

sfackler:
Appears to be no longer relevant.

PR #284, opened 2015-10-09
AES CFB-mode feature

### LINKED ISSUE #190: Support for AES 128 CFB8
It's already on OpenSSL (source: https://github.com/luqmana/mcchat/blob/master/crypto.rs) but not supported here for unknown reasons.

Examples of applications using this right now: Minecraft.

It would help us on https://github.com/PistonDevelopers/hematite_server/issues/30 development, currently I've ported Go's AES impl to just get this working.

bheart:
Straight forward optional AES CFB support, closes #190.

### DIFF
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -417,6 +417,9 @@ extern "C" {
     #[cfg(feature = "aes_ctr")]
     pub fn EVP_aes_128_ctr() -> *const EVP_CIPHER;
     // fn EVP_aes_128_gcm() -> EVP_CIPHER;
+    pub fn EVP_aes_128_cfb1() -> *const EVP_CIPHER;
+    pub fn EVP_aes_128_cfb128() -> *const EVP_CIPHER;
+    pub fn EVP_aes_128_cfb8() -> *const EVP_CIPHER;
     pub fn EVP_aes_256_cbc() -> *const EVP_CIPHER;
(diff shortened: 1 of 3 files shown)


openssl-sys/src/lib.rs
```diff
@@ -417,6 +417,8 @@ extern "C" {
     #[cfg(feature = "aes_ctr")]
     pub fn EVP_aes_128_ctr() -> *const EVP_CIPHER;
     // fn EVP_aes_128_gcm() -> EVP_CIPHER;
+    #[cfg(feature = "aes_cfb")]
+    pub fn EVP_aes_128_cfb() -> *const EVP_CIPHER;
```
sfackler: If I'm reading the OpenSSL changelogs right, I don't think this needs to be behind a feature since it was added way back in version 0.9.7.

bheart: I figured as much

bheart: I guess it's missing from the Windows distribution after all?

sfackler:
The linker seems to be complaining on the Windows side: https://ci.appveyor.com/project/sfackler/rust-openssl/build/1.0.196/job/vvv10ism91660t7g

bheart:
Turns out EVP_aes_128_cfb wasn't in the Windows distribution and that it's really just an alias of EVP_aes_128_cfb128 (oversight on my part, whoops). So I changed that and slapped the other CFB modes on.

sfackler:
Thanks!

PR #285, opened 2015-10-11
Update README with OSX 10.11 instructions

winding-lines:
These instructions are from one of the open issues on the project. They are working for me, I am not sure if there are some possible side effects from the brew link --force.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,10 +29,12 @@ rust-openssl to a separate installation. OSX releases starting at 10.11, "El
 Capitan", no longer include OpenSSL headers which will prevent the `openssl`
 crate from compiling.
 
-For OSX 10.11 you can use brew to install OpenSSL.
+For OSX 10.11 you can use brew to install OpenSSL and then set the environment variables
+as described below.
 ```bash
 brew install openssl
-brew link openssl --force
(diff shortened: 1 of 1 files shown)


sfackler:
Force linking OpenSSL is a somewhat dangerous thing to do and I am pretty hesitant to recommend that course of action. It seems like a safer bet to use the environment variables already described to tell the build script specifically where to find the brew installed OpenSSL.

winding-lines:
I've changed the README as per your feedback. 

One issue is that all the consumers of the openssl crate may have to setup the environment variables. Is this correct? If so could we integrate this in build.rs to reduce the development friction?

sfackler:
Integrate it in what way? Adding `/usr/local/opt/openssl` to the search paths?

winding-lines:
Yes, is that possible/desirable?

Marius
---

sfackler:
Looks good to me with a rebase.

erickt:
@winding-lines: got time to do this rebase?

winding-lines:
I think I rebased :)

gabeio:
:+1: just used that method to get this working thanks :smile:

nox:
Couldn't build.rs set these environment variables itself?

sfackler:
@nox the build.rs is what consumes them.

PR #286, opened 2015-10-12
Add set_certificate_chain_file()

jedisct1:
SSL_CTX_use_certificate_chain_file() is preferred over SSL_CTX_use_certificate_file().

It allows the use of complete certificate chains instead of loading only the first certificate in a PEM file.


sfackler:
Thanks!

PR #287, opened 2015-10-12
Add certs.pem in cert probe list

radare:
It turns out that some distributions use /etc/ssl/certs.pem, which was causing some troubles.
Related issue https://github.com/rust-lang/cargo/issues/1978#issuecomment-147515236


alexcrichton:
@sfackler would you be ok releasing a new version of openssl-sys with this?

PR #294, opened 2015-10-22
Get nonblocking tests working on OSX/Windows

alexcrichton:
Follow-up after #272


sfackler:
The 32 bit Windows build failed, but it looks unrelated?

alexcrichton:
Ah yeah that should be fixed in the next nightly I believe

PR #296, opened 2015-10-23
expose buffer for crypter

cheme:
Hello,
I added function for crypter to let user use it with a single writeable buffer (usefull in my case because I write my data in a Write stream and do note require a vec<u8> for every update).
(I also removed the extra block size in update function which does not seem usefull)


sfackler:
The symm implementation has been rewritten to allow for buffer reuse.

PR #309, opened 2015-11-24
Add support for Server Name indication (SNI) on the server's side

Geal:
This is implemented through a callback, like the certificate verification

### DIFF
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -534,6 +534,7 @@ extern "C" {
     pub fn SSL_write(ssl: *mut SSL, buf: *const c_void, num: c_int) -> c_int;
     pub fn SSL_get_ex_data_X509_STORE_CTX_idx() -> c_int;
     pub fn SSL_get_SSL_CTX(ssl: *mut SSL) -> *mut SSL_CTX;
+    pub fn SSL_set_SSL_CTX(ssl: *mut SSL, ctx: *mut SSL_CTX) -> *mut SSL_CTX;
     pub fn SSL_get_current_compression(ssl: *mut SSL) -> *const COMP_METHOD;
     pub fn SSL_get_peer_certificate(ssl: *mut SSL) -> *mut X509;
     pub fn SSL_get_ssl_method(ssl: *mut SSL) -> *const SSL_METHOD;
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -135,6 +135,8 @@ pub const PKCS5_SALT_LEN: c_int = 8;
 pub const SSL_CTRL_OPTIONS: c_int = 32;
 pub const SSL_CTRL_CLEAR_OPTIONS: c_int = 77;
 
+pub const SSL_CTRL_SET_TLSEXT_SERVERNAME_CB:  c_int = 53;
+pub const SSL_CTRL_SET_TLSEXT_SERVERNAME_ARG: c_int = 54;
 pub const SSL_CTRL_SET_TLSEXT_HOSTNAME: c_int = 55;
 pub const SSL_CTRL_EXTRA_CHAIN_CERT: c_int = 14;
 
@@ -538,6 +540,8 @@ extern "C" {
     pub fn SSL_state_string(ssl: *mut SSL) -> *const c_char;
     pub fn SSL_state_string_long(ssl: *mut SSL) -> *const c_char;
 
+    pub fn SSL_get_servername(ssl: *const SSL, name_type: c_long) -> *const c_char;
+
     pub fn SSL_COMP_get_name(comp: *const COMP_METHOD) -> *const c_char;
 
     pub fn SSL_CTX_new(method: *const SSL_METHOD) -> *mut SSL_CTX;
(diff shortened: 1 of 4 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -297,6 +303,54 @@ extern fn raw_verify_with_data<T>(preverify_ok: c_int,
...
+        let callback: Option<ServerNameCallback> = mem::transmute(callback);
+        let mut s = Ssl { ssl: ssl };
+
+        let res = match callback {
+            None => ffi::SSL_TLSEXT_ERR_ALERT_FATAL,
+            Some(callback) => callback(&mut s, ad)
+        };
+
+        // Allows dropping the Ssl instance without calling SSL_FREE on the SSL object
+        s.ssl = ptr::null_mut() as *mut ffi::SSL;
```
sfackler: `mem::forget(s)` seems a bit cleaner of an option here.

openssl/src/ssl/mod.rs
```diff
@@ -297,6 +303,54 @@ extern fn raw_verify_with_data<T>(preverify_ok: c_int,
...
+
+        let data: Box<T> = mem::transmute(arg);
+
+        let res = match callback {
+            None => ffi::SSL_TLSEXT_ERR_ALERT_FATAL,
+            Some(callback) => callback(&mut s, ad, &*data)
+        };
+
+        // Allows dropping the Ssl instance without calling SSL_FREE on the SSL object
+        s.ssl = ptr::null_mut() as *mut ffi::SSL;
```
sfackler: See above

openssl/src/ssl/mod.rs
```diff
@@ -297,6 +303,54 @@ extern fn raw_verify_with_data<T>(preverify_ok: c_int,
...
+extern fn raw_sni_with_data<T>(ssl: *mut ffi::SSL, ad: &mut c_int, arg: *mut c_void) -> c_int
+                                  where T: Any + 'static {
+    unsafe {
+        let ssl_ctx = ffi::SSL_get_SSL_CTX(ssl);
+
+        let callback = ffi::SSL_CTX_get_ex_data(ssl_ctx, SNI_IDX);
+        let callback: Option<ServerNameCallbackData<T>> = mem::transmute(callback);
+        let mut s = Ssl { ssl: ssl };
+
+        let data: Box<T> = mem::transmute(arg);
```
sfackler: You can transmute to a `&T` here to avoid the mem::forget below.

openssl/src/ssl/mod.rs
```diff
@@ -485,6 +544,36 @@ impl SslContext {
...
 
+    /// Configures the server name indication (SNI) callback for new connections
+    ///
+    /// obtain the server name with `get_servername` then set the corresponding context
+    /// with `set_ssl_context`
+    pub fn set_servername_callback(&mut self, callback: Option<ServerNameCallback>) {
+        unsafe {
+            ffi::SSL_CTX_set_ex_data(self.ctx, SNI_IDX,
+                                     mem::transmute(callback));
+            //let f: extern fn(c_int, *mut ffi::X509_STORE_CTX) -> c_int = raw_sni;
```
sfackler: Dead code?

openssl/src/ssl/mod.rs
```diff
@@ -886,6 +976,29 @@ impl Ssl {
...
+
+    /// change the context corresponding to the current connection
+    pub fn set_ssl_context(&self, ctx: &SslContext) -> SslContext {
+        SslContext { ctx: unsafe { ffi::SSL_set_SSL_CTX(self.ssl, ctx.ctx) } }
+    }
+
+    /// obtain the context corresponding to the current connection
+    pub fn get_ssl_context(&self) -> SslContext {
+        let ssl_ctx = unsafe { ffi::SSL_get_SSL_CTX(self.ssl) };
+        SslContext { ctx: ssl_ctx }
```
sfackler: What happens when this is SSL_free'd?

openssl-sys/src/lib.rs
```diff
@@ -567,6 +572,8 @@ extern "C" {
     pub fn SSL_CTX_set_cipher_list(ssl: *mut SSL_CTX, s: *const c_char) -> c_int;
 
     pub fn SSL_CTX_ctrl(ssl: *mut SSL_CTX, cmd: c_int, larg: c_long, parg: *mut c_void) -> c_long;
+    pub fn SSL_CTX_callback_ctrl(ssl: *mut SSL_CTX, cmd: c_int, callback: Option<extern fn()>) -> c_long;
```
sfackler: According to the man page this is something that should only be invoked through macros? https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_ctrl.html

Those get shimmed out in openssl-sys-extras.

openssl/src/ssl/mod.rs
Geal: Following our conversation on IRC, I was wary of potential memory leaks using `mem::forget` here, but I just ran a test for this, and I see no evidence of a leak, so let's go with that solution.

openssl/src/ssl/mod.rs
Geal: I see, nice!

openssl/src/ssl/mod.rs
Geal: Fixed

openssl/src/ssl/mod.rs
Geal: Unfortunately, same issue as when `Ssl` is dropped. The clean solution would be to increase the refcount of SSL_CTX, but since it is an opaque structure, it is not possible from Rust. Maybe in a shim?

openssl-sys/src/lib.rs
Geal: It's done, thanks for the tip

Geal:
Hi! Any news on this? Could you launch again the Appveyor pass? It seems the fail is not related to the code.

sfackler:
The Appveyor failure isn't a big deal - the server hosting the Windows installers seems to be pretty unreliable.

The issue with `get_ssl_context` needs to be worked out before it can merge, since it's a massive memory safety issue otherwise.

Geal:
Would a small part in C to increase the refcount before returning a `SslContext` be alright?

sfackler:
Yeah, a bit of C is totally fine. You can look at openssl-sys-extras to see how to get the header include paths.

Geal:
Alright, this will fix the refcount issue

Geal:
Thanks!

PR #317, opened 2015-12-06
Use OPENSSL_INCLUDE_DIR as fallback in build script

### LINKED ISSUE #316: 'fatal error: 'openssl/hmac.h' file not found' on OSX 10.11 when building `openssl-sys-extras`
When trying to build `openssl-sys-extras`, the build seems to fail with `fatal error: 'openssl/hmac.h' file not found` on OSX 10.11, even if `OPENSSL_INCLUDE_DIR` is set to `/usr/local/opt/openssl/include`.

## The Fix

Looking at the build script of `openssl-sys-extras`, you will see that it uses `DEP_OPENSSL_INCLUDE` to determine additional include directories. Thus running cargo with something like this ...

```bash
OPENSSL_INCLUDE_DIR=/usr/local/opt/openssl/include DEP_OPENSSL_INCLUDE=/usr/local/opt/openssl/include cargo build
```

... will do the trick.

## Proposition

To prevent others from running into the issue as well, one could use `OPENSSL_INCLUDE_DIR` as secondary option in the build script.

Byron:
fixes #316

### DIFF
--- a/openssl-sys-extras/build.rs
+++ b/openssl-sys-extras/build.rs
@@ -10,9 +10,12 @@ fn main() {
     let options_shim_file = generate_options_shim();
     let mut config = gcc::Config::new();
 
-    if let Some(paths) = env::var_os("DEP_OPENSSL_INCLUDE") {
-        for path in env::split_paths(&paths) {
-            config.include(PathBuf::from(path));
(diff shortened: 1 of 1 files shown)


Byron:
Please feel free to close this one - additional notes are in the linked issue.

Issue #314, opened 2015-11-30
Unable to build openssl 0.7.x for ARM


operutka:
I've been trying to build the new 0.7.x release for ARM using rust 1.4.0 cross-compiler; unfortunately I always get the following errors:

```
...
/home/xxx/.cargo/registry/src/github.com-0a35038f75765ae4/openssl-0.7.0/src/bn/mod.rs:92:50: 92:64 error: mismatched types:
 expected `*const u8`,
    found `*const i8`
(expected u8,
    found i8) [E0308]
/home/xxx/.cargo/registry/src/github.com-0a35038f75765ae4/openssl-0.7.0/src/bn/mod.rs:92             try_ssl!(ffi::BN_dec2bn(v.raw_ptr(), c_str.as_ptr()));
                                                                                                                                              ^~~~~~~~~~~~~~
/home/xxx/.cargo/registry/src/github.com-0a35038f75765ae4/openssl-0.7.0/src/macros.rs:24:19: 24:39 note: in this expansion of try_ssl_if! (defined in /home/xxx/.cargo/registry/src/github.com-0a35038f75765ae4/openssl-0.7.0/src/macros.rs)
/home/xxx/.cargo/registry/src/github.com-0a35038f75765ae4/openssl-0.7.0/src/bn/mod.rs:92:13: 92:67 note: in this expansion of try_ssl! (defined in /home/xxx/.cargo/registry/src/github.com-0a35038f75765ae4/openssl-0.7.0/src/macros.rs)
/home/xxx/.cargo/registry/src/github.com-0a35038f75765ae4/openssl-0.7.0/src/bn/mod.rs:92:50: 92:64 help: run `rustc --explain E0308` to see a detailed explanation
...
```

Note that this is only an example. There is a lot of errors like this but it's always type mismatch between `*const u8` and `*const i8`. Building it directly for x86/x86_64 works fine as well as building the 0.6.x release for ARM, x86 and x86_64. I use the same version of Rust on all mentioned architectures (it is 1.4.0 8ab8581f6 2015-10-27).

My guess is that CStr and CString use different mapping for c_char on ARM but I don't know why there is no problem with the 0.6.x release.

sfackler:
We upgraded libc versions from 0.6 to 0.7 that fixed the definition of c_char on ARM. Should be an easy fix to add a cast there.

operutka:
Great! As I mentioned before, this is not the only problematic spot. Here is the complete error output:

```
[...]

PR #322, opened 2015-12-16
Fix regression of c_char type mismatches on ARM

operutka:
The problem with signed/unsigned c_char (issue #314) reappeared in the latest release. This patch fixes it.


sfackler:
Ugh, thanks. I'll try to get Travis set up to run a cross compilation build to ARM to avoid regressions like this in the future.

sfackler:
Released v0.7.3

operutka:
thx

PR #324, opened 2015-12-18
Fix Cargo.toml to actually depend on gdi32-sys and user32-sys

gentoo90:
`.dependencies` was missing, so it didn't even depend on them


sfackler:
Thanks! Kind of surprised Cargo was okay with a bunch of random unused fields

PR #328, opened 2016-01-05
Add support for RSA PEM files.

Cyberunner23:
Adds support for RSA PEM files. I needed to encrypt some data with an RSA PEM key so I added some functions to do that.


sfackler:
Can you add some tests? Seems good to me other than that.

PR #330, opened 2016-01-09
Add a interface to RSA structs

esclear:
This is my take at making the PKey / RSA struct available in the binding and making the BigNums of keys accessible.
This will be used in a crate to support JWK (JSON Web Keys).

It is currently used by casting the pointer returned by `PKey::get_handle`:

```rust
let gen = X509Generator::new();

let (cert, pkey) = gen.generate().unwrap();

unsafe {
    let evppkey = pkey.get_handle();
    let rsa = (*evppkey).pkey as *mut RSA;
    let n : BigNum = (*rsa).get_n();
    println!("n == {:?}", n);
}
```

I would also like to hide the `BigNum::new_from_ffi` function, because it does not need to be exported publicly.

### DIFF
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -65,9 +65,9 @@ unsafe impl Sync for BIO_METHOD {}
 
 #[repr(C)]
 pub struct RSA {
-    pad: c_int,
-    version: c_long,
-    meth: *const c_void,
+    pub pad: c_int,
+    pub version: c_long,
+    pub meth: *const c_void,
 
     pub engine: *mut c_void,
     pub n: *mut BIGNUM,
@@ -79,17 +79,17 @@ pub struct RSA {
     pub dmq1: *mut BIGNUM,
     pub iqmp: *mut BIGNUM,
 
-    ex_data: *mut c_void,
-    references: c_int,
-    flags: c_int,
+    pub ex_data: *mut c_void,
+    pub references: c_int,
+    pub flags: c_int,
 
-    _method_mod_n: *mut c_void,
-    _method_mod_p: *mut c_void,
-    _method_mod_q: *mut c_void,
+    pub _method_mod_n: *mut c_void,
+    pub _method_mod_p: *mut c_void,
+    pub _method_mod_q: *mut c_void,
 
-    bignum_data: *mut c_char,
-    blinding: *mut c_void,
-    mt_blinding: *mut c_void,
+    pub bignum_data: *mut c_char,
+    pub blinding: *mut c_void,
+    pub mt_blinding: *mut c_void,
 }
 
 #[repr(C)]
@@ -98,10 +98,10 @@ pub struct EVP_PKEY {
     pub save_type: c_int,
     pub references: c_int,
     pub ameth: *const c_void,
-    engine: *mut ENGINE,
+    pub engine: *mut ENGINE,
     pub pkey: *mut c_void,
-    save_parameters: c_int,
-    attributes: *mut c_void,
+    pub save_parameters: c_int,
+    pub attributes: *mut c_void,
 }
 
 #[repr(C)]
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -22,9 +22,7 @@ pub type ENGINE = c_void;
 pub type EVP_CIPHER = c_void;
 pub type EVP_CIPHER_CTX = c_void;
 pub type EVP_MD = c_void;
-pub type EVP_PKEY = c_void;
 pub type EVP_PKEY_CTX = c_void;
-pub type RSA = c_void;
 pub type SSL = c_void;
 pub type SSL_CTX = c_void;
 pub type SSL_METHOD = c_void;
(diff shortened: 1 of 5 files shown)


openssl-sys/src/lib.rs
```diff
@@ -66,6 +64,47 @@ pub struct BIO_METHOD {
 unsafe impl Sync for BIO_METHOD {}
 
 #[repr(C)]
+pub struct RSA {
+    pad: c_int,
```
sfackler: We'll want all of these to be public since this crate just contains the raw FFI bindings.

openssl/src/crypto/rsa.rs
```diff
@@ -0,0 +1,39 @@
+use ffi;
+use bn::BigNum;
+use std::fmt;
+
+pub struct RSA {
+    pub rsa_obj : ffi::RSA
```
sfackler: I'd prefer this to be private, and methods to be added to pull the relevant fields.

openssl/src/crypto/rsa.rs
```diff
@@ -0,0 +1,39 @@
...
+
+    pub unsafe fn get_p(&self) -> BigNum {
+        BigNum::new_from_ffi(self.rsa_obj.p).unwrap()
+    }
+
+    pub unsafe fn get_q(&self) -> BigNum {
+        BigNum::new_from_ffi(self.rsa_obj.q).unwrap()
+    }
+
+    pub fn get_type(&self) -> &str {
```
sfackler: What's the use case for this?

openssl/src/crypto/rsa.rs
```diff
@@ -0,0 +1,39 @@
...
+        "rsa"
+    }
+}
+
+impl fmt::Debug for RSA {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Currently no debug output. Sorry :(")
```
sfackler: Can this just be `write!(f, "RSA")`?

openssl/src/bn/mod.rs
```diff
@@ -102,6 +102,18 @@ impl BigNum {
         })
     }
 
+    pub unsafe fn new_from_ffi(orig: *mut ffi::BIGNUM) -> Result<BigNum, SslError> {
+        if orig.is_null() {
+            panic!("Null Pointer was supplied to BigNum::new_from_ffi");
+        }
+        let r = ffi::BN_dup(orig);
+        if r.is_null() {
+            panic!("Unexpected null pointer from BN_dup(..)")
```
sfackler: Why does this method return a `Result` if it panics when errors are encountered?

openssl/src/crypto/rsa.rs
```diff
@@ -0,0 +1,39 @@
+use ffi;
+use bn::BigNum;
+use std::fmt;
+
+pub struct RSA {
+    pub rsa_obj : ffi::RSA
+}
+
+impl RSA {
+    pub unsafe fn get_n(&self) -> BigNum {
```
sfackler: Rust convention is to not include the `get_` prefix for getter functions.

openssl/src/bn/mod.rs
esclear: I was and am not quite sure what kind of error to return.

openssl/src/crypto/rsa.rs
esclear: Actually, there is none. This was just left over from playing with the changes and JWK generation.

openssl/src/crypto/rsa.rs
esclear: I wanted to add output of the e, p and q parameters of the key.
Or would this be too verbose?

sfackler: Oh, a more robust implementation is absolutely better, but a minimal impl should be `RSA` rather than `Currently no debug output. Sorry :(`.

openssl/src/bn/mod.rs
sfackler: `SslError::get()` will return the contents of the error stack that OpenSSL will set when this returns NULL.

sfackler:
Don't worry about the failing windows test, by the way. The build configuration's messed up on there and I haven't gotten around to fixing it.

jethrogb:
I think somewhere in the refactoring it became impossible to instantiate this struct.

esclear:
@jethrogb: Which struct are you talking about?

jethrogb:
@esclear `openssl::crypto::rsa::RSA`. It has private fields and there are no constructors.

esclear:
@jethrogb I see. It is a problem indeed. Kind of.
It is not intended to be manually instanciated, but rather created by casting pointers as shown in the pull requests comment.
But I might add some kind of generator-type possibilities of constructing RSA structs.

Problem is, you never exactly know, what else would be required to generate or instanciate native RSA objects.

Any proposals?

jethrogb:
I see. The standard interface used in most of the other wrappers in this library is to store a `*mut ffi::RSA` in this struct and add various methods to instantiate it. For example, you would add `unsafe fn from_ffi(*mut ffi::RSA) -> RSA` which would just give you a pointer. Then in `PKey`, you could add a method `to_rsa(&self) -> Option<RSA>` which calls `ffi::EVP_PKEY_get1_RSA`. Remember to add a `Drop` impl.

esclear:
Sounds like a good idea.
I will try that out tomorrow.

esclear:
@jethrogb Looking at it further, I think I have found some problems:
There does not seem to be any intended way of correctly copying / cloning OpenSSL RSA objects.
Therefor thay cannot exist without PKey objects.

I might have overlooked something, but that's what I've found so far.

sfackler:
I just pushed a commit that adds constructors while fixing a bug in another set of methods: https://github.com/sfackler/rust-openssl/commit/4e58fd10de3daa3eca08e906a4b2e4239ac6d37d

PR #334, opened 2016-01-18
ssl: fix refcounting of SslContext when set_ssl_context is used

### LINKED ISSUE #333: Calling set_ssl_context() twice with the same context causes Undefined Behavior
```
pub fn set_ssl_context(&self, ctx: &SslContext) -> SslContext
```

One passes in a `&SslContext`, but gets back an owned `SslContext`. This allows double-freeing of any SslContext, just call `set_ssl_context()` twice with the same context. Can be the same `Ssl`/`self` or a different one.

Right now I'm avoiding this by doing:
```
mem::forget(ssl.set_ssl_context(&ctx));
```

But this probably isn't perfectly correct (I track all my contexts in `Arc`s, and really need to convert them back to `Arc`s to decrease the ref count), but leaking memory is quite a bit better than crashing.

jmesmon:
Additionally impl Clone for SslContext to both allow us to use it &
allow external users to take advantage of SslContext's internal
refcount.

Maintain the existing signature for set_ssl_context(), but
add inline comments recommending changing it.

Fixes #333

Also adds an attempt at testing for this, which could be expanded.

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -481,12 +481,20 @@ fn wrap_ssl_result(res: c_int) -> Result<(), SslError> {
 }
 
 /// An SSL context object
+///
+/// Internally ref-counted, use `.clone()` in the same way as Rc and Arc.
 pub struct SslContext {
     ctx: *mut ffi::SSL_CTX,
 }
 
 unsafe impl Send for SslContext {}
 unsafe impl Sync for SslContext {}
+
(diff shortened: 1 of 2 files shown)


sfackler:
Wow, OpenSSL's behavior in SSL_set_SSL_CTX is really bizarre. I would have expected it to return the old context, and of course there is literally zero documentation anywhere. We'll remove the return value in the next major version bum.

PR #336, opened 2016-01-20
impl Clone for PKey and X509 by using their 'references' member

jmesmon:
This is based on the existing code for handling `SslContexts`.

Note that as soon as we start working with `X509` like this, there is quite a bit of restructuring that is (probably) a good idea (notably, getting rid of the 'owned' boolean in favor of always owning/refcounting) unless there are perf concerns.

### DIFF
--- a/openssl/src/c_helpers.c
+++ b/openssl/src/c_helpers.c
@@ -7,3 +7,7 @@ void rust_SSL_clone(SSL *ssl) {
 void rust_SSL_CTX_clone(SSL_CTX *ctx) {
     CRYPTO_add(&ctx->references,1,CRYPTO_LOCK_SSL_CTX);
 }
+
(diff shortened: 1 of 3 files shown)


sfackler:
Yep, definitely agree that some refactoring is in order.

sfackler:
It turns out that OpenSSL 1.1 makes most of its types opaque so the current strategy of adjusting the refcount isn't going to work. I think I'm going to have to back this change out unfortunately.

jmesmon:
It also (afaik) adds functions that increment ref counts on (at least) SSL_CTX, X509, and EVP_PKEY (via `*_up_ref` functions).

We'll probably need to select between `CRYPT_add` opencoding and `*_up_ref` functions depending on their availability.

sfackler:
Ah great thanks

PR #340, opened 2016-01-26
Local Certificate & PKey accessors for SslContext

jmesmon:
Allow retrieval of the local private key and local certificate from an SslContext. This uses reference counting machinery added previously for `.clone()`.

Also adjust the existing `peer_certificate()`'s lifetime (extend to 'static) to avoid strange lifetime inference issue. I'm pretty sure this should be a non-breaking change (because the lifetime is just being expanded).

### DIFF
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -677,6 +677,9 @@ extern "C" {
 
     pub fn SSL_CTX_set_cipher_list(ssl: *mut SSL_CTX, s: *const c_char) -> c_int;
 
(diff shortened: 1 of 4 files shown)


jmesmon:
Note: In this PR I've named the functions that construct instances from pointers by incrimenting their reference count `from_ref`. For `SslContext` (in a previous PR) I called the same thing (but for `SslContext`) `new_ref`. It probably would be a good idea to pick one naming convention, but I'm not sure which is a better choice. With some input here, I'd be happy to adjust the naming to match a common form.

sfackler:
`ref` normally implies a shared reference - `from_ptr` or `from_raw_parts` are more canonical names I believe.

PR #341, opened 2016-01-26
x509: allow setting & getting validity times to use time::Tm and std::time::Duration

jmesmon:
To support this, `Asn1Time` grows the ability to be constructed from a `time::Tm` and converted to a `time::Tm`.

Additionally, a few extra methods were added to `X509Generator` and the way it tracks validity prior to cert signing was adjusted.

For the `Asn1Time` <=> `time::Tm` conversion, I had to add a parser for ASN.1 formatted times.
I don't really like this, but the API the openssl exposes doesn't appear provide another way to do this (without precision & 32bit hazards via `time_t`).

Note that I am mixing `time::Tm` and `std::time::Duration` here. Ideally I'd use all `std` data-structures, but there isn't anything comparable to `time::Tm` (yet, anyhow) so I've stuck with using `std` where possible but falling back to the `time` crate.

### DIFF
--- a/openssl/src/asn1/mod.rs
+++ b/openssl/src/asn1/mod.rs
@@ -1,9 +1,14 @@
-use libc::c_long;
+use libc::{c_long, time_t, c_char};
 use std::ptr;
 
+use time;
 use ffi;
 use ssl::error::SslError;
+use std::ffi::CString;
+use std::slice;
 
+#[cfg(test)]
+mod tests;
 
 pub struct Asn1Time {
     handle: *mut ffi::ASN1_TIME,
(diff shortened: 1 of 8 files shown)


jmesmon:
Hold off on this, looks like I might have missed a few ASN.1 time edge cases.

sfackler:
Looks like at least some portion of this will need to be defined behind Cargo features: https://travis-ci.org/sfackler/rust-openssl/jobs/104926719#L108

openssl/src/asn1/mod.rs
```diff
@@ -46,3 +103,143 @@ impl Drop for Asn1Time {
...
+            t.tm_mday,
+            t.tm_hour,
+            t.tm_min,
+            t.tm_sec,
+            t.tm_nsec,
+            tz_hours,
+            tz_min)
+}
+
+fn parse_n_digit<I>(b: &mut I, ct: usize, name: &str) -> Result<i32, String>
```
sfackler: Can't this be replaced with `i32`'s `FromStr` impl on `&str[..ct]`?

sfackler:
The time crate is somewhat deprecated - does chrono have an appropriate type for this?

jmesmon:
It looks like the `chrono` crate (hadn't heard about it before now) has `DateTime` which would fit the need here. I'm guessing `chrono` is preferred over `time`?

As for the feature flag: I'm actually supprised the ASN1_TIME functions I'm using are so new (added in Feb 2015), given the difficulty in using them, I had presumed they were much older.

I suppose the options wrt lacking `ASN1_TIME_set_string` are:

 - don't expose a bunch of functionality if we lack a "asn1time" feature flag (of some sort), OR
 - provide a shim for ASN1_TIME_set_string in openssl-sys-extras on versions of openssl that lack it (not sure how easy it would be to tell, I haven't looked at whether there is probing for function existence).

Generally I'd prefer the second option (assuming it is possible). Any preference there?

On using FromStr to parse numbers: looks like it could work. I'll try it out.

sfackler:
I'd personally lean towards the first option, actually - probably expose wrappers for `ASN1_TIME_set` and `ASN1_TIME_adj`?

sfackler:
Closing due to inactivity, but feel free to open with a rebase/those questions resolved!

PR #343, opened 2016-01-28
Preserve X.509 extension insertion order

### LINKED ISSUE #327: X509 cert/key pair generation is nondeterministic
Given the following program:

```rust
extern crate openssl;

use openssl::nid::Nid;
use openssl::x509::X509Generator;
use openssl::x509::extension::Extension;

fn main() {
    let ca_gen = X509Generator::new()
        .add_extension(Extension::OtherNid(Nid::SubjectKeyIdentifier, "hash".to_owned()))
        .add_extension(Extension::OtherNid(Nid::AuthorityKeyIdentifier, "keyid:always".to_owned()));

    ca_gen.generate().expect("failed to generate CA cert and key");
}
```

Running the program sometimes completes successfully and sometimes panics with:

```
thread '<main>' panicked at 'failed to generate CA cert and key: OpenSslErrors([UnknownError { library: "X509 V3 routines", function: "V2I_AUTHORITY_KEYID", reason: "unable to get issuer keyid" }])', src/libcore/result.rs:757
```

It seems that the order of operations is not deterministic, because OpenSSL is not always able to find a key ID to use for authorityKeyIdentifier. (In this case, it's for a self-signed CA certificate so the subjectKeyIdentifier and AuthorityKeyIdentifier should be the same value: the ID of the CA's own public key.) I've tried it with the two `add_extension` calls in reverse order, and the behavior is the same.

jimmycuadra:
Ensures that extensions that are order-dependent are inserted in the same order when calling out to OpenSSL during certificate signing.

Fixes #327.

### DIFF
--- a/openssl/src/x509/mod.rs
+++ b/openssl/src/x509/mod.rs
@@ -146,8 +146,7 @@ pub struct X509Generator {
     bits: u32,
     days: u32,
     names: Vec<(String, String)>,
-    // RFC 3280 §4.2: A certificate MUST NOT include more than one instance of a particular extension.
-    extensions: HashMap<ExtensionType, Extension>,
+    extensions: Extensions,
     hash_type: HashType,
 }
 
@@ -166,7 +165,7 @@ impl X509Generator {
             bits: 1024,
             days: 365,
             names: vec![],
-            extensions: HashMap::new(),
+            extensions: Extensions::new(),
             hash_type: HashType::SHA1,
         }
     }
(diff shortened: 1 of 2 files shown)


openssl/src/x509/mod.rs
```diff
@@ -618,6 +620,72 @@ impl Drop for X509Req {
     }
 }
 
+/// A collection of X.509 extensions.
+///
+/// Upholds the invariant that a certificate MUST NOT include more than one
+/// instance of a particular extension, according to RFC 3280 §4.2.
```
sfackler: Could you also note in the comment here that addition order matters, which is why we have this extra logic in the first place?

sfackler:
Looks good other than one doc tweak.

jimmycuadra:
Updated the doc block as requested. Thanks for the review!

PR #347, opened 2016-02-01
Update El Capitan section

ctulek:
I used the instructions on the following comment and it worked for me:
https://github.com/phusion/passenger/issues/1630#issuecomment-147527656

This should help to fix the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,12 @@ rust-openssl to a separate installation. OSX releases starting at 10.11, "El
 Capitan", no longer include OpenSSL headers which will prevent the `openssl`
 crate from compiling.
 
+Try the following commands to fix the issue (assumes Homebrew is used for package management):
+```
+brew install openssl
+brew link openssl --force
+```
+
 ### Windows
 
 On Windows, consider building with [mingw-w64](http://mingw-w64.org/).


README.md
```diff
@@ -31,6 +31,12 @@ rust-openssl to a separate installation. OSX releases starting at 10.11, "El
 Capitan", no longer include OpenSSL headers which will prevent the `openssl`
 crate from compiling.
 
+Try the following commands to fix the issue (assumes Homebrew is used for package management):
+```
+brew install openssl
+brew link openssl --force
```
sfackler: I am not sure I'm on board with doing this - there is a reason that brew does not link openssl by default. A better recommendation could be to set `OPENSSL_LIB_DIR` to `/usr/local/opt/openssl/lib` and `OPENSSL_INCLUDE_DIR` to `/usr/local/opt/openssl/include`.

ctulek: I hear you about the --force part, that may be hard to make as a general recommendation. By `OPENSSL_LIB_DIR` do you mean setting them in Cargo.toml or somewhere like that or in bash profile? I am not experienced about the former, but the latter does not look to me any different than `brew link openssl --force`.

I am curious about your answer. According to that I will close this PR/

sfackler: It'd go in your `~/.bashrc` or equivalent for other shells. The `OPENSSL_LIB_DIR` and `OPENSSL_INCLUDE_DIR` variables are only used by this library's logic, while `brew link openssl --force` is a global action. The downside of that is that any programs that expect to be running against the OpenSSL 0.9.8 that OSX ships with could end up pulling in the modern one installed by homebrew and blowing up, or finding the *headers* for the homebrew version but linking against the stock version. 0.9.8 and 1.0.2 aren't binary compatible, so this will cause mysterious breakage.

ctulek: Thanks a lot for explaining it. Closing this PR

ctulek:
The suggestion was not a good one.

PR #349, opened 2016-02-02
Fix Nid::UID value

jwilm:
`Nid::UID` (userId) previously held the value of `Nid::uid` (uniqueIdentifier).

I think the NID values were potentially taken from the wrong list in openssl. There's an [old open PR](https://github.com/sfackler/rust-openssl/pull/84) on here that mentioned reading them from `include/openssl/objects.h`, but that list doesn't contain all nid values, and they conflict (mildly) with the list in `include/openssl/obj_mac.h`. This being the [second PR](https://github.com/sfackler/rust-openssl/pull/232) to correct some NID values seems to support this claim.


openssl/src/nid.rs
```diff
@@ -104,7 +104,7 @@ pub enum Nid {
     G,
     S,
     I,
-    UID,
+    uid, // uniqueIdentifier
```
sfackler: It might be better to rename this to `UniqueIdentifier`?

sfackler: Actually, it seems like this should keep its name and the new variant should be called `UserId` or something like that to preserve back compat, yes?

jwilm: Well, openssl references them as `uid` and `UID` as included in this PR. However, there are probably issues with many more of the Nid values, and perhaps we should plan to update them all at once for the next breaking release. In the mean time, I can just change the new variant to UserId. What do you think?

sfackler: Yeah, these variants should definitely be renamed - `G`, `S`, and `I` are not particularly reasonable to work with :). The `breaks` branch has the in-progress breaking changes that'll land in the next major release if you'd like to open a PR with the big rename against that.

In the mean time, let's add this as `UserId`.

jwilm: > In the mean time, let's add this as `UserId`.

Done.

> these variants should definitely be renamed 

It's more than renaming. There are missing variants, and the values have been shown in the past to be incorrect.

PR #350, opened 2016-02-06
Update gcc dependency version

fabricedesre:
This is needed to pick up a recent update to gcc-rs that brings support for armv7 targets.


fabricedesre:
@sfackler It looks like the failure is unrelated to the change, but I can't retrigger the appveyor build.

sfackler:
I don't think any change needs to be made to the repo itself - a `cargo update` on your end should cause cargo to use the newest version of `gcc`.

fabricedesre:
You're right. Thanks!

PR #356, opened 2016-02-29
Add support for SSL_CIPHER

erikjohnston:
Also, adds SSL_get_version

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -769,6 +769,46 @@ impl SslContext {
     }
 }
 
+pub struct SslCipher {
+    cipher: *const ffi::SSL_CIPHER,
+}
+
+impl SslCipher {
+    pub fn name(&self) -> &'static str {
+        let name = unsafe {
+            let ptr = ffi::SSL_CIPHER_get_name(self.cipher);
+            CStr::from_ptr(ptr as *const _)
+        };
+
+        str::from_utf8(name.to_bytes()).unwrap()
+    }
+
+    pub fn version(&self) -> &'static str {
+        let version = unsafe {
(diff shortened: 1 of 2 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -769,6 +769,67 @@ impl SslContext {
...
+        str::from_utf8(version.to_bytes()).unwrap()
+    }
+
+    /// Returns the number of secret bits used for the cipher.
+    ///
+    /// The first element is the number of secret bits used for the cipher.
+    ///
+    /// The second element, if not None, is the number of bits processed by
+    /// the chosen algorithm,
+    pub fn bits(&self) -> (i32, Option<i32>) {
```
sfackler: Could you make this return a struct instead? Something as simple as

```rust
pub struct CipherBits {
    pub secret: i32,
    pub algorithm: Option<i32>,
}
```

erikjohnston: Yes, that would be much nicer.

erikjohnston:
I should also add that I've assumed the `SSL_CIPHER` returned by `SSL_get_current_cipher` is only valid for the lifetime of the `SSL` object, but I couldn't actually find anything in the openssl docs about it.

sfackler:
Yep, OpenSSL's documentation is pretty awful :) I'll jump into the source tonight and make sure, but I'd guess your assumption is correct.

erikjohnston:
Yes it is awful :)

Thanks, I managed to double check that all the static strings are in fact static strings by looking at the source, but then I started running a fever and collapsed (for, like, a week). I've only just recovered, so I'm quite happy to leave it to you to check that last detail!

PR #358, opened 2016-03-10
(maint) Update download version for OpenSSL in appveyor to latest

ajroetker:
This commit updates the download version for OpenSSL in appveyor,
previous to this commit the version that was installed was mismatched
from the download version.


sfackler:
Oops, thanks!

PR #361, opened 2016-03-18
Allow Rust to infer the type of the argument to SSL_CIPHER_description.

Ms2ger:
This allows the code to compile on Android, where an unsigned char is
expected.


Ms2ger:
Please publish to crates.io too

sfackler:
Published 0.7.8

PR #365, opened 2016-03-23
Use bitflags 0.5

saurvs:
Part of https://github.com/servo/servo/issues/10077


sfackler:
Bitflags 0.5 seems to break back-compat: https://travis-ci.org/sfackler/rust-openssl/builds/117955331. I don't think we can do this until a major version bump. If you repoen the PR against the `breaks` branch, we can get it ready for the next major release.

saurvs:
Okay.

sfackler:
Covered by #385

PR #367, opened 2016-03-26
Use bitflags 0.5 in the breaks branch

saurvs:
Part of https://github.com/servo/servo/issues/10077.

### DIFF
--- a/openssl/Cargo.toml
+++ b/openssl/Cargo.toml
@@ -29,7 +29,7 @@ pkcs5_pbkdf2_hmac = ["openssl-sys/pkcs5_pbkdf2_hmac"]
 nightly = []
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


sfackler:
Doesn't appear to compile: https://travis-ci.org/sfackler/rust-openssl/builds/118684997

gkoz:
Bitflags 0.5 makes everything private by default. you can't just bump the version, you need to change `flags Foo: ...` to `pub flags Foo: ...`.

saurvs:
I added `pub` but it doesn't fix it because it the `ffi_extras` and `ffi` namespaces aren't being resolved for some reason. It works fine if the constants from those modules are replaced with literals. Any idea why this is happening?

sfackler:
Covered by #385

PR #371, opened 2016-04-04
Implement Send for all structs in sys crate

### LINKED ISSUE #370: Send for BIO_METHOD?
Hi @sfackler 

Ran into an issue updating Solicit to rust-openssl 0.7 from 0.6. It looks like the BIO_METHOD type now includes a raw pointer which is `!Send`. Is there any reason the BIO_METHOD is actually `!Send`? It currently has a `Sync` impl.

```
src/http/client/tls.rs:179:6: 179:22 note: `*const i8` cannot be sent between threads safely
src/http/client/tls.rs:179:6: 179:22 note: required because it appears within the type `openssl_sys::BIO_METHOD`
src/http/client/tls.rs:179:6: 179:22 note: required because it appears within the type `openssl::ssl::SslStream<std::net::tcp::TcpStream>`
```

jwilm:
The higher level wrappers for these types provide the safe API.

Resolves #370.

---
This change is [Reviewable](https://reviewable.io/reviews/sfackler/rust-openssl/371)

### DIFF
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -60,6 +60,8 @@ pub struct BIO_METHOD {
     pub callback_ctrl: Option<unsafe extern "C" fn(*mut BIO, c_int, bio_info_cb) -> c_long>,
 }
 
+unsafe impl Send for BIO_METHOD {}
+
 // so we can create static BIO_METHODs
 unsafe impl Sync for BIO_METHOD {}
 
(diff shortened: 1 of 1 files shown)


sfackler:
Apologies for contradicting myself, but after some discussion on IRC, I actually think we want to go in the other direction: https://botbot.me/mozilla/rust-libs/2016-04-04/?msg=63496420&page=1. Ideally, basically all types in the -sys crate would be `!Send` and `!Sync`. Code that wraps them to expose a safe interface would then opt-into implementing those traits where appropriate via unsafe impls. In this case, that would either be a type in the openssl crate itself or in Solicit.

Does that make sense?

jwilm:
Yeah that sounds way more appropriate. Will continue discussion in original issue.

PR #372, opened 2016-04-04
Add safe wrapper BioMethod for ffi::BIO_METHOD

### LINKED ISSUE #370: Send for BIO_METHOD?

jwilm:
Adds a wrapper for `ffi::BIO_METHOD` located at `ssl::bio::BioMethod`. This enables `SslStream` to be `Send` without doing an unsafe impl on the ffi struct.

Resolves #370.

---
This change is [Reviewable](https://reviewable.io/reviews/sfackler/rust-openssl/372)

### DIFF
--- a/openssl/src/ssl/bio.rs
+++ b/openssl/src/ssl/bio.rs
@@ -1,5 +1,5 @@
 use libc::{c_char, c_int, c_long, c_void, strlen};
-use ffi::{BIO, BIO_METHOD, BIO_CTRL_FLUSH, BIO_TYPE_NONE, BIO_new};
+use ffi::{self, BIO, BIO_CTRL_FLUSH, BIO_TYPE_NONE, BIO_new};
 use ffi_extras::{BIO_clear_retry_flags, BIO_set_retry_read, BIO_set_retry_write};
 use std::any::Any;
 use std::io;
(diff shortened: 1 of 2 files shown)


jwilm:
It might make sense to annotate this type with `#[doc(hidden)]` since it's not really usable outside of where it's defined in the current form.

openssl/src/ssl/bio.rs
```diff
@@ -17,19 +18,38 @@ pub struct StreamState<S> {
...
+                create: Some(create),
+                destroy: Some(destroy::<S>),
+                callback_ctrl: None,
+            },
+            _p: PhantomData
+        }
+    }
+}
+
+unsafe impl<S: Send> Send for BioMethod<S> {}
```
sfackler: I think this impl can be unqualified (i.e. just `impl<S> Send ...`)

openssl/src/ssl/mod.rs
```diff
@@ -1117,8 +1119,7 @@ make_LibSslError! {
 /// A stream wrapper which handles SSL encryption for an underlying stream.
 pub struct SslStream<S> {
     ssl: Ssl,
-    _method: Arc<ffi::BIO_METHOD>, // NOTE: this *must* be after the Ssl field so things drop right
-    _p: PhantomData<S>,
```
sfackler: And then I'd leave this here to pick up a `!Send` if `S: !Send`.

openssl/src/ssl/bio.rs
```diff
@@ -17,19 +18,38 @@ pub struct StreamState<S> {
...
-        bread: Some(bread::<S>),
-        bputs: Some(bputs::<S>),
-        bgets: None,
-        ctrl: Some(ctrl::<S>),
-        create: Some(create),
-        destroy: Some(destroy::<S>),
-        callback_ctrl: None,
-    });
+/// Safe wrapper for BIO_METHOD
+pub struct BioMethod<S> {
```
sfackler: Actually, we can just make this non-generic, except for the constructor.

openssl/src/ssl/bio.rs
jwilm: Could you elaborate on that? That would cause SslStream to always be Send regardless of the stream type.

sfackler: Taken together, the changes from the couple of comments I left scattered around would be to have this:

```
pub struct SslStream<S> {
    ssl: Ssl,
    _method: Arc<BioMethod>,
    _p: PhantomData<S>,
}

struct BioMethod(ffi::BIO_METHOD);

unsafe impl Send for BioMethod {}

impl BioMethod {
    pub fn new<S>() -> BioMethod {
        // ...
    }
}
```

openssl/src/ssl/mod.rs
jwilm: If we keep the `S: Send` qualification on `BioMethod` instead, we can remove the `unsafe impl Send` on `SslStream` since it will automatically be `Send` due to `Ssl` and `BioMethod` both being `Send`. Send via composition rather than `unsafe impl` feels safer.

openssl/src/ssl/bio.rs
jwilm: Happy to go this route if you prefer. Just wanted to check first that you saw my thoughts on your other comment before making the changes.

https://github.com/sfackler/rust-openssl/pull/372#discussion_r58461806

sfackler: In both cases we don't have an unsafe impl on SslStream. In they layout above, we make BioMethod non-generic and unconditionally Send since all it is is a struct of function pointers. The only thing in an SslStream that may cause us to be !Send is S, which is taken care of by the _p field. Does that make sense?

sfackler:
I don't think we have to worry about #[doc(hidden)] since the bio module is private.

openssl/src/ssl/bio.rs
jwilm: D'oh, right. I was confused because there's currently an unsafe impl on the SslStream. Changes incoming.

jwilm:
> I don't think we have to worry about #[doc(hidden)] since the bio module is private.

Oh, great!

The recommended changes have been pushed. Thanks for the review and entertaining my questions :).

PR #374, opened 2016-04-06
change PKey::clone to use EVP_PKEY_copy_parameters

kcking:
Right now PKey::clone just increments the reference count, falsely giving full ownership of the same struct to two owners. Code that simply `clone`s and repeatedly calls `gen` from multiple threads will segfault.

Here is a branch with just the test showing that it fails: https://github.com/kcking/rust-openssl/tree/kcking-pkey-clone-does-not-copy

(I didn't get a chance to take a look at other Clone impls since my current project only uses PKey)

### DIFF
--- a/openssl/src/c_helpers.c
+++ b/openssl/src/c_helpers.c
@@ -8,10 +8,6 @@ void rust_SSL_CTX_clone(SSL_CTX *ctx) {
     CRYPTO_add(&ctx->references,1,CRYPTO_LOCK_SSL_CTX);
 }
 
-void rust_EVP_PKEY_clone(EVP_PKEY *pkey) {
-    CRYPTO_add(&pkey->references,1,CRYPTO_LOCK_EVP_PKEY);
-}
-
 void rust_X509_clone(X509 *x509) {
     CRYPTO_add(&x509->references,1,CRYPTO_LOCK_X509);
 }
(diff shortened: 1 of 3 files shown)


openssl/src/c_helpers.c
```diff
@@ -8,8 +8,12 @@ void rust_SSL_CTX_clone(SSL_CTX *ctx) {
     CRYPTO_add(&ctx->references,1,CRYPTO_LOCK_SSL_CTX);
 }
 
-void rust_EVP_PKEY_clone(EVP_PKEY *pkey) {
-    CRYPTO_add(&pkey->references,1,CRYPTO_LOCK_EVP_PKEY);
+EVP_PKEY* rust_EVP_PKEY_clone(EVP_PKEY *pkey) {
+    EVP_PKEY* new_pkey = EVP_PKEY_new();
+    if (EVP_PKEY_copy_parameters(new_pkey, pkey) != 0) {
```
sfackler: `EVP_PKEY_copy_parameters` looks to be a real function rather than a macro, so it can be defined in openssl-sys and then used directly from the rust Clone implementation. This C file only exists since `CRYPTO_add` is a macro.

kcking: Ah good point, just fixed that :)

sfackler:
Thanks!

PR #376, opened 2016-04-10
Switch to using DER encode/decode to clone PKey (since EVP_PKEY_copy_parameters doesn't work)

kcking:
I clearly did not test my last change enough -- accessing `n()`, `q()`, or `p()` on the cloned PKey still causes a segfault (at least it fails fast now ;) ).

According to this (not so populated) SO post, encoding and decoding is an accepted way to copy both private and public keys (http://stackoverflow.com/questions/6449443/how-can-we-copy-an-evp-pkey-which-includes-rsa-key). I figure safety should be the priority here so I used this method.

Let me know what you think!

[failing branch](https://github.com/kcking/rust-openssl/tree/kcking-failing-pkey-clone-copies-private)

### DIFF
--- a/openssl/src/crypto/pkey.rs
+++ b/openssl/src/crypto/pkey.rs
@@ -609,11 +609,19 @@ impl Drop for PKey {
 
 impl Clone for PKey {
     fn clone(&self) -> Self {
-        unsafe {
-            let new_evp = ffi::EVP_PKEY_new();
-            assert!(ffi::EVP_PKEY_copy_parameters(new_evp, self.evp) == 0);
-            PKey::from_handle(new_evp, self.parts)
+        let mut pkey = PKey::from_handle(unsafe { ffi::EVP_PKEY_new() }, self.parts);
+        //  copy by encoding to DER and back
+        match self.parts {
+            Parts::Public => {
+                pkey.load_pub(&self.save_pub()[..]);
+            },
+            Parts::Both => {
+                pkey.load_priv(&self.save_priv()[..]);
+            },
+            Parts::Neither => {
+            },
         }
+        pkey
     }
 }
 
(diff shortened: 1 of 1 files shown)


openssl-sys/src/lib.rs
```diff
@@ -548,7 +548,6 @@ extern "C" {
     pub fn EVP_PKEY_new() -> *mut EVP_PKEY;
     pub fn EVP_PKEY_free(k: *mut EVP_PKEY);
     pub fn EVP_PKEY_assign(pkey: *mut EVP_PKEY, typ: c_int, key: *const c_void) -> c_int;
-    pub fn EVP_PKEY_copy_parameters(to: *mut EVP_PKEY, from: *const EVP_PKEY) -> c_int;
```
sfackler: We'll want to leave this here since it's made it into a release.

sfackler:
OpenSSL continues to amaze.

kcking:
:)

Switching to stable now

openssl/src/crypto/pkey.rs
```diff
@@ -609,11 +609,19 @@ impl Drop for PKey {
...
     fn clone(&self) -> Self {
-        unsafe {
-            let new_evp = ffi::EVP_PKEY_new();
-            assert!(ffi::EVP_PKEY_copy_parameters(new_evp, self.evp) == 0);
-            PKey::from_handle(new_evp, self.parts)
+        let mut pkey = PKey::from_handle(unsafe { ffi::EVP_PKEY_new() }, self.parts);
+        //  copy by encoding to DER and back
+        match self.parts {
+            Parts::Public => {
+                pkey.load_pub(self.save_pub().as_slice());
```
sfackler: Can you change the `foo.as_slice()`s to `&foo` here and below?

PR #379, opened 2016-04-13
Add ability to set session ID context on an SSL context

rix0rrr:
This is necessary to make authentication with client certificates work
without session restarts.


sfackler:
Thanks!

PR #381, opened 2016-04-14
Add 1DES symm ciphers (des-cbc, des-ecb, des-cfb, des-ofb)

chaaz:
1DES is well and truly dead for actual sensitive information, (its
keysize is too small for modern purposes), but it can still find use in
backwards compatiblity or educational applications.

### DIFF
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -529,8 +529,6 @@ extern "C" {
 
     pub fn EVP_des_cbc() -> *const EVP_CIPHER;
     pub fn EVP_des_ecb() -> *const EVP_CIPHER;
-    pub fn EVP_des_cfb() -> *const EVP_CIPHER;
-    pub fn EVP_des_ofb() -> *const EVP_CIPHER;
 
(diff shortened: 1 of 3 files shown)


chaaz:
Hmm.. regarding this AppVeyor failure:

```note: C:\projects\rust-openssl\openssl\target\debug\openssl-9bed4e01438f8395.0.o: In function `ZN7openssl6crypto13symm_internal4evpcE':
C:\projects\rust-openssl/openssl\src\crypto/symm_internal.rs:31: undefined reference to `EVP_des_cfb'```

Is cargo failing to pick up the additions to openssl-sys? Or does EVP_des_cfb not exist in Win32 ssleay32 / libeay32 libraries for some reason?

I'm open to ideas here, but I might end up dropping des_cfb() (and des_ofb() ?) if that's what it takes.

sfackler:
That's a linker error indicating that ssleay32 doesn't define that function. You can put it behind a feature gate like we do with some other features that aren't always present in OpenSSL.

chaaz:
Huh, doesn't seem to be in openssl linux (1.0.1f) either. I'm just going to take it out for now.

PR #385, opened 2016-04-29
Upgrade to work with bitflags 0.5 and 0.6

mbrubeck:


sfackler:
Could you upgrade the version of rust we build against on appveyor? Looks like it's too old for `#[no_std]` support: https://ci.appveyor.com/project/sfackler/rust-openssl/build/1.0.533

sfackler:
You may need to do the same on travis.

mbrubeck:
Updated Travis and Appveyor config to use Rust 1.8.0.

Ryman:
This has rather unfortunate effects for downstream crates that don't have their own version requirement for bitflags (i.e. they don't depend on it), as then cargo is happy to pull in the latest version which leads to the Rust 1.8 requirement noted above. 

It's kind of a shame to have such a major crate in the ecosystem relying on essentially the bleeding edge version of rust. For a concrete example, this is going to make anything depending on hyper now require Rust 1.8 if they want to use ssl.

mbrubeck:
As a workaround, downstream crates that aren't on Rust 1.8 can do `cargo update -p bitflags --precise 0.5.0` to downgrade bitflags.  (I'm not sure which Rust version is required for bitflags 0.5.0.)  I made the constraints as weak as possible for exactly this reason.

Unfortunately Cargo can't automatically choose the best version based on the Rust compiler version, unless something like rust-lang/rfcs#303 is implemented.

mbrubeck:
bitflags 0.5.0 is compatible with all versions of Rust ≥ 1.0, so all projects can still build with the latest openssl crate as long they pin bitflags to 0.5.0 in Cargo.lock.

PR #388, opened 2016-05-02
Remove unnecessary explicit lifetimes.

frewsxcv:


sfackler:
I like to keep the explicit lifetimes when they're annotations on a "normal" (i.e. non-`str` or `[T]`) type since you otherwise can't see from the signature that a borrow is taking place. The change to get_name seems good though.

frewsxcv:
Done

PR #389, opened 2016-05-03
expose rsa from raw private key and rsa sign and verify

cmsd2:
the private key constructor is useful for Json web keys which are not pem files but instead package up the public and private exponent bignums.

the sign and verify should be general purpose, but I've been testing it with a sha256 digest for use with JWS. I've tested it produces output compatible with rfc7515 appendix A2

I couldn't see the openssl NID values in openssl-sys so I created an enum in openssl, which doesn't feel right. Where's the best place to put this?


openssl/src/crypto/rsa.rs
```diff
@@ -7,6 +7,24 @@ use std::io::{self, Read};
...
 use bio::MemBio;
 
+#[derive(Copy, Clone, Debug)]
+pub enum PKCSHashType {
+    SHA256,
+    SHA384,
+    SHA512
+}
+
+/// https://github.com/openssl/openssl/blob/master/include/openssl/obj_mac.h#L2790
```
sfackler: These are NIDs, so you should be able to use `nid::Nid`.

cmsd2:
let me know if you have any more suggestions.

sfackler:
Thanks. It's a bit unfortunate to pass a raw Nid in for the hash_id but I'm working on some change to Nid that will make that a more reasonable.

PR #391, opened 2016-05-09
(maint) Use wget with appveyor (to enable msvc testing)

ajroetker:
This commit re-enables msvc testing by using wget with appveyor instead
of Start-FileDownload. This commit also fixes an issue where the version
of OpenSSL and Rust were specified twice in appveyor's config meaning
that it was possible to update one of the version for a release but not
the other.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,13 +5,14 @@ environment:
   matrix:
   - TARGET: i686-pc-windows-gnu
     BITS: 32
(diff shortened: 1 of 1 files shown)


sfackler:
What's the problem with Start-FileDownload?

ajroetker:
Looks like Start-FileDownload works on msvc again, the first time I tried (some weeks ago) this it was failing with "inappropriate arguments passed to Start-FileDownload". Thanks for taking a look.

PR #402, opened 2016-05-31
DSA bindings

bbatha:
Tests do not pass currently. The `DSA_*` functions are also probably the wrong methods to bind in the long term, the `PKey` module should be re-written to use the high level `EVP` api including `EVP_SIgnInit` and friends.Though the high level `EVP` api does not support all of the same functionality as the low-level crypto apis, so we will need to support some way to access the lower level api.


bbatha:
This is now ready for review. The mac issues look like a problem with test setup. I'll get on my mac this weekend and take a look.

bbatha:
@sfackler this is now ready to go.

openssl/src/crypto/dsa.rs
```diff
@@ -0,0 +1,351 @@
...
+            Ok(result == 1)
+        }
+    }
+
+    pub fn as_ptr(&self) -> *mut ffi::DSA {
+        self.0
+    }
+
+    // The following getters are unsafe, since BigNum::new_from_ffi fails upon null pointers
+    pub fn p(&self) -> Result<BigNum, SslError> {
```
sfackler: It seems like we can coalesce `p` and `has_p` by just having this return a `Result<Option<BigNum>, SslError>`. This mirrors the existing RSA API, though, so that should probably happen in a follow up on the breaks branch.

sfackler:
Looks good to me, thanks!

openssl/src/crypto/dsa.rs
bbatha: I'd be happy to make this change, should I cherry-pick this change to the breaks branch or are you keeping that rebased on master?

sfackler: It's up to date right now. Feel free to make it!

PR #408, opened 2016-06-13
Document most of BigNum

jonas-schievink:
Also adds a few examples and hides the `unchecked` module from the docs, since it appears empty anyways.


sfackler:
Thanks!

PR #410, opened 2016-06-14
Password callbacks

jonas-schievink:
Tests are missing (both manual and automated), but I wanted to get this ready for an initial review, since I'm really not sure how to implement these callbacks safely (but I think I like this pattern).


openssl/src/crypto/pkey.rs
```diff
@@ -93,6 +94,32 @@ impl PKey {
         }
     }
 
+    /// Read a private key from PEM, supplying a password callback to be invoked if the private key
+    /// is encrypted.
+    ///
+    /// The callback will be passed the password buffer and should return the number of characters
+    /// placed into the buffer.
+    pub fn private_key_from_pem_cb<R, F>(reader: &mut R, pass_cb: F) -> Result<PKey, SslError>
+        where R: Read, F: FnMut(&mut [i8]) -> usize
```
sfackler: c_char isn't always signed - maybe just use a `&mut [c_char]` instead here?

sfackler:
This looks pretty nice!

catch_unwind and resume_unwind weren't stabilized until Rust 1.9 so I think we might want to put these behind a Cargo features for now - I like lagging behind by one release to give people time to upgrade.

jonas-schievink:
Okay, I think everything is fine now.

openssl/src/crypto/rsa.rs
```diff
@@ -277,4 +303,22 @@ mod test {
 
         assert!(result);
     }
+
+    #[test]
+    pub fn test_password() {
```
sfackler: This will need to be flagged under the catch_unwind feature - I fixed the travis builds so if you rebase onto master it should be ok.

jonas-schievink: Done

openssl/src/crypto/pkey.rs
```diff
@@ -93,6 +98,33 @@ impl PKey {
         }
     }
 
+    /// Read a private key from PEM, supplying a password callback to be invoked if the private key
+    /// is encrypted.
+    ///
+    /// The callback will be passed the password buffer and should return the number of characters
+    /// placed into the buffer.
```
sfackler: The docs for these methods should mention that they require the `catch_unwind` feature.

openssl/src/crypto/util.rs
```diff
@@ -0,0 +1,58 @@
...
+                                      _rwflag: c_int,
+                                      cb_state: *mut c_void)
+                                      -> c_int
+                                      where F: FnMut(&mut [i8]) -> usize {
+    let result = panic::catch_unwind(|| {
+        // build a `i8` slice to pass to the user callback
+        let pass_slice = unsafe { slice::from_raw_parts_mut(buf, size as usize) };
+        let callback = unsafe { &mut *(cb_state as *mut CallbackState<F>) };
+
+        callback.cb.take().unwrap()(pass_slice)
```
sfackler: Doesn't seem like this option dance is necessary if we're only using `FnMut`.

sfackler:
Made two more comments, and I think we're good then.

openssl/src/crypto/util.rs
jonas-schievink: Aaah I think `FnMut` was easier to work with before I had `CallbackState`. I'll switch to `FnOnce`, since that seems like the better thing to use there.

jonas-schievink:
Great! Done.

jonas-schievink:
Ping @sfackler

sfackler:
Oops, thanks!

PR #412, opened 2016-06-16
Extend symmetric crypter API to allow for buffer reuse

koute:


sfackler:
The symm implementation has been rewritten in a way that allows buffer reuse.

PR #420, opened 2016-07-15
Signature Generation/Verification Support

posborne:
Previously, there was basic support for signing and verifying signatures on buffers in PKey.  This change exposes two new structs, `Signer` and `Verifier`, that allow for signatures to be more easily generated/verified when performing streaming operations.

For my initial use case, I am performing signature verification of the entire filesystem of an embedded device which cannot fit into RAM.

### DIFF
--- a/openssl/src/crypto/signature.rs
+++ b/openssl/src/crypto/signature.rs
@@ -107,15 +107,21 @@ impl<'a> Signer<'a> {
     /// Signer (e.g. `SHA512` will return a Vector with 64 bytes).
     #[inline]
     pub fn finish(&mut self) -> Result<Vec<u8>, SslError> {
-        let mut sigbuf = [0u8; 8 * 1024];
-        let mut siglen = sigbuf.len() as libc::size_t;
+        let mut siglen: libc::size_t = 0;
+        let mut sigbuf: Vec<u8>;
         unsafe {
+            // get siglen to determine buffer size we need
+            try_ssl_if!(1 != ffi::EVP_DigestSignFinal(self.ctx, ptr::null_mut(), &mut siglen));
+
+            // actually do the signature into the properly sized buffer
+            sigbuf = Vec::with_capacity(siglen);
+            sigbuf.set_len(siglen);
             try_ssl_if!(1 !=
                         ffi::EVP_DigestSignFinal(self.ctx,
-                                                 (&mut sigbuf[..]).as_mut_ptr(),
+                                                 sigbuf.as_mut_ptr(),
                                                  &mut siglen));
         }
-        Ok(Vec::from(&sigbuf[..siglen]))
+        Ok(sigbuf)
     }
 }
 
--- a/openssl/src/crypto/signature.rs
+++ b/openssl/src/crypto/signature.rs
@@ -1,0 +1,261 @@
+// Copyright (c) 2016, The rust-openssl developers.
+
+//! Interface for computing/verifying message signatures
+//!
+//! The `Signer` allows for the computation of crytographic signatures
+//! of data given a private key.  The corresponding public key can then
+//! be used to verify the integrity and authenticity of signed messages
+//! using the `Verifier`.
+//!
+//! # Example
+//!
+//! ```rust
+//! use openssl::crypto::signature::{Signer, Verifier};
+//! use openssl::crypto::pkey::PKey;
+//! use openssl::crypto::hash;
+//!
+//! // Generate public/private keypairs
+//! let mut keypair = PKey::new();
+//! let mut privkey = PKey::new();
+//! let mut pubkey = PKey::new();
+//! keypair.gen(1024);
+//! privkey.load_priv(&keypair.save_priv()[..]);
+//! pubkey.load_pub(&keypair.save_pub()[..]);
+//!
+//! // generate some data
+//! let data1: Vec<u8> = (0..25).cycle().take(1024).collect();
+//! let data2: Vec<u8> = (100..150).cycle().take(1024 * 5).collect();
+//!
+//! // sign some content with private key
+//! let mut signer = Signer::new(hash::Type::SHA512, &privkey);
+//! signer.update(&data1[..]);
+//! signer.update(&data2[..]);
+//! let signature = signer.sign();
+//!
+//! // verify content using signature and public key
+//! let mut verifier = Verifier::new(hash::Type::SHA512, &pubkey);
+//! verifier.update(&data1[..]);
+//! verifier.update(&data2[..]);
+//! assert!(verifier.verify(&signature[..]));
+//! ```
+
+use std::ptr;
+use libc;
+
+use ffi;
+use crypto::pkey::PKey;
+use crypto::hash;
+
+/// Signature Generation Context
+pub struct Signer<'a> {
+    ctx: *mut ffi::EVP_MD_CTX,
+    pkey: &'a PKey,
+}
+
+/// Signature Verification Context
+pub struct Verifier<'a> {
+    ctx: *mut ffi::EVP_MD_CTX,
+    pkey: &'a PKey,
+}
+
+macro_rules! openssl_assert {
+    ($expression:expr, $msg:expr) => (
+        if ! $expression {
+            let error = ::ffi::ERR_get_error();
+            let bytes = ::std::ffi::CStr::from_ptr(::ffi::ERR_error_string(error, ::std::ptr::null_mut())).to_bytes();
+            panic!("{}: {}", $msg, ::std::str::from_utf8(bytes).unwrap());
+        }
+    )
+}
+
+impl<'a> Signer<'a> {
+    /// Create and initialize a new Signer
+    ///
+    /// The digest type specified will determine the final form  of the
+    /// signature returned by `finalize`.
+    ///
+    /// # Panics
+    ///
+    /// If the provided pkey does not contain a private key.
+    pub fn new(digest_type: hash::Type, pkey: &'a PKey) -> Signer {
+        ffi::init();
+
+        let ctx = unsafe {
+            let r = ffi::EVP_MD_CTX_create();
+            openssl_assert!(!r.is_null(), "EVP_MD_CTX_create failed");
+            r
+        };
+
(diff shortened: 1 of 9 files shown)


posborne:
CC @nastevens

posborne:
After looking at other APIs a bit more I'm considering the following:

* Implement `Write` trait
* Expose `finish` instead of `sign/verify` to match API of `Hasher`

I would also be very open to additional test cases that would be worth adding.  I don't have extensive experience with OpenSSL's APIs.

posborne:
@sfackler @alexcrichton Feedback on OSX failure would be nice.  I tested on OSX but am using a brew linked version locally.  Should this be feature gated in some way or is there some other fix?

alexcrichton:
I think I've actually seen those symbols before and they're what I think are the canonical reason of why system openssl doesn't work but a homebrew openssl does (or at least one way or another). So yeah this looks like it may be a difference between homebrew and system openssl

posborne:
Ok, I placed the new signature code behind the `signing` feature flag and CI seems happy now and I have updated the API to be more similar to `Hasher`.  Other items I believe would be worth considering during review:

* I added a macro for printing out more useful panic messages local to the module.  Is there a better alternative to this?  Is this worth extracting out to error or some other place?
* Should any of the places I am panicking actually be worth returning a result type?  For instance, If a key is provided for signing/verifying does not contain the required private/public portion respectively.
* Implementing `Write` makes sense.  Would it also make sense to expose `update` in addition which makes it clear that the call cannot fail?

openssl/src/crypto/signature.rs
```diff
@@ -0,0 +1,291 @@
...
+//! // verify content using signature and public key
+//! let mut verifier = Verifier::new(hash::Type::SHA512, &pubkey);
+//! verifier.write_all(&data1[..]).unwrap();
+//! verifier.write_all(&data2[..]).unwrap();
+//! assert!(verifier.finish(&signature[..]));
+//! ```
+
+// pkey not used in structure but we need to ensure that the PKey
+// lives as long as the structure
+#![allow(dead_code)]
```
sfackler: Why?

openssl/src/crypto/signature.rs
```diff
@@ -0,0 +1,291 @@
...
+
+use std::ptr;
+use std::io::{self, Write};
+use libc;
+
+use ::ffi;
```
sfackler: The leading `::` isn't necessary - `use` paths are absolute by default.

openssl/src/crypto/signature.rs
```diff
@@ -0,0 +1,291 @@
...
+        }
+
+        signer
+    }
+
+    /// Feed bytes to be added to the signature calculation
+    #[inline]
+    fn update(&mut self, data: &[u8]) {
+        unsafe {
+            let r = ffi::EVP_DigestUpdate(self.ctx, data.as_ptr(), data.len() as u32);
```
sfackler: This downcast seems dangerous.

openssl/src/crypto/signature.rs
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    /// Finalize the signature calculation and return the signature
+    ///
+    /// The signature form will be determined by the hashing type for this
+    /// Signer (e.g. `SHA512` will return a Vector with 64 bytes).
+    #[inline]
+    pub fn finish(&mut self) -> Vec<u8> {
+        let mut sigbuf = [0u8; 8 * 1024];
+        let mut siglen = sigbuf.len() as libc::size_t;
```
sfackler: You don't need to explicitly cast to `size_t` - it's guaranteed to be `usize`.

sfackler:
I would personally drop all of the asserts in favor of an API that returns a `Result`.

The compilation errors you ran into are because OSX's default system OpenSSL is version 0.9.8, while this API was added in version 1.0.0.

openssl/src/crypto/signature.rs
posborne: > Why?

This has to do with my note about the unused `pkey` reference that is present to avoid a potential use-after-free bug.  This `ctx` has a pointer to that pkey -- as such, having the signer/validator live longer than the pkey would be a bug.

I am quite open to better options for ensuring that the pkey lives at least as long as the validator/signer.

openssl/src/crypto/signature.rs
posborne: > The leading :: isn't necessary - use paths are absolute by default.

Sure, will update.

openssl/src/crypto/signature.rs
posborne: I suppose we could map things into multiple underlying update calls, although I sure hope nobody is trying pass chunks longer than can be represented with 32 bits of length (~4GB) to update.  That is not generally a good way to do things (maybe OK with memory mapped IO?).

posborne:
> I would personally drop all of the asserts in favor of an API that returns a Result.

I will update to try to make this a reality and agree that it is probably the right way to handle things.  I wrote things this way as that is how things are handled by `Hasher` and `Crypter`.  It's quite possible that these should also be exposing result types (the exception possibly being in the case of allocation where matching `std` behavior seems OK).

posborne:
Ok, I've pushed a few more commits that do the follow now (and also squashed fixup commits, etc.):

* The PKey ref is now removed in favor of using a [PhantomData]() to express the lifetime constraint without taking any space.
* We now return Result types pervasively.
* Addressed concern related to usize -> u32 conversion at the write implementation (public interface) level.

posborne:
@sfackler Any open items on this one that you believe need to be addressed?

sfackler:
Sorry for missing this!

This looks good to go, but needs a rebase (the error type has changed).

openssl/src/crypto/signature.rs
```diff
@@ -0,0 +1,289 @@
...
+    /// Create and initialize a new Signer
+    ///
+    /// The digest type specified will determine the final form  of the
+    /// signature returned by `finalize`.
+    pub fn new(digest_type: hash::Type, pkey: &'a PKey) -> Result<Signer, SslError> {
+        ffi::init();
+
+        let ctx = unsafe {
+            let r = ffi::EVP_MD_CTX_create();
+            assert!(!r.is_null(), "EVP_MD_CTX_create failed");
```
sfackler: This should probably just use `try_ssl_null!`.

openssl/src/crypto/signature.rs
```diff
@@ -0,0 +1,289 @@
...
+            r
+        };
+
+        let md = digest_type.evp_md();
+
+        unsafe {
+            try_ssl_if!(1 !=
```
sfackler: Can be simplified to just `try_ssl!`.

openssl/src/crypto/signature.rs
```diff
@@ -0,0 +1,289 @@
...
+            ctx: ctx,
+            pkey: PhantomData,
+        })
+    }
+
+    /// Feed bytes to be added to the signature calculation
+    #[inline]
+    fn update(&mut self, data: &[u8]) -> Result<(), SslError> {
+        unsafe {
+            lift_ssl_if!(1 != ffi::EVP_DigestUpdate(self.ctx, data.as_ptr(), data.len() as u32))
```
sfackler: Same

openssl/src/crypto/signature.rs
```diff
@@ -0,0 +1,289 @@
...
+    ///
+    /// The signature form will be determined by the hashing type for this
+    /// Signer (e.g. `SHA512` will return a Vector with 64 bytes).
+    #[inline]
+    pub fn finish(&mut self) -> Result<Vec<u8>, SslError> {
+        let mut siglen: libc::size_t = 0;
+        let mut sigbuf: Vec<u8>;
+        unsafe {
+            // get siglen to determine buffer size we need
+            try_ssl_if!(1 != ffi::EVP_DigestSignFinal(self.ctx, ptr::null_mut(), &mut siglen));
```
sfackler: Same

openssl/src/crypto/signature.rs
```diff
@@ -0,0 +1,289 @@
...
+
+            // actually do the signature into the properly sized buffer
+            sigbuf = Vec::with_capacity(siglen);
+            sigbuf.set_len(siglen);
+            try_ssl_if!(1 !=
```
sfackler: ^

openssl/src/crypto/signature.rs
sfackler: This should loop over `data` in case it's longer than u32 bytes. Check out the hash module for a reference.

openssl/src/crypto/signature.rs
```diff
@@ -0,0 +1,289 @@
...
+
+        Ok(Signer {
+            ctx: ctx,
+            pkey: PhantomData,
+        })
+    }
+
+    /// Feed bytes to be added to the signature calculation
+    #[inline]
+    fn update(&mut self, data: &[u8]) -> Result<(), SslError> {
```
sfackler: I'd make this public so people don't have to mix IO errors and SSL errors.

openssl/src/crypto/signature.rs
posborne: > Can be simplified to just try_ssl!.

Unfortunately, it cannot with the current implementation of `try_ssl` (See https://github.com/sfackler/rust-openssl/blob/master/openssl/src/macros.rs#L24).  That implementation considers anything that is not a 0 success.  For these APIs, anything that is not 1 is an error -- that's an important difference; several of the calls do return negative numbers on error and the default `try_ssl` would consider these success.

openssl/src/crypto/signature.rs
posborne: > This should loop over data in case it's longer than u32 bytes. Check out the hash module for a reference.

Agreed, if I make this public.  Previously, this was dealt with at the public interface layer (`Write` impl).

openssl/src/crypto/signature.rs
sfackler: Ah right. It might make sense to update `try_ssl` to `$e != 1` instead of `$e == 0`.

sfackler:
Continued in #474.

PR #423, opened 2016-07-24
Set SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER flag

taheris:
Hi,

Following on from https://github.com/hyperium/hyper/issues/872, this PR sets the [`SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER`](https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_set_mode.html) flag so that `SSL_write()` retries are accepted at a different pointer location (if the contents of the buffer are still the same).

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -567,7 +567,7 @@ impl SslContext {
         let ctx = SslContext { ctx: ctx };
 
         // this is a bit dubious (?)
-        try!(ctx.set_mode(ffi::SSL_MODE_AUTO_RETRY));
+        try!(ctx.set_mode(ffi::SSL_MODE_AUTO_RETRY | ffi::SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER));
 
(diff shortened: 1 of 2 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -568,6 +568,7 @@ impl SslContext {
 
         // this is a bit dubious (?)
         try!(ctx.set_mode(ffi::SSL_MODE_AUTO_RETRY));
+        try!(ctx.set_mode(ffi::SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER));
```
sfackler: You can coalesce these into `try!(ctx.set_mode(ffi::SSL_MODE_AUTO_RETRY | ffi::SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER))`.

taheris: Quite right. PR updated.

sfackler:
Thanks!

PR #424, opened 2016-07-27
Add MidHandshakeSslStream

alexcrichton:
Allows recognizing when a stream is still in handshake mode and can gracefully
transition when ready. The blocking usage of the API should still be the same,
just helps nonblocking implementations!


alexcrichton:
This is intended to eventually go against the `breaks` branch, but right now it's against master. I ended up deleting some code instead of updating it if it looked like it was deleted on the `breaks` branch.

alexcrichton:
Closing to retarget against the `breaks` branch

PR #426, opened 2016-07-29
Add more support for non-blocking DTLS sockets.

rfw:
This introduces three new methods on SslStream:

- get_dtlsv1_timeout: Gets the next DTLS handshake timeout.
- handle_dtlsv1_timeout: Handles DTLS handshake timeout expiry.
- do_handshake: Performs the initial TLS/SSL handshake.

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -17,7 +17,10 @@ use std::any::Any;
 use libc::{c_uchar, c_uint};
 #[cfg(any(feature = "npn", feature = "alpn"))]
 use std::slice;
-use std::marker::PhantomData;
+#[cfg(any(feature = "dtlsv1", feature = "dtlsv1_2"))]
+use std::time;
+#[cfg(any(feature = "dtlsv1", feature = "dtlsv1_2"))]
+use libc::timeval;
 
 use ffi;
 use ffi_extras;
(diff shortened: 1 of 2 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -1434,6 +1438,53 @@ impl<S> SslStream<S> {
...
+    ///
+    /// `handle_dtlsv1_timeout` should be called when the timeout expires to
+    /// ask OpenSSL to retransmit.
+    #[cfg(any(feature = "dtlsv1", feature = "dtlsv1_2"))]
+    pub fn get_dtlsv1_timeout(&mut self) -> Option<time::Duration> {
+        let mut timeout: timeval = unsafe { mem::zeroed() };
+        let ret = unsafe { ffi::SSL_ctrl(self.ssl.ssl, ffi::DTLS_CTRL_GET_TIMEOUT, 0,
+                                         &mut timeout as *mut _ as *mut _) };
+        if ret > 0 {
+            Some(time::Duration::new(timeout.tv_sec as u64, timeout.tv_usec as u32))
```
sfackler: `Duration` stores seconds and nanoseconds, not seconds and microseconds.

openssl/src/ssl/mod.rs
```diff
@@ -1434,6 +1438,53 @@ impl<S> SslStream<S> {
...
+        } else {
+            Err(self.make_error(ret as i32))
+        }
+    }
+
+    /// Performs a TLS/SSL handshake.
+    ///
+    /// This is particularly useful with a nonblocking socket, where the error
+    /// value will identify if OpenSSL is waiting on read or write readiness.
+    pub fn do_handshake(&mut self) -> Result<(), Error> {
```
sfackler: Does this ever need to be called after the initial handshake completes?

rfw: No, but it can be called multiple times for nonblocking sockets – it can return WantRead/WantWrite until calling it again succeeds.

openssl/src/ssl/mod.rs
```diff
@@ -1434,6 +1438,56 @@ impl<S> SslStream<S> {
         }
     }
 
+    /// Gets the next DTLS handshake timeout.
+    ///
+    /// This is intended to be used with a nonblocking socket, where the value
+    /// can be used to set timeouts on reading and performing handshakes.
+    ///
+    /// `handle_dtlsv1_timeout` should be called when the timeout expires to
+    /// ask OpenSSL to retransmit.
```
sfackler: Can you note in the documentation that this method and the one below requires the dtls features?

openssl/src/ssl/mod.rs
sfackler: Right now that isn't strictly necessary since we rely on `SSL_read` and `SSL_write` being able to step through that internally. This will be changing in the next breaking release - `connect` and `accept` will return a `MidHandshakeSslStream<T>` if the socket's nonblocking that you can use to step through the handshake until it completes and you get a normal `SslStream` out of it.

Want to leave this method out? The others look good to go with one little addition to the docs.

rfw: Sounds good, I'll leave it out for now.

PR #427, opened 2016-07-29
Implement save_der for X509 and X509Req

onur:
Hi. I've been trying to write a letsencrypt client and I realized openssl crate doesn't support DER encoding for X509 and X509Req. I decided to implement this.

Test is a bit weak but save_der is working as intended.


sfackler:
Thanks!

PR #431, opened 2016-07-30
WIP: use libc types instead of rust native ones for better portability

bbatha:
I intend to expand this to fix most of the ffi signatures.


openssl/src/crypto/rsa.rs
```diff
@@ -162,23 +162,23 @@ impl RSA {
         }
     }
 
-    pub fn size(&self) -> Result<u32, SslError> {
+    pub fn size(&self) -> Result<isize, SslError> {
```
bbatha: This may make more sense to return `usize` as all uses of this method cast it to `usize` or `c_uint`.

bbatha:
Refiling against the breaks branch

PR #432, opened 2016-07-31
Add MidHandshakeSslStream

alexcrichton:
Allows recognizing when a stream is still in handshake mode and can gracefully
transition when ready. The blocking usage of the API should still be the same,
just helps nonblocking implementations!

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -5,6 +5,7 @@ use std::ffi::{CStr, CString};
 use std::fmt;
 use std::io;
 use std::io::prelude::*;
+use std::error as stderror;
 use std::mem;
 use std::str;
 use std::path::Path;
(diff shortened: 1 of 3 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -1137,6 +1135,66 @@ impl<S: Read + Write> SslStream<S> {
...
+    Failure(Error),
+    /// The handshake was interrupted midway through.
+    Interrupted(MidHandshakeSslStream<S>),
+}
+
+/// An SSL stream midway through the handshake process.
+#[derive(Debug)]
+pub struct MidHandshakeSslStream<S> {
+    stream: SslStream<S>,
+    is_accept: bool,
```
sfackler: I think we can call [SSL_do_handshake](https://www.openssl.org/docs/manmaster/ssl/SSL_do_handshake.html) after the initial call to accept/connect and avoid having to track this state.

openssl/src/ssl/mod.rs
```diff
@@ -1137,6 +1135,66 @@ impl<S: Read + Write> SslStream<S> {
     }
 }
 
+/// An error or intermediate state after a TLS handshake attempt.
+#[derive(Debug)]
+pub enum HandshakeError<S> {
```
sfackler: This should implement Error for ease of use in the blocking case.

alexcrichton:
Done!

sfackler:
https://travis-ci.org/sfackler/rust-openssl/jobs/148742545#L4032 I think these are legit failures

alexcrichton:
I think fixed? Did some of these errors exist ahead of time?

sfackler:
Some of them did, yeah. I'm fixing those.

PR #433, opened 2016-08-01
BigNum binary operators with different lifetimes.

tmiasko:
Modify BigNum binary operators so that references with different lifetimes can be used together.


sfackler:
Thanks!

Issue #435, opened 2016-08-03
X509_get_notAfter, X509_get_notBefore


dweinstein:
Looks like bindings for these haven't been added yet. Just curious if there's any particular reason why before I try to take a stab at it.

sfackler:
Go for it!

dweinstein:
So I started out thinking I could just go and add the missing functions to the openssl-sys/lib.rs but doesn't seem as easy as following what was done for e.g., `X509_get_subject_name`.

somewhat unfortunately:

```
± rabin2 -s /usr/local/Cellar/openssl/1.0.2h_1/lib/libssl.1.0.0.dylib | grep -i x509_get
vaddr=0x00037180 paddr=0x00037180 ord=1081 fwd=NONE sz=0 bind=LOCAL type=FUNC name=imp.X509_get_issuer_name
vaddr=0x00037186 paddr=0x00037186 ord=1082 fwd=NONE sz=0 bind=LOCAL type=FUNC name=imp.X509_get_pubkey
vaddr=0x0003718c paddr=0x0003718c ord=1083 fwd=NONE sz=0 bind=LOCAL type=FUNC name=imp.X509_get_signature_nid
vaddr=0x00037192 paddr=0x00037192 ord=1084 fwd=NONE sz=0 bind=LOCAL type=FUNC name=imp.X509_get_subject_name

```

looks like (at least on my mac) X509_get_notBefore/notAfter are defined as preprocessor macros:

```
± grep "X509_get_not"  /usr/local/Cellar/openssl/1.0.2h_1/include/openssl/x509.h
# define         X509_get_notBefore(x) ((x)->cert_info->validity->notBefore)
# define         X509_get_notAfter(x) ((x)->cert_info->validity->notAfter)
```

so... I'm pretty new to rust FFI stuff but won't those macros be useless to us in openssl-sys and we'll need to do the pointer manipulation ourselves since there's no export to use? if you can nudge me or confirm I'm in the right direction I appreciate it.

sfackler:
So right now we have some C code that we use to wrap all of those macros in C functions, but I want to start moving away from that where possible. I'd define those as Rust function in openssl-sys like this: https://github.com/sfackler/rust-openssl/blob/08e27f31ed851873f7684ac806b837e8cff4a28f/openssl-sys/src/lib.rs#L458

dweinstein:
[...]

PR #436, opened 2016-08-03
Add support for getting X509 notBefore, notAfter

dweinstein:
- [x] todo: tests

Related to https://github.com/sfackler/rust-openssl/issues/435


openssl/src/asn1/mod.rs
```diff
@@ -2,8 +2,12 @@ use libc::c_long;
 use std::ptr;
 
 use ffi;
+use ffi_extras;
 use error::ErrorStack;
 
+// use ssl::error::SslError;
+
+#[derive(Debug)]
```
dweinstein: this is temporary for debugging

openssl/src/asn1/mod.rs
```diff
@@ -18,6 +22,13 @@ impl Asn1Time {
         }
     }
 
+    pub fn unowned(handle: *mut ffi::ASN1_TIME) -> Asn1Time {
```
dweinstein: definitely would like some feedback on this...

openssl/src/asn1/mod.rs
```diff
@@ -2,8 +2,12 @@ use libc::c_long;
 use std::ptr;
 
 use ffi;
+use ffi_extras;
 use error::ErrorStack;
 
+// use ssl::error::SslError;
```
dweinstein: this was raising a build error for me atm:
```
/Users/user/rust/rust-openssl/openssl/src/asn1/mod.rs:8:5: 8:25 error: unresolved import `ssl::error::SslError`. There is no `SslError` in `ssl::error`. Did you mean to use `StdError`? [E0432]
/Users/user/rust/rust-openssl/openssl/src/asn1/mod.rs:8 use ssl::error::SslError;
                                                            ^~~~~~~~~~~~~~~~~~~~
/Users/user/rust/rust-openssl/openssl/src/asn1/mod.rs:8:5: 8:25 help: run `rustc --explain E0432` to see a detailed explanation

```

dweinstein: also didn't seem to be in use atm anyway...

openssl/src/asn1/mod.rs
sfackler: If we're going to add something like this we'll definitely need to add a lifetime paramter to `Asn1Time` to enforce that it can't be used after the structure that does own the `ASN1_TIME` has been destroyed. Take a look at the `x509` module - I overhauled that yesterday to have a `X509Ref<'a>` for unowned use cases and a `X509` type for owned use cases which derefs to `X509Ref<'static>`.

sfackler:
Needs a rebase

dweinstein:
@sfackler rebased, lemme know what you think.

sfackler:
Hey, sorry for the delay, I've been prepping then 0.8 release. Will take a look tonight!

openssl/src/x509/mod.rs
```diff
@@ -424,6 +424,22 @@ impl<'a> X509Ref<'a> {
         }
     }
 
+    /// Returns Issuer validity notAfter
+    pub fn not_after(&self) -> Asn1TimeRef {
```
dweinstein: I was thinking these should probably return a `Result`

openssl/src/x509/mod.rs
```diff
@@ -424,6 +424,22 @@ impl<'a> X509Ref<'a> {
...
+        unsafe {
+            let date = ffi_extras::X509_get_notAfter(self.handle());
+            Asn1TimeRef::new(date)
+        }
+    }
+
+    /// Returns Issuer validity notBefore
+    pub fn not_before(&self) -> Asn1TimeRef {
```
dweinstein: `Result` here as well?

dweinstein:
no worries! looks like you've been busy cleaning stuff up and I'll need to rebase and shuffle things around a bit to match w/ openssl-sys-extras going away. If you have any general thoughts on how `Asn1TimeRef` let me know.

openssl/src/x509/mod.rs
sfackler: The macro's load directly from the internal structures, so that's not necessary. It probably should return an `Option` though, unless notBefore and notAfter are required to be present.

openssl/src/asn1/mod.rs
```diff
@@ -25,15 +30,44 @@ impl Asn1Time {
...
     pub fn handle(&self) -> *mut ffi::ASN1_TIME {
         self.0
     }
 }
 
+impl<'a> fmt::Display for Asn1TimeRef<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mem_bio = try!(MemBio::new());
+        let as_str = unsafe {
+            ffi::ASN1_TIME_print(mem_bio.handle(), self.0);
```
sfackler: This should have a `try_ssl!` around it.

openssl/src/asn1/mod.rs
```diff
@@ -25,15 +30,44 @@ impl Asn1Time {
...
+        &self.0
+    }
+}
+
+/// A borrowed Asn1Time
+pub struct Asn1TimeRef<'a>(*mut ffi::ASN1_TIME, PhantomData<&'a ()>);
+
+impl<'a> Asn1TimeRef<'a> {
+    /// Creates a new `Asn1TimeRef` wrapping the provided handle.
+    pub unsafe fn new(handle: *mut ffi::ASN1_TIME) -> Asn1TimeRef<'a> {
```
sfackler: Can you rename this to `from_ptr`?

openssl/src/asn1/mod.rs
```diff
@@ -25,15 +30,44 @@ impl Asn1Time {
...
+        Asn1TimeRef(handle, PhantomData)
+    }
 
     /// Returns the raw handle
     pub fn handle(&self) -> *mut ffi::ASN1_TIME {
```
sfackler: And this to `as_ptr`?

openssl/src/x509/mod.rs
sfackler: It sounds like these are always present, so let's just have an assert that checks for null.

dweinstein:
@sfackler does anything using `c_helpers` need to be behind a feature flag for now? I did what I thought was obvious, to move the ASN1 shims I added earlier to `c_helpers.c`, but now I'm getting:

```
/Users/user/rust/./rust-openssl/openssl/src/x509/mod.rs:439:24: 439:63 error: failed to resolve. Maybe a missing `extern crate c_helpers`? [E0433]
/Users/user/rust/./rust-openssl/openssl/src/x509/mod.rs:439             let date = ::c_helpers::rust_0_8_X509_get_notAfter(self.0);
                                                                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/
```

sfackler:
Yep, they're all disabled by default so we don't require OpenSSL headers for the default feature set. Take a look here for reference for another feature that requires them:

https://github.com/sfackler/rust-openssl/blob/master/openssl/Cargo.toml#L35
https://github.com/sfackler/rust-openssl/blob/master/openssl/src/dh/mod.rs#L14-L24

dweinstein:
@sfackler ok test is passing and I believe I've addressed all your comments. Let me know what you think. Thanks!

openssl/src/x509/mod.rs
```diff
@@ -433,6 +433,26 @@ impl<'a> X509Ref<'a> {
         }
     }
 
+    /// Returns Issuer validity notAfter
```
sfackler: Can you mention in these docs that the `x509_expiry` feature is required?

sfackler:
You'll need to update https://github.com/sfackler/rust-openssl/blob/master/openssl/test/run.sh#L7 to include your new feature so it's tested on Travis.

Other than that and the doc thing, this looks good to go!

openssl/src/x509/mod.rs
```diff
@@ -433,6 +433,28 @@ impl<'a> X509Ref<'a> {
         }
     }
 
+    /// Returns certificate Not Before validity period.
+    /// Requires the `x509_expiry` feature.
+    #[cfg(feature = "x509_expiry")]
+    pub fn not_after(&self) -> Asn1TimeRef {
```
dweinstein: I wonder if these should be returning an `Asn1TimeRef` with a lifetime, like:

```diff
-    pub fn not_after(&self) -> Asn1TimeRef {
+    pub fn not_after<'b>(&'b self) -> Asn1TimeRef<'b> {
         unsafe {
             let date = ::c_helpers::rust_0_8_X509_get_notAfter(self.0);
             assert!(!date.is_null());
-            Asn1TimeRef::from_ptr(date)
+            Asn1TimeRef(date, PhantomData)
         }

```

sfackler: The lifetime is elided - what's actually happening is identical to that patch. I like the explicitness though - can you add the `'b` lifetime in as well? The constructor change isn't necessary.

sfackler:
Looks good to me - just have to wait on CI!

openssl/src/x509/mod.rs
```diff
@@ -10,7 +10,7 @@ use std::collections::HashMap;
 use std::marker::PhantomData;
 
 use HashTypeInternals;
-use asn1::Asn1Time;
+use asn1::{Asn1Time, Asn1TimeRef};
```
dweinstein: looks like `Asn1TimeRef` is causing a warning here in CI when we're not using the `x509_expiry` feature. would it be wise to wrap the `use` stmt here w/ a cfg thing? Not sure what would be idiomatic rust here or your preference.

sfackler:
Could you possibly rebase to pull in your CI fix from the other PR?

sfackler:
Thanks!

PR #437, opened 2016-08-05
Add more support for non-blocking DTLS sockets.

rfw:
This introduces two new methods on MidHandshakeSslStream:
- get_dtlsv1_timeout: Gets the next DTLS handshake timeout.
- handle_dtlsv1_timeout: Handles DTLS handshake timeout expiry.

This is against the breaks branch which contains MidHandshakeSslStream.

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -18,6 +18,10 @@ use libc::{c_uchar, c_uint};
 #[cfg(any(feature = "npn", feature = "alpn"))]
 use std::slice;
 use std::marker::PhantomData;
+#[cfg(any(feature = "dtlsv1", feature = "dtlsv1_2"))]
+use std::time;
+#[cfg(any(feature = "dtlsv1", feature = "dtlsv1_2"))]
+use libc::timeval;
 
 use ffi;
 use ffi_extras;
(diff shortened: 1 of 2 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -1223,6 +1227,44 @@ impl<S> MidHandshakeSslStream<S> {
...
+    /// This should be called when the timeout from `get_dtlsv1_timeout`
+    /// expires, to retransmit the handshake.
+    ///
+    /// This method requires either the `dtlsv1` or `dtlsv1_2` feature.
+    #[cfg(any(feature = "dtlsv1", feature = "dtlsv1_2"))]
+    pub fn handle_dtlsv1_timeout(&mut self) -> Result<(), Error> {
+        let ret = unsafe { ffi::SSL_ctrl(self.stream.ssl.ssl, ffi::DTLS_CTRL_HANDLE_TIMEOUT, 0,
+                                         ptr::null_mut()) };
+        if ret > 0 {
+            Ok(())
```
sfackler: It looks like the different positive return values may be meaningful? https://github.com/rbit/pydtls/blob/master/dtls/openssl.py#L719

openssl/src/ssl/mod.rs
```diff
@@ -1223,6 +1227,44 @@ impl<S> MidHandshakeSslStream<S> {
...
+        } else {
+            Err(self.stream.make_error(ret as i32))
```
sfackler: I think you'll probably just want to build an ErrorStack here.

openssl/src/ssl/mod.rs
rfw: Ick, I was reading off the boringssl headers for what the return values are: https://boringssl.googlesource.com/boringssl.git/+/master/include/openssl/ssl.h#523

Unfortunately, the documentation in the area is extremely sparse for actual OpenSSL, and that seems to be the most authoritative source on it. However, the OpenSSL library is pretty hard-and-fast on checking for > 0 rather than == 0 here, so I believe this is the correct behavior: https://github.com/openssl/openssl/blob/16938284cf5bba32006925281cb28b04d166a234/ssl/record/rec_layer_d1.c#L388

openssl/src/ssl/mod.rs
rfw: ... but I guess the above doesn't matter if I'm using wrap_ssl_result :)

sfackler: Hmm, reading the BoringSSL docks you linked above, I wonder if the `make_error` call was actually correct...

sfackler:
Is there any good way to test these? I'm not particularly familiar with DTLS unfortunately.

sfackler:
Ping

PR #439, opened 2016-08-14
Fix OS X instructions

CamJN:
These instructions are a better use of the cargo build system and don't require the user to remember to set env vars or introduce an incompatible openssl into the system library load paths.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,12 +31,16 @@ rust-openssl to a separate installation. OSX releases starting at 10.11, "El
 Capitan", no longer include OpenSSL headers which will prevent the `openssl`
 crate from compiling.
 
-For OSX 10.11 you can use brew to install OpenSSL and then set the environment variables
-as described below.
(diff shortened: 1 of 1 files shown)


sfackler:
Nice! That part of the README's a bit out of date, and I haven't gotten around to fixing it yet. It should probably say something along the lines of:

* You can build against the native OpenSSL for the default feature set of openssl 0.8 (headers no longer required).
* Native OpenSSL is 0.9.8, so if you need more modern features or features that require headers and you're on El Cap, install with Homebrew, and then either:
    * Do the build script override, or
    * Set the environment variables

I still want to mention the environment variables as an option because if you set them in your e.g. .bashrc things will "just work" for anything that uses openssl-sys.

We should also be sure to mention that the build script overrides goes in `.cargo/config`, not `Cargo.toml`.

sfackler:
The 0.9 release overhauls the library detection process, and 0.9.8 isn't supported any more.

PR #440, opened 2016-08-17
Improve build script

dweinstein:
- try and fallback to a mirror when openssl.org is down
- check the sha1 of the downloaded tarball


sfackler:
Thanks! openssl.org is weirdly unreliable.

dweinstein:
looks like `ssl::tests::test_read_nonblocking` test failure on appveyor is unrelated to the change in this PR, so should be safe otherwise.

sfackler:
Yeah that test's been flickering for a while.

PR #456, opened 2016-09-24
Improved perfomance of vector allocation

iamnotacake:
Set length to capacity instead of filling vector with zeros


sfackler:
Do you have any data comparing the performance before and after this change?

sfackler:
Ping

iamnotacake:
Pong. Sorry for so long silence. I performed _some_ tests and there was no difference. Maybe Rust optimizes the code. Or RAM I/O is very fast...

sfackler:
Cool! I'd prefer to avoid messing around with uninitialized buffers unless there's a concrete performance need.

Issue #460, opened 2016-10-04
Add private/public encrypt/decrypt to RSA


aoprisan:
I am in need to use RSA public key encryption and I saw that the functions are missing from rust-openssl while the openssl-sys has them already. Browsing through history I saw they were removed from PKey where they initially lived. Would it be a good idea to add them to RSA ? I can make a PR in the next couple of days as I have a fork already.

sfackler:
Please do! I must have forgotten to move the methods over. They didn't make sense on PKey since they didn't actually work with non-RSA things.

aoprisan:
Will do the next couple of days, thanks.

PR #461, opened 2016-10-05
Added public/private encrypt/decrypt functions to RSA

aoprisan:
This is about issue #460 . I have ported partially the code in the commit to PKey, including tests. I added the following functions https://github.com/thommay/rust-openssl/blob/11e3b1b56317ca1e24cfc1c0a3805123fa73bfb8/openssl/src/crypto/pkey.rs#L393, though I am not sure all are needed. 
I also added two macros to handle OpenSSL methods which return size values (-1 is the error code).

### DIFF
--- a/openssl/src/crypto/rsa.rs
+++ b/openssl/src/crypto/rsa.rs
@@ -11,6 +11,24 @@ use HashTypeInternals;
 use crypto::hash;
 use crypto::util::{CallbackState, invoke_passwd_cb};
 
+/// Type of encryption padding to use.
+#[derive(Copy, Clone)]
+pub enum EncryptionPadding {
+    NoPadding,
+    OAEP,
+    PKCS1v15
+}
+
+impl EncryptionPadding {
+    pub fn openssl_padding_code(&self) -> c_int {
+        match self {
+            &EncryptionPadding::NoPadding => ffi::RSA_NO_PADDING,
+            &EncryptionPadding::OAEP => ffi::RSA_PKCS1_OAEP_PADDING,
+            &EncryptionPadding::PKCS1v15 => ffi::RSA_PKCS1_PADDING
+        }
+    }
+}
+
 pub struct RSA(*mut ffi::RSA);
 
 impl Drop for RSA {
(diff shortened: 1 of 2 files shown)


bbatha:
Instead of the new macro you can use `try_ssl_if!(err_code == -1)`

openssl/src/crypto/rsa.rs
```diff
@@ -162,6 +180,125 @@ impl RSA {
         }
     }
 
+    pub fn max_data(&self) -> Option<u32> {
+        // 41 comes from RSA_public_encrypt(3) for OAEP
+        self.size().map(|len| len - 41)
```
sfackler: This was copied from the old impl, right? There's a (now closed) issue talking about how this is wrong for non-OAEP padding, so we shouldn't have the method unless we can implement it correctly.

openssl/src/crypto/rsa.rs
```diff
@@ -11,6 +11,24 @@ use HashTypeInternals;
...
+/// Type of encryption padding to use.
+#[derive(Copy, Clone)]
+pub enum EncryptionPadding {
+    NoPadding,
+    OAEP,
+    PKCS1v15
+}
+
+impl EncryptionPadding {
+    pub fn openssl_padding_code(&self) -> c_int {
```
sfackler: Can we make this private?

openssl/src/crypto/rsa.rs
```diff
@@ -162,6 +180,125 @@ impl RSA {
...
+
+           Ok(to)
+        }
+    }
+
+    /**
+      * Encrypts data with the public key, using OAEP padding, returning the encrypted data. The
+      * supplied data must not be larger than max_data().
+      */
+     pub fn encrypt(&self, s: &[u8]) -> Result<Vec<u8>, ErrorStack> { self.public_encrypt_with_padding(s, EncryptionPadding::OAEP) }
```
sfackler: I would just not have this and force people to specify their padding personally.

sfackler:
Thanks! Couple of little tweaks and this should be good to go.

openssl/src/crypto/rsa.rs
aoprisan: Sure, will do.

openssl/src/crypto/rsa.rs
aoprisan: Yep, I will remove it for now.

aoprisan: Fixing it would involve passing the encoding and taking that into account ?

aoprisan:
`try_ssl_if` returns Unit and in my case I need the returned value. Should I leave `lift_ssl_returns_size` or refactor `try_ssl_if` ?

openssl/src/crypto/rsa.rs
sfackler: Yep, I believe so.

sfackler:
I've been meaning to refactor the error conversion macros, don't worry about that for now.

sfackler:
Thanks!

PR #464, opened 2016-10-11
Add support for OpenSSL 1.1.0

### LINKED ISSUE #452: Doesn't support OpenSSL 1.1.0
If I'm correct, OpenSSL 1.1.0 doesn't include some methods relied upon by this crate, e.g. CRYPTO_set_locking_callback.

alexcrichton:
This commit is relatively major refactoring of the `openssl-sys` crate as well
as the `openssl` crate itself. The end goal here was to support OpenSSL 1.1.0,
and lots of other various tweaks happened along the way. The major new features
are:

* OpenSSL 1.1.0 is supported
* OpenSSL 0.9.8 is no longer supported (aka all OSX users by default)
* All FFI bindings are verified with the `ctest` crate (same way as the `libc`
  crate)
* CI matrixes are vastly expanded to include 32/64 of all platforms, more
  OpenSSL version coverage, as well as ARM coverage on Linux
* The `c_helpers` module is completely removed along with the `gcc` dependency.
* The `openssl-sys` build script was completely rewritten
  * Now uses `OPENSSL_DIR` to find the installation, not include/lib env vars.
  * Better error messages for mismatched versions.
  * Better error messages for failing to find OpenSSL on a platform (more can be
    done here)
  * Probing of OpenSSL build-time configuration to inform the API of the `*-sys`
    crate.
* Many Cargo features have been removed as they're now enabled by default.

As this is a breaking change to both the `openssl` and `openssl-sys` crates this
will necessitate a major version bump of both. There's still a few more API
questions remaining but let's hash that out on a PR!

Closes #452


openssl-sys/build.rs
```diff
@@ -1,86 +1,323 @@
...
-                vec!["ssl32", "eay32"]
-            }
-        } else {
-            vec!["ssl", "crypto"]
+    try_pkg_config();
+
+    // TODO: try to be more helpful in this message, e.g.:
+    //
+    // * if we're on linux, recommend apt-get something or other
+    // * if we're on Windows, link to a README explaining more
```
alexcrichton: I'd love to handle this specifically (compiling on Windows) before landing. Do you perhaps have suggestions about what to put in this error though?

openssl-sys/src/ossl110.rs
```diff
@@ -0,0 +1,142 @@
...
+pub const OPENSSL_CFLAGS: c_int = 1;
+pub const OPENSSL_BUILT_ON: c_int = 2;
+pub const OPENSSL_PLATFORM: c_int = 3;
+pub const OPENSSL_DIR: c_int = 4;
+
+pub const CRYPTO_EX_INDEX_SSL: c_int = 0;
+pub const CRYPTO_EX_INDEX_SSL_CTX: c_int = 1;
+
+pub fn init() {
+    // TODO: fill this in, probably need to at least do something?
```
alexcrichton: Oh I never did get around to filling this out. Do you know if OpenSSL 1.1.0 actually needs any initialization?

openssl/src/ssl/bio.rs
```diff
@@ -78,24 +66,16 @@ pub unsafe fn get_mut<'a, S: 'a>(bio: *mut BIO) -> &'a mut S {
...
 }
 
-#[cfg(feature = "nightly")]
 fn catch_unwind<F, T>(f: F) -> Result<T, Box<Any + Send>>
     where F: FnOnce() -> T
 {
     ::std::panic::catch_unwind(::std::panic::AssertUnwindSafe(f))
 }
 
-#[cfg(not(feature = "nightly"))]
```
alexcrichton: I took the liberty to remove this here

openssl/src/ssl/bio.rs
```diff
@@ -175,22 +155,128 @@ unsafe extern "C" fn ctrl<S: Write>(bio: *mut BIO,
...
+#[allow(bad_style)]
+mod compat {
+    use std::io::{Read, Write};
+
+    use libc::c_int;
+    use ffi;
+    pub use ffi::{BIO_set_init, BIO_set_flags, BIO_set_data, BIO_get_data};
+
+    pub unsafe fn BIO_set_num(_bio: *mut ffi::BIO, _num: c_int) {
+        // TODO: what should this do?
```
alexcrichton: I don't even know what the `num` field was previously, @sfackler do you know if it's important to fill this in?

openssl/src/ssl/mod.rs
```diff
@@ -73,74 +72,11 @@ bitflags! {
...
-                x if x == ffi::DTLSv1_2_method() => Some(SslMethod::Dtlsv1_2),
-                _ => None,
-            }
-        }
-    }
+    // TODO: support more methods
+    /// Support the TLS protocol
+    Tls,
+    /// Support DTLS protocol
+    Dtls,
```
alexcrichton: The size of this enum both before and after worries me. Do you have thoughts though of how to proceed here?

openssl-sys/src/ossl110.rs
jwilm: [`OPENSSL_init_ssl()`](https://www.openssl.org/docs/man1.1.0/ssl/OPENSSL_init_ssl.html) and [`OPENSSL_init_crypto()`](https://www.openssl.org/docs/man1.1.0/crypto/OPENSSL_init_crypto.html) both say

> As of version 1.1.0 OpenSSL will automatically allocate all resources that it needs so no explicit initialisation is required. Similarly it will also automatically deinitialise as required.

It seems like as long as the defaults are ok, then no explicit call is necessary. Those man pages list the defaults.

jwilm: Just kidding... now that I've looked more closely, this is actually just for 1.1.0.

alexcrichton: Oh yay! Sounds like we don't actually have to do anything in that case, thanks @jwilm!

openssl/src/crypto/rsa.rs
```diff
@@ -42,14 +42,14 @@ impl Drop for RSA {
 impl RSA {
     /// only useful for associating the key material directly with the key, it's safer to use
     /// the supplied load and save methods for DER formatted keys.
-    pub fn from_public_components(n: BigNum, e: BigNum) -> Result<RSA, ErrorStack> {
+    pub fn from_public_components(n: BigNum,
```
sfackler: Wait, `d` is the private exponent. I don't think it makes sense to pass it in here.

openssl/src/ssl/bio.rs
sfackler: I believe it's just a utility field that the BIO implementation can use to store stuff. If we're not using it it should be fine to ignore.

openssl/src/ssl/mod.rs
sfackler: Are the other methods no longer exposed in 1.1?

openssl/src/ssl/mod.rs
```diff
@@ -1412,3 +1343,107 @@ impl<'a> IntoSsl for &'a SslContext {
...
+    pub unsafe fn get_new_ssl_idx(f: ffi::CRYPTO_EX_free) -> c_int {
+        ffi::CRYPTO_get_ex_new_index(ffi::CRYPTO_EX_INDEX_SSL,
+                                     0,
+                                     ptr::null_mut(),
+                                     None,
+                                     None,
+                                     Some(f))
+    }
+
+    pub unsafe fn get_method(method: SslMethod) -> *const ffi::SSL_METHOD {
```
sfackler: Is this used?

openssl/src/ssl/tests/mod.rs
```diff
@@ -50,18 +51,28 @@ struct Server {
...
     fn spawn(args: &[&str], input: Option<Box<FnMut(ChildStdin) + Send>>) -> (Server, SocketAddr) {
+        static CERT: &'static [u8] = include_bytes!("../../../test/cert.pem");
+        static KEY: &'static [u8] = include_bytes!("../../../test/key.pem");
+
+        let mut me = env::current_exe().unwrap();
+        me.pop();
+        let cert = me.join("cert.pem");
+        let key = me.join("key.pem");
+
+        File::create(&cert).unwrap().write_all(CERT).unwrap();
```
sfackler: Could we stick these in a tmpdir?

openssl/src/ssl/tests/mod.rs
```diff
@@ -226,10 +237,10 @@ run_test!(new_sslstream, |method, stream| {
     SslStream::connect(&SslContext::new(method).unwrap(), stream).unwrap();
 });
 
-run_test!(get_ssl_method, |method, _| {
-    let ssl = Ssl::new(&SslContext::new(method).unwrap()).unwrap();
-    assert_eq!(ssl.ssl_method(), method);
-});
+// run_test!(get_ssl_method, |method, _| {
```
sfackler: Dead code?

systest/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+authors = ["Alex Crichton <alex@alexcrichton.com>"]
+build = "build.rs"
+
+[dependencies]
+libc = "0.2"
+openssl-sys = { path = "../openssl-sys" }
+
+[build-dependencies]
+ctest = { git = "https://github.com/alexcrichton/ctest" }
```
sfackler: It'd be kind of nice if this were published.

systest/build.rs
```diff
@@ -0,0 +1,111 @@
...
+        s == "PasswordCallback" ||
+            s == "bio_info_cb" ||
+            s.starts_with("CRYPTO_EX_")
+    });
+    cfg.skip_struct(|s| {
+        s == "ProbeResult"
+    });
+    cfg.skip_fn(move |s| {
+        s == "CRYPTO_memcmp" ||                 // uses volatile
+            s == "DH_new_from_params" ||        // not defined
```
sfackler: This should just be removed. I missed it by accident a while ago.

systest/build.rs
```diff
@@ -0,0 +1,111 @@
...
+            s == "SSL_get_ex_new_index" ||
+            s == "SSL_CTX_get_ex_new_index" ||
+            s == "CRYPTO_get_ex_new_index"
+        })
+    });
+    cfg.skip_field_type(|s, field| {
+        (s == "EVP_PKEY" && field == "pkey") ||      // union
+            (s == "GENERAL_NAME" && field == "d")    // union
+    });
+    cfg.skip_signededness(|s| {
```
sfackler: Why are we skipping signedness checks?

sfackler:
Worth noting that this also drops 1.0.0 support, but that also seems fine since it's no longer supported either.

jwilm:
@sfackler I haven't reviewed this PR in detail, but at the surface, it appears to support 1.0.x given that files like [openssl-sys/src/ossl10x.rs](https://github.com/sfackler/rust-openssl/pull/464/files#diff-e54f7c19f943634f9bee0771e488e7c1) are included. How do you figure otherwise?

sfackler:
I'm talking about 1.0.0 specifically, not 1.0.1 or 1.0.2. See here for the check: https://github.com/sfackler/rust-openssl/pull/464/files#diff-aca849e774a4e9d79fcc036a9de0d596R212

jwilm:
Ah, thanks for the clarification.

openssl-sys/build.rs
```diff
@@ -1,86 +1,323 @@
...
     };
+    f.read_to_string(&mut version_header).unwrap();
 
-    let mode = if env::var_os("OPENSSL_STATIC").is_some() {
-    	"static"
-    } else {
-    	"dylib"
+    // Do a bit of string parsing to find `#define OPENSSL_VERSION_TEXT "..."`
+    let version_line = version_header.lines().find(|l| {
+        l.contains("OPENSSL_VERSION_TEXT") && !l.contains("fips")
```
cuviper: There are multiple lines, e.g. I have:

```
# define OPENSSL_VERSION_NUMBER  0x1010002fL
# ifdef OPENSSL_FIPS
#  define OPENSSL_VERSION_TEXT    "OpenSSL 1.1.0b-fips  26 Sep 2016"
# else
#  define OPENSSL_VERSION_TEXT    "OpenSSL 1.1.0b  26 Sep 2016"
# endif
```

But it might be easier to work with `OPENSSL_VERSION_NUMBER`.

sfackler: Ah that makes sense. Yeah, working with VERSION_NUMBER does sound a bit more reliable.

openssl-sys/build.rs
sfackler: I actually have never had success getting a MinGW-installed OpenSSL to work - the compiler seems to have troubles finding linux-style libraries or something. The README currently points people to http://slproweb.com/products/Win32OpenSSL.html to obtain binaries.

openssl/src/ssl/mod.rs
alexcrichton: All of the `FOOvX` methods were deprecated, but still exist. I believe the SSLv2 method was outright deleted.

openssl-sys/Cargo.toml
```diff
@@ -10,38 +10,12 @@ documentation = "https://sfackler.github.io/rust-openssl/doc/v0.7.17/openssl_sys
...
-ecdh_auto = []
-hmac_clone = []
-
 [dependencies]
 libc = "0.2"
 
 [build-dependencies]
 pkg-config = "0.3"
 
-[target.le32-unknown-nacl.dependencies]
```
alexcrichton: For now, yeah, I figured this was just from a time long since passed.

openssl-sys/build.rs
```diff
@@ -1,86 +1,323 @@
...
-            for path in mingw_paths {
-                println!("cargo:rustc-link-search=native={}", path);
-            }
+        let homebrew = Path::new("/usr/local/opt/openssl@1.1");
+        if homebrew.exists() {
+            return homebrew.to_path_buf().into()
+        }
+
+        let system = Path::new("/usr/lib/libssl.0.9.8.dylib");
+        if system.exists() {
```
alexcrichton: Hm good point, we should probably do that. `pkg-config` is somewhat nonstandard on OSX I think, but no harm in running it.

openssl-sys/build.rs
```diff
@@ -1,86 +1,323 @@
...
+
+    let lib = match pkg_config::find_library("openssl") {
+        Ok(lib) => lib,
+        Err(_) => return,
+    };
+
+    if lib.include_paths.len() == 0 {
+        panic!("
+
+Used pkg-config to discover the OpenSSL installation, but pkg-config did not
```
alexcrichton: In theory `PKG_CONFIG_ALLOW_SYSTEM_CFLAGS` set above will force `pkg-config` to print all those out, but if that doesn't work then yeah this'll be a weird failure.

openssl/src/ssl/mod.rs
sfackler: Cool, that makes sense. We should make sure to add the other `SSL_OP_NO_*` constants and have people configure protocol selection through those.

openssl-sys/build.rs
sfackler: The picture I have in my head is that the OSX codepath behaves just like the Linux codepath except that it first checks for a Homebrew install.

openssl-sys/src/lib.rs
```diff
@@ -374,30 +154,21 @@ pub const SSL_VERIFY_NONE: c_int = 0;
...
-pub const SSL_OP_NO_COMPRESSION: c_long =                           0x00020000;
-pub const SSL_OP_ALLOW_UNSAFE_LEGACY_RENEGOTIATION: c_long =        0x00040000;
-pub const SSL_OP_SINGLE_ECDH_USE: c_long =                          0x00080000;
-pub const SSL_OP_SINGLE_DH_USE: c_long =                            0x00100000;
-pub const SSL_OP_CIPHER_SERVER_PREFERENCE: c_long =                 0x00400000;
-pub const SSL_OP_TLS_ROLLBACK_BUG: c_long =                         0x00800000;
-pub const SSL_OP_NO_SSLv2: c_long =                                 0x01000000;
-pub const SSL_OP_NO_SSLv3: c_long =                                 0x02000000;
-pub const SSL_OP_NO_TLSv1: c_long =                                 0x04000000;
+#[cfg(not(ossl101))]
```
alexcrichton: I was thinking that for a particular OpenSSL version this gives you what *can* be the API of that version, although it'd ideally be trimmed to the precise API of what's exactly on the system.

In that sense I'm definitely in favor of conditionally compiling away all the OpenSSL 1.1.0 APIs when compiling for OpenSSL 1.0.1, for example, to weed out errors sooner.

(also this is required to pass systest)

openssl-sys/src/lib.rs
```diff
@@ -1098,14 +719,30 @@ extern "C" {
                         -> c_int;
     pub fn PKCS12_free(p12: *mut PKCS12);
 
-    pub fn sk_free(st: *mut _STACK);
-    pub fn sk_pop_free(st: *mut _STACK, free: Option<unsafe extern "C" fn (*mut c_void)>);
-    pub fn sk_pop(st: *mut _STACK) -> *mut c_char;
+    #[cfg_attr(any(ossl101, ossl102), link_name = "sk_free")]
```
alexcrichton: How so? It's a bit of a weird attribute but allows us to work with OpenSSL 1.1.0 idioms while still working against older versions.

openssl-sys/src/ossl10x.rs
```diff
@@ -0,0 +1,562 @@
...
+    pub seedlen: c_int,
+    pub counter: *mut ::BIGNUM,
+    pub references: c_int,
+    pub ex_data: ::CRYPTO_EX_DATA,
+    pub meth: *const ::DH_METHOD,
+    pub engine: *mut ::ENGINE,
+}
+
+#[repr(C)]
+pub struct X509 {
```
alexcrichton: These were needed for defining various functions here and there, and they're all validated as being correct. It may be possible to configure OpenSSL in a number of configurations perhaps to test, but the definitions here were all at least copied from the actual header file.

openssl-sys/src/ossl110.rs
```diff
@@ -0,0 +1,142 @@
...
+pub enum EVP_MD_CTX {}
+pub enum EVP_CIPHER {}
+pub enum HMAC_CTX {}
+pub enum BIGNUM {}
+pub enum OPENSSL_STACK {}
+pub enum DH {}
+pub enum X509 {}
+pub enum SSL_CTX {}
+
+pub const SSL_OP_MICROSOFT_SESS_ID_BUG: c_ulong =                   0x00000000;
```
alexcrichton: Apparently, yes!

openssl/src/bn/mod.rs
```diff
@@ -185,53 +185,54 @@ impl<'a> BigNumRef<'a> {
...
+            if ffi::BN_add_word(self.as_ptr(), w as ffi::BN_ULONG) == 1 {
                 Ok(())
             } else {
                 Err(ErrorStack::get())
             }
         }
     }
 
-    pub fn sub_word(&mut self, w: c_ulong) -> Result<(), ErrorStack> {
+    pub fn sub_word(&mut self, w: u32) -> Result<(), ErrorStack> {
```
alexcrichton: This was actually an incorrect definition of the `BN_ULONG` type alias. (evidenced by the changes in the sys crate above). The definition of this is `c_uint` on Windows which is only 32-bits.

I figured it's safe at least to consume the *smallest* bit width used across platforms and then we'd return the largest bit width any platform returns. In that sense there shouldn't be a downcast?

openssl/src/crypto/hash.rs
```diff
@@ -123,22 +121,20 @@ impl Hasher {
...
-        while !data.is_empty() {
-            let len = cmp::min(data.len(), c_uint::max_value() as usize);
-            unsafe {
-                try_ssl!(ffi::EVP_DigestUpdate(self.ctx, data.as_ptr(), len as c_uint));
-            }
-            data = &data[len..];
+        unsafe {
+            try_ssl!(ffi::EVP_DigestUpdate(self.ctx,
+                                           data.as_ptr() as *mut _,
+                                           data.len()));
```
alexcrichton: I believe so, yeah. The systest crate turned this up at least and the actual definition uses `size_t` on all versions.

openssl/src/crypto/hmac.rs
```diff
@@ -66,7 +64,7 @@ use self::State::*;
 /// assert_eq!(res, spec);
 /// ```
 pub struct HMAC {
```
alexcrichton: Due to the size of this, perhaps that could be done in a follow-up PR?

openssl/src/crypto/hmac.rs
```diff
@@ -202,6 +190,73 @@ pub fn hmac(t: Type, key: &[u8], data: &[u8]) -> Result<Vec<u8>, ErrorStack> {
...
+
+#[cfg(ossl10x)]
+#[allow(bad_style)]
+mod compat {
+    use std::mem;
+    use std::cell::UnsafeCell;
+
+    use ffi;
+
+    pub struct HMAC_CTX {
```
alexcrichton: I figured it's good hygiene to return `*mut T` from a `&self`, but not required other than that.

openssl/src/crypto/rsa.rs
alexcrichton: Oh I have no idea what it is, we just needed to pass something to the API on 1.1.0. I guess that should be null?

openssl/src/ssl/mod.rs
alexcrichton: In one location yeah to do this conversion when calling `SSL_CTX_new`

openssl/src/crypto/hmac.rs
sfackler: Yeah, seems fine.

openssl/src/ssl/tests/mod.rs
alexcrichton: Oh I left this in as I figure we'd want to sort out what to do with the `ssl_method` method, which this was testing.

openssl/src/crypto/rsa.rs
sfackler: That's what I'd guess?

systest/build.rs
alexcrichton: Oh this is either a bug or a feature in ctest depending on how you look at it. It tries to test the signedness of all `type foo = bar` items in Rust, but these types skipped are all function pointers, which have no signedness.

openssl-sys/build.rs
alexcrichton: Hm the MinGW packages seem to be working on AppVeyor at least? I suppose I can otherwise just point to the README on Windows.

openssl-sys/src/lib.rs
sfackler: It doesn't look like the usage of these function's can't be unified due to the different representation of `_STACK`, so it doesn't seem like munging the name of the function on 1.0.x buys us anything other than confusing people searching for the 1.0.x name.

openssl-sys/build.rs
sfackler: Oh great, if that works it seems preferable.

openssl/src/ssl/tests/mod.rs
sfackler: Ah right. Let's leave ssl_method out entirely for now and see if anyone asks for it back. The implementation always seemed a bit sketchy to me with the unwrap.

openssl/src/ssl/mod.rs
sfackler: Oh right

openssl/src/ssl/tests/mod.rs
alexcrichton: You ok picking that up as a dep? I figured this was as good a place as any without picking up new dependencies.

openssl/src/crypto/hmac.rs
sfackler: Oh right, the struct is stored inline rather than being allocated by OpenSSL.

All of this stuff'll be deleted when we switch over to the PKey impl so it doesn't really matter.

openssl/src/ssl/tests/mod.rs
sfackler: It's just a dev dependency. w/e

openssl-sys/build.rs
```diff
@@ -64,46 +64,69 @@ fn find_openssl_dir(target: &str) -> OsString {
...
 
-Failed to find an approprate OpenSSL installation. It looks like you're
-compiling on macOS, where the system contains a version of OpenSSL 0.9.8. This
-crate no longer supports OpenSSL 0.9.8.
+It looks like you're compiling on macOS, where the system contains a version of
+OpenSSL 0.9.8. This crate no longer supports OpenSSL 0.9.8.
 
 As a consumer of this crate, you can fix this error by using Homebrew to
 install the `openssl` package, or as a maintainer you can use the openssl-sys
 0.7 crate for support with OpenSSL 0.9.8.
```
sfackler: You'd actually want to use openssl-sys 0.8, since 0.7 required headers.

alexcrichton:
Ok, lots of commits addressing comments pushed. I'm curious though, could you elaborate on this point?

> Cool, that makes sense. We should make sure to add the other SSL_OP_NO_* constants and have people configure protocol selection through those.

openssl-sys/build.rs
sfackler: Oh right derp, was thing of openssl not openssl-sys.

openssl-sys/build.rs
alexcrichton: 👍

openssl-sys/build.rs
alexcrichton: The "wrong version found" message is now a bit worse, but the version detection is likely more robust.

openssl-sys/src/lib.rs
alexcrichton: Ok, I can conditionally compile it

sfackler:
I'm talking about these constants: https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/src/lib.rs#L398-L400.

OpenSSL also defines ones to turn of TLSv1.1 and TLSv1.2, and some DTLS variants. We'd want to add those as well so people can e.g. only support TLSv1.2

sfackler:
Oh right, they're sitting just below commented out: https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/src/lib.rs#L402

sfackler:
(The'll also need to then be added to the bitflags thing in the ssl module).

sfackler:
Seems like some cfgs are missing based off of travis.

alexcrichton:
Ok adding cfgs to try to fix tests, updated README, addressed more comments, still missing SSL_OP business.

If you know what to do there it may be best if you just send a commit for that?

alexcrichton:
@sfackler I had to add [this commit](https://github.com/sfackler/rust-openssl/commit/989b28012a912b8175e9a142320a087e4b5697c4) to prevent segfaults in the `test_alpn_server_select_none` test which is failing like:

```
failures:

---- ssl::tests::test_alpn_server_select_none stdout ----

	thread 'ssl::tests::test_alpn_server_select_none' panicked at 'bad reason: 336151648', openssl/src/error.rs:142

failures:

    ssl::tests::test_alpn_server_select_none
```

Any ideas?

sfackler:
Not sure what'd be going on there. It could be that error handling is different in 1.1

sfackler:
Opened a PR to add the remaining constants.

openssl-sys/src/lib.rs
```diff
@@ -264,35 +73,10 @@ pub struct X509V3_CTX {
...
-    pub dmax: c_int,
-    pub neg: c_int,
-    pub flags: c_int,
-}
-
-impl Copy for BIGNUM {}
-impl Clone for BIGNUM {
-    fn clone(&self) -> BIGNUM { *self }
-}
+#[cfg(target_pointer_width = "64")]
```
sfackler: The conditions in openssl's definition seem different/more complex. Are these equivalent? https://github.com/openssl/openssl/blob/6f0ac0e2f27d9240516edb9a23b7863e7ad02898/crypto/bn/asm/x86_64-gcc.c#L67

alexcrichton: Yeah this wasn't 100% the same from what I could tell but I figured it was close enough in terms of ctest should turn up any problems.

openssl/Cargo.toml
```diff
@@ -12,40 +12,33 @@ build = "build.rs"
...
-aes_xts = ["openssl-sys/aes_xts"]
-aes_ctr = ["openssl-sys/aes_ctr"]
-npn = ["openssl-sys/npn"]
-alpn = ["openssl-sys/alpn"]
-rfc5114 = ["openssl-sys/rfc5114"]
-ecdh_auto = ["openssl-sys/ecdh_auto"]
-pkcs5_pbkdf2_hmac = ["openssl-sys/pkcs5_pbkdf2_hmac"]
-hmac_clone = ["openssl-sys/hmac_clone"]
+aes_xts = []
+aes_ctr = []
```
sfackler: These were added in 1.0.2 as well?

alexcrichton:
Ok I pushed one last commit to just ignore the failing test on OpenSSL 1.1.0, I may not have enough time to track it down right now :(

I think that should be everything though?

openssl/Cargo.toml
```diff
@@ -12,40 +12,33 @@ build = "build.rs"
...
-x509_clone = ["c_helpers"]
-x509_generator_request = ["c_helpers"]
-x509_expiry = ["c_helpers"]
-ssl_context_clone = ["c_helpers"]
-hmac = ["c_helpers"]
-dh_from_params = ["c_helpers"]
+# Added in OpenSSL 1.0.2
+rfc5114 = []
+
+# TODO: what to do about these features?
```
sfackler: We can delete these I believe.

.travis.yml
```diff
@@ -1,28 +1,86 @@
...
+  - curl https://static.rust-lang.org/rustup.sh |
+    sh -s -- --add-target=$TARGET --disable-sudo -y --prefix=`rustc --print sysroot`
 script:
-- ./openssl/test/run.sh
+  - ./openssl/test/run.sh
+
+cache:
+  cargo: true
+  directories:
+    - $HOME/openssl
```
alexcrichton: Oh this was actually just caching the OpenSSL we compiled. Figured it'd make the builds a bit faster and less noisy to avoid always doing that.

sfackler: Oh right.

sfackler:
The ALPN failures are somewhat concerning but it seems okay to land this and figure those out later.

sfackler:
I've started a project to track other work that needs to happen before we can release: https://github.com/sfackler/rust-openssl/projects/1

sfackler:
I've added you as a collaborator on the repo so feel free to add stuff to that as needed.

sfackler:
Everything's working now except for DTLS tests, which are failing on Windows and ARM. Not sure what's going on there.

sfackler:
DTLS tests have never been run on Windows, so I'm not particularly worried about them. Not sure what's going on with the ARM failures though.

alexcrichton:
Looks like ARM failures are all related to [this timeout](https://github.com/sfackler/rust-openssl/blob/master/openssl/src/ssl/tests/mod.rs#L121). I increased that to 1s and the tests passed.

Perhaps we could test the client and server code at the same time? Is there a critical reason we *must* test against `openssl` itself?

sfackler:
Ah, the timeout thing makes sense!

I don't think there's any intrinsic reason to use `openssl` - it's probably that way because the tests were written before the server side logic existed.

sfackler:
Woo, merging!

Issue #459, opened 2016-10-02
Need a cfg to support OPENSSL_NO_COMP builds


Arnavion:
`SSL_get_current_compression` and `SSL_COMP_get_name` are not defined in OPENSSL_NO_COMP builds.

The full list is:

```
ssleay:
SSL_get_current_compression
SSL_get_current_expansion
SSL_COMP_add_compression_method
SSL_COMP_free_compression_methods
SSL_COMP_free_compress_methods
SSL_COMP_get_compress_methods
SSL_COMP_get_compression_methods
SSL_COMP_get_name
SSL_COMP_set0_compress_methods
SSL_COMP_set0_compression_methods

libeay:
BIO_f_zlib
COMP_compress_block
COMP_CTX_free
COMP_CTX_new
COMP_expand_block
COMP_rle
COMP_zlib
COMP_zlib_cleanup
ERR_load_COMP_strings
```

but only those two are currently defined in openssl-sys/src/lib.rs

sfackler:
Thanks for the report! We can stick this behind a feature flag, but we'll have to enable it by default for backwards compatibility which is a bit unfortunate.

sfackler:
I'm a bit confused here, actually. It appears that OpenSSL defines these methods even if `OPENSSL_NO_COMP` is passed, they just return `NULL`: https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/ssl/ssl_ciph.c#L1934.

What compilation error are you running into?

sfackler:
It appears that these are always defined, at least in 1.0.1 and up, which is all the next release is going to support. Feel free to reopen if that's not the case!

Arnavion:
The .def file is generated based on configuration options. If it's configured with `no-comp` the functions will not be exported.

sfackler:
Ah, so it's specifically a Windows issue?

Arnavion:
I'm on Windows yes. I'm using the openssl build we build as part of [gtk-win32](https://github.com/hexchat/gtk-win32) where we configure with no-comp.

Arnavion:
(You can generate a list like the one in the OP for any arbitrary feature-set by looking at `util/libeay.num` and `util/ssleay.num`. I believe `Configure` should generate these even on Linux. The format is described in `util/mkdef.pl`.)

PR #470, opened 2016-10-15
Support OpenSSL conditional compilation

sfackler:
I'm a bit conflicted as to how to handle the OPENSSL_NO_CONF case. There are 3 options I see:

* We remove the `Ssl::compression` method automatically when `OPENSSL_NO_CONF` is detected.
* We have people opt-in to compression with a Cargo feature.
* We have people opt-out of compression with a Cargo feature.

I went with option 3 for now. It's pretty rare to disable compression, and it's a breaking change if we discover that we've missed some other conditional API if we go the opt-in route. On the other hand, opt-out features aren't composable in the way that Cargo would like features to be. Option 1 might be the best approach, but it may be unclear as to why some methods aren't available.

We do have people opt-in to features from newer OpenSSL versions via Cargo features, for reference.

Thoughts, @alexcrichton?

### DIFF
--- a/openssl-sys/build.rs
+++ b/openssl-sys/build.rs
@@ -283,6 +283,7 @@ The build is now aborting due to this version mismatch.
 
         // Look for `#define OPENSSL_FOO`, print out everything as our own
         // #[cfg] flag.
+        let mut vars = vec![];
         for line in conf_header.lines() {
             let i = match line.find("define ") {
                 Some(i) => i,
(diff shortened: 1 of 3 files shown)


alexcrichton:
My personal preference is to detect whether OpenSSL has this method, an if it does we call it and if it doesn't we return a runtime error. That is, I find runtime errors with proper error handling much easier to deal with than vast matrixes of compile time options and how they interact.

sfackler:
This appears that it's not actually necessary maybe? https://github.com/sfackler/rust-openssl/issues/459#issuecomment-253959124

sfackler:
Turns out it's not necessary! Backing the compression stuff out.

alexcrichton:
👍

PR #474, opened 2016-10-15
Signature/Digest API

sfackler:
Also drop the HMAC module in favor of OpenSSL's HMAC `PKey` support.


posborne:
Thanks for picking this back up @sfackler.  I had been looking at taking another shot at this with the recent 1.10 changes being merged in (and other recent activity) and just noticed this.

I'll try to provide a code review and perform some testing on the project I have (not public) that was using the new APIs to verify that this works in that context.

CC @nastevens

openssl/src/crypto/sign.rs
```diff
@@ -0,0 +1,274 @@
...
+            let mut buf = vec![0; len];
+            try_ssl_if!(ffi::EVP_DigestSignFinal(self.0, buf.as_mut_ptr() as *mut _, &mut len)
+                    != 1);
+            Ok(buf)
+        }
+    }
+}
+
+impl<'a> Write for Signer<'a> {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
```
posborne: As discussed in #420 this (and `update`) would have problems if a write of a buffer greater than `u32::max_value()` bytes in size are attempted.  The [ostrich algorithm](https://en.wikipedia.org/wiki/Ostrich_algorithm) might be fine for this case, but making a comment as it is one difference with the implementation from the previous PR.

openssl/src/crypto/sign.rs
```diff
@@ -0,0 +1,274 @@
+use ffi;
```
posborne: Might be nice to consider adding back something like the examples in the docs from #420: https://github.com/sfackler/rust-openssl/pull/420/files#diff-35c3c6c147bd45b93218a395c7901036R3

sfackler:
Cool! I believe the API here should be basically identical to the one in your PR, but let me know if something's busted.

openssl/src/crypto/sign.rs
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    pub fn update(&mut self, buf: &[u8]) -> Result<(), ErrorStack> {
+        unsafe {
+            try_ssl_if!(ffi::EVP_DigestUpdate(self.0, buf.as_ptr() as *const _, buf.len()) != 1);
+            Ok(())
+        }
+    }
+
+    pub fn finish(&self, signature: &[u8]) -> Result<(), ErrorStack> {
```
posborne: This API does not make sense.  When you are verifying, you need to be able to get the actual verification result.  That is, the return type should be `... -> Result<bool, ErrorStack>` similar to #420 (https://github.com/sfackler/rust-openssl/pull/420/files#diff-35c3c6c147bd45b93218a395c7901036R190).

That is, it needs to be possible to differentiate a true error calculating the signature from the signature failing (probably because either the signature was bad or the contents are corrupt/tampered with).

openssl-sys/src/lib.rs
```diff
@@ -489,6 +491,23 @@ extern {
     pub fn EVP_DigestFinal(ctx: *mut EVP_MD_CTX, res: *mut u8, n: *mut u32) -> c_int;
     pub fn EVP_DigestFinal_ex(ctx: *mut EVP_MD_CTX, res: *mut u8, n: *mut u32) -> c_int;
 
+    pub fn EVP_DigestSignInit(ctx: *mut EVP_MD_CTX,
```
sfackler: Yep, these APIs were added in 1.0.0 so we can include them unconditionally.

openssl/src/crypto/sign.rs
sfackler: That issue no longer applies - the signature of `EVP_DigestUpdate` was previously wrong - the size argument is a size_t not an int.

openssl/src/crypto/sign.rs
sfackler: However, I have had a hard time actually getting OpenSSL to produce a "real" error. Passing in an empty signature, for example, still produces a return code of 0.

sfackler: The OpenSSL docs imply that the `Result<(), ErrorStack>` form may be preferrable?

> EVP_DigestVerifyFinal() returns 1 for success; any other value indicates failure. A return value of zero indicates that the signature did not verify successfully (that is, tbs did not match the original data or the signature had an invalid form), while other values indicate a more serious error (and sometimes also indicate an invalid signature form).

openssl/src/crypto/sign.rs
sfackler: Added!

openssl/src/crypto/sign.rs
posborne: Ah, there we go!

PR #478, opened 2016-10-18
Implement new feature setup

sfackler:
The basic idea here is that there is a feature for each supported
OpenSSL version. Enabling multiple features represents support for
multiple OpenSSL versions, but it's then up to you to check which
version you link against (probably by depending on openssl-sys and
making a build script similar to what openssl does).

@alexcrichton this is basically the structure you proposed for scenarios.


alexcrichton:
👍

PR #506, opened 2016-10-31
Implemented a generic Stack API and use it to deal with StackOf(X509) and StackOf(GENERAL_NAME)

simias:
I've only tested with OpenSSL 1.0.x but I've added the code for 1.1.0.

### DIFF
--- a/openssl/src/stack.rs
+++ b/openssl/src/stack.rs
@@ -1,0 +1,319 @@
+use std::ops::{Deref, DerefMut, Index, IndexMut};
+use std::iter;
+use std::borrow::Borrow;
+use std::convert::AsRef;
+
+#[cfg(ossl110)]
+use libc::c_int;
+
+use ffi;
+use types::{OpenSslType, Ref};
+
+/// Trait implemented by stackable types. This must *only* be
+/// implemented on opaque types that can be directly casted into their
+/// `CType`.
+pub trait Stackable: OpenSslType {
+    /// C stack type for this element. Generally called
+    /// `stack_st_{ELEMENT_TYPE}`, normally hidden by the
+    /// `STACK_OF(ELEMENT_TYPE)` macro in the OpenSSL API.
+    type StackType;
+}
+
+/// An owned stack of `T`.
+pub struct Stack<T: Stackable>(*mut T::StackType);
+
+impl<T: Stackable> Stack<T> {
+    /// Return a new Stack<T>, taking ownership of the handle
+    pub unsafe fn from_ptr(stack: *mut T::StackType) -> Stack<T> {
+        Stack(stack)
+    }
+}
+
+impl<T: Stackable> Drop for Stack<T> {
+    #[cfg(ossl10x)]
+    fn drop(&mut self) {
+        unsafe {
+            loop {
+                let ptr = ffi::sk_pop(self.as_stack());
+
+                if ptr.is_null() {
+                    break;
+                }
+
+                // Build the owned version of the object just to run
+                // its `drop` implementation and delete the item.
+                T::from_ptr(ptr as *mut _);
+            }
+
+            ffi::sk_free(self.0 as *mut _);
+        }
+    }
+
+    #[cfg(ossl110)]
+    fn drop(&mut self) {
+        unsafe {
+            loop {
+                let ptr = ffi::OPENSSL_sk_pop(self.as_stack());
+
+                if ptr.is_null() {
+                    break;
+                }
+
+                // Build the owned version of the object just to run
+                // its `drop` implementation and delete the item.
+                T::from_ptr(ptr as *mut _);
+            }
+
+            ffi::OPENSSL_sk_free(self.0 as *mut _);
+        }
+    }
+}
+
+impl<T: Stackable> AsRef<Ref<Stack<T>>> for Stack<T> {
+    fn as_ref(&self) -> &Ref<Stack<T>> {
+        &*self
+    }
+}
+
+impl<T: Stackable> Borrow<Ref<Stack<T>>> for Stack<T> {
+    fn borrow(&self) -> &Ref<Stack<T>> {
+        &*self
+    }
+}
+
+unsafe impl<T: Stackable> OpenSslType for Stack<T> {
+    type CType = T::StackType;
+
+    unsafe fn from_ptr(ptr: *mut T::StackType) -> Stack<T> {
+        Stack(ptr)
+    }
+
+    fn as_ptr(&self) -> *mut T::StackType {
+        self.0
+    }
+}
+
+impl<T: Stackable> Deref for Stack<T> {
+    type Target = Ref<Stack<T>>;
+
+    fn deref(&self) -> &Ref<Stack<T>> {
+        unsafe { Ref::from_ptr(self.0) }
+    }
+}
+
+impl<T: Stackable> DerefMut for Stack<T> {
+    fn deref_mut(&mut self) -> &mut ::types::Ref<Stack<T>> {
+        unsafe { Ref::from_ptr_mut(self.0) }
+    }
+}
+
+impl<T: Stackable> Ref<Stack<T>> {
+    /// OpenSSL stack types are just a (kinda) typesafe wrapper around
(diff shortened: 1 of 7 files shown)


simias:
I considered adding an `OpaqueType` trait or something like that that could be used on all those "Ref" types which are just typing sugar around a raw C pointer. This trait could hold the underlying C type and maybe the "destructor" function to get rid of them.

Then Stackable could require the type to be an `OpaqueType` and I wouldn't have to add the type and destructor value in the `Stackable` trait. We could also implement the `from_ptr`/`from_ptr_mut`/`as_ptr` in the `OpaqueType` implementation instead of rewriting them everywhere.

That seemed like a pretty deep change though so I didn't want to attempt it without getting some feedback first.

simias:
I updated the tests.

openssl/src/stack.rs
```diff
@@ -0,0 +1,275 @@
...
+pub trait Stackable {
+    /// Underlying C type for this object.
+    type CType;
+    /// C stack type for this element. Generally called
+    /// `stack_st_{ELEMENT_TYPE}`, normally hidden by the
+    /// `STACK_OF(ELEMENT_TYPE)` macro in the OpenSSL API.
+    type StackType;
+    /// When the stack is owned this destructor is used to free each
+    /// element.
+    fn destructor() -> unsafe extern fn(*mut Self::CType);
```
sfackler: I think we can avoid having to add this by manually popping and dropping all values in the destructor.

openssl/src/stack.rs
```diff
@@ -0,0 +1,275 @@
...
+
+use libc::c_void;
+
+use ffi;
+use opaque::Opaque;
+
+/// Trait implemented by stackable types. This must *only* be
+/// implemented on opaque types that can be directly casted into their
+/// `CType`.
+pub trait Stackable {
```
sfackler: I think this can be replaced with the newly added `OpenSslType` trait.

openssl/src/stack.rs
```diff
@@ -0,0 +1,275 @@
...
+    /// `STACK_OF(ELEMENT_TYPE)` macro in the OpenSSL API.
+    type StackType;
+    /// When the stack is owned this destructor is used to free each
+    /// element.
+    fn destructor() -> unsafe extern fn(*mut Self::CType);
+}
+
+/// Reference to a stack of `T`. No destructors will be run when this
+/// is dropped. This is akin to a Rust slice.
+pub struct StackRef<T: Stackable> {
```
sfackler: And this can be replaced by the new `Ref<T>` type.

openssl/src/stack.rs
```diff
@@ -0,0 +1,275 @@
...
+    pub fn iter_mut(&mut self) -> IterMut<T> {
+        IterMut {
+            stack: self,
+            pos: 0,
+        }
+    }
+
+    /// Returns a reference to the element at the given index in the
+    /// stack or `None` if the index is out of bounds
+    pub fn get(&self, idx: usize) -> Option<&T> {
```
sfackler: Whenever we're returning `&T`, we should probably be returning `&Ref<T>`, and same for `&mut T`.

openssl/src/x509/mod.rs
```diff
@@ -78,6 +78,29 @@ impl X509StoreContextRef {
     pub fn error_depth(&self) -> u32 {
         unsafe { ffi::X509_STORE_CTX_get_error_depth(self.as_ptr()) as u32 }
     }
+
+    pub fn get_chain(&self) -> Option<&mut StackRef<X509Ref>> {
```
sfackler: We probably don't want a mutable reference. I believe the return type here will end up as `Option<&Ref<Stack<X509>>`.

simias:
Ah yes, this Ref type is what I had in mind when I was talking about an `OpaqueType`. Turns out I should've been a little more patient :)

openssl/src/stack.rs
simias: Well there's one caveat: the `Stackable` trait contains the CType *and* the stack type (`st_stack_`...) this way I can have Rust check that it gets passed a correctly typed pointer in the constructor.

We could do without of course but we'd lose a bit of type-safety. And after all not all `OpenSslType`s are stackable, so maybe it would make sense to have a specific trait for the subset that is? Something like:

```rust
pub trait Stackable: OpenSslType {
    type StackType;
}
```

openssl/src/stack.rs
simias: Oh yeah, I missed the `ToOwned` implementation. I guess I should require Stackable elements to implement that.

openssl/src/x509/mod.rs
simias: That looked weird to me as well but the pointer returned by the OpenSSL method is not const. You think that's a mistake? I must admit that OpenSSL seems rather inconsistent with its uses of consts.

sfackler:
No worries! This PR is what inspired me to actually do the work to split out Ref :)

openssl/src/x509/mod.rs
sfackler: OpenSSL is indeed very inconsistent about its const use; I can only think of a small handful of functions that actually take const pointers to things. I find it better to go off of the logical mutability of the operation rather than the const or non constness of the pointers. In particular, it's very rare to have a function that gets a mutable reference from a shared reference.

openssl/src/stack.rs
```diff
@@ -0,0 +1,312 @@
...
+        let rem = self.stack.len() - self.pos;
+
+        (rem, Some(rem))
+    }
+}
+
+impl<'a, T: Stackable> iter::ExactSizeIterator for Iter<'a, T> {
+}
+
+/// A mutable iterator over the stack's contents. Unfortunately we can't
```
sfackler: Same

openssl/src/stack.rs
```diff
@@ -0,0 +1,312 @@
...
+    /// Return a new Stack<T>, taking ownership of the handle
+    pub unsafe fn from_ptr(stack: *mut T::StackType) -> Stack<T> {
+        Stack(stack)
+    }
+}
+
+impl<T: Stackable> Drop for Stack<T> {
+    #[cfg(ossl10x)]
+    fn drop(&mut self) {
+        unsafe {
```
sfackler: Let's do the popping manually to avoid requiring the destructor method in Stackable:

```rust
fn drop(&mut self) {
    unsafe {
        loop {
            let ptr = OPENSSL_sk_pop(self.0 as *mut _);
            if ptr.is_null() {
                break;
            }
            T::from_ptr(ptr);
        }
        OPENSSL_sk_free(self.0 as *mut _);
    }
}
```

simias: Ah, I didn't think of simply building the `T` like that. Duh.

simias:
Done

openssl/src/stack.rs
```diff
@@ -0,0 +1,326 @@
...
+            }
+
+            ffi::OPENSSL_sk_free(self.0 as *mut _);
+        }
+    }
+}
+
+impl<T: Stackable> Clone for Stack<T> {
+    fn clone(&self) -> Stack<T> {
+        self.to_owned()
```
sfackler: I'm probably missing something, but where is to_owned actually implemented here?

simias: Uh, not sure why I left that here. Probably a remnant of a copy/paste. I'm going to remove it.

I'm not sure if it's worth actually implementing cloning stacks. I guess it could be implemented if T is Clone. But honestly most of the time you'll be better off collecting the stack into a Vec if you want a copy.

If there's a use case where we might want to create a stack in rust code in order to send it to OpenSLL it might be interesting to implement `Clone` and `IntoIterator`. But for the time being I think read-only is fine.

simias:
I removed the `Clone` implementation. I think for the purpose of this PR a read-only implementation of the `Stack` is sufficient. The mutable operations can be implemented when the need arises.

sfackler:
Yep, I don't see any reason to need to clone for now.

sfackler:
Thanks!

PR #508, opened 2016-11-01
Add method to encode a public key as a DER blob

simias:
I used a similar documentation as `public_key_to_pem` but wouldn't it be better to say "encode" instead of "store"? it's not really storing anything IMO.


sfackler:
Yeah, "encode" does sound better to me. Feel free to change! Looks good to me otherwise.

simias:
Done.

sfackler:
Thanks! Will merge once the builds finish.

PR #510, opened 2016-11-01
Implement base64 encoding and decoding

### LINKED ISSUE #509: Expose base64 decode/encode functions
I could implement it but I'm not sure what would be the best way to do that.

OpenSSL implements base64 through its BIO abstraction, using `BIO_f_base64`. I'm not sure if rust-openssl should have a generic BIO handling interface (a bit like the current `Hash`/`MessageDigest` API) of if base64 is specific enough to warrant a straightforward `base64::encode` `base64::decode` implementation.

I don't know if there would be other uses for exposing the `BIO` interface. I tried to find other useful methods besides `BIO_f_base64` but since the OpenSSL doc is garbage I didn't find much of interest. If `base64` is the only useful use case then we might as well just implement that.

simias:
Closes #509

### DIFF
--- a/openssl/src/base64.rs
+++ b/openssl/src/base64.rs
@@ -1,0 +1,191 @@
+use std::ptr;
+use std::slice;
+
+use ffi;
+use error::ErrorStack;
+use {cvt_p, cvt};
+
+/// Specify the behavior of the encoder and decoder regarding the
+/// handling of newlines.
+#[derive(PartialEq, Eq, Copy, Clone)]
+pub enum LineMode {
+    /// Tell the encoder not to generate newlines. Tell the decoder to
+    /// generate an error if a newline is encountered.
+    SingleLine,
+    /// Allow the encoder to generate newlines if the output is longer
+    /// than an arbitrary limit (usually 64 characters). Tell the
+    /// decoder to ignore newlines.
+    MultiLine,
+}
+
+/// Encode `data` in base64. If `newlines` is true then the data is
+/// encoded with line breaks.
+pub fn encode(data: &[u8], line_mode: LineMode) -> Result<Vec<u8>, ErrorStack> {
+    if data.is_empty() {
+        return Ok(Vec::new());
+    }
+
+    let bio = unsafe { try!(cvt_p(ffi::BIO_new(ffi::BIO_f_base64()))) };
+
+    let bio = Bio(bio);
+
+    if line_mode == LineMode::SingleLine {
+        unsafe {
+            ffi::BIO_set_flags(bio.as_ptr(), ffi::BIO_FLAGS_BASE64_NO_NL);
+        }
+    }
+
+    unsafe {
+        // Add some memory to hold the result
+        let mem = try!(cvt_p(ffi::BIO_new(ffi::BIO_s_mem())));
+
+        try!(cvt_p(ffi::BIO_push(bio.as_ptr(), mem)));
+
+        // Write `data` to the BIO
+        try!(cvt(ffi::BIO_write(bio.as_ptr(),
+                                data.as_ptr() as *const _,
+                                data.len() as _)));
+
+        // Flush output
+        try!(cvt(ffi::BIO_flush(bio.as_ptr())));
+
+        // Build a slice from the output
+        let mut ptr = ptr::null_mut();
+        let len = ffi::BIO_get_mem_data(bio.as_ptr(), &mut ptr);
+
+        let s =
+            slice::from_raw_parts(ptr as *const _ as *const _, len as usize);
+
+        Ok(s.to_owned())
+    }
+}
+
+/// Decode `data` from base64. If `nl` is true then the decoder will
+/// accept newlines in the base64 data.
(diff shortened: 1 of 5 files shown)


openssl/src/base64.rs
```diff
@@ -0,0 +1,170 @@
...
+        try!(cvt(ffi::BIO_write(bio.as_ptr(),
+                                data.as_ptr() as *const _,
+                                data.len() as _)));
+
+        // Flush output
+        try!(cvt(ffi::BIO_flush(bio.as_ptr())));
+
+        // Build a slice from the output
+        let mut ptr = ptr::null_mut();
+        let len = ffi::BIO_get_mem_data(bio.as_ptr(), &mut ptr);
```
sfackler: This should pass in `mem, not `bio.as_ptr()`, right?

openssl/src/base64.rs
```diff
@@ -0,0 +1,170 @@
...
+    fn drop(&mut self) {
+        unsafe {
+            ffi::BIO_free_all(self.0);
+        }
+    }
+}
+
+/// Encode `data` in base64. If `newlines` is true then the data is
+/// encoded with line breaks.
+pub fn encode(data: &[u8], newlines: bool) -> Result<Vec<u8>, ErrorStack> {
```
sfackler: I'd make an `enum Newline { Yes, No }` rather than passing in a boolean here - it's not super clear what `base64::encode(data, true)` means.

openssl/src/base64.rs
```diff
@@ -0,0 +1,170 @@
...
+    unsafe {
+        // Add some memory to hold the result
+        let mem = try!(cvt_p(ffi::BIO_new(ffi::BIO_s_mem())));
+
+        try!(cvt_p(ffi::BIO_push(bio.as_ptr(), mem)));
+
+        // Write `data` to the BIO
+        try!(cvt(ffi::BIO_write(bio.as_ptr(),
+                                data.as_ptr() as *const _,
+                                data.len() as _)));
```
simias: I'll admit that I was a bit optimistic here. What's the proper way to handle short read/writes? Can I just return `ErrorStack::get()` in these situations?

simias:
As a side note, I noticed some functions (`Error::get`, `MemBio::new`, ...) call `ffi::init()`. When should this be called exactly? Should I call it here in the `encode`/`decode` functions?

sfackler:
Oh yeah, for init, you'll want to make sure it's been called before calling into any OpenSSL functions. So we would want it in encode and decode.

openssl/src/base64.rs
sfackler: Nah, just mark off how much you've finished and write the rest until it's all gone.

openssl/src/base64.rs
```diff
@@ -0,0 +1,170 @@
...
+            ffi::BIO_set_flags(bio.as_ptr(), ffi::BIO_FLAGS_BASE64_NO_NL);
+        }
+    }
+
+    // Allocate output buffer, we know that base64 data will be at
+    // least 4/3 bigger than the raw input
+    let mut data = vec![0; (data.len() * 3 + 3) / 4];
+
+    unsafe {
+        let len = try!(cvt(ffi::BIO_read(bio.as_ptr(),
```
simias: The problem is that I don't know how much data I expect precisely, at least not without implementing parts of a base64 decoder just to figure it out. The exact length depends on the number of '=' signs at the end and of course if there are newlines in the input those should be stripped away.

The size of the `vec!` is an upper bound so I know it should all fit in a single read but I'm not sure how to identify short reads.

openssl/src/base64.rs
sfackler: Oh right, duh. Short writes can't be a thing then. We just need to handle the very long input case with a loop.

simias:
Alternatively if this feels overkill we might just give up and rely on the base64 crate instead. After all base64 encoding/decoding is not exactly rocket science, there might not be any clear benefit to using OpenSSL for that.

openssl/src/base64.rs
sfackler: Oh, actually, let's use `BIO_write_ex`, which handles the short write issue and takes a size_t for the length, avoiding the loop entirely: https://www.openssl.org/docs/manmaster/man3/BIO_read.html

sfackler: And then the same thing for read.

sfackler:
Yeah, it also seems fine to just not support base64 here. It's potentially nice since you avoid an extra dependency if you're already pulling in this, but it also doesn't seem particularly crucial.

simias:
Yeah I'm kind of torn, it's quite a lot of unsafe code for such a trivial thing. I'm going to implement the changes you requested and then let you decide whether to push the button or not!

jwilm:
> Yeah, it also seems fine to just not support base64 here. It's potentially nice since you avoid an extra dependency if you're already pulling in this, but it also doesn't seem particularly crucial.

Adding to this sentiment, the base64 handling in openssl should be considered 100% unsafe code. Discouraging its use by omitting this API may be the right choice. There are several high quality, pure Rust implementations available, after all.

simias:
Okay, I agree with that, I'm giving up on this.

PR #513, opened 2016-11-06
Provide a tailored error message on Linux

alexcrichton:
I just ran into a case where I installed OpenSSL in a docker container but I
forgot to install pkg-config. Right now openssl-sys relies on pkg-config, so
print out a nice error about this.


sfackler:
Thanks!

PR #519, opened 2016-11-08
Actually support AES GCM

### LINKED ISSUE #518: AES GCM requires some extra functionality on Crypter
https://wiki.openssl.org/index.php/Manual:EVP_EncryptInit(3)#GCM_and_OCB_Modes

sfackler:
This is an AEAD cipher, so we need some extra functionality. As another
bonus, we no longer panic if provided an IV with a different length than
the cipher's default.

Closes #518

### DIFF
--- a/openssl/src/symm.rs
+++ b/openssl/src/symm.rs
@@ -135,8 +135,7 @@ impl Crypter {
     ///
     /// # Panics
     ///
-    /// Panics if an IV is required by the cipher but not provided, or if the
-    /// IV's length does not match the expected length (see `Cipher::iv_len`).
+    /// Panics if an IV is required by the cipher but not provided.
     pub fn new(t: Cipher,
                mode: Mode,
                key: &[u8],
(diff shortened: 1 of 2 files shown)


openssl/src/symm.rs
```diff
@@ -135,8 +135,7 @@ impl Crypter {
     ///
     /// # Panics
     ///
-    /// Panics if an IV is required by the cipher but not provided, or if the
-    /// IV's length does not match the expected length (see `Cipher::iv_len`).
+    /// Panics if an IV is required by the cipher but not provided.
```
jwilm: Can this constraint just be encoded via the type system? Maybe have `new` and `new_with_iv` where `new` takes a subset of `Cipher` that doesn't require an initialization vector?

openssl/src/symm.rs
```diff
@@ -534,4 +588,42 @@ mod tests {
 
         cipher_test(super::Cipher::des_ecb(), pt, ct, key, iv);
     }
+
+    #[test]
+    fn test_aes128_gcm() {
+        let key = "0e00c76561d2bd9b40c3c15427e2b08f";
```
jwilm: Why not format the test vectors as `[u8]`? The tests would be a bit more straightforward not converting to/from hex everywhere.

sfackler: The hex strings are copied directly out of the NIST test vectors, where they're provided hex encoded. I could stick the decode right on the string literals I guess.

sfackler:
It would certainly be nice to have higher level APIs, but I think we always want to have this low level base around.

sfackler:
There's an old abandoned rewrite of the symm module that attempted to do something like that: https://github.com/sfackler/rust-openssl/pull/173.

PR #524, opened 2016-11-13
More elliptic curve functionality

sfackler:
cc @jwilm


sfackler:
One thing that came up while implementing: Why are there separate constructors/accessors for GFp and GF2^m curves? The implementations appear to be identical.

openssl/src/ec_key.rs
```diff
@@ -1,37 +1,336 @@
...
+                                  b.as_ptr(),
+                                  ctx.as_ptr()))
+                .map(|_| ())
+        }
+    }
+
+    /// Computes `generator * n + q * m`, storing the result in `self`.
+    ///
+    /// If `n` is `None`, `q * m` will be computed instead.
+    pub fn mul(&mut self,
```
sfackler: I wonder if this should be split into two methods?

sfackler:
I'm considering moving the point methods onto group - might make more sense that way?

sfackler:
Doin it

sfackler:
On second thought, I don't like that setup

jwilm:
Finally have some time to look at this; will leave some comments

sfackler:
I should note that I'm not thrilled about having to pass an `EcGroupRef` around everywhere, but I'm not sure what a better option would be.

systest/build.rs
```diff
@@ -52,7 +52,7 @@ fn main() {
             format!("bio_info_cb*")
         } else if s == "_STACK" {
             format!("struct stack_st")
-        } else if is_struct && s.chars().next().unwrap().is_lowercase() {
+        } else if is_struct && s != "point_conversion_form_t" && s.chars().next().unwrap().is_lowercase() {
```
jwilm: A comment to explain the special casing here would be nice

openssl/src/ec_key.rs
```diff
@@ -1,37 +1,338 @@
...
             init();
             cvt_p(ffi::EC_KEY_new_by_curve_name(nid.as_raw())).map(EcKey)
         }
     }
 
+    /// Generates a new public/private key pair on the specified curve.
+    pub fn generate(group: &EcGroupRef) -> Result<EcKey, ErrorStack> {
+        unsafe {
+            let key = EcKey(try!(cvt_p(ffi::EC_KEY_new())));
+            try!(cvt(ffi::EC_KEY_set_group(key.as_ptr(), group.as_ptr())));
```
jwilm: Does `EC_KEY_set_group` really just borrow an `EC_GROUP` for only the duration of the call?

openssl/src/ec_key.rs
```diff
@@ -1,37 +1,338 @@
...
+                None
+            } else {
+                Some(BigNumRef::from_ptr(ptr as *mut _))
+            }
+        }
+    }
+
+    /// Checks the key for validity.
+    pub fn check_key(&self) -> Result<(), ErrorStack> {
+        unsafe { cvt(ffi::EC_KEY_check_key(self.as_ptr())).map(|_| ()) }
```
jwilm: Seems like this should return a bool. Otherwise, an `Err` indicates the key is either invalid or openssl encountered an error.

openssl/src/ec_key.rs
jwilm: Seems like a good idea; one case for calling `EC_POINT_mul` really just requires a second point. Although.... I see there's no way to retrieve an `EcGroup` given an `EcPoint` currently.

openssl/src/ec_key.rs
```diff
@@ -1,37 +1,338 @@
...
+            let len = ffi::EC_POINT_point2oct(group.as_ptr(),
+                                              self.as_ptr(),
+                                              form.0,
+                                              ptr::null_mut(),
+                                              0,
+                                              ctx.as_ptr());
+            if len == 0 {
+                return Err(ErrorStack::get());
+            }
+            let mut buf = vec![0; len];
```
jwilm: Why use a zeroing vec construction? The values will be overwritten immediately. I'm guessing this is to avoid manually managing the capacity and length, but the perf hit is a bummer. Maybe in the future we can figure out a less error-prone way to initialize a Vec without zeroing. Actually, maybe it would be worth pushing the burden of allocation to the caller.

openssl/src/ec_key.rs
```diff
@@ -1,37 +1,338 @@
...
+                                                   ctx.as_ptr())));
+            Ok(res == 0)
+        }
+    }
+}
+
+impl EcPoint {
+    /// Creates a new point on the specified curve.
+    pub fn new(group: &EcGroupRef) -> Result<EcPoint, ErrorStack> {
+        unsafe { cvt_p(ffi::EC_POINT_new(group.as_ptr())).map(EcPoint) }
```
jwilm: Asked this in another place, but is the EC_GROUP really just borrowed, or does EC_POINT expect ownership?

sfackler: From the implementation, the point only stores the group's method, which is has a static lifetime: https://github.com/openssl/openssl/blob/master/crypto/ec/ec_lib.c#L556

openssl/src/ec_key.rs
sfackler: It only returns a 0 or 1, unfortunately, and it seems useful to know why it failed to validate rather than not know.

openssl/src/ec_key.rs
sfackler: The group is duplicated: https://github.com/openssl/openssl/blob/6f0ac0e2f27d9240516edb9a23b7863e7ad02898/crypto/ec/ec_key.c#L424

openssl/src/ec_key.rs
sfackler: I like to avoid the extra unsafety of uninitialized values unless it's clear that there is a measurable/significant performance difference.

openssl/src/ec_key.rs
jwilm: ah cool; maybe worth adding a comment about it?

sfackler:
Maybe. I haven't done that elsewhere (e.g. `ssl::SslContext`) though unfortunately.

PR #530, opened 2016-12-09
Add Blowfish support

0xa:


sfackler:
Thanks!

Looks like one of these is a macro which the tests are not happy with: https://github.com/openssl/openssl/blob/6f0ac0e2f27d9240516edb9a23b7863e7ad02898/include/openssl/evp.h#L732

sfackler:
Cool, looks like tests are happy. Could you add some tests to make sure these behave as expected? Looks like there are some test vectors you can use here: https://www.schneier.com/code/vectors.txt

0xa:
Those are without padding I think and would require calling .pad(false) somewhere, how would you prefer that done?

sfackler:
That seems fine to disable padding if those test vectors require it.

sfackler:
Thanks!

PR #533, opened 2016-12-12
Clarify documentation for OSX install

adrianbrink:
This clarification is badly needed, since people like me always spend hours trying to find the answer or give up extremely frustrated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,13 +56,10 @@ homebrew:
 ```bash
 brew install openssl
 ```
-This will install opensll as a keg-only version, which means that cargo won't be
-able to discover it. Furthermore running
-```bash
-brew link --force openssl
-```
-won't solve your problem as brew will rightfully tell you that this move is very
-insecure.
+
+#### Note for users of the hyper crate
+```brew install``` won't directly work and neither will ```brew link --force openssl```.
+It will  fail because hyper still depends on an older version of this crate.
 The solution is to export two environment variables which allow cargo to locate
 the correct openssl libraries installed by brew.
 ```
--- a/README.md
+++ b/README.md
@@ -56,6 +56,21 @@ homebrew:
 ```bash
 brew install openssl
 ```
+This will install opensll as a keg-only version, which means that cargo won't be
+able to discover it. Furthermore running
+```bash
+brew link --force openssl
+```
+won't solve your problem as brew will rightfully tell you that this move is very
+insecure.
+The solution is to export two environment variables which allow cargo to locate
+the correct openssl libraries installed by brew.
+```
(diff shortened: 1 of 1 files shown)


sfackler:
Homebrew installs are automatically recognized in the current release. Documentation similar to this is in the old 0.7.x README, which is linked from the current one here: https://github.com/sfackler/rust-openssl#warning

adrianbrink:
Mh, I would still leave it in there for now, since when installing [hyper](https://github.com/hyperium/hyper/blob/master/Cargo.toml) you will still need to fix it by adding the env vars.

And most people won't be digging through the actual version numbers. They either want it to work like with NPM or find a super simple explanation on the front-page. :-)

sfackler:
What does "working like NPM" mean? I don't really know how NPM would avoid the issue of libraries changing over time.

adrianbrink:
Working like NPM means: it should be easy and convenient to use and just work :-)

That's why I would include this even in the current documentation.

adrianbrink:
I have amended it so that it highlights that this only applies to users of the hyper crate. Once hyper updates its version of this crate we should remove this extensions.

sfackler:
@seanmonstar seriously, can you upgrade this freaking dependency? I am going to start forwarding all 0.7 support to you.

README.md
```diff
@@ -57,6 +57,18 @@ homebrew:
...
 
+#### Note for users of the hyper crate
+```brew install``` won't directly work and neither will ```brew link --force openssl```.
+It will  fail because hyper still depends on an older version of this crate.
+The solution is to export two environment variables which allow cargo to locate
+the correct openssl libraries installed by brew.
+```
+export OPENSSL_LIB_DIR=/usr/local/opt/openssl/lib
+export OPENSSL_INCLUDE_DIR=/usr/local/opt/openssl/include
+```
```
sfackler: No, `DEP_OPENSSL_INCLUDE` is set by cargo.

adrianbrink:
If you don't want to merge it, please just close it. :-(

golddranks:
I just – after debugging this for n:th time – came here to add the thing about "This formula is keg-only, which means it was not symlinked into /usr/local.", but it seems that the documentation fix is already underway! (Or then Hyper will finally update its dependency, which makes the additional note in the docs needless... but that could still take weeks, and having a temporary note in the documentation is only a few-row change... do it, please!)

adrianbrink:
If you want to merge it I'll update the PR today.

sfackler:
I don't want to be peppering the README with random portions of the version for random old versions. Hyper will be cutting a release soon decoupling it from 0.7.

Issue #534, opened 2016-12-13
Failure "unable to get local issuer certificate" on Windows


Philipp91:
I'm trying to run the example code from the documentation: http://sfackler.github.io/rust-openssl/doc/v0.9.2/openssl/ssl/index.html. I mostly include the detailed error message here for other users to find, because currently the only result in the issue tracker is the dangerous issue explaining how to disable certificate validation -- which is not how one should react to this error.

It works on Ubuntu but fails on Windows:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Failure(MidHandshakeSslStream { stream: SslStream { stream: TcpStream { addr: V6([2003:6a:8f3e:ce34:7542:e1b1:3d39:e107]:61961), peer: V6([2a00:1450:4001:81b::200e]:443), socket: 308 }, ssl: Ssl { state: "error", verify_result: X509VerifyError { code: 20, error: "unable to get local issuer certificate" } } }, error: Ssl(ErrorStack([Error { code: 337047686, library: "SSL routines", function: "tls_process_server_certificate", reason: "certificate verify failed" }])) })', ../src/libcore\result.rs:837
```
This is the line where it fails: `connector.connect("google.com", stream).unwrap()`.

The reason it does not work is because the OpenSSL binaries recommended by rust-openssl (http://slproweb.com/products/Win32OpenSSL.html) do not contain trusted roots. On my command line, I have two versions of openssl.exe. One is bundled with those binaries and it throws the same error when I execute `openssl s_client  -connect google.com:443`. The other is bundled with my Git installation and that one has a `C:\Git\usr\ssl\certs\ca-bundle.crt` file, so it works fine.

To help its users solve this issue, rust-openssl should either recommend different OpenSSL binaries (if any exist), or those binaries or rust-openssl should provide instructions on how to install root certificates, or rust-openssl should explain how to pin the correct certificate (and point users there from the main documentation where the code sample can be found).

sfackler:
I'd be happy to take a PR improving the README docs for Windows!

Philipp91:
See #535

unordered:
my ubuntu:bionic also
but [it's solved](https://original.unordered.org/timelines/5a57e94425c00000)
🎉🎉🎉

PR #535, opened 2016-12-14
#534 Explain how to install trusted root certificates

Philipp91:
The slproweb.com OpenSSL distribution does not contain root certificates, so they need to be downloaded and installed manually to avoid certificate warnings when making requests.


sfackler:
Thanks!

Issue #521, opened 2016-11-12
LibreSSL support


chneukirchen:
The version check for OPENSSL_VERSION_NUMBER in build.rs bails out for LibreSSL, where the value is 0x20000000L. The crate seems to build fine else, assuming LibreSSL is like OpenSSL 1.0.1.

sfackler:
I'd be more than happy to take a patch adding LibreSSL support!

chneukirchen:
Hm, just lying its 1.0.1 breaks systest because it tries to look into structures opaque in LibreSSL.

I don't really know these APIs, help wanted. :)

sfackler:
It'll probably be a somewhat significant amount of work, though I'm not sure how much the LibreSSL people have changed at the ABI level. The general approach would be to use the systest crate to figure out which definitions are wrong and add a new submodule with the correct definitions like we currently do for OpenSSL 1.1.0 and 1.0.x, and then when that's all working, making any necessary changes in the main crate. We'd also want CI builds added with whichever LibreSSL versions are supported.

sfackler:
It sounds like it's API compatible with 1.0.1, but not ABI compatible, so I'm guessing some struct definitions will need to be adjusted: https://github.com/libressl-portable/portable#compatibility-with-openssl

chneukirchen:
It would be great to have this in a release, as cargo 0.15 needs rust-openssl 0.9.3 and thus doesn't build on libressl platforms at the moment.

sfackler:
Done!

PR #538, opened 2016-12-20
Add LibreSSL support

semarie:
This PR adds LibreSSL support to rust-openssl.

It passed all tests with LibreSSL 2.5.1 (openssl, openssl-sys and systest).
Please note I had to disabling `dtlsv1` tests due to timeout issues.

Fixes: #521

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,6 +72,8 @@ matrix:
           packages:
             - gcc-multilib
 
+    # LibreSSL
+    - env: BUILD_LIBRESSL_VERSION=2.5.0
 
 before_install:
   - ./openssl/test/build.sh
@@ -83,4 +85,5 @@ script:
 cache:
   cargo: true
   directories:
+    - $HOME/libressl
     - $HOME/openssl
--- a/openssl-sys/build.rs
+++ b/openssl-sys/build.rs
@@ -247,6 +247,15 @@ fn validate_headers(include_dirs: &[PathBuf],
     } else if version_text.contains("0x10100") {
         println!("cargo:rustc-cfg=ossl110");
         println!("cargo:version=110");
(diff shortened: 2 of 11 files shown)


sfackler:
Awesome!

We'll want a Travis build against libressl - could you set that up?

semarie:
I am unsure about to be able to setup Travis (I work on OpenBSD, not Linux). I could take a look, but there is no guarantee.

semarie:
I will take a look tomorrow to see why my local build (under OpenBSD and with clang) doesn't have the errors present under Linux. I should have missed something.

sfackler:
Sounds good, thanks!

Just to make sure, LibreSSL plans to continue being API compatible with some release of OpenSSL in the future, is that right?

semarie:
For what I understood, LibreSSL should keep some API compatibility with OpenSSL. But some interfaces could be not implemented (if considered insane or unneed), or just enough for the API compatibility: as example, most of the flags values mismatch in Travis are commented as `Obsolete flags kept for compatibility. No sane code should use them.` in LibreSSL. I will ask for confirmation.

I wonder if the most straightforward approch shouldn't to start a `openssl-sys/src/libressl.rs` based on `ossl10x.rs`. It could take `ossl10x.rs` more readable, and will permit integrate LibreSSL changes independently from OpenSSL. But it also means duplicating code.

sfackler:
Cool, that makes sense. Let's have a separate set of cfgs and feature flags for LibreSSL then.

semarie:
@sfackler do you think this rewrite is good enough ?

sfackler:
This seems good, thanks!

Keruspe:
Could we get a release for this? :)

sfackler:
Done!

PR #539, opened 2016-12-20
Handle zero-length writes in SSL_write

alexcrichton:
This commit adds some short-circuits for zero-length reads/writes to
`SslStream`. Because OpenSSL returns 0 on error, then we could mistakenly
confuse a 0-length success as an actual error, so we avoid writing or reading 0
bytes by returning quickly with a success.

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -1506,6 +1506,15 @@ impl<S: Read + Write> SslStream<S> {
     /// This is particularly useful with a nonblocking socket, where the error
     /// value will identify if OpenSSL is waiting on read or write readiness.
     pub fn ssl_read(&mut self, buf: &mut [u8]) -> Result<usize, Error> {
+        // The intepretation of the return code here is a little odd with a
+        // zero-length write. OpenSSL will likely correctly report back to us
+        // that it read zero bytes, but zero is also the sentinel for "error".
+        // To avoid that confusion short-circuit that logic and return quickly
+        // if `buf` has a length of zero.
+        if buf.len() == 0 {
+            return Ok(0)
+        }
+
         let ret = self.ssl.read(buf);
         if ret > 0 {
             Ok(ret as usize)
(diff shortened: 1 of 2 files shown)


alexcrichton:
This is a fix for https://github.com/tokio-rs/tokio-tls/issues/15 on Linux. Before this commit the weird error in the post shows up, but after it seems to work as expected.

alexcrichton:
Note that it seems [`SSL_write_ex`](https://www.openssl.org/docs/manmaster/man3/SSL_write_ex.html) solves this ambiguity problem, but I'm unsure about the portability of that across openssl revisions

sfackler:
Is the other end actually hanging up here? It may also be caused by https://github.com/sfackler/rust-openssl/blob/master/openssl/src/ssl/connector.rs#L38 interacting weirdly with nonblocking IO. If removing that mode flag also fixes things I think I'd prefer to do that.

alexcrichton:
Ah ok. I tried removing that flag though and it comes back with the same error as before :(

alexcrichton:
To be clear I'm not actually sure what's happening at the protocol level. I took at look at `strace` and I didn't see a literal return of `Ok(0)` from `write` or `sendto` or anything like that. The best I could guess is that there was some SSL protocol thing indicating that there was no more data or something like that?

sfackler:
The remote can shut down the connection, but that's a permanent thing - with this change, do things just "fail properly", or is it a transient 0-byte write and things continue on after that.

SSL_write_ex appears to have existed for a long time, so let's use that rather than trying to guess like we are now.

alexcrichton:
Yeah I don't really know how but everything starts working with this patch. "starts working" for me entails that I get a massive amount of output to the console in the script in the linked issue, no errors happen. 

There's a few layers of abstraction between here and the console though: rust-openssl -> native-tls -> tokio-tls -> tokio-core -> tokio-proto -> hyper -> actual application. I haven't dug thoroughly into what's going on and why this fixes things, though.

alexcrichton:
Some digging has revealed that the real problem here was a zero-length write and how we'd conflate writing zero bytes with an error. I've updated the PR appropriately along with the description as well.

sfackler:
👍

PR #540, opened 2016-12-24
Use metadeps to specify pkg-config dependencies declaratively

joshtriplett:
This makes it easier for distribution packaging tools to generate
appropriate package dependencies.

### DIFF
--- a/openssl-sys/Cargo.toml
+++ b/openssl-sys/Cargo.toml
@@ -14,8 +14,11 @@ build = "build.rs"
 libc = "0.2"
 
 [build-dependencies]
-pkg-config = "0.3"
+metadeps = "1"
 
 [target.'cfg(windows)'.dependencies]
 user32-sys = "0.2"
 gdi32-sys = "0.2"
+
+[package.metadata.pkg-config]
+openssl = "1.0.1"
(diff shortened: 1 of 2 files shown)


sfackler:
Neat! It looks like this will bump the minimum supported Rust version to 1.13 due to the error-chain dependency.

joshtriplett:
@sfackler Interesting; I didn't realize that error-chain required 1.13.  I reported that as https://github.com/brson/error-chain/issues/101 .

If that issue gets addressed, hopefully this will work.  If it doesn't, I may abandon error-chain in metadeps.  (I like it, but I already think I may disable the `backtrace` feature that pulls in many dependencies. If it also prevents widespread use by projects that need to run on older Rust, then I'll have to drop it.)

joshtriplett:
@sfackler OK, I've gotten a change into error-chain so it works with Rust 1.10, and updated metadeps to work with that.  As soon as error-chain 0.7.2 gets released to crates.io, I'll re-push this so it re-runs CI and confirms the functional build on Rust 1.10.

sfackler:
Looks like that was published and the builds are happy. Thanks!

PR #548, opened 2017-01-04
Added SSL_SESSION and SSL_get_session

halmohri:
I have modified libressl.rs, ossl10x.rs, and ossl110.rs in openssl-sys to include SSL_SESSION which is a struct representing the current session. Also, I added SSL_get_session which is a function returning a pointer to SSL_SESSION. This is a useful addition that provides access to session data including the session key.

### DIFF
--- a/openssl-sys/src/libressl.rs
+++ b/openssl-sys/src/libressl.rs
@@ -2,7 +2,7 @@ use std::sync::{Mutex, MutexGuard};
 use std::sync::{Once, ONCE_INIT};
 use std::mem;
 
-use libc::{c_int, c_char, c_void, c_long, c_uchar, size_t, c_uint, c_ulong};
+use libc::{c_int, c_char, c_void, c_long, c_uchar, size_t, c_uint, c_ulong, uint32_t};
 use libc::time_t;
 
 #[repr(C)]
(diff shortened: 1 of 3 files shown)


halmohri:
Steven, 

Could you help with understanding the failed checks? It seems like SSL_get_session has a return type problem..?

openssl-sys/src/libressl.rs
```diff
@@ -525,6 +579,7 @@ extern {
                                                                 is_export: c_int,
                                                                 keylength: c_int)
                                                                 -> *mut ::EC_KEY);
+    pub fn SSL_get_session(ssl: *mut ::SSL) -> *mut ::SSL_SESSION;
```
halmohri: Yes! I missed that. Thanks.
There's another issue with the size of SSL_SESSION and crypto_ex_data_st. Four bits off from their C representations.

halmohri: This is how I'm translating C to Rust: 

```
struct crypto_ex_data_st {
	    STACK_OF(void) *sk;
};

```
```
#[repr(C)]
pub struct crypto_ex_data_st {
    pub sk: *mut stack_st_void,
}
```

halmohri: Seems like this PR won't be done with errors.

PR #550, opened 2017-01-05
LibreSSL support improvements

Keruspe:
Makes test pass with libressl 2.4.4 and export set_ecdh_auto

### DIFF
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -1206,15 +1206,15 @@ pub const X509_V_ERR_UNSUPPORTED_EXTENSION_FEATURE: c_int = 45;
 pub const X509_V_ERR_UNSUPPORTED_NAME_SYNTAX: c_int = 53;
 pub const X509_V_OK: c_int = 0;
 
-#[cfg(not(ossl101))]
+#[cfg(not(any(ossl101, libressl)))]
 pub const X509_CHECK_FLAG_ALWAYS_CHECK_SUBJECT: c_uint = 0x1;
-#[cfg(not(ossl101))]
+#[cfg(not(any(ossl101, libressl)))]
 pub const X509_CHECK_FLAG_NO_WILDCARDS: c_uint = 0x2;
-#[cfg(not(ossl101))]
+#[cfg(not(any(ossl101, libressl)))]
 pub const X509_CHECK_FLAG_NO_PARTIAL_WILDCARDS: c_uint = 0x4;
-#[cfg(not(ossl101))]
+#[cfg(not(any(ossl101, libressl)))]
 pub const X509_CHECK_FLAG_MULTI_LABEL_WILDCARDS: c_uint = 0x8;
-#[cfg(not(ossl101))]
+#[cfg(not(any(ossl101, libressl)))]
 pub const X509_CHECK_FLAG_SINGLE_LABEL_SUBDOMAINS: c_uint = 0x10;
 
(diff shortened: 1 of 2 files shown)


openssl-sys/src/lib.rs
```diff
@@ -1206,15 +1206,15 @@ pub const X509_V_ERR_UNSUPPORTED_EXTENSION_FEATURE: c_int = 45;
 pub const X509_V_ERR_UNSUPPORTED_NAME_SYNTAX: c_int = 53;
 pub const X509_V_OK: c_int = 0;
 
-#[cfg(not(ossl101))]
+#[cfg(not(any(ossl101, libressl)))]
```
sfackler: I don't think this is necessary since we set `ossl101` for LibreSSL?

openssl/src/ssl/mod.rs
```diff
@@ -774,12 +774,12 @@ impl SslContextBuilder {
     /// curve.
     ///
     /// Requires the `v102` feature and OpenSSL 1.0.2.
-    #[cfg(all(feature = "v102", ossl102))]
+    #[cfg(any(all(feature = "v102", ossl102), libressl))]
```
sfackler: LibreSSL backported this I'm assuming? I'd still want it behind a feature flag.

openssl-sys/src/lib.rs
Keruspe: either we don't or that's not sufficient, without the first commit, on a fresh clone of master I get that: https://gist.github.com/Keruspe/12e48a0432b43196c9a6e5de80cd4698

openssl/src/ssl/mod.rs
Keruspe: fine, changing that then

openssl/src/ssl/mod.rs
```diff
@@ -774,12 +774,12 @@ impl SslContextBuilder {
     /// curve.
     ///
     /// Requires the `v102` feature and OpenSSL 1.0.2.
-    #[cfg(all(feature = "v102", ossl102))]
+    #[cfg(all(feature = "v102", any(ossl102, libressl)))]
```
sfackler: It seems like we'll want to add a separate set of Cargo features for libressl I think.

Keruspe: What would the feature be named? Wouldn't it be better to try to be as seamless as possible? (I probably don't realize what all the implications are)

sfackler: `librev23` corresponding to LibreSSL 2.3 maybe, and similarly for the other versions we support? I'm not sure when `ecdh_auto` was added, but if it's always been around in every release of LibreSSL we can probably then just have an unversioned `libre` feature.

The reason I'm worrying about this is that LibreSSL mirrors OpenSSL 1.0.1 APIs, but then apparently with some other additions like this, so it doesn't really correspond to existing feature.

Keruspe: set_ecdh_auto has been there since at least libressl 2.1.2, for what it's worth.

Having features specific to libressl, wouldn't that make things harder for downstreams to be compatible with both libressl and openssl?

sfackler: I mean, we can't pretend that features exist when they actually don't. It's hard either way.

Keruspe: Indeed, the whole situation upstreams sucks :(

PR #558, opened 2017-01-20
Support AES IGE

### LINKED ISSUE #523: Add AES IGE support
OpenSSL already implements AES IGE, could you add it in this package? I'd be willing to help out as well

sfackler:
This is a special snowflake used only by Telegram apparently.

Closes #523

### DIFF
--- a/openssl/src/aes.rs
+++ b/openssl/src/aes.rs
@@ -1,0 +1,115 @@
+//! Low level AES functionality
+//!
+//! The `symm` module should be used in preference to this module in most cases.
+use ffi;
+use std::mem;
+use libc::c_int;
+
+use symm::Mode;
+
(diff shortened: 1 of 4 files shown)


openssl/src/aes.rs
```diff
@@ -0,0 +1,112 @@
+use ffi;
+use std::mem;
+use libc::c_int;
+
+use symm::Mode;
+
+#[derive(Debug)]
+pub struct KeyError(());
```
habnabit: Not an `ErrorStack`?

sfackler: These functions don't adjust the error stack: https://github.com/openssl/openssl/blob/master/crypto/aes/aes_core.c#L640-L643

Issue #525, opened 2016-11-15
DTLS tests are super flaky on Travis


sfackler:
It's all over UDP so I guess it's not that surprising. It may make sense to turn these off on Travis.

PR #567, opened 2017-01-25
Fix cross compilation from Linux to Windows by MinGW

Frederick888:


Frederick888:
Test failed due to #525.

sfackler:
Could you rebase? Looks good otherwise!

sfackler:
Hmm, actually, is this going to do the right thing? We're appending rather than prepending this pkg-config search path so it won't be selected if there's another openssl in a previous `PKG_CONFIG_PATH` entry. In addition, we're enabling pkg-config even if the cross compilation openssl doesn't exist in which case we'll incorrectly try to link to the host OpenSSL.

Frederick888:
@sfackler 
1. (prepend or append) I had a careful consideration about this and I thought our "guess" shouldn't be prioritized over user configurations. For example, the user has two different builds of OpenSSL, one in `/usr/x86_64-w64-mingw32/lib/pkgconfig` and `/opt/mingw` for the other, and s/he intends to link with the latter one. In case of existing `PKG_CONFIG_PATH` that would lead to a wrong result, users could simply either clear the env variable or manually specify it, it's kinda beyond our control.
0. It's a decision on whether to check the pkg-config file before hand or let pkg-config itself check for us. Both will fail in the end, both are fine to me.

sfackler:
The issue with unconditionally using pkg-config is that if the user doesn't actually have a mingw cross compiled OpenSSL in the search path, pkg-config will happily return the global system one, which will then cause mysterious linker failures down the line since it's the wrong format for Windows.

Frederick888:
@sfackler Then the only way to completely avoid it is to check if `openssl.pc` is present on our own. Otherwise, if the user specifies a wrong path, s/he will still get the "mysterious" weird issue.

It could be something like:

    if target.contains("windows-gnu") && host.contains("windows") {
        env::set_var("PKG_CONFIG_ALLOW_CROSS", "1");
    } else if target.contains("windows-gnu") && host.contains("linux") {
        let mut pkg_path = env::var("PKG_CONFIG_PATH").unwrap_or(String::new()).to_string();
        if target.contains("x86_64") {
            pkg_path.push_str(":/usr/x86_64-w64-mingw32/lib/pkgconfig");
        } else if target.contains("i686") {
            pkg_path.push_str(":/usr/i686-w64-mingw32/lib/pkgconfig");
        }
        let pkg_found = pkg_path.split(':')
            .map(|p| Path::new(&format!("{}/openssl.pc", p)).exists())
            .fold(false, |a, b| a || b);
        if pkg_found {
            env::set_var("PKG_CONFIG_ALLOW_CROSS", "1");
        } else {
            return;
        }
    } else if target.contains("windows") {
        return
    }

What do you think about it?

sfackler:
This will still pick up non-target OpenSSLs if the user happens to have it in their `PKG_CONFIG_PATH`. The heuristics here seem a bit sketchy :(

Frederick888:
@sfackler That depends on how deep you'd like to go, and actually I think it's kinda what the user should take care of.
If you only want to recognise Windows and prevent users from linking libraries for other platforms to Windows, it could be easily done by checking the first several bytes. Otherwise, for example, I've got one toolchain for ARMv6 and another for ARMv7/8, and I can't tell which platform the library is exactly for even by `objdump` (and I never expect the build script to do that for me).

PR #568, opened 2017-01-26
X509 request information

mredlek:
I added X509ReqRef.subject_name and X509ReqRef.version.

My aim is to add all properties on a csr to be able to build a new certificate with the information from the csr.

This is my first pull request, so I kept the change small.

### DIFF
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -51,7 +51,6 @@ pub enum X509_CRL {}
 pub enum X509_EXTENSION {}
 pub enum X509_NAME {}
 pub enum X509_NAME_ENTRY {}
-pub enum X509_REQ {}
 pub enum X509_STORE {}
 pub enum X509_STORE_CTX {}
 pub enum bio_st {}
@@ -1387,6 +1386,7 @@ extern {
     pub fn AES_set_decrypt_key(userKey: *const c_uchar, bits: c_int, key: *mut AES_KEY) -> c_int;
     pub fn AES_ige_encrypt(in_: *const c_uchar, out: *mut c_uchar, length: size_t, key: *const AES_KEY, ivec: *mut c_uchar, enc: c_int);
 
+    pub fn ASN1_INTEGER_get(dest: *mut ASN1_INTEGER) -> c_long;
     pub fn ASN1_INTEGER_set(dest: *mut ASN1_INTEGER, value: c_long) -> c_int;
     pub fn ASN1_GENERALIZEDTIME_free(tm: *mut ASN1_GENERALIZEDTIME);
     pub fn ASN1_GENERALIZEDTIME_print(b: *mut BIO, tm: *const ASN1_GENERALIZEDTIME) -> c_int;
@@ -1917,6 +1917,8 @@ extern {
 
     pub fn ASN1_STRING_free(x: *mut ASN1_STRING);
     pub fn ASN1_STRING_length(x: *const ASN1_STRING) -> c_int;
+
+    pub fn ASN1_INTEGER_free(x: *mut ASN1_INTEGER);
 
     pub fn X509_STORE_new() -> *mut X509_STORE;
     pub fn X509_STORE_free(store: *mut X509_STORE);
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -1935,6 +1935,9 @@ extern {
 
     pub fn X509_REQ_add_extensions(req: *mut X509_REQ, exts: *mut stack_st_X509_EXTENSION) -> c_int;
     pub fn X509_REQ_sign(x: *mut X509_REQ, pkey: *mut EVP_PKEY, md: *const EVP_MD) -> c_int;
+    pub fn X509_REQ_set_version(x: *mut X509_REQ, version: c_long) -> c_int;
+    pub fn X509_REQ_set_subject_name(req: *mut X509_REQ, name: *mut ::X509_NAME) -> c_int;
+
 
     #[cfg(not(ossl101))]
     pub fn X509_VERIFY_PARAM_free(param: *mut X509_VERIFY_PARAM);
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -1386,7 +1386,7 @@ extern {
     pub fn AES_set_decrypt_key(userKey: *const c_uchar, bits: c_int, key: *mut AES_KEY) -> c_int;
     pub fn AES_ige_encrypt(in_: *const c_uchar, out: *mut c_uchar, length: size_t, key: *const AES_KEY, ivec: *mut c_uchar, enc: c_int);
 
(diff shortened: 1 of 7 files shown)


openssl/src/asn1.rs
```diff
@@ -75,6 +75,28 @@ impl Asn1StringRef {
     }
 }
 
+type_!(Asn1Integer, Asn1IntegerRef, ffi::ASN1_INTEGER, ffi::ASN1_INTEGER_free);
+
+impl Asn1IntegerRef {
+    pub fn get(&self) -> i64 {
+        unsafe {
+            return ::ffi::ASN1_INTEGER_get(self.as_ptr());
```
sfackler: You'll want to cast this to i64 for 32 bit platforms.

I'd also remove the explicit return.

openssl/src/asn1.rs
```diff
@@ -75,6 +75,28 @@ impl Asn1StringRef {
...
+        }
+    }
+
+    pub fn set(&self, value: i64) -> Result<(), ErrorStack>
```
sfackler: This should take `&mut self`.

openssl/src/asn1.rs
```diff
@@ -75,6 +75,28 @@ impl Asn1StringRef {
...
+    {
+        unsafe {
+            let res = ::ffi::ASN1_INTEGER_set(self.as_ptr(), value);
```
sfackler: You can simplify this to `cvt(ffi::ASN1_INTEGER_set(self.as_ptr(), value as c_long)).map(|_| ())`.

openssl/src/asn1.rs
sfackler: Hmm, taking an i64 is a bit unfortunate since that's lossy on 32 bit platforms.

@alexcrichton any ideas what the right approach is here? We could take an i32 here, but then it's asymmetric with get...

openssl/src/x509/mod.rs
```diff
@@ -599,6 +599,21 @@ impl X509Req {
 impl X509ReqRef {
     to_pem!(ffi::PEM_write_bio_X509_REQ);
     to_der!(ffi::i2d_X509_REQ);
+
+    pub fn version(&self) -> i64
+    {
+        unsafe {
+            let version = compat::X509_REQ_get_version(self.as_ptr());
```
sfackler: You'll need to cast here.

openssl/src/x509/mod.rs
```diff
@@ -599,6 +599,21 @@ impl X509Req {
...
+            version
+        }
+    }
+
+    pub fn subject_name(&self) -> &X509NameRef {
+        unsafe {
+            let name = compat::X509_REQ_get_subject_name(self.as_ptr());
```
sfackler: Will the subject name ever be missing?

mredlek:
Thank you for your review. I will apply the suggested changes, after I know the best return type for Asn1IntegerRef.get. I also found in the automatic build that ASN1_INTEGER_get doesn't take a *mut ASN1_INTEGER but a *const ASN1_INTEGER.

openssl/src/x509/mod.rs
mredlek: I copied it from X509Ref.subject_name and these function seem to share the same documentation.

openssl/src/asn1.rs
mredlek: I will add the mut.

I share the doubth about the integer type.

openssl/src/asn1.rs
mredlek: I will remove the explicit return.

openssl-sys/src/lib.rs
```diff
@@ -1387,6 +1386,7 @@ extern {
     pub fn AES_set_decrypt_key(userKey: *const c_uchar, bits: c_int, key: *mut AES_KEY) -> c_int;
     pub fn AES_ige_encrypt(in_: *const c_uchar, out: *mut c_uchar, length: size_t, key: *const AES_KEY, ivec: *mut c_uchar, enc: c_int);
 
+    pub fn ASN1_INTEGER_get(dest: *mut ASN1_INTEGER) -> c_long;
```
mredlek: I will change the *mut to *const.

mredlek:
I think I added your suggestions. Do you want me to squash the commits into one?

For now I let Asn1Integer get an i64 (such that every value can be read) and set an i32 (such that every value can be written).

openssl/src/x509/mod.rs
```diff
@@ -599,6 +599,33 @@ impl X509Req {
...
+            X509NameRef::from_ptr(name)
+        }
+    }
+
+    pub fn set_subject_name(&self, value: &X509NameRef) -> Result<(), ErrorStack> {
```
sfackler: This should take `&mut self`.

openssl/src/x509/mod.rs
sfackler: I'm not sure we can rely on this.

sfackler:
Ping

mredlek:
Sorry, I missed your review. Thanks for your review.

I will change the second point.

About the first point: a quick grep doesn't result in NULL-checks around that function, which suggest the return in not nullable. However, I also didn't find any deferences of the result. It usually ends in generic functions which does a NULL check. Of course, I could just add the Option<>, but if the field is really not-null this would be an unnessecairy check for the user. If this field turn out to be nullable, I also expect X509_get_subject_name to be nullable. I will do an additional investigation for this point.

I have a couple of generic questions:
* Should I squash the commits?
* Should I rebase or merge (if neccesairy)?
* The point about the conversion from c_long to i32 / i64 is still open.
* Where can I ask questions about a pull request I am still working on?

sfackler:
Cool - if other code calling the function doesn't check for null I think it's reasonable to just return the reference directly. Could you add an assert in the method body that checks that the result is non-null just in case?

I don't feel strongly about rebase/merge/squash - whatever you prefer.

Let's the `Asn1Integer` getters/setters as is for now. If necessary we can always deprecate these and add new ones in the future.

Feel free to ping me on IRC, or just open the PR in an unfinished state and we can look over it there.

sfackler:
Could you rebase/merge in master so the AppVeyor build starts working again?

openssl/src/x509/mod.rs
```diff
@@ -629,6 +629,34 @@ impl X509Req {
...
     to_der!(ffi::i2d_X509_REQ);
+
+    pub fn version(&self) -> i32
+    {
+        unsafe {
+            compat::X509_REQ_get_version(self.as_ptr()) as i32
+        }
+    }
+
+    pub fn set_version(&mut self, value: i32) -> Result<(), ErrorStack>
```
sfackler: `X509_REQ` is reference counted, so we unfortunately can't mutate it. These are covered in a separate PR: https://github.com/sfackler/rust-openssl/pull/515/files#diff-3c34fc20efa0aafbeb82758e99f718e1R719

sfackler:
I'm going to land this and then adjust things to work with #515. Thanks!

PR #579, opened 2017-02-11
X509 signature access

### LINKED ISSUE #576: Get signature from certificate
I feel like I might be missing something obvious here. After loading a certificate with one of the available X.509 methods (e.g. `X509::from_pem()`), how can I get access to (the bytes of) the certificates signature?

sfackler:
Does this provide what you're looking for @brooss?

Closes #576


brooss:
> Does this provide what you're looking for @brooss?

Yes, exactly what I was looking for. Thank you.

PR #583, opened 2017-02-19
Add note about needing pkg-config

SilverWingedSeraph:
I didn't have it and it took me a bit to figure out what the problem was.


sfackler:
Thanks!

PR #592, opened 2017-03-06
Fix for len() == isize::max() for stacks that are unallocated

Byron:
The culprit is that [`sk_num(stack)` can return -1](http://docs.huihoo.com/doxygen/openssl/1.0.1c/include_2openssl_2stack_8h.html#aefb72413520398ef3cf5cdbada71d492) as c_int if there is no stack allocated.

Previously, thanks to unsafe casts, this would result in a isize::max() for len() and iteration size if there was no stack.

Now this case is handled specifically, which fixes the issue.

_NOTE_: I ran into this issue when implementing [TLS for tokio-cassandra](https://github.com/nhellwig/tokio-cassandra/issues/5), which explains why the password for the keystore in the unit-test is `cassandra`.

There is at least [one other place](https://github.com/sfackler/rust-native-tls/blob/391170043fd87a4e3c909057cb079c3349fb9f25/src/imp/openssl.rs#L123) that is vulnerable this this issue.


sfackler:
The problem is I think a bit deeper than this - we can't be storing a null pointer as a `Stack` since `&StackRef` can never be null. I believe we'll want to change the `Pkcs12::decode` to allocate an empty stack if necessary for now, and then have a FIXME to change that field to an `Option<Stack>` in the next breaking release.

Byron:
Thanks for your reply! What I hear is that in addition to what I did, `Pkcs12::decode` should be adjusted to not allow that case, for now.
I will get right to work.

Please note that I am strongly believing the `stack_len()` function must handle the case that `-1` is returned, either way. We should not be the only ones creating those stacks during `Pkcs12::decode`, or should not assume this will be the case forever.

sfackler:
-1 can only be returned if the stack pointer is null, which is a thing that is not allowed. You're already in undefined behavior land.

Byron:
I just checked the code and saw the decoding is done by openssl. Thus I would conclude that for that library, it's fine to leave empty stacks unallocated, which makes sense for all kinds of vectors I guess (Rust certainly does the same). Getting -1 for a length seems like something that can happen all the time, as is indicated as possibility in [the openssl API docs](http://docs.huihoo.com/doxygen/openssl/1.0.1c/stack_8c_source.html#l00294).

Not acknowledging this oddity is setting this library up for all kinds of wrong behaviour - see the endless loop one gets when iterating over `isize::max_value()` non-existing items.

sfackler:
`len` is defined on `StackRef`. A `&StackRef` is constructed directly from an `*mut _STACK` - it is simply a pointer to that memory region. If the `*mut _STACK` is `NULL`, then you're calling `len` on a null `&StackRef`, which *cannot exist*. It is undefined behavior to even produce a value like that. Some function returning -1 is the least of your worries at that point.

Byron:
Thanks for taking the time to explain this matter in greater detail.

I think I get it now - the problem is the stack itself is not allocated, which should never happen.
Do you think it's worth it to protect from that right [where it happens](https://github.com/byron/rust-openssl/blob/ec2685347c8eba32234a14c4fdc81d6ce83dc97d/openssl/src/stack.rs#L89-L89), or assert for not-null?
In any case, the change you requested I can do now - I still wonder though whether a similar issue can be found in other places too - right now a `Stack` created from a null-ptr behaves incorrectly.

Byron:
Alright - I think I have implemented it as you envisioned it.

In the current version I have added an assertion in `Stack` to prevent that kind of thing from happening in the first place.

Unfortunately not all tests run on my machine (OSX) so I will rely on Travis for now - these 4 take forever/a long time.
```
test ssl::tests::get_peer_certificate::dtlsv1 has been running for over 60 seconds
test ssl::tests::ssl_verify_callback::dtlsv1 has been running for over 60 seconds
test ssl::tests::test_read_dtlsv1 has been running for over 60 seconds
test ssl::tests::test_write_dtlsv1 has been running for over 60 seconds
```

Byron:
@sfackler This seems to work - is there anything else that should be done?

sfackler:
Looks good - sorry for the delay!

The DTLS tests are a bit iffy - they depend on the `openssl` shell program which will default to one that's too old to support DTLS on OSX.

PR #598, opened 2017-03-13
(maint) Recreate ability to pass in OPENSSL_LIBS variable

ajroetker:
Prior to this commit in 43c951f743e68fac5f45119eda7c994882a1d489 the
ability to pass OPENSSL_LIBS was removed from the build.rs of
openssl-sys. This commit adds the ability to pass custom names for the
OPENSSL_LIBS back in. This is useful for when building openssl across
linux and windows with the same lib names (ssl:crypto) and the default
names provided by the build script are not valid.

### DIFF
--- a/openssl-sys/build.rs
+++ b/openssl-sys/build.rs
@@ -65,16 +65,23 @@ fn main() {
 
     let version = validate_headers(&[include_dir.clone().into()]);
 
-    let libs = match version {
-        Version::Openssl101 | Version::Openssl102 if target.contains("windows") => {
-            ["ssleay32", "libeay32"]
+    let libs_env = env::var("OPENSSL_LIBS").ok();
+    let libs = match libs_env {
+        Some(ref v) => v.split(":").collect(),
+        None => {
+            match version {
+                Version::Openssl101 | Version::Openssl102 if target.contains("windows") => {
+                    vec!["ssleay32", "libeay32"]
+                }
+                Version::Openssl110 if target.contains("windows") => vec!["libssl", "libcrypto"],
+                _ => vec!["ssl", "crypto"],
+            }
         }
-        Version::Openssl110 if target.contains("windows") => ["libssl", "libcrypto"],
-        _ => ["ssl", "crypto"],
     };
 
+
     let kind = determine_mode(Path::new(&lib_dir), &libs);
-    for lib in libs.iter() {
+    for lib in libs.into_iter() {
         println!("cargo:rustc-link-lib={}={}", kind, lib);
     }
 }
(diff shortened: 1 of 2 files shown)


ajroetker:
This is the relevant code that was deleted https://github.com/sfackler/rust-openssl/commit/43c951f743e68fac5f45119eda7c994882a1d489#diff-aca849e774a4e9d79fcc036a9de0d596L33

ajroetker:
@alexcrichton Was there a reason this functionality was removed or was the functionality of it deemed unnecessary?

alexcrichton:
Ah yeah I deleted this logic in 0.9 because openssl-sys got far more aggressive at detecting libs on the system. I'd consider it quite unfortunately to have to tell users "to build this crate please set this env var" when that's arguably logic that openssl-sys should include instead. 

I'd personally be more in favor of placing whatever necessary logic is required in openssl-sys to find the right names than to support the env var again, but that's mainly up to @sfackler

ajroetker:
@alexcrichton It seems like the problem I have is that we have nonstandard names of our openssl libs on windows w/ openssl 1.0.2 (ssl:crypto). Would your suggestion be to do something more clever than allowing an override, something like "find the name of openssl libraries directly" (I'm not quite sure how you'd do that reliably)?

openssl-sys/build.rs
```diff
@@ -63,18 +63,24 @@ fn main() {
...
-    let version = validate_headers(&[include_dir.clone().into()]);
-
-    let libs = match version {
-        Version::Openssl101 | Version::Openssl102 if target.contains("windows") => {
-            ["ssleay32", "libeay32"]
+    let libs_env = env::var("OPENSSL_LIBS").ok();
+    let libs = match libs_env {
+        Some(ref v) => v.split(":").collect(),
+        None => {
+            let version = validate_headers(&[include_dir.clone().into()]);
```
sfackler: This needs to be run unconditionally since it also discovers the build configuration.

sfackler:
Yep, I have no problem adding this functionality back IIRC we mostly removed it because we didn't think it was actually used.

ajroetker:
@sfackler thanks for taking a look, I amended the commit to move the validate_headers out of the conditional

sfackler:
Thanks! One last thing - could you update the README down [here](https://github.com/sfackler/rust-openssl#manual-configuration) to mention the new variable?

ajroetker:
@sfackler no problem, added a note to the README!

ajroetker:
@sfackler not sure what's going on with travis, I assume the failures are transients

sfackler:
Yeah, those are just some failures from travis-side caching.

metajack:
We just ran into this in Servo. Thanks for adding this back!

Issue #597, opened 2017-03-13
Deadlock on armv7-unknown-linux-gnueabihf


pgerber:
On armv7-unknown-linux-gnueabihf the following code leads to a deadlock:

```rust
extern crate hyper_openssl;

use hyper_openssl::OpensslClient;

fn main() {
    OpensslClient::new().unwrap();
}
```

Backtrace:
```
(gdb) bt
#0  0x76dae780 in __lll_lock_wait (futex=futex@entry=0x76817150, private=<optimized out>) at ../ports/sysdeps/unix/sysv/linux/arm/nptl/lowlevellock.c:46
#1  0x76da91a4 in __GI___pthread_mutex_lock (mutex=0x76817150) at pthread_mutex_lock.c:79
#2  0x54ab5260 in std::sys::imp::mutex::{{impl}}::lock (self=0x76817150) at /checkout/src/libstd/sys/unix/mutex.rs:67
#3  0x54ab4dac in std::sys_common::mutex::{{impl}}::lock (self=0x76817150) at /checkout/src/libstd/sys_common/mutex.rs:40
#4  0x54ab5bd4 in std::sync::mutex::{{impl}}::lock<()> (self=0x7682c260) at /checkout/src/libstd/sync/mutex.rs:221
#5  0x54ab8c9c in openssl_sys::ossl10x::locking_function (mode=9, n=12, _file=0x76fa065c "ssl_cert.c", _line=146) at /home/pg/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-sys-0.9.8/src/ossl10x.rs:720
#6  0x76f93ab8 in SSL_get_ex_data_X509_STORE_CTX_idx () from /usr/lib/arm-linux-gnueabihf/libssl.so.1.0.0
#7  0x76f91380 in SSL_CTX_new () from /usr/lib/arm-linux-gnueabihf/libssl.so.1.0.0
#8  0x54ab3188 in openssl::ssl::{{impl}}::new (method=...) at /home/pg/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-0.9.8/src/ssl/mod.rs:598
#9  0x54ab2ac8 in openssl::ssl::connector::ctx (method=...) at /home/pg/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-0.9.8/src/ssl/connector.rs:23
[...]

PR #601, opened 2017-03-16
Panic if lock managed by `locking_function` is doubly unlocked

pgerber:
Trying to unlock an unlocked lock is always an error and should
be treated as such.

This is related to #597.


sfackler:
Thanks!

PR #608, opened 2017-03-29
show help message when pkg-config is missing

BusyJay:
The [error message](https://github.com/sfackler/rust-openssl/blob/42ad50ae670397fae9db956e931ed935301f713b/openssl-sys/build.rs#L144) is supposed to be displayed when pkg-config is missing, but now it never shows up because the result is unwrapped directly when running pkg-config.


sfackler:
Ah, thank you!

PR #613, opened 2017-04-05
Add new EC/PKEY methods to permit deriving shared secrets.

aosmond:
WebPush [1] uses EC to derive a shared secret in order to generate a shared symmetric encryption key. I've added the missing API bits to do this:

[1] https://tools.ietf.org/html/draft-ietf-webpush-encryption-08#section-2


openssl/src/pkey.rs
```diff
@@ -216,18 +234,28 @@ impl PKeyCtxRef {
         };
         Ok(Padding::from_raw(pad))
     }
-}
 
-impl ForeignTypeRef for PKeyCtxRef {
-    type CType = ffi::EVP_PKEY_CTX;
+    pub fn derive_from_peer(&mut self, peer: &PKeyRef) -> Result<Vec<u8>, ErrorStack> {
+        unsafe {
+            try!(cvt(ffi::EVP_PKEY_derive_init(self.as_ptr())));
```
sfackler: Seems like it might make sense to define a method for each of these functions? From reading the man pages, it seems like there are some things you may want to do after setting the peer at the very least.

aosmond: Sounds good, I had the blinders on for my own use cases :).

sfackler:
Thanks!

Issue #600, opened 2017-03-15
Failing to compile against Openssl 1.0.2k built with no-ec2m


ajroetker:
platform: x86_64-pc-windows-gnu
rust: 1.15.1
openssl: 1.0.2k (Configured with no-ec2m)

Not sure what I'm doing wrong here but when trying to compile against rust-openssl-0.9.9 I'm seeing the following error at the final gcc calls:

```
= note: C:\cygwin64\var\tmp\tmp.20T4aqC2hk\puppet-code\target\debug\deps\libopenssl-3e666053462ffb12.rlib(openssl-3e666053462ffb12.0.o): In function `openssl::ec::{{impl}}::components_gf2m':
C:\Users\Administrator\.cargo\registry\src\github.com-1ecc6299db9ec823\openssl-0.9.9\src/ec.rs:76: undefined reference to `EC_GROUP_get_curve_GF2m'
```

As far as I can tell that symbol *isn't* defined because our openssl is built with no-ec2m aka the OPENSSL_NO_EC is defined and those symbols don't make it into the header files (they're wrapped in an `ifndef OPENSSL_NO_EC`.

It looks like the crate handles `OPENSSL_NO_EC` in some cases, e.g. https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/src/ossl10x.rs#L421. Do some of EC functions in https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/src/lib.rs#L1530 need to also be conditionally compiled (similarly with their calls in openssl/src/ec.rs) or am I offbase and my error has come from elsewhere?

Thanks for the help!

ajroetker:
Also I confirmed that building openssl without the no-ec2m specified solves the problem.

sfackler:
I'm assuming the behavior difference you're seeing is due to differences in how the Windows and Linux linkers work - the Linux one is apparently okay with an unresolved symbol in a function that is never called, and Windows isn't.

The general policy right now is to deal with `OPENSSL_NO_FOO` on an on-demand basis - there are basically an infinite number of such toggles and I don't really want to try to deal with them all :). I'd be more than happy to take a PR adding the right cfgs. You'll want to cfg-off both the extern definitions in the sys crate as well as the wrappers in the non-sys crate.

PR #615, opened 2017-04-11
(issues-600) Avoid compiling ec code against no-ec2m openssl

ajroetker:
This commit avoids defining code that leads to undefined references when
compiling against an openssl built with no ec2m.

### DIFF
--- a/openssl-sys/build.rs
+++ b/openssl-sys/build.rs
@@ -15,6 +15,7 @@ const DEFINES: &'static [&'static str] = &[
     "OPENSSL_NO_BUF_FREELISTS",
     "OPENSSL_NO_COMP",
     "OPENSSL_NO_EC",
+    "OPENSSL_NO_EC2M",
     "OPENSSL_NO_ENGINE",
     "OPENSSL_NO_KRB5",
     "OPENSSL_NO_NEXTPROTONEG",
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -1521,13 +1521,16 @@ extern {
     pub fn EC_KEY_check_key(key: *const EC_KEY) -> c_int;
     pub fn EC_KEY_free(key: *mut EC_KEY);
 
+    #[cfg(not(osslconf = "OPENSSL_NO_EC2M"))]
     pub fn EC_GF2m_simple_method() -> *const EC_METHOD;
 
     pub fn EC_GROUP_new(meth: *const EC_METHOD) -> *mut EC_GROUP;
     pub fn EC_GROUP_new_curve_GFp(p: *const BIGNUM, a: *const BIGNUM, b: *const BIGNUM, ctx: *mut BN_CTX) -> *mut EC_GROUP;
+    #[cfg(not(osslconf = "OPENSSL_NO_EC2M"))]
     pub fn EC_GROUP_new_curve_GF2m(p: *const BIGNUM, a: *const BIGNUM, b: *const BIGNUM, ctx: *mut BN_CTX) -> *mut EC_GROUP;
     pub fn EC_GROUP_new_by_curve_name(nid: c_int) -> *mut EC_GROUP;
     pub fn EC_GROUP_get_curve_GFp(group: *const EC_GROUP, p: *mut BIGNUM, a: *mut BIGNUM, b: *mut BIGNUM, ctx: *mut BN_CTX) -> c_int;
+    #[cfg(not(osslconf = "OPENSSL_NO_EC2M"))]
     pub fn EC_GROUP_get_curve_GF2m(group: *const EC_GROUP, p: *mut BIGNUM, a: *mut BIGNUM, b: *mut BIGNUM, ctx: *mut BN_CTX) -> c_int;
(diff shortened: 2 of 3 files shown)


ajroetker:
Wrt https://github.com/sfackler/rust-openssl/issues/600

sfackler:
I think you'll also need to flag off some methods in `SslContextBuilder` as well that reference `EcKey`s.

ajroetker:
@sfackler I wasn't quite sure if I should wrap individual ffi calls in ec/ec_key with `#[cfg(not(osslconf = "OPENSSL_NO_EC"))]` or if wrapping the module defines would be sufficient.

sfackler:
Disabling `ec` and `ec_key` is part, but then there are still methods defined elsewhere like https://docs.rs/openssl/0.9.10/openssl/ssl/struct.SslContextBuilder.html#method.set_tmp_ecdh or https://docs.rs/openssl/0.9.10/openssl/pkey/struct.PKey.html#method.from_ec_key

ajroetker:
Compiling this against a no-ec2m doesn't work atm looks like I need to add an OPENSSL_NO_EC2M rather than OPENSSL_NO_EC

ajroetker:
@sfackler So I have indeed verified that this fixes the ec2m compilation error, but I'm still seeing some weird behavior even with this patch.

```
    Finished dev [unoptimized + debuginfo] target(s) in 0.1 secs
     Running target\debug\deps\puppet_code-d63094cbd2c32330.exe
error: test failed
C:/tools/pl-build-tools/bin/cargo.exe: error while loading shared libraries: ?: cannot open shared object file: No such file or directory
```

And if I use static linking I also get a bunch of undefined ref errors

```
  = note: C:\cygwin64\var\tmp\tmp.Urf68jH1cm\puppet-code\target\debug\deps\libopenssl_sys-9270f3fe0429c564.rlib(rand_win.o):rand_win.c:(.text+0xd8b): undefined reference to `__imp_GetDeviceCaps'
          C:\cygwin64\var\tmp\tmp.Urf68jH1cm\puppet-code\target\debug\deps\libopenssl_sys-9270f3fe0429c564.rlib(rand_win.o):rand_win.c:(.text+0xdb8): undefined reference to `__imp_CreateCompatibleBitmap'
          C:\cygwin64\var\tmp\tmp.Urf68jH1cm\puppet-code\target\debug\deps\libopenssl_sys-9270f3fe0429c564.rlib(rand_win.o):rand_win.c:(.text+0xdce): undefined reference to `__imp_GetObjectW'
          C:\cygwin64\var\tmp\tmp.Urf68jH1cm\puppet-code\target\debug\deps\libopenssl_sys-9270f3fe0429c564.rlib(rand_win.o):rand_win.c:(.text+0xe91): undefined reference to `__imp_GetDIBits'
          C:\cygwin64\var\tmp\tmp.Urf68jH1cm\puppet-code\target\debug\deps\libopenssl_sys-9270f3fe0429c564.rlib(rand_win.o):rand_win.c:(.text+0xf20): undefined reference to `__imp_DeleteObject'
          ld: C:\cygwin64\var\tmp\tmp.Urf68jH1cm\puppet-code\target\debug\deps\libopenssl_sys-9270f3fe0429c564.rlib(rand_win.o): bad reloc address 0x0 in section `.pdata'
          ld: final link failed: Invalid operation
```

I verified that compiling openssl without no-ec2m does fix these errors, I'm not quite sure what's going wrong. Any suggestions?

sfackler:
Uh, that seems weird. Does your copy of OpenSSL work when linked to other things?

ajroetker:
@sfackler Yep the same copy of openssl works fine when linked to rust-openssl 0.7

ajroetker:
Also if I build and try and run the executable (vs running tests) I see this weird error
```
C:/cygwin64/var/tmp/tmp.Urf68jH1cm/puppet-code/target/debug/puppet-code.exe: error while loading shared libraries: SSLEAY32.dll: cannot open shared object file: No such file or directory
```

My env variables look like
```
echo $OPENSSL_LIBS
ssl:crypto

echo $OPENSSL_DIR
C:/ProgramFiles64Folder/PuppetLabs/Client/tools
```

sfackler:
Is C:/ProgramFiles64Folder/PuppetLabs/Client/tools/lib in your `PATH`?

ajroetker:
It was not @sfackler but that does not appear to make any difference

sfackler:
Uh, wait, where is the `SSLEAY32.dll` its referring to? The stuff in that lib directory seem different.

ajroetker:
@sfackler I'm not sure what's trying to look up `SSLEAY32.dll`, my understanding is that it should only be using libssl and libcrypto if I pass in `OPENSSL_LIBS=ssl:crypto`

```
$ nm ./target/debug/puppet-code.exe | grep SSLEAY
0000000000c63014 I _head_SSLEAY32_dll
0000000000c684d4 I SSLEAY32_dll_iname
```

ajroetker:
@sfackler Looks like my issue was that `C:/ProgramFiles64Folder/PuppetLabs/Client/tools/bin` wasn't on my path for Windows builds, putting that in fixed the issue above. So this PR should be good to go, thanks a ton for the help.

sfackler:
Thanks!

PR #616, opened 2017-04-12
Don't force allocation for message digests

sfackler:
@bascule

### DIFF
--- a/openssl/src/hash.rs
+++ b/openssl/src/hash.rs
@@ -1,5 +1,7 @@
 use std::io::prelude::*;
 use std::io;
+use std::ops::{Deref, DerefMut};
+use std::fmt;
 use ffi;
 
 #[cfg(ossl110)]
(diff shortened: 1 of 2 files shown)


openssl/src/hash.rs
```diff
@@ -192,34 +202,95 @@ impl Drop for Hasher {
...
+    #[inline]
+    fn as_ref(&self) -> &[u8] {
+        self.deref()
+    }
+}
+
+impl PartialEq for DigestBytes {
+    #[inline]
+    fn eq(&self, other: &DigestBytes) -> bool {
+        &**self == &**other
```
Philipp91: constant-time comparison?

sfackler: Ah sure, good call. I'm not sure how useful this impl is anyway since it seems a bit rare to be comparing two `DigestBytes` rather than e.g. a `DigestBytes` and a `Vec<u8>`. I only added it for testing convenience but that's easy to fix. Maybe it's best to just remove it?

PR #622, opened 2017-04-24
try to find openssl libraries in a vcpkg ports tree

mcgoo:
Optionally find libs/dlls for MSVC ABI builds in a ports tree built by Microsoft's newish [vcpkg](https://github.com/Microsoft/vcpkg)

I'll send you a PR for curl-sys shortly which is a little less straightforward, and we can discuss there.

cc @alexcrichton

### DIFF
--- a/openssl-sys/build.rs
+++ b/openssl-sys/build.rs
@@ -1,4 +1,6 @@
 extern crate pkg_config;
+#[cfg(target_env = "msvc")]
+extern crate vcpkg;
 extern crate gcc;
 
 use std::collections::HashSet;
@@ -98,6 +100,7 @@ fn find_openssl_dir(target: &str) -> OsString {
     }
 
     try_pkg_config();
+    try_vcpkg();
 
     let mut msg = format!("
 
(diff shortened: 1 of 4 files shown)


openssl-sys/build.rs
```diff
@@ -217,6 +219,33 @@ fn try_pkg_config() {
...
 
+/// Attempt to find OpenSSL through vcpkg.
+///
+/// Note that if this succeeds then the function does not return as vcpkg
+/// should tell us all the information that we need.
+fn try_vcpkg() {
+
+    // vcpkg will decline if it can not handle the target
+    let lib = match vcpkg::Config::new()
+        .lib_name("libeay32")
```
sfackler: These were renamed to `libcrypto` and `libssl` in OpenSSL 1.1.0 to match up with other platforms. Does vcpkg support OpenSSL 1.1.0 right now?

sfackler:
Awesome! Is vcpkg mature enough that it'd make sense to update the README to make it the recommended route to get OpenSSL?

mcgoo:
I’m not positive, but it looks pretty good to me. It’s only 6 months old but a few months ago it suddenly hit the point of always working for my tree (a nightmare mix of boost, several database drivers, protobuf3, zlib, wt, qt5, zeromq) and people are adding ports pretty rapidly now.

I know they are working on making it work with the standalone Visual C++ Build Tools but at present it’s documentation says it requires VS2015 or VS2017 on the machine. I’m not sure what impact that would have on your CI infrastructure, or user base. 

The current version of the build helper does nothing if there is no vcpkg tree on the machine, so the CI tests that just ran are not exercising anything but the “I give up” case.

If you are interested in trying it out you either need to install vcpkg, install some ports (`vcpkg install openssl:x64-windows openssl:x64-windows-static`) and run `vcpkg integrate install` or set the VCPKG_ROOT environment variable to point to your vcpkg tree. (`vcpkg integrate install` will add it to the default include/lib paths for any msbuild/visual studio which may or may not be what you want if you ever use those tools.)

openssl-sys/build.rs
mcgoo: It is at 1.0.2k right now. There is only a single version of each port in the tree. https://github.com/Microsoft/vcpkg/issues/218 for upgrade discussion.

I can try and futureproof it by probing for the 1.1.0 names also?

mcgoo:
Hi @sfackler. Here is the PR I was talking about for the curl-sys changes. https://github.com/alexcrichton/curl-rust/pull/166.

I think this PR should be complete now - it works for me in the context of the curl stuff and also pq-sys which was the motivation for doing it in the first place, so unlike the other PRs it would be great if you are willing to merge this.  

With this it's pretty easy to get a working openssl (1.0.2k) on msvc, with static or dynamic linkage.

Now I look it over again, I guess it would probably be better to have the 1.1.0 instructions above the 1.0.2.

(The CI is currently broken because it can't download the .pem file.)

Thanks for taking the time to consider this.

mcgoo:
Cleaned up and squashed. I rearranged the docs to put use of 1.1.0 back up the top and added an appveyor build that exercises the vcpkg functionality. The appveyor build links dynamically, but you can also do a fully static build with `RUSTFLAGS=-Ctarget-feature=+crt-static`.

sfackler:
Oops, sorry for leaving this for so long!

Issue #625, opened 2017-05-03
Doesn't compile with target i686-linux-android


chshawkn:
On MacOS 10.12.4
rustup 1.2.0 (70faf07 2017-04-08)
rustc 1.17.0 (56124baa9 2017-04-24)

when `cargo build --target=i686-linux-android` got error:
```bash
error[E0308]: mismatched types
  --> /Users/user/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-0.9.11/src/pkcs12.rs:38:40
   |
38 |                                        pass.as_ptr(),
   |                                        ^^^^^^^^^^^^^ expected i8, found u8
   |
   = note: expected type `*const i8`
              found type `*const u8`
   = help: here are some functions which might fulfill your needs:
           - .offset(...)
           - .wrapping_offset(...)

error[E0308]: mismatched types
  --> /Users/user/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-0.9.11/src/string.rs:52:40
   |
52 |             let slice = CStr::from_ptr(self.as_ptr()).to_bytes();
   |                                        ^^^^^^^^^^^^^ expected u8, found i8
   |
   = note: expected type `*const u8`
              found type `*mut i8`

error[E0308]: mismatched types
  --> /Users/user/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-0.9.11/src/version.rs:57:29
   |
57 |     unsafe { CStr::from_ptr(OpenSSL_version(OPENSSL_VERSION)).to_str().unwrap() }
   |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected u8, found i8
   |
   = note: expected type `*const u8`
              found type `*const i8`
   = help: here are some functions which might fulfill your needs:
           - .offset(...)
           - .wrapping_offset(...)

error[E0308]: mismatched types
  --> /Users/user/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-0.9.11/src/version.rs:63:29
   |
63 |     unsafe { CStr::from_ptr(OpenSSL_version(OPENSSL_CFLAGS)).to_str().unwrap() }
   |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected u8, found i8
   |
   = note: expected type `*const u8`
              found type `*const i8`
   = help: here are some functions which might fulfill your needs:
           - .offset(...)
           - .wrapping_offset(...)

error[E0308]: mismatched types
[...]

PR #626, opened 2017-05-03
fix cross compile with target i686-linux-android

chshawkn:
fix https://github.com/sfackler/rust-openssl/issues/625


sfackler:
I believe this is due to a fix in libc's definition of c_char on Android. Does the build work on using nightly Rust?

chshawkn:
I didn't try nightly Rust.

sfackler:
Everything should work fine as is if you use a rustc that includes https://github.com/rust-lang/rust/pull/40317.

chshawkn:
@sfackler Thanks. You are right, it works!

PR #634, opened 2017-05-12
Make set_ecdh_auto available when using OpenSSL 1.1.0

badboy:
I tried compiling [sozu and failed due to having OpenSSL 1.1.0 locally](https://github.com/sozu-proxy/sozu/issues/178).

With these minimal fixes I was able to build sozu against OpenSSL 1.1.0.
I just copied over the necessary definitions from `ossl10x.rs`.

I am no expert in OpenSSL or this Rust wrapper, so I wonder if we need to bring `ossl10x.rs` and `ossl110.rs` closer together again to avoid such problems, where APIs are available in both versions.

I run `cargo test` locally.

Happy  to help working this out correctly.

### DIFF
--- a/openssl-sys/src/ossl110.rs
+++ b/openssl-sys/src/ossl110.rs
@@ -1,4 +1,5 @@
 use libc::{c_int, c_void, c_char, c_uchar, c_ulong, c_long, c_uint, size_t};
+use std::ptr;
 
 pub enum BIGNUM {}
 pub enum BIO {}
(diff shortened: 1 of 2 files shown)


sfackler:
`SSL_CTRL_SET_ECDH_AUTO` doesn't exist in OpenSSL 1.1.0.

badboy:
I see, thanks. I'll check what it actually does and see what's the way to go with 1.1.0

sfackler:
set_ecdh_auto only existed in 1.0.2. The behavior is always enabled in 1.1.0.

badboy:
Oh, that makes it even easier to fix for sozu then, thanks!

PR #635, opened 2017-05-12
use appveyor DownloadFile to download cacert.pem

mcgoo:
Neither Start-FileDownload nor curl seem to work but appveyor DownloadFile does.


sfackler:
Weird, thanks.

PR #641, opened 2017-05-26
Expose PSK via a SslContextBuilder::set_psk_callback method

luser:
This patch is definitely not in a state to be landed, but I figured I'd submit it for feedback. I needed TLS-PSK for a little side project I'm working on, and OpenSSL seemed to be the only thing I could use from Rust at the moment, but I needed to expose an API to set the PSK identity and key. This was fairly straightforward to do by adding a `SslContextBuilder::set_psk_callback` method. I'm not entirely sure if I covered all the necessary bases with the `openssl-sys` bits, and I didn't yet add documentation for the new method. I'm also not confident that the code I wrote in the `raw_psk` callback to fill the out parameters is written in the most sensible way.

I was able to get this to work for my use case, but I did have to explicitly call `set_cipher_list("PSK-AES128-CBC-SHA")`, which was annoying to track down, so maybe the API could be better designed to allow those ciphers in this case? Also I wound up calling `danger_connect_without_providing_domain_for_certificate_verification_and_server_name_indication`, which is not really the nicest thing, but I understand why it's named that. :)

If we wanted to make this friendlier I guess we could give `SslConnector` something like `fn connect_psk(&self, psk_identity: &str, psk: &[u8], stream: S) -> ...` and have it handle setting the callback and such internally? That might be more effort than is worthwhile for such a niche feature.

### DIFF
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -2252,14 +2252,15 @@ extern "C" {
                                                               arg: *mut c_void)
                                                               -> c_int,
                                             arg: *mut c_void);
+    #[cfg(not(osslconf = "OPENSSL_NO_PSK"))]
     pub fn SSL_CTX_set_psk_client_callback(ssl: *mut SSL_CTX,
-                                           psk_client_cb: extern "C" fn(*mut SSL,
-                                                                        *const c_char,
-                                                                        *mut c_char,
-                                                                        c_uint,
-                                                                        *mut c_uchar,
-                                                                        c_uint)
-                                                                        -> c_uint);
+                                           psk_client_cb: Option<extern "C" fn(*mut SSL,
+                                                                               *const c_char,
+                                                                               *mut c_char,
+                                                                               c_uint,
+                                                                               *mut c_uchar,
+                                                                               c_uint)
+                                                                               -> c_uint>);
     pub fn SSL_select_next_proto(out: *mut *mut c_uchar,
                                  outlen: *mut c_uchar,
                                  inbuf: *const c_uchar,
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -2252,6 +2252,14 @@ extern "C" {
                                                               arg: *mut c_void)
                                                               -> c_int,
                                             arg: *mut c_void);
+    pub fn SSL_CTX_set_psk_client_callback(ssl: *mut SSL_CTX,
+                                           psk_client_cb: extern "C" fn(*mut SSL,
+                                                                        *const c_char,
+                                                                        *mut c_char,
+                                                                        c_uint,
+                                                                        *mut c_uchar,
+                                                                        c_uint)
+                                                                        -> c_uint);
     pub fn SSL_select_next_proto(out: *mut *mut c_uchar,
                                  outlen: *mut c_uchar,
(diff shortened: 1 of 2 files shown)


openssl-sys/src/lib.rs
```diff
@@ -2252,6 +2252,14 @@ extern "C" {
                                                               arg: *mut c_void)
                                                               -> c_int,
                                             arg: *mut c_void);
+    pub fn SSL_CTX_set_psk_client_callback(ssl: *mut SSL_CTX,
+                                           psk_client_cb: extern "C" fn(*mut SSL,
```
sfackler: This should take an `Option<extern "C" fn(...)>`.

openssl/src/ssl/mod.rs
```diff
@@ -307,6 +307,39 @@ extern "C" fn raw_verify<F>(preverify_ok: c_int, x509_ctx: *mut ffi::X509_STORE_
     }
 }
 
+extern "C" fn raw_psk<F>(ssl: *mut ffi::SSL,
+                         hint: *const c_char,
+                         identity: *mut c_char,
+                         _max_identity_len: c_uint,
+                         psk: *mut c_uchar,
+                         _max_psk_len: c_uint) -> c_uint
+    where F: Fn(&mut SslRef, &str) -> Result<(String, Vec<u8>), ErrorStack> + Any + 'static + Sync + Send
```
sfackler: I would probably expose the OpenSSL API more directly - something like `Fn(&mut SslRef, hint: Option<&[u8]>, identity: &mut [u8], psk: &mut [u8]) -> Result<(usize, usize), ErrorStack>`.

openssl-sys/src/lib.rs
luser: I wasn't entirely sure of the convention here. Is `Option<...>` used for nullable function pointers?

openssl/src/ssl/mod.rs
luser: I wrapped it in a more Rust-friendly API because it seems like it would be error-prone to expose it at that level. Specifically, `identity` needs to be null-terminated. (Passing `hint` as an `Option` is much better though, I don't know why I didn't think of that!)

I pushed my code that's using this to GitHub so you can see what it looks like in practice with the current implementation:
https://github.com/luser/steamstreaming/blob/004611fdc8db949020e4e25011df361f01e1e39a/src/lib.rs#L86

I'm happy to do whatever you think is best here, though.

openssl-sys/src/lib.rs
sfackler: Yeah, function pointers aren't nullable so `Option` wrapping is the FFI-correct way to represent a nullable function pointer to C. It's a bit weird...

luser:
I changed the callback API to match what you suggested, and wrapped the PSK functions in what I hope is the right `#[cfg]` incantation to make things not break in configurations where the SSL library is built without PSK support. I also added documentation for the new API.

luser:
The AppVeyor failures look like they all failed to download the OpenSSL installer. I don't know if that's intermittent or if the URL is really broken there.

sfackler:
Sorry for forgetting about this, looks good!

luser:
Thanks! This wasn't really blocking me from doing work, it's just a side project of mine anyway. :)

PR #642, opened 2017-06-04
Add pkg-config to readme

oussama:
error: failed to run custom build command for `openssl-sys v0.9.13`
run pkg_config fail: "Failed to run `\"pkg-config\" \"--libs\" \"--cflags\" \"openssl\"`: No such file or directory (os error 2)"

Problem encountered on WSL (Ubuntu on windows).
Fixed by installing pkg-config

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ on them.
 
 ```bash
 # On Ubuntu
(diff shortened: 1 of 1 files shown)


sfackler:
Thanks!

PR #647, opened 2017-06-07
remove unused dependencies on windows

mcgoo:
These are never pulled in with `extern crate` so I don't think they are required.


sfackler:
These were added quite a long time ago - I believe because they're required at runtime but OpenSSL dists didn't pull them in automatically?

alexcrichton:
FWIW these do indeed have no effect if they're not linked into the crate (e.g. `extern crate foo`), so in that sense I guess it's ok nowadays? (unsure myself)

sfackler:
Well, might as well figure out!

PR #655, opened 2017-06-25
Enable build on LibreSSL 5.6.0 development branch.

ltratt:
Without this, openssl-sys can't compile on OpenBSD-current. As far as I can tell, the only differences with respect to LibreSSL 5.5.4 are additional exposed functions: there do not appear to be any breaking changes. Certainly all the test suites in the repository succeed with this commit.

### DIFF
--- a/openssl-sys/build.rs
+++ b/openssl-sys/build.rs
@@ -237,8 +237,10 @@ fn validate_headers(include_dirs: &[PathBuf]) -> Version {
 #include <openssl/opensslv.h>
 #include <openssl/opensslconf.h>
 
-#if LIBRESSL_VERSION_NUMBER >= 0x20505000
+#if LIBRESSL_VERSION_NUMBER >= 0x20601000
 RUST_LIBRESSL_NEW
+#elif LIBRESSL_VERSION_NUMBER >= 0x20600000
+RUST_LIBRESSL_260
 #elif LIBRESSL_VERSION_NUMBER >= 0x20504000
 RUST_LIBRESSL_254
 #elif LIBRESSL_VERSION_NUMBER >= 0x20503000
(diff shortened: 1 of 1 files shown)


sfackler:
Thanks! Can you bump the Travis build that's currently using 2.5.4 to 2.6.0 as well? https://github.com/sfackler/rust-openssl/blob/master/.travis.yml#L75

sfackler:
Looks like there isn't a 2.6.0 source tarball at that location yet. Might need to grab it from somewhere else?

ltratt:
Ah, right, I did wonder if that's what changing that variable might lead to!

I think I may not have been very clear in the original PR: sorry! 2.6.0 isn't out yet AFAIK, but the development branch (which is what those of us following OpenBSD's beta releases are now seeing) registers itself as 2.6.0. As things stand, that means folk like me can't build rust-openssl (or things that depend on it e.g. rustc) on our development boxes. But it does also mean that what "counts as" 2.6.0 could yet change...

sfackler:
Oh, sure. I'm totally on board with supporting 2.6.0, but I think you'll just have to tweak the test infrastructure a bit to pull a 2.6.0 tarball from wherever it's hosted.

ltratt:
I don't think LibreSSL does intermediate tarball releases (presumably 2.6.0 is officially scheduled for Sep/Oct-ish), so I'm not really sure how to go about doing this. We could, I suppose, do a git/cvs checkout of the development branch?

sfackler:
Hmm, we could do that, but maybe it's too much work at that point. It seems fine to land this without CI then - 2.6.0 is in beta after all :)

ltratt:
OK, shall I revert the travis patch?

sfackler:
Yep!

ltratt:
OK, done. It'll take a while for the tests, I guess.

sfackler:
Thanks!

ltratt:
Thanks!

Issue #659, opened 2017-07-09
Human readable error code handling


64:
I'm not sure if there's a better way of doing this, but currently I have some code which looks like this:

```rust
let e: openssl::error::Error = /* some error */;
if openssl_sys::ERR_GET_REASON(e.code()) == 396 /* SSL_R_VERSION_TOO_LOW */ {
   // Handle version too low
} else {
    // Handle other error
}
```

Could perhaps the error reason messages (such as those found [here](https://github.com/openssl/openssl/blob/master/include/openssl/sslerr.h)) be added to `openssl_sys` so that there are some human-readable error codes? I understand that the `e.reason()` function could be used to the same effect, but then you just end up with a bunch of string comparisons (which are also a pain to hard-code).

This crate is really good, I just think that the error handling is a bit unintuitive and could be improved in general.

sfackler:
Yeah, those definitely seem reasonable to pull in!

64:
In that case I might open a pull request later today, unless you're already on it. There are quite a lot of `xerr.h` files (such as that `sslerr.h` file I linked), should we pull in all the values from all of them?

sfackler:
The general approach is to pull things in on demand, but if we're pulling in those error constants it'd probably make sense to grab all of them.

PR #660, opened 2017-07-10
Add error reason and function codes to openssl-sys

64:
Fixes https://github.com/sfackler/rust-openssl/issues/659.

I wrote a small script ([here](https://gist.github.com/64/8ac13019f4faa491018aab6b5c265141)) to pull all the error reason and function codes from the `include/openssl` folder. You just give it the path to the folder where the headers are located and it generates an output file containing the list of error codes. If you wanted you could integrate the script into the `build.rs` file somehow but the issue is that my script only works off the back of https://github.com/openssl/openssl/commit/52df25cf2e656146cb3b206d8220124f0417d03f which separates the error codes into their own header files, so you won't be able to rely on it for older versions of OpenSSL.

Another issue is that there are some reason error code `#defines` in [`err.h`](https://github.com/openssl/openssl/blob/master/include/openssl/err.h#L170) which can't be converted by my tool yet since their values rely on other preprocessor macros.

Thoughts?


sfackler:
Thanks! It looks like the systest infrastructure need to be updated to pull in some extra headers that define these constants. I'm betting you'll also need to condition some that don't exist on older versions of OpenSSL.

64:
Ah yes, didn't see that. Even after including the relevant files, I still have about 225 errors which I presume are due to additions since the release of the version on my system (1.1.0f) and now (where I pulled the error codes from). 

Not entirely sure how to proceed with this, but I think a good option would be to disable testing for all the error codes this PR adds (if that's possible). Unless newer versions of OpenSSL have removed or changed the value of error codes from previous OpenSSL versions then there *should* be no issues with doing that; you'll just end up with a bunch of constants that may or may not actually exist on your actual OpenSSL installation.

Alternatively, it might be possible to adapt my tool to generate the above codes automatically for the version of OpenSSL that the user chooses to use. Either way, it's a tricky situation.

sfackler:
I'm not particularly confident that none of these codes have changed values over time - the most reliable approach would probably be to grab the definitions for each release we care about and then figure out what the diffs are and put things in the right places wrt conditional compilation.

64:
I'll try and get a list of the error codes for OpenSSL 1.0.1, 1.0.2 and 1.1.0 (latest patches each) and I'll get back to you then.

sfackler:
Cool, thanks.

I keep around builds of each of 1.0.1, 1.0.2 and 1.1.0 and switch between them with `OPENSSL_DIR` when adding stuff - it makes it a bit less painful since you have the source + headers for everything there for you.

64:
Just did a small tweak to my tool to get it to spit out values from older OpenSSL versions. Here are the resultant files:

https://gist.github.com/64/5b10ad1e2142970fefecdaefa2b339b9

So, diff you say? Or we could just straight up `include!` the appropriate file. Testing this is probably going to be an issue too.

sfackler:
I'd mildly prefer figuring out what's shared by all of them and not duplicating that bit.

64:
https://gist.github.com/64/c84e9e95eb23746dae16b6027d26775d

There are four files here:

* The intersection of all the files (at the bottom of the gist)
* What you need to add to the intersection to get OpenSSL 1.0.1
* What you need to add to the intersection to get OpenSSL 1.0.2
* What you need to add to the intersection to get OpenSSL 1.1.0

Does that sound good?

sfackler:
Yep!

64:
Alright, now I just need to get the tests to work.

64:
Looks like the AppVeyor failures aren't related. Not entirely sure why the Travis build failed too.

sfackler:
Yeah, don't worry about the Appveyor failures - they're having some firewall configuration problems right now.

There might possibly be a non-header file in the directory you're scanning?

PR #668, opened 2017-07-22
Mention Debian for apt-get instructions

mgeisler:
Since Debian is the upstream for Ubuntu (and many other distributions), I think it makes sense to mention it explicitly.


sfackler:
Thanks!

PR #673, opened 2017-08-01
Added methods to convert Asn1TimeRef to a Unix time stamp.

illegalprime:
I have to verify a bunch of SSL certificates and I couldn't find another way to do it.

### DIFF
--- a/openssl/src/asn1.rs
+++ b/openssl/src/asn1.rs
@@ -12,6 +12,12 @@ use error::ErrorStack;
 use nid::Nid;
 use string::OpensslString;
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct TimeDiff {
+    pub days: c_int,
+    pub secs: c_int,
+}
+
 foreign_type! {
     type CType = ffi::ASN1_GENERALIZEDTIME;
     fn drop = ffi::ASN1_GENERALIZEDTIME_free;
@@ -53,12 +59,12 @@ impl fmt::Display for Asn1TimeRef {
 
 impl Asn1TimeRef {
     #[cfg(any(ossl102, ossl110))]
-    pub fn as_unix(&self) -> Result<time_t, ErrorStack> {
+    pub fn diff_from_epoch(&self) -> Result<TimeDiff, ErrorStack> {
         Asn1Time::from_unix(0)?.diff(Some(self))
     }
 
     #[cfg(any(ossl102, ossl110))]
-    pub fn diff(&self, compare: Option<&Self>) -> Result<time_t, ErrorStack> {
+    pub fn diff(&self, compare: Option<&Self>) -> Result<TimeDiff, ErrorStack> {
         let mut days = 0;
         let mut seconds = 0;
         let other = compare.map(Self::as_ptr).unwrap_or_else(ptr::null_mut);
@@ -69,7 +75,10 @@ impl Asn1TimeRef {
 
         match err {
             0 => Err(ErrorStack::get()),
-            _ => Ok(days as time_t * 24 * 60 * 60 + seconds as time_t),
+            _ => Ok(TimeDiff {
+                days: days,
+                secs: seconds,
+            }),
         }
     }
 }
@@ -90,6 +99,8 @@ impl Asn1Time {
     }
 
     pub fn from_unix(time: time_t) -> Result<Self, ErrorStack> {
+        ffi::init();
+
         unsafe {
             let asntime = ffi::ASN1_TIME_set(ptr::null_mut(), time);
             match asntime.is_null() {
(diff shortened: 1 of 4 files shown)


openssl/src/asn1.rs
```diff
@@ -51,6 +51,25 @@ impl fmt::Display for Asn1TimeRef {
     }
 }
 
+impl Asn1TimeRef {
+    pub fn as_unix(&self) -> Result<i64, ErrorStack> {
```
sfackler: I think it might make more sense to wrap these OpenSSL functions more directly.

openssl/src/asn1.rs
```diff
@@ -51,6 +51,25 @@ impl fmt::Display for Asn1TimeRef {
...
+        let mut days = 0;
+        let mut seconds = 0;
+
+        unsafe {
+            let epoch = ffi::ASN1_TIME_set(ptr::null_mut(), 0);
```
sfackler: This also leaks `epoch`.

illegalprime: :+1:

openssl/src/asn1.rs
illegalprime: as in split it up into separate `set` and `diff` functions?

illegalprime:
@sfackler some checks seem to be failing at the setup steps:

https://circleci.com/gh/sfackler/rust-openssl/674?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

sfackler:
That failure is weird - I can repro it locally but that install was working fine as of 3 days ago... I'm poking around.

sfackler:
Ah, you should be good to go if you rebase against current master.

illegalprime:
yup that did it! Thanks!

openssl/src/asn1.rs
```diff
@@ -51,6 +51,29 @@ impl fmt::Display for Asn1TimeRef {
...
 }
 
+impl Asn1TimeRef {
+    #[cfg(any(ossl102, ossl110))]
+    pub fn as_unix(&self) -> Result<time_t, ErrorStack> {
+        Asn1Time::from_unix(0)?.diff(Some(self))
+    }
+
+    #[cfg(any(ossl102, ossl110))]
+    pub fn diff(&self, compare: Option<&Self>) -> Result<time_t, ErrorStack> {
```
sfackler: Can this return a `struct TimeDiff { days, secs }` instead? You can't generally assume days are the same length.

openssl/src/asn1.rs
```diff
@@ -65,6 +88,16 @@ impl Asn1Time {
     pub fn days_from_now(days: u32) -> Result<Asn1Time, ErrorStack> {
         Asn1Time::from_period(days as c_long * 60 * 60 * 24)
     }
+
+    pub fn from_unix(time: time_t) -> Result<Self, ErrorStack> {
+        unsafe {
+            let asntime = ffi::ASN1_TIME_set(ptr::null_mut(), time);
```
sfackler: This needs to call `ffi::init` first.

openssl/src/asn1.rs
illegalprime: Then `as_unix` would be difficult to implement, also a constant like this is used here: https://github.com/sfackler/rust-openssl/blob/master/openssl/src/asn1.rs#L66

openssl/src/asn1.rs
illegalprime: :+1:

openssl/src/asn1.rs
sfackler: Why would it be difficult to implement? This function would return exactly the same information as `ASN1_time_diff` does.

sfackler: That constructor should probably not exist.

openssl/src/asn1.rs
```diff
@@ -51,6 +57,32 @@ impl fmt::Display for Asn1TimeRef {
     }
 }
 
+impl Asn1TimeRef {
+    #[cfg(any(ossl102, ossl110))]
+    pub fn diff_from_epoch(&self) -> Result<TimeDiff, ErrorStack> {
+        Asn1Time::from_unix(0)?.diff(Some(self))
+    }
+
+    #[cfg(any(ossl102, ossl110))]
```
sfackler: These should be flagged on features like here: https://github.com/sfackler/rust-openssl/blob/master/openssl/src/ssl/mod.rs#L1415

illegalprime: I think the link broke, what do you mean?

sfackler: https://github.com/sfackler/rust-openssl/blob/master/openssl/src/ssl/mod.rs#L1186-L1203

clementsnahum:
Is there a reason this PR was closed? It looks like useful functionality.

clementsnahum:
@sfackler if I were to resurrect this PR would you have any concerns about getting it merged? It looks as though you might be considering alternatives for comparing `Asn1Time` (https://github.com/sfackler/rust-openssl/issues/1157)

sfackler:
Please do!

PR #674, opened 2017-08-09
Add elliptic curve functions

bradleybeddoes:
I've been working with Rust and JWK, ([RFC 7517](https://tools.ietf.org/html/rfc7517)), extensively over the past few months and finally got around to looking into use Elliptic Curves for various signing/encryption activities.

This PR adds support for: 

* Using affine coordinates with `EcKeyBuilderRef` to set the public key
* Using a `BigNum` with `EcKeyBuilderRef` to set the private key

A useful source JWA example is provided in RFC 7517 within [Appendix A.2](https://tools.ietf.org/html/rfc7517#appendix-A.2) which has formed the basis of the tests provided here.

This is my first excursion into Rust FFI so please excuse any missteps here. I'm more than happy to address any feedback to help with merge.

Thanks!


openssl/src/ec.rs
```diff
@@ -459,12 +459,35 @@ impl EcKeyBuilderRef {
     pub fn generate_key(&mut self) -> Result<&mut EcKeyBuilderRef, ErrorStack> {
         unsafe { cvt(ffi::EC_KEY_generate_key(self.as_ptr())).map(|_| self) }
     }
+
+    /// Sets the public key based on affine coordinates.
+    pub fn set_public_key_affine_coordinates(&mut self,
+                                             x: &mut BigNumRef,
```
sfackler: Are these inputs modified by `EC_KEY_set_public_key_affine_coordinate`? The documentation doesn't appear to say one way or another.

sfackler:
Seems reasonable to me other than the mutability question!

openssl/src/ec.rs
bradleybeddoes: My checks seem to indicate that `x` and `y` aren't modified, at least not right now.

Unfortunately, the OpenSSL definition does not use `const` pointers on this function for `x` and `y`. I changed the Rust signature to reflect this.

n.b. I originally used `const` pointers in the extern and non-mutable references for `x` and `y` here but that led to failures in CI due to pointer mismatch.

sfackler: Oh yeah, I wouldn't treat const-ness at the C API level as much of an indicator - the usage there is pretty inconsistent. Source diving is usually the only way to figure out mutability/ownership semantics of functions :(

bradleybeddoes: My reading of how the implementation calls out within OpenSSL (`EC_POINT_set_affine_coordinates_GFp` /`EC_POINT_set_affine_coordinates_GF2m`) is that no modification will be made to `x` / `y` and my local checks backed that up.

I changed the Rust signature out of an abundance of caution. However, if the approach has previously been to make the Rust level more reflective of what is actually going on in OpenSSL regardless of the C API I am happy with that.

Let me know your preference here and I'll fix this up.

sfackler: Cool - let's go with `&BigNumRef`s then.

sfackler:
Thanks!

PR #675, opened 2017-08-09
added cms decryption

sdemos:
I needed cms decryption for a project so I added the bindings and figured I should send them upstream. 

I added bindings for the `CMS_decrypt` function, as well as the `SMIME_read_CMS` and `CMS_ContentInfo_free` functions for creating and freeing the `CMS_ContentInfo` struct. I'm using it in my project and everything works as expected. 

This is my first experience with the ffi so I just emulated how some of the other files did it. Let me know if there is anything else I should do to help get it upstream.

### DIFF
--- a/openssl/src/cms.rs
+++ b/openssl/src/cms.rs
@@ -1,0 +1,60 @@
+//! CMS archive
+
+use ffi;
+use foreign_types::{ForeignType, ForeignTypeRef};
+use std::ptr;
+use error::ErrorStack;
+
+use bio::{MemBio, MemBioSlice};
+
+use x509::X509;
+use pkey::PKeyRef;
+
+use cvt;
+use cvt_p;
+
+foreign_type! {
+    type CType = ffi::CMS_ContentInfo;
+    fn drop = ffi::CMS_ContentInfo_free;
+
+    pub struct CmsContentInfo;
+    pub struct CmsContentInfoRef;
+}
+
+impl CmsContentInfoRef {
+    pub fn decrypt(&self, pkey: &PKeyRef, cert: &X509) -> Result<Vec<u8>, ErrorStack> {
+        unsafe {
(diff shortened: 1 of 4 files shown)


sdemos:
Apparently `libressl` doesn't have cms decryption, at least I couldn't find it after a little searching, so I added the cfg macro above all the cms functions I added so the functions don't exist when using `libressl`.

openssl/src/cms.rs
```diff
@@ -0,0 +1,63 @@
...
+
+use cvt;
+use cvt_p;
+
+#[cfg(not(libressl))]
+foreign_type! {
+    type CType = ffi::CMS_ContentInfo;
+    fn drop = ffi::CMS_ContentInfo_free;
+
+    pub struct Cms;
```
sfackler: Do you think these should be called `Cms` or `CmsContentInfo`? I usually like to use the OpenSSL name when possible.

openssl/src/cms.rs
```diff
@@ -0,0 +1,63 @@
...
+
+use bio::{MemBio, MemBioSlice};
+
+use x509::X509;
+use pkey::PKeyRef;
+
+use cvt;
+use cvt_p;
+
+#[cfg(not(libressl))]
```
sfackler: I'd tag the entire module with this instead of sticking it on everything in here.

openssl/src/cms.rs
sdemos: I didn't have any particular reason for choosing `Cms`. It's probably better to stick to the OpenSSL name.

sdemos:
I fixed the things you mentioned. It looks like one of the tests failed, but I don't see how it could be related to my changes.

sfackler:
Oh, don't worry about the DTLS tests - they're super flaky. I actually got rid of them last night. Thanks!

PR #683, opened 2017-08-21
Add more elliptic curve functions 

bradleybeddoes:
Provides ability to get affine coordinates from an EcPoint.

The initial use case here is creating JWK representations as defined within RFC 7517 from an EcKey created via a PEM source.

### DIFF
--- a/openssl/src/ec.rs
+++ b/openssl/src/ec.rs
@@ -273,7 +273,7 @@ impl EcPointRef {
     }
 
     /// Place affine coordinates of a curve over a prime field in the provided x and y BigNum's
-    pub fn get_affine_coordinates_gfp(
+    pub fn affine_coordinates_gfp(
         &self,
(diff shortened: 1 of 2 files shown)


openssl/src/ec.rs
```diff
@@ -271,6 +271,45 @@ impl EcPointRef {
             Ok(res == 0)
         }
     }
+
+    /// Place affine coordinates of a curve over a prime field in the provided x and y BigNum's
+    pub fn get_affine_coordinates_gfp(
```
sfackler: Could you remove the `get_` prefix? Rust APIs typically omit it.

sfackler:
Thanks! LGTM other than a little naming change.

bradleybeddoes:
No worries - should have been that way from the start, all sorted now.

sfackler:
Thanks!

PR #684, opened 2017-08-22
Support builds of OpenSSL from vendored source

### LINKED ISSUE #580: Consider adding the option to building openssl

sfackler:
Closes #580.

### DIFF
--- a/test/build_openssl.sh
+++ b/test/build_openssl.sh
@@ -5,8 +5,6 @@ if [ -d "${OPENSSL_DIR}" ]; then
     exit 0
 fi
 
-apt-get install -y --no-install-recommends curl
-
 case "${LIBRARY}" in
 "libressl")
     URL1="http://ftp3.usa.openbsd.org/pub/OpenBSD/LibreSSL/libressl-${VERSION}.tar.gz"
@@ -15,6 +13,9 @@ case "${LIBRARY}" in
 "openssl")
     URL1="https://openssl.org/source/openssl-${VERSION}.tar.gz"
     URL2="http://mirrors.ibiblio.org/openssl/source/openssl-${VERSION}.tar.gz"
+    ;;
+"")
+    exit 0
     ;;
 esac
 
@@ -32,6 +33,8 @@ case "${TARGET}" in
     export CC=arm-linux-gnueabihf-gcc
     ;;
 esac
+
+apt-get install -y --no-install-recommends curl
 
 mkdir -p /tmp/build
 cd /tmp/build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,7 @@ job: &JOB
   - run: ./test/build_openssl.sh
   - *SAVE_OPENSSL
   - *RESTORE_DEPS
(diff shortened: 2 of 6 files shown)


sfackler:
cc @alexcrichton

test/build_openssl.sh
```diff
@@ -33,6 +34,8 @@ case "${TARGET}" in
     ;;
 esac
 
+apt-get install -y --no-install-recommends curl
```
alexcrichton: Shouldn't this logic (and most of the logic in this script) end up getting removed in favor of the src crate?

sfackler: The src crate can only build 1.1.0.

alexcrichton: Bah :(

alexcrichton:
Some things I'd personally do before merging:

* [ ] Add documentation in the README for the feature
* [ ] Make sure there's a system in place for updating openssl-src -- I'd specifically like to talk to the infra team in Rust about this.

sfackler:
One other thing we should figure out is the upgrade story. 1.1.0 EOLs in just over a year: https://www.openssl.org/policies/releasestrat.html. We could simply say that we track the newest release?

alexcrichton:
Yes I'd imagine that the crate provides no guarantee really about what version it provides, other than:

* The version is compatible with the `openssl` crate
* The version is supported by the OpenSSL project at the time of publication
* The most recent version of the crate uses a version of OpenSSL with no known security vulnerabilities.

That is, I'd imagine 1.0.2 and 1.1.0 would both be fine to use today in openssl-src.

sfackler:
It is a bit nontrivial due to version-specific features. For example, if we went with 1.0.2, people would presumably want to use [`set_ecdh_auto`](https://github.com/sfackler/rust-openssl/blob/master/openssl/src/ssl/mod.rs#L625), but that's been removed in newer versions. Even for things that aren't removed, you'd need to enable the `v111` or `v120` or whatever feature to retain access.

We could disable those features when vendored is enabled? Or just not worry about it since it'll only happen once every couple of years.

alexcrichton:
Hm that's a good point. I wouldn't be too too worried about this yeah but it seems like enabling vendoring probably shouldn't opt you in to a specific version, you'd have to still select it manually? We could then, if necessary, give a nicer error when you require vendoring and features but there's a version mismatch.

sfackler:
Yeah, you'd still have to manually opt-in. It seems fine to just note our upgrade policy in the documentation. If people want to opt into version specific features they'll just need to be careful around upgrade times.

sfackler:
One other thing to decide here is how the vendored build is activated. It's currently done via a Cargo feature, but it's not a super great use of that. Only the root crate is going to be the thing that cares, and it normally isn't going to be directly depending on openssl-sys. The other option is to control via an environment variable, but we then have to unconditionally depend on the source package. Seems like not the worst thing though?

alexcrichton:
Perhaps we could take a leaf out of rayon's book and require `RUSTFLAGS='--cfg openssl_sys_unstable'`?

PR #692, opened 2017-09-11
Allow overriding lib/include dirs when cross-compiling.

jessicah:
Needed when cross-compiling cargo, to use provided shared libraries from Haiku packages.

### DIFF
--- a/openssl-sys/build.rs
+++ b/openssl-sys/build.rs
@@ -47,8 +47,9 @@ fn main() {
     let mut env_include_dir = "OPENSSL_INCLUDE_DIR".to_string();
 
     if target != host {
-        env_lib_dir = "TARGET_OPENSSL_LIB_DIR".to_string();
-        env_include_dir = "TARGET_OPENSSL_INCLUDE_DIR".to_string();
+	let prefix = target.to_uppercase().replace("-", "_");
+        env_lib_dir = format!("{}_OPENSSL_LIB_DIR", prefix);
+        env_include_dir = format!("{}_OPENSSL_INCLUDE_DIR", prefix);
     }
 
     println!("cargo:rerun-if-env-changed={}", env_lib_dir);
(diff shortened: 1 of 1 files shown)


jessicah:
In https://github.com/alexcrichton/git2-rs/pull/248#issuecomment-328403345, Alex suggested an alternative that I shall explore. Will close this PR if it works out.

sfackler:
Is this dealing with issues when using openssl in both a build script and the crate itself? That makes sense, but I might lean towards putting the target itself in the environment variable (e.g. `X86_64_UNKNOWN_LINUX_GNU_OPENSSL_DIR`) rather than just `TARGET_OPENSSL_DIR`.

jessicah:
I don't think the approach mentioned in git2-rs is really workable, as it requires embedding paths to the cross-compiler's environment in the Cargo.toml, which is just silly IMO.

This is primarily used for building the crate itself in a cross-compilation environment. I don't see how it would have other uses. I can see how if building for multiple targets, that the variable could get re-used erroneously, so I'll make this change :)

sfackler:
The overrides live in `.cargo/config` rather than `Cargo.toml` so it's a bit less weird, but I'm on board with the proposed change here regardless.

jessicah:
Shall I also include commit to bump to 0.9.18?

sfackler:
We'll need to fall back to the unprefixed versions for backwards compatibility. I'd probably also be a good idea to support the prefix for all the environment variables we use. Might be best to split the logic out to a method that looks something like this:

```rust
fn env(name: &str) -> Option<OsString> {
    let prefix = env::var("TARGET").unwrap().to_uppercase().replace("-", "_");
    let prefixed = format!("{}_{}", prefix, name);
    println!("cargo:rerun-if-env-changed={}", prefixed);
    if let Some(var) = env::var_os(&prefixed) {
        return Some(var);
    }
    println!("cargo:rerun-if-env-changed={}", name);
    env::var_os(name)
}
```

sfackler:
Nah, I'll do that when cutting the release.

sfackler:
Thanks!

PR #695, opened 2017-09-14
openssl-sys: support libressl 2.6.1

Keruspe:
systest currently fails as cms.h doesn't exist anymore, not sure how to make it conditional

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,9 +78,9 @@ openssl_101: &OPENSSL_101
 libressl_250: &LIBRESSL_250
   LIBRARY: libressl
   VERSION: 2.5.0
-libressl_260: &LIBRESSL_260
+libressl_261: &LIBRESSL_261
   LIBRARY: libressl
-  VERSION: 2.6.0
+  VERSION: 2.6.1
 
 x86_64: &X86_64
   TARGET: x86_64-unknown-linux-gnu
@@ -139,10 +139,10 @@ jobs:
     <<: *JOB
     environment:
       <<: [*LIBRESSL_250, *X86_64, *BASE]
-  x86_64-libressl-2.6.0:
+  x86_64-libressl-2.6.1:
     <<: *JOB
     environment:
-      <<: [*LIBRESSL_260, *X86_64, *BASE]
+      <<: [*LIBRESSL_261, *X86_64, *BASE]
 workflows:
   version: 2
   tests:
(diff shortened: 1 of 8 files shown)


sfackler:
The systest crate should be able to get access to the `libressl261` in its build script.

Is there any semantic meaning at all to major, minor, and patch version bumps to the libressl folks? It's not clear to me why removing a public API header would happen in a patch release.

Keruspe:
If you're speaking of doing what my second commit does, I already tried and it doesn't seem to work either. Am I missing something there?

sfackler:
They aren't passed as cfgs, but rather as environment variables: https://github.com/sfackler/rust-openssl/blob/master/openssl/build.rs

sfackler:
Can you bump this build to use 2.6.1 rather than 2.6.0? https://github.com/sfackler/rust-openssl/blob/master/.circleci/config.yml#L81

Keruspe:
@sfackler I can't seem to find where the env variables are set. Currently I see that DEP_OPENSSL_LIBRESSL is true, and DEP_OPENSSL_VERSION is 101 which doesn't help me special-casing libressl 2.6.1

Keruspe:
I got it to build, now looking in a few tests failures:

```
bad SSL_OP_ALL value at byte 0: rust: 20 (0x14) != c 4 (0x4)                                              
bad SSL_OP_ALL value at byte 3: rust: 128 (0x80) != c 0 (0x0)                                             
bad SSL_OP_SINGLE_ECDH_USE value at byte 2: rust: 8 (0x8) != c 0 (0x0)                                    
bad SSL_OP_TLSEXT_PADDING value at byte 0: rust: 16 (0x10) != c 0 (0x0)                                   
bad SSL_OP_CRYPTOPRO_TLSEXT_BUG value at byte 3: rust: 128 (0x80) != c 0 (0x0)      
```

sfackler:
Cool - you'll have to conditionally adjust those constants in openssl-sys sounds like.

Keruspe:
This last commit might need some further digging? I don't see it specifically mentioned in the release notes

sfackler:
Yeah, that seems a bit suspicious.

sfackler:
Seems like it could be this line in the changelog?

> * Cleaned up the EC key/curve configuration handling.

Keruspe:
The man page for set_tmp_ecdh now states that the this function is now a noop, so I guess that last change is right after all

sfackler:
Ok - thanks!

PR #697, opened 2017-09-15
Set SSL_MODE_RELEASE_BUFFERS by default

### LINKED ISSUE #696: SSL_MODE_RELEASE_BUFFERS on by default?
I've been informed that it's quite common in almost all production codebases to turn this option on by default and apparently [Python also switch to do this](https://bugs.python.org/issue25672) awhile ago!

Although maybe this is explicitly left out?

alexcrichton:
Closes #696

### DIFF
--- a/openssl/src/ssl/connector.rs
+++ b/openssl/src/ssl/connector.rs
@@ -5,6 +5,7 @@ use error::ErrorStack;
 use ssl::{self, SslMethod, SslContextBuilder, SslContext, Ssl, SSL_VERIFY_PEER, SslStream,
           HandshakeError};
 use pkey::PKeyRef;
+use version;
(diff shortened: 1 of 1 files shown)


sfackler:
My only worry here is around people running super old versions of OpenSSL becoming vulnerable to CVE 2014-0198. They've had 3 years to upgrade, though, so maybe that's not a huge concern at this point?

alexcrichton:
It actually may be easy enough to fix that, according to the Python thread it was fixed in 1.0.1h. Do you know what hex number that corresponds to to use `::version::number()`?

sfackler:
It'd be `0x1000108f` I think?

alexcrichton:
Updated!

PR #729, opened 2017-09-17
Don't build openssl in parallel

aidanhs:
See https://github.com/rust-lang/rust/pull/41563 - historically it's caused problems, and although it should be better in 1.1 I'm still not sure I'd recommend it. Up to you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ tar xf openssl-1.1.0f.tar.gz
 cd openssl-1.1.0f
 export CC=...
 ./Configure --prefix=... linux-x86_64 -fPIC
-make -j$(nproc)
+make -j1 # OpenSSL has historically not supported parallel builds
 make install
 ```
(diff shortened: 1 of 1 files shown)


sfackler:
I'd just change the line to `make` - it matches up with openssl's own INSTALL file.

aidanhs:
The problem with that is the `MAKEFLAGS` environment variable, which is the way it caused problems in Rust. My concern was that people are going to put this into scripts called by other scripts (e.g. openssl-src-rs) that then inherit any parallelism flags.

However, I've looked back through the issues and it appears openssl has put reasonable some work into making parallel builds not break. Realistically the only consumer to see this will be Rust itself and we'll know what caused as soon as it happens, so let's just let this percolate back to Rust via openssl-src-rs and see what happens.

cc @alexcrichton for the record.

PR #734, opened 2017-09-26
AES module level docs and example

### LINKED ISSUE #698: Improve documentation of the `aes` module
This is a tracking issue for the impl period for the `openssl` crate, specifically for improving the documentation of the `aes` module. Documentation should include:

* [ ] Documentation on all types contained in the module
* [ ] Documentation for all methods on these types
* [ ] Documentation for all free functions
* [ ] Module documentation explaining what it is and why you'd use it
* [ ] At least one code example in the module documentation

The documentation may also wish to link to the official OpenSSL documentation where possible to ensure that there's the most up-to-date info available!

AndyGauge:
Fixes #698

### DIFF
--- a/openssl/src/aes.rs
+++ b/openssl/src/aes.rs
@@ -9,10 +9,6 @@
 //! create a new key with [`new_encrypt`] and perform an encryption/decryption
 //! using that key with [`aes_ige`].
 //!
-//! AES is a 128-bit (16 byte) block cipher.  The rust implmentation will panic
-//! if the input or output does not meet this 16-byte boundry.  Attention must
-//! be made in this low level implementation to pad the value to the 128-bit boundry.
-//!
 //! [`new_encrypt`]: struct.AesKey.html#method.new_encrypt
 //! [`aes_ige`]: fn.aes_ige.html
 //!
@@ -119,6 +115,10 @@ impl AesKey {
 /// block ciphers implemented for AES requires an initalization vector.  The IGE mode
 /// allows a stream of blocks to be encrypted or decrypted without having the entire
 /// plaintext available.  For more information, visit [AES IGE Encryption].
+///
+/// This block cipher uses 16 byte blocks.  The rust implmentation will panic
+/// if the input or output does not meet this 16-byte boundry.  Attention must
+/// be made in this low level implementation to pad the value to the 128-bit boundry.
 ///
 /// [AES IGE Encryption]: http://www.links.org/files/openssl-ige.pdf
 ///
--- a/openssl/src/aes.rs
+++ b/openssl/src/aes.rs
@@ -6,6 +6,10 @@
 //! create a new key with [`new_encrypt`] and perform an encryption/decryption
 //! using that key with [`aes_ige`].
 //!
+//! AES is a 128-bit (16 byte) block cipher.  The rust implmentation will panic
+//! if the input or output does not meet this 16-byte boundry.  Attention must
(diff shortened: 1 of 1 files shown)


openssl/src/aes.rs
```diff
@@ -63,6 +107,14 @@ impl AesKey {
 
 /// Performs AES IGE encryption or decryption
 ///
+/// AES IGE (Infinite Garble Extension) is the form of AES block cipher utilized in
```
sfackler: It's *a* form, not *the* form. An obscure one at that.

openssl/src/aes.rs
```diff
@@ -1,15 +1,59 @@
 //! Low level AES functionality
 //!
-//! The `symm` module should be used in preference to this module in most cases.
+//! Advanced Encryption Standard (AES) provides symmetric key cipher that
```
sfackler: This should probably mention that most people looking for symmetric encryption should be using the [`Crypter`](https://docs.rs/openssl/0.9.19/openssl/symm/struct.Crypter.html) type. This module maps to the lower-level and more obscure options that aren't covered by the Crypter interface.

openssl/src/aes.rs
```diff
@@ -1,15 +1,66 @@
...
+//! AES ECB, CBC, XTS, CTR, CFB, GCM and other conventional symmetric encryption
+//! modes are found in [`symm`].  This is the implementation of AES IGE.
+//!
+//! Advanced Encryption Standard (AES) provides symmetric key cipher that
+//! the same key is used to encrypt and decrypt data.  This implementation
+//! uses 128, 192, or 256 bit keys.  This module provides functions to
+//! create a new key with [`new_encrypt`] and perform an encryption/decryption
+//! using that key with [`aes_ige`].
+//!
+//! AES is a 128-bit (16 byte) block cipher.  The rust implmentation will panic
```
sfackler: There are 128 and 256 bit variants of AES.

I'm not sure what the panic note is referring to - aes_ige doesn't have any asserts around the lengths of the input and output buffers.

BrianOn99: @sfackler I think it is refering to those lines
``` rust
/// Performs AES IGE encryption or decryption
///
/// # Panics
///
/// Panics if `in_` is not the same length as `out`, if that length is not a multiple of 16, or if
/// `iv` is not at least 32 bytes.
pub fn aes_ige(in_: &[u8], out: &mut [u8], key: &AesKey, iv: &mut [u8], mode: Mode) {
    unsafe {
        assert!(in_.len() == out.len());
        assert!(in_.len() % ffi::AES_BLOCK_SIZE as usize == 0);
        assert!(iv.len() >= ffi::AES_BLOCK_SIZE as usize * 2);
```

Doesn't it panic if the buffer lengths is not multiple of ffi::AES_BLOCK_SIZE ?

sfackler:
Oh jeez, I totally misread in_ as iv, sorry!

sfackler:
I would move those details to the function they describe though.

BrianOn99:
Agree, and the line `128-bit (16 byte) block cipher` need change.

BrianOn99:
Sorry. that line is correct.  Please ignore my prevoius comment.

openssl/src/aes.rs
AndyGauge: I can see the confusion.  I'll try to incorporate both of those details at the top (the key is 128-bit, 192-bit, or 256-bit and that the input must be bounded to 16-bytes)

sfackler:
Looks good, thanks!

PR #738, opened 2017-09-28
Add instructions for adding OpenSSL DLLs to PATH if needed

johnthagen:
@alexcrichton Per our discussion, instructions that are needed if user takes non-default option while installing pre-compiled OpenSSL binaries on Windows


alexcrichton:
👍

PR #742, opened 2017-09-28
Issue #704, module `dh`

ejmg:
closes [#704](https://github.com/sfackler/rust-openssl/issues/704)

### DIFF
--- a/openssl/src/dh.rs
+++ b/openssl/src/dh.rs
@@ -23,10 +23,17 @@ use bn::BigNum;
 
 foreign_type! {
     type CType = ffi::DH;
+
     fn drop = ffi::DH_free;
 
+    /// Diffie-Hellman Key agreement object
+    /// A key object that is based on the diffie-hellman algorithm
+    ///
+    /// As described here https://wiki.openssl.org/index.php/Manual:Dh(3)
+    /// and here: https://wiki.openssl.org/index.php/Diffie_Hellman
     pub struct Dh;
 
+    /// a borrowed reference of the dh type, diffie-hellman key agreement object
     pub struct DhRef;
 }
 
--- a/openssl/src/dh.rs
+++ b/openssl/src/dh.rs
@@ -1,3 +1,15 @@
+//! dh.rs
+//! 
+//! summary: bindings for dh library of OpenSSL
+//! author: steven 
+//!
+//! dh is the diffie-hellman (dh) algorithm which allows for two parties to derive the key used for
+//! communication. in dh, the two parties agree to some large prime, `p`, and a generator, `g`,
+//! where 0 < g < p.
+//!
+//! For a full summary, please read https://wiki.openssl.org/index.php/Diffie_Hellman
+
+#![deny(missing_docs)]
 use error::ErrorStack;
 use ffi;
 use foreign_types::ForeignTypeRef;
@@ -21,7 +33,17 @@ impl DhRef {
     to_der!(ffi::i2d_DHparams);
 }
 
+
+// TODO Implement overloaded DH initialization for equivalent methods in 1.1.x
 impl Dh {
+    /// DH object contains three parameters: p, q, and g.
+    ///
+    /// This method will set the c pointers for the DH object, p, q, and g. If the parameters have
+    /// not been set, they will default to NULL.
+    /// 
+    /// This corresponds to DH_get0_pqg() on OpenSSL 1.1.0 and inits DH's fields in 1.1.x.
+    /// Note that currently we assume that you provide all three parameters. In openSSL, there is
+    /// overloading for when q, public key, private key is or isn't provided.
     pub fn from_params(p: BigNum, g: BigNum, q: BigNum) -> Result<Dh, ErrorStack> {
         unsafe {
             let dh = Dh(cvt_p(ffi::DH_new())?);
@@ -35,6 +57,7 @@ impl Dh {
             Ok(dh)
         }
     }
+
 
     from_pem!(Dh, ffi::PEM_read_bio_DHparams);
     from_der!(Dh, ffi::d2i_DHparams);
--- a/openssl/src/dh.rs
+++ b/openssl/src/dh.rs
@@ -1,13 +1,15 @@
 //! dh.rs
 //! 
 //! summary: bindings for dh library of OpenSSL
-//! author: steven 
+//! author: steven fackler
+//! version: 
 //!
 //! dh is the diffie-hellman (dh) algorithm which allows for two parties to derive the key used for
 //! communication. in dh, the two parties agree to some large prime, `p`, and a generator, `g`,
 //! where 0 < g < p.
 //!
-//! For a full summary, please read https://wiki.openssl.org/index.php/Diffie_Hellman
+//! For a full summary of the algo, please read https://wiki.openssl.org/index.php/Diffie_Hellman
+//! manpages: https://www.openssl.org/docs/manpages.html
 
 #![deny(missing_docs)]
 use error::ErrorStack;
@@ -36,14 +38,14 @@ impl DhRef {
 
 // TODO Implement overloaded DH initialization for equivalent methods in 1.1.x
 impl Dh {
-    /// DH object contains three parameters: p, q, and g.
-    ///
+    
     /// This method will set the c pointers for the DH object, p, q, and g. If the parameters have
     /// not been set, they will default to NULL.
-    /// 
-    /// This corresponds to DH_get0_pqg() on OpenSSL 1.1.0 and inits DH's fields in 1.1.x.
-    /// Note that currently we assume that you provide all three parameters. In openSSL, there is
-    /// overloading for when q, public key, private key is or isn't provided.
+    ///
+    /// This calls DH_get0_pqg() on OpenSSL 1.1.0 and inits DH's fields manually in 1.1.x.
+    /// Note that currently we assume that you only provide the parameters p, q, and g. In openSSL,
+    /// there is overloading for when q, public key, private key is or isn't provided.
+    /// openssl: https://www.openssl.org/docs/man1.1.0/crypto/DH_set0_pqg.html
     pub fn from_params(p: BigNum, g: BigNum, q: BigNum) -> Result<Dh, ErrorStack> {
         unsafe {
             let dh = Dh(cvt_p(ffi::DH_new())?);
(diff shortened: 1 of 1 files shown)


ejmg:
Ah, it appears the `foreign_type!` declaration needs to be documented because `#![deny(missing_docs)]` failed it. Dumb mistake.

ejmg:
Do I even need to document this block? Otherwise I'll just take out the `#![deny(missing_docs)]` macro which in of itself probably doesn't need to stay in the codebase.
``` rust
foreign_type! {
    type CType = ffi::DH;
    fn drop = ffi::DH_free;

    pub struct Dh;

    pub struct DhRef;
}
```

openssl/src/dh.rs
```diff
@@ -1,3 +1,17 @@
+//! dh.rs
+//! 
+//! summary: bindings for dh library of OpenSSL
+//! author: steven fackler
+//! version:
```
sfackler: I don't think the first couple of lines here are really relevant for someone reading this documentation.

openssl/src/dh.rs
```diff
@@ -9,10 +23,17 @@ use bn::BigNum;
 
 foreign_type! {
     type CType = ffi::DH;
+
     fn drop = ffi::DH_free;
 
+    /// Diffie-Hellman Key agreement object
+    /// A key object that is based on the diffie-hellman algorithm
```
sfackler: This line will be folded up to the first line by Markdown.

openssl/src/dh.rs
```diff
@@ -21,7 +42,17 @@ impl DhRef {
     to_der!(ffi::i2d_DHparams);
 }
 
+
+// TODO Implement overloaded DH initialization for equivalent methods in 1.1.x
 impl Dh {
+    
+    /// This method will set the c pointers for the DH object, p, q, and g. If the parameters have
+    /// not been set, they will default to NULL.
```
sfackler: The second sentence doesn't seem super relevant - we are setting them.

openssl/src/dh.rs
```diff
@@ -1,3 +1,17 @@
+//! dh.rs
+//! 
+//! summary: bindings for dh library of OpenSSL
+//! author: steven fackler
+//! version: 
+//!
+//! dh is the diffie-hellman (dh) algorithm which allows for two parties to derive the key used for
```
sfackler: Capitalization

openssl/src/dh.rs
```diff
@@ -1,3 +1,17 @@
...
+//! communication. in dh, the two parties agree to some large prime, `p`, and a generator, `g`,
+//! where 0 < g < p.
+//!
+//! For a full summary of the algo, please read https://wiki.openssl.org/index.php/Diffie_Hellman
```
sfackler: Can the links be shortened up instead of being raw URLs in the documentation? I.e. `For a full summary of Diffie-Hellman, refer to the [OpenSSL wiki](https://wiki.openssl.org/index.php/Diffie_Hellman).`

openssl/src/dh.rs
```diff
@@ -1,3 +1,17 @@
...
+//! manpages: https://www.openssl.org/docs/manpages.html
```
sfackler: Why is this here?

openssl/src/dh.rs
```diff
@@ -9,10 +23,17 @@ use bn::BigNum;
 
 foreign_type! {
     type CType = ffi::DH;
+
     fn drop = ffi::DH_free;
 
+    /// Diffie-Hellman Key agreement object
```
sfackler: Punctuation?

openssl/src/dh.rs
```diff
@@ -9,10 +23,17 @@ use bn::BigNum;
...
+    /// A key object that is based on the diffie-hellman algorithm
+    ///
+    /// As described here https://wiki.openssl.org/index.php/Manual:Dh(3)
+    /// and here: https://wiki.openssl.org/index.php/Diffie_Hellman
     pub struct Dh;
 
+    /// a borrowed reference of the dh type, diffie-hellman key agreement object
```
sfackler: Capitalization, punctuation.

I like the format [here](https://github.com/sfackler/rust-openssl/pull/743/files#diff-3399783e0db3cafe8354809089223aaeR57) for documentation for the `*Ref` types.

openssl/src/dh.rs
```diff
@@ -21,7 +42,17 @@ impl DhRef {
...
 
+
+// TODO Implement overloaded DH initialization for equivalent methods in 1.1.x
 impl Dh {
+    
+    /// This method will set the c pointers for the DH object, p, q, and g. If the parameters have
+    /// not been set, they will default to NULL.
+    ///
+    /// This calls DH_get0_pqg() on OpenSSL 1.1.0 and inits DH's fields manually in 1.1.x.
+    /// Note that currently we assume that you only provide the parameters p, q, and g. In openSSL,
```
sfackler: "OpenSSL"

openssl/src/dh.rs
```diff
@@ -21,7 +42,17 @@ impl DhRef {
...
 }
 
+
+// TODO Implement overloaded DH initialization for equivalent methods in 1.1.x
 impl Dh {
+    
+    /// This method will set the c pointers for the DH object, p, q, and g. If the parameters have
+    /// not been set, they will default to NULL.
+    ///
+    /// This calls DH_get0_pqg() on OpenSSL 1.1.0 and inits DH's fields manually in 1.1.x.
```
sfackler: "initializes"

openssl/src/dh.rs
```diff
@@ -36,10 +67,15 @@ impl Dh {
         }
     }
 
+
     from_pem!(Dh, ffi::PEM_read_bio_DHparams);
     from_der!(Dh, ffi::d2i_DHparams);
 
+    /// return a DH object for the IETF RFC 5114 value.
```
sfackler: Capitalization

openssl/src/dh.rs
```diff
@@ -48,7 +84,11 @@ impl Dh {
         }
     }
 
+    /// return a DH object for the IETF RFC 5114 value.
```
sfackler: See above

openssl/src/dh.rs
sfackler: The phrasing here could be a bit more clear - maybe something like "Returns the standard 1024 bit MODP group with 160-bit prime order subgroup specified in [RFC 5114](https://tools.ietf.org/html/rfc5114#section-2.1)"

openssl/src/dh.rs
```diff
@@ -57,7 +97,11 @@ impl Dh {
...
```
sfackler: See above

openssl/src/dh.rs
```diff
@@ -67,11 +111,13 @@ impl Dh {
     }
 }
 
+/// sets rust-openssl to call DH_set0_pqg as defined by fn from_params if using openSSL 1.1.x
```
sfackler: These docs won't be rendered anywhere since this is a private module, FYI.

ejmg:
Hey, @sfackler I'm sorry for failing on standards. I'll fix them first thing tomorrow morning. Thanks for the feedback.

openssl/src/dh.rs
ejmg: +1

openssl/src/dh.rs
ejmg: +1, didn't even know this was possible in rustdoc

openssl/src/dh.rs
ejmg: I felt bad for having this one link to a different source of documentation, the wiki, rather than the manpages as I use everywhere else in this file so I thought to directly link to the general manpage docs. I'll yank the line out.

openssl/src/dh.rs
ejmg: @sfackler When I read the code, I read it as we're taking values in, regardless of what they are, and setting them as the ptrs. Do we not inherit the behavior of defaulting to `NULL` when someone calls this function and provides nothing for the actual values for `p`, `g`, and `q`?

openssl/src/dh.rs
ejmg: Does the following work?
```rust
    /// Return a DH object for the standard 1024 bit MODP group with 160-bit prime order subgroup
    /// specified in [IETF RFC 5114](https://tools.ietf.org/html/rfc5114#section-2.1).
    ///
    /// get_1024_160 corresponds to DH_get_1024_160() in 1.0.2 and 1.1.x
    ///
    /// Requires the `v102` or `v110` features and OpenSSL 1.0.2 or OpenSSL 1.1.0.
    /// [OpenSSL Documentation](https://www.openssl.org/docs/man1.1.0/crypto/DH_get_1024_160.html).
```

openssl/src/dh.rs
ejmg: Am I to assume this method is for the 2047 bit MODP group with 224-bit prime order subgroup (and so on)? I want to be sure since this is well outside of my realm of knowledge right now.

openssl/src/dh.rs
ejmg: Does doing something like [Dh] in RustDoc link to the type? I didn't know about this feature at all!

openssl/src/dh.rs
ejmg: +1

openssl/src/dh.rs
ejmg: +1

openssl/src/dh.rs
ejmg: +1

openssl/src/dh.rs
ejmg: Okay, pulled.

openssl/src/dh.rs
sfackler: The `[Foo]` part is a reference to the link defined as `[Foo]: https://foo.com` below.

openssl/src/dh.rs
sfackler: The user can't provide nothing for those values.

openssl/src/dh.rs
sfackler: Yep!

openssl/src/dh.rs
sfackler: Yeah, they all come from that same RFC, just with different sizes.

sfackler:
Any update here?

ejmg:
@sfackler sorry for completely ghosting on you. Classes have not been the easiest on me so far this semester. I rebased and pushed my changes. Please let me know if all is well this time round. I can get back to you quicker this time if something is awry again.

PR #743, opened 2017-09-28
Doc asn1 module

### LINKED ISSUE #699: Improve documentation on the `asn1` module
This is a tracking issue for the impl period for the `openssl` crate, specifically for improving the documentation of the `asn1` module. Documentation should include:

* [ ] Documentation on all types contained in the module
* [ ] Documentation for all methods on these types
* [ ] Documentation for all free functions
* [ ] Module documentation explaining what it is and why you'd use it
* [ ] At least one code example in the module documentation

The documentation may also wish to link to the official OpenSSL documentation where possible to ensure that there's the most up-to-date info available!

AndyGauge:
Fixes #699

### DIFF
--- a/openssl/src/asn1.rs
+++ b/openssl/src/asn1.rs
@@ -155,6 +155,11 @@ impl Asn1StringRef {
     }
 
     /// Return the string as an array of bytes
+    ///
+    /// The bytes do not directly corespond to UTF-8 encoding.  To interact with
+    /// strings in rust, it is preferable to use [`as_utf8`]
+    ///
+    /// [`as_utf8`]: struct.Asn1String.html#method.as_utf8
     pub fn as_slice(&self) -> &[u8] {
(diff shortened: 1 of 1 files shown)


openssl/src/asn1.rs
```diff
@@ -88,10 +154,12 @@ impl Asn1StringRef {
         }
     }
 
+    /// Return the string as an array of bytes
```
sfackler: Can we note here that the bytes may not be in a standard UTF8 compatible encoding and point to as_utf8?

openssl/src/asn1.rs
```diff
@@ -137,7 +237,17 @@ foreign_type! {
     type CType = ffi::ASN1_OBJECT;
     fn drop = ffi::ASN1_OBJECT_free;
 
+    /// Object Identifier
+    ///
+    /// Use OBJ_nid2obj() ffi to create objects, not Asn1Object directly.
```
sfackler: We don't bind to ASN1_OBJECT_new, so I'm not sure how relevant this is.

Issue #703, opened 2017-09-15
Deprecate/remove `crypto` module?


alexcrichton:
It's only got one item and it's deprecated!

alexcrichton:
@sfackler mind commenting about the fate of this module?

sfackler:
Yeah it only exists for back compat.

johnthagen:
I believe #745 closed this issue (unless this issue is left open for the removal of the module). @sfackler can you confirm?

sfackler:
Yep, we should be good to go here. Next time we're ready to cut a breaking release I'll just make a pass and delete all of the deprecated bits, so no need to keep issues open for everything.

PR #745, opened 2017-09-30
Hide deprecated crypto module from docs

johnthagen:
At least partially addresses #703

@sfackler @alexcrichton Should I also add `#![deprecated ]` for the module?

### DIFF
--- a/openssl/src/crypto.rs
+++ b/openssl/src/crypto.rs
@@ -1,3 +1,4 @@
+#![doc(hidden)]
 use string::OpensslString;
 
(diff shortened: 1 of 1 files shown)


sfackler:
Yeah, let's deprecate the entire module.

johnthagen:
@sfackler Done. I deprecated it "since" the next patch release (0.9.20) assuming it would be part of the next patch release. Is that how that would normally work?

sfackler:
Yep, seems good!

PR #747, opened 2017-10-03
Symm documentation

BrianOn99:
Referring to issue #723, adding additional docs to symm

### DIFF
--- a/openssl/src/symm.rs
+++ b/openssl/src/symm.rs
@@ -1,3 +1,4 @@
+//! High level interface to certain symmetric ciphers.
 use std::cmp;
 use std::ptr;
 use libc::c_int;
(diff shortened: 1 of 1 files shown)


BrianOn99:
It would be nice if there is a url showing which cipher need an iv but I cannot find one.

openssl/src/symm.rs
```diff
@@ -12,6 +36,10 @@ pub enum Mode {
     Decrypt,
 }
 
+/// Represent a particular cipher algorithm.  See OpenSSL doc at [`EVP_EncryptInit`] for more
```
sfackler: The first paragraph of documentation should be a short single sentence since it is shown in the parent page of the documentation. Can you move the other sentences here and below down to their own paragraph?

BrianOn99: Please check if it is better now.

sfackler:
Thanks!

PR #749, opened 2017-10-03
Document conf module

johnthagen:
@sfackler When learning about this module, it seems like the `CONF` system can do more than what is exposed in the Rust API. For example, there are `NCONF_load` and `NCONF_get_string` functions to actually parse things from a configuration file.

Are these intentionally left out? I'm having a hard time coming up with examples in Rust as it doesn't seem to me you can actually do much with the `Conf` objects in Rust, unless I'm missing something.

If these were left out intentionally, we should add a note about that in the docs.

### DIFF
--- a/openssl/src/conf.rs
+++ b/openssl/src/conf.rs
@@ -8,14 +8,11 @@ pub struct ConfMethod(*mut ffi::CONF_METHOD);
 
 impl ConfMethod {
     /// Retrieve handle to the default OpenSSL configuration file processing function.
-    ///
-    /// # OpenSSL Implementation Details
-    ///
-    /// `NCONF` stands for "New Conf", as described in crypto/conf/conf_lib.c. This is
-    /// a newer API than the "CONF classic" functions.
     pub fn default() -> ConfMethod {
         unsafe {
             ffi::init();
+            // `NCONF` stands for "New Conf", as described in crypto/conf/conf_lib.c. This is
+            // a newer API than the "CONF classic" functions.
             ConfMethod(ffi::NCONF_default())
         }
     }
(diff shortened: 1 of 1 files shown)


sfackler:
OpenSSL's APIs are approximately infinitely large :). We add bindings to this library on-demand. Up to now, no one's needed anything more than the very basic functionality to interface with some of the X509 extension stuff.

openssl/src/conf.rs
```diff
@@ -6,17 +7,25 @@ use error::ErrorStack;
 pub struct ConfMethod(*mut ffi::CONF_METHOD);
 
 impl ConfMethod {
+    /// Retrieve handle to the default OpenSSL configuration file processing function.
+    ///
+    /// # OpenSSL Implementation Details
+    ///
+    /// `NCONF` stands for "New Conf", as described in crypto/conf/conf_lib.c. This is
```
sfackler: I don't think we expose the name `NCONF` anywhere, right?

johnthagen:
> We add bindings to this library on-demand.

Totally reasonable. I just wanted to make sure I wasn't missing anything. Are the level of docs sufficient then for what we do expose?

openssl/src/conf.rs
johnthagen: Correct, this is an artifact of someone coming in completely unfamiliar with the OpenSSL conf module. My first question when looking at the `ffi` calls was "What does `NCONF` even stand for?" There is no documentation for the module, so it took me a while of digging in the source code to even figure that out.

I had to actually read part of a Google Book, to figure out what OpenSSL was actually exposing here.

This could be changed to a non-doc comment as it's more useful for those actually looking at the implementation and fixing/extending it.

Thoughts?

sfackler: Ah gotcha, that makes sense! I'd change this to a normal comment, then, since it's something that's relevant to someone reading the source but not the API docs.

johnthagen: Done.

sfackler:
Thanks!

PR #753, opened 2017-10-07
Added module-level documentation for the `sha` module.

zsck:
The documentation included describes what the SHA family of hash functions is,
what hash functions are for, and a little bit about why one may want to use the
SHA family of hash functions.  I have also included a couple of examples demonstrating
how to create a hasher and update it, as well as how to hash bytes directly.


sfackler:
Thanks!

PR #754, opened 2017-10-07
Add documentation to the `hash` module

arcrose:
This PR contains additions to the `hash` module's documentation.  There is now module-level documentation containing an example of how to use the module, as well as documentation for each non-deprecated function and method.


openssl/src/hash.rs
```diff
@@ -1,3 +1,36 @@
...
+//! use hex::ToHex;
+//!
+//! fn main() {
+//!     let mut hasher = hash::Hasher::new(hash::MessageDigest::sha512()).unwrap();
+//!     let bytes_hashed = hash_stream(&mut hasher).unwrap();
+//!     let hash = hasher.finish2().unwrap();
+//!     println!("Hashed {} bytes to {}", bytes_hashed, hash.to_hex());
+//! }
+//!
+//! fn hash_stream<W: Write>(hasher: &mut W) -> io::Result<usize> {
```
sfackler: I think it'd probably make more sense to have this take a `Hasher` and use `update` rather than `write`.

openssl/src/hash.rs
```diff
@@ -16,39 +49,66 @@ use error::ErrorStack;
 pub struct MessageDigest(*const ffi::EVP_MD);
 
 impl MessageDigest {
+    /// Construct a `MessageDigest` that will hash bytes supplied to `Hasher.update` using the MD5
```
sfackler: Let's shorten the first sentences here up a bit - maybe something like "Returns a `MessageDigest` corresponding to the MD5 algorithm".

openssl/src/hash.rs
```diff
@@ -16,39 +49,66 @@ use error::ErrorStack;
...
         unsafe { MessageDigest(ffi::EVP_ripemd160()) }
     }
 
+    /// Obtains a pointer to the underlying identifier for the message digest algorithm in use.
     pub fn as_ptr(&self) -> *const ffi::EVP_MD {
         self.0
     }
 }
 
+/// Represents the state of a message digest, which moves through three states as it is created,
```
sfackler: This is a private type, so no need to document it.

openssl/src/hash.rs
```diff
@@ -89,9 +149,12 @@ use self::State::*;
...
 /// # Warning
 ///
-/// Don't actually use MD5 and SHA-1 hashes, they're not secure anymore.
+/// The MD5 and SHA-1 hash algorithms are no longer considered secure and should not be used in new
+/// software. Use of these algorithms should be restricted only to cases where compatibility with
+/// legacy systems is required.
 ///
-/// Don't ever hash passwords, use the functions in the `pkcs5` module or bcrypt/scrypt instead.
+/// The hash algorithms available here are not suited to password hashing. For such applications,
+/// prefer instead to use either the `pkcs5` module or, more simply, either bcrypt or scrypt.
```
sfackler: I don't know if I'd say that bcrypt or scrypt are more simple. scrypt is in the `pkcs5` module, and pbkdf2 (also in the `pkcs5` module) is a decent third option.

openssl/src/hash.rs
zsck: I see your point, though I wanted to demonstrate the usefulness of `Write` being implemented.  With that in mind, would you still prefer just calling `update`?

sfackler: Having an example of why the Write implementation is useful is a good idea, but I don't think this does that - you can replace `W` with `Hasher` and `write` with `update` and nothing else really changes.

One option could be to use `io::copy` to hash the contents of a file.

zsck: Duly noted! I think you'll like the changes I just pushed up.

openssl/src/hash.rs
```diff
@@ -89,9 +155,13 @@ use self::State::*;
...
 ///
-/// Don't actually use MD5 and SHA-1 hashes, they're not secure anymore.
+/// The MD5 and SHA-1 hash algorithms are no longer considered secure and should not be used in new
+/// software. Use of these algorithms should be restricted only to cases where compatibility with
+/// legacy systems is required.
 ///
-/// Don't ever hash passwords, use the functions in the `pkcs5` module or bcrypt/scrypt instead.
+/// The hash algorithms available here are not suited to password hashing. For such applications,
+/// prefer instead to use the `pkcs5` module more which includes more suitable options, such as
+/// bcrypt and scrypt.
```
sfackler: There isn't a bcrypt implementation in the pkcs5 module.

openssl/src/hash.rs
```diff
@@ -1,3 +1,51 @@
...
+//! ```rust
+//! extern crate openssl;
+//! extern crate hex;
+//! 
+//! use openssl::hash;
+//! use hex::ToHex;
+//! 
+//! fn main() {
+//!     let mut hasher = hash::Hasher::new(hash::MessageDigest::sha512()).unwrap();
+//!     hasher.update(b"HTTP/1.1 200 OK\r\n");
```
sfackler: We shouldn't be ignoring the errors here.

PR #757, opened 2017-10-16
Add an example of making a CA and certs and verifying.

bvinc:
I thought this crate could really use an example to bring together the information scattered through the documentation and tests and show how to perform common tasks.

I tried to use all of the most common, but secure, algorithms and extensions, in a simple and understandable way.

Would you be interested in including this?


openssl/examples/mk_certs.rs
```diff
@@ -0,0 +1,143 @@
...
+        SubjectKeyIdentifier::new().build(&cert_builder.x509v3_context(None, None))?;
+    cert_builder.append_extension(subject_key_identifier)?;
+
+    cert_builder.sign(&privkey, MessageDigest::sha256())?;
+    let cert = cert_builder.build();
+
+    Ok((cert, privkey))
+}
+
+/// Make a X509 request with the given number of bits and the given common name
```
sfackler: Which number of bits?

openssl/examples/mk_certs.rs
```diff
@@ -0,0 +1,143 @@
...
+/// Make a certificate and private key signed by the given CA cert and private key
+fn mk_ca_signed_cert(ca_cert: &X509Ref, ca_privkey: &PKeyRef) -> Result<(X509, PKey), ErrorStack> {
+    let rsa = Rsa::generate(2048)?;
+    let privkey = PKey::from_rsa(rsa)?;
+
+    let req = mk_request(&privkey)?;
+
+    let mut cert_builder = X509::builder()?;
+    cert_builder.set_version(2)?;
+    let serial_number = ::openssl::bn::BigNum::from_dec_str("456")?;
```
sfackler: It seems a bit iffy to be hardcoding the serial number since it has to be unique per certificate per issuer.

bvinc:
Good catches.  I've updated the pull request.

openssl/examples/mk_certs.rs
```diff
@@ -0,0 +1,153 @@
...
+/// Make a certificate and private key signed by the given CA cert and private key
+fn mk_ca_signed_cert(ca_cert: &X509Ref, ca_privkey: &PKeyRef) -> Result<(X509, PKey), ErrorStack> {
+    let rsa = Rsa::generate(2048)?;
+    let privkey = PKey::from_rsa(rsa)?;
+
+    let req = mk_request(&privkey)?;
+
+    let mut cert_builder = X509::builder()?;
+    cert_builder.set_version(2)?;
+    let serial_number = {
```
sfackler: I think it's probably be best to mirror what `openssl x509` does here: https://github.com/openssl/openssl/blob/208fb891e36f16d20262710c70ef0ff3df0e885c/apps/apps.c#L1499.

I believe the equivalent here would be

```rust
let mut serial = BigNum::new()?;
serial.rand(159, MSB_MAYBE_0, false)?;
```

sfackler:
Thanks!

PR #636, opened 2017-05-12
Fix LibreSSL versions other than 2.5.0

sfackler:
Turns out LibreSSL breaks ABI compatibility across minor releases. Fun!

Big props to @alexcrichton again - ctest made the debugging here easy instead of a nightmare!

I do not want to invest the CI time in testing every minor version of LibreSSL, so we're just going to test the oldest and newest supported versions and cross our fingers for the rest.

Closes #632 
Closes #619

PR #759, opened 2017-10-23
Libressl v2.6.3

feeeeb:
this adds support for LibreSSL 2.6.3. tests look good on my local machine.

i based this commit off of:

commit e0efd1d438fbf426a7e1006e4983b0f352c630f6
commit 1308cb2b526355c66854ee22ea03de93bae326c1
commit 701d3781dd8febcc582c801ba63527fcb5c328fb

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,6 +82,9 @@ libressl_250: &LIBRESSL_250
 libressl_261: &LIBRESSL_262
   LIBRARY: libressl
   VERSION: 2.6.2
+libressl_263: &LIBRESSL_263
+  LIBRARY: libressl
+  VERSION: 2.6.3
 
 x86_64: &X86_64
   TARGET: x86_64-unknown-linux-gnu
@@ -144,6 +147,10 @@ jobs:
     <<: *JOB
     environment:
       <<: [*LIBRESSL_262, *X86_64, *BASE]
+  x86_64-libressl-2.6.3:
+    <<: *JOB
+    environment:
+      <<: [*LIBRESSL_263, *X86_64, *BASE]
 workflows:
   version: 2
   tests:
@@ -159,3 +166,4 @@ workflows:
     - armhf-openssl-1.0.1
     - x86_64-libressl-2.5.0
     - x86_64-libressl-2.6.2
+    - x86_64-libressl-2.6.3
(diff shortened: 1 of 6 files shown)


feeeeb:
i think the new circleci test failed due to 2.6.3 not being available in the ftp mirrors yet, from looking at what that script does: https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/

my laptop runs OpenBSD 6.2 which includes LibreSSL 2.6.3 by default.

sfackler:
According to its website,

> The latest stable release is 2.5.5
> The latest development release is 2.6.2

feeeeb:
it's in there:
https://www.openbsd.org/62.html
https://github.com/openbsd/src/blob/70f155023853bdbfcd08e322323f5bfc8a542d3d/lib/libcrypto/opensslv.h

i understand preferring waiting for it to be released more officially though

sfackler:
Cool, yeah, I'd prefer to wait until there's a tarball we can test against.

j-t-d:
2.6.3 is released as of November 6th - https://www.libressl.org/releases.html

sfackler:
Sweet - looks like this just needs a rebase and we should be good to go then.

jirutka:
LibreSSL 2.6.3 is already released. We’ve already upgraded LibreSSL package in Alpine Linux to the latest version and now cannot build cargo package due to rust-openssl incompatibility.

Could you please explain me why do you fix rust-openssl to exact versions of LibreSSL, so it breaks with every update of LibreSSL just due to detection?! This is really unusable for distributions.

sfackler:
@jirutka Because libressl breaks ABI compatibility on minor version releases.

sfackler:
In particular, 2.5.1 did: https://github.com/sfackler/rust-openssl/pull/636

ncopa:
libressl does not follow traditional [semver](http://semver.org/). libressl 2.5.x was unstable up to 2.5.2. This means that 2.5.0, 2.5.1, 2.5.2 was unstable and all broke ABI. From 2.5.3 and onwards it has been stable.

The same thing applies to 2.6.x. The 2.6.0 to 2.6.2 releases were all development releases and ABI incompatible. The [2.6.3](https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-2.6.3-relnotes.txt) release was the first stable and the next releases in 2.6.x series will all be ABI compatible.

It is not known which release will be the next stable til after it has been announced.

feeeeb:
rebased, let me know if this looks ok

sfackler:
@ncopa cool - sounds like we can whitelist 2.6.x then.

sfackler:
Thanks!

PR #768, opened 2017-11-04
Update data-encoding major version

ia0:


openssl/LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright 2011 Google Inc.
+Copyright 2011-2017 Google Inc.
```
sfackler: Why?

ia0: As a Google employee, Google claims copyright over my patches [1]. However, it is not explicitly said whether the copyright year matters or not. I can ask on Monday if you want.

[1] https://opensource.google.com/docs/patching/

sfackler: Ah cool, that makes sense :)

sfackler:
Thanks!

ia0:
Thanks for the quick review!

Issue #772, opened 2017-11-06
Mark types that are Send and Sync


gbip:
Hello,
It seems that [OpenSSL is thread safe](https://stackoverflow.com/questions/33777302/is-openssl-thread-safe#33780647) but rsa::Rsa and symm::Crypter are not marked Send nor Sync.
Would it be possible to implement this ?

sfackler:
Yep, we should be able to mark all of our wrapper types as Send + Sync.

gbip:
Does this just involve implementing Send and Sync for most of the type, or is there some specific behavior to watch for ?

sfackler:
I think we can implement Send/Sync for all types created via `foreign_types!{}`.

gbip:
So all there is to do is to modify the macro ?
If it is that simple, I might want to implement it and make a PR.

sfackler:
That macro is defined in another crate, but you could make a macro in this crate that wraps it and adds the Send/Sync impls.

gbip:
Ok, I will take a look at that.

PR #773, opened 2017-11-09
Impl Send and Sync for all types which are created by `foreign_type` !

gbip:
Hello, 
This PR follows #772. I have a few problems while implementing the changes.
First of all, I can't seem to figure how the `foreign_type!` macro uses the provided (or not) `clone` function. It just seems to ignore it when I give it a function. This is the reason why the wrapper is missing the `$(fn clone = $clone:expr;)*` line. Can you explain to me how should it work, and how to wrap correctly `foreign_type!` ?

Next, the compiler is complaining about missing doc, and I can't figure why. For more details, take a look at the [output of the compiler.](https://gist.github.com/gbip/ed6d8aa46ecb7bfeca8e7b11a901d73b)

### DIFF
--- a/openssl/src/macros.rs
+++ b/openssl/src/macros.rs
@@ -214,3 +214,35 @@ macro_rules! from_pem {
             from_pem, $t, $f);
     }
 }
+
+
+macro_rules! foreign_type_and_impl_send_sync {
+    (
+        $(#[$impl_attr:meta])*
+        type CType = $ctype:ty;
+        fn drop = $drop:expr;
+        $(fn clone = $clone:expr;)*
+
+        $(#[$owned_attr:meta])*
+        pub struct $owned:ident;
+        $(#[$borrowed_attr:meta])*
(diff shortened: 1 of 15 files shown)


openssl/src/macros.rs
```diff
@@ -214,3 +214,29 @@ macro_rules! from_pem {
...
+        $(#[$owned_attr:meta])*
+        pub struct $owned:ident;
+        $(#[$borrowed_attr:meta])*
+        pub struct $borrowed:ident;
+    )
+        => {
+            foreign_type! {
+                type CType = $ctype;
+                fn drop = $drop;
+                pub struct $owned;
```
sfackler: You need to forward the attributes into foreign_type.

gbip: Okay I will look into that, thanks !

gbip:
I think that this is now ok.

sfackler:
Thanks!

PR #778, opened 2017-11-21
Update foreign-types to 0.3

nox:


openssl/src/stack.rs
```diff
@@ -72,12 +71,6 @@ impl<T: Stackable> AsRef<StackRef<T>> for Stack<T> {
     }
 }
 
-impl<T: Stackable> Borrow<StackRef<T>> for Stack<T> {
```
sfackler: `Stack<T>` doesn't have a `foreign_type!`-generated impl, so isn't this still necessary?

nox: Oops, restored.

sfackler:
Thanks!

nox:
This is a breaking change btw.

sfackler:
Oh right, derp. Will revert for now, and re-land some time in the next month or so when 0.10 is ready to land.

PR #780, opened 2017-11-24
PKCS7 Support

maennchen:
I'm really new to Rust. Any help to make this PR good is appreciated.

### DIFF
--- a/openssl/src/crypto/pkcs7/pk7_smime.rs
+++ b/openssl/src/crypto/pkcs7/pk7_smime.rs
@@ -1,0 +1,257 @@
+use x509::{X509, X509Ref};
+use x509::store::X509Store;
+use ffi;
+use bio::{MemBio, MemBioSlice};
+use error::ErrorStack;
+use stack::Stack;
+use foreign_types::ForeignType;
+use symm::Cipher;
+use pkey::PKeyRef;
+use libc::c_int;
+use std::ptr::null_mut;
+use foreign_types::ForeignTypeRef;
+
+pub struct PKCS7(*mut ffi::pkcs7_st);
+
+bitflags! {
+    pub struct PKCS7Flags: c_int {
+        const PKCS7_TEXT = ffi::PKCS7_TEXT;
+        const PKCS7_NOCERTS = ffi::PKCS7_NOCERTS;
+        const PKCS7_NOSIGS = ffi::PKCS7_NOSIGS;
+        const PKCS7_NOCHAIN = ffi::PKCS7_NOCHAIN;
+        const PKCS7_NOINTERN = ffi::PKCS7_NOINTERN;
+        const PKCS7_NOVERIFY = ffi::PKCS7_NOVERIFY;
+        const PKCS7_DETACHED = ffi::PKCS7_DETACHED;
+        const PKCS7_BINARY = ffi::PKCS7_BINARY;
+        const PKCS7_NOATTR = ffi::PKCS7_NOATTR;
+        const PKCS7_NOSMIMECAP = ffi::PKCS7_NOSMIMECAP;
+        const PKCS7_NOOLDMIMETYPE = ffi::PKCS7_NOOLDMIMETYPE;
+        const PKCS7_CRLFEOL = ffi::PKCS7_CRLFEOL;
+        const PKCS7_STREAM = ffi::PKCS7_STREAM;
+        const PKCS7_NOCRL = ffi::PKCS7_NOCRL;
+        const PKCS7_PARTIAL = ffi::PKCS7_PARTIAL;
+        const PKCS7_REUSE_DIGEST = ffi::PKCS7_REUSE_DIGEST;
+        #[cfg(not(any(ossl101, ossl102, libressl)))]
+        const PKCS7_NO_DUAL_CONTENT = ffi::PKCS7_NO_DUAL_CONTENT;
+    }
+}
+
+impl PKCS7 {
+    pub fn smime_write(&self, input: &[u8], flags: PKCS7Flags) -> Result<Vec<u8>, ErrorStack> {
+        ffi::init();
+
+        unsafe {
+            let input_bio = MemBioSlice::new(input)?;
+
+            let output = MemBio::new()?;
+
+            if ffi::SMIME_write_PKCS7(output.as_ptr(), self.0, input_bio.as_ptr(), flags.bits) == 1 {
+                Ok(output.get_buf().to_owned())
+            } else {
+                Err(ErrorStack::get())
+            }
+        }
+    }
+
+    pub fn smime_read(input: &[u8], bcount: &mut Vec<u8>) -> Result<PKCS7, ErrorStack> {
+        ffi::init();
+
+        let input_bio = MemBioSlice::new(input)?;
+
+        let mut bcount_bio = null_mut();
+
+        let pkcs7 = unsafe { ffi::SMIME_read_PKCS7(input_bio.as_ptr(), &mut bcount_bio) };
+
+        bcount.clear();
+
+        if !bcount_bio.is_null() {
+            let bcount_bio = MemBio::from_ptr(bcount_bio);
+            bcount.append(&mut bcount_bio.get_buf().to_vec());
+        }
+
+        if pkcs7.is_null() {
+            Err(ErrorStack::get())
+        } else {
+            Ok(PKCS7(pkcs7))
+        }
+    }
+
+    pub fn decrypt(&self, pkey: &PKeyRef, cert: &X509Ref) -> Result<Vec<u8>, ErrorStack> {
+        ffi::init();
+
+        let output = MemBio::new()?;
+
+        unsafe {
+            if ffi::PKCS7_decrypt(self.0, pkey.as_ptr(), cert.as_ptr(), output.as_ptr(), 0) == 1 {
+                Ok(output.get_buf().to_owned())
+            } else {
+                Err(ErrorStack::get())
+            }
+        }
+    }
+
+    pub fn encrypt(certs: &Stack<X509>, input: &[u8], cypher: Cipher, flags: PKCS7Flags) -> Result<PKCS7, ErrorStack> {
+        ffi::init();
+
+        let input_bio = MemBioSlice::new(input)?;
+
+        let pkcs7 = unsafe { ffi::PKCS7_encrypt(certs.as_ptr(), input_bio.as_ptr(), cypher.as_ptr(), flags.bits) };
+
+        if pkcs7.is_null() {
+            Err(ErrorStack::get())
+        } else {
+            Ok(PKCS7(pkcs7))
+        }
+    }
+
+    pub fn sign(signcert: &X509Ref, pkey: &PKeyRef, certs: &Stack<X509>, input: &[u8], flags: PKCS7Flags) -> Result<PKCS7, ErrorStack> {
+        ffi::init();
+
+        let input_bio = MemBioSlice::new(input)?;
+
+        let pkcs7 = unsafe { ffi::PKCS7_sign(signcert.as_ptr(), pkey.as_ptr(), certs.as_ptr(), input_bio.as_ptr(), flags.bits) };
+
+        if pkcs7.is_null() {
+            Err(ErrorStack::get())
+        } else {
+            Ok(PKCS7(pkcs7))
+        }
+    }
+
+    pub fn verify(&self, certs: &Stack<X509>, store: &X509Store, indata: Option<&[u8]>, out: Option<&mut Vec<u8>>, flags: PKCS7Flags) -> Result<bool, ErrorStack> {
+        ffi::init();
+        
+        let out_bio = MemBio::new()?;
+
+        let result = match indata {
+            Some(data) => {
+                let indata_bio = MemBioSlice::new(data)?;
+                unsafe { ffi::PKCS7_verify(self.0, certs.as_ptr(), store.as_ptr(), indata_bio.as_ptr(), out_bio.as_ptr(), flags.bits) }
+            },
+            None => unsafe { ffi::PKCS7_verify(self.0, certs.as_ptr(), store.as_ptr(), null_mut(), out_bio.as_ptr(), flags.bits) }
+        };
+
+        if let Some(data) = out {
+            data.clear();
+            data.append(&mut out_bio.get_buf().to_vec());
+        }
+
+        if result == 1 {
+            Ok(true)
+        } else {
+            Err(ErrorStack::get())
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use x509::X509;
+    use x509::store::X509StoreBuilder;
+    use symm::Cipher;
+    use crypto::pkcs7::pk7_smime::PKCS7_STREAM;
+    use crypto::pkcs7::pk7_smime::PKCS7_DETACHED;
+    use crypto::pkcs7::pk7_smime::PKCS7;
+    use pkey::PKey;
+    use stack::Stack;
+
+    #[test]
+    fn encrypt_decrypt_test() {
+        let cert = include_bytes!("../../../test/certs.pem");
+        let cert = X509::from_pem(cert).unwrap();
+        let mut certs = Stack::new().unwrap();
+        certs.push(cert.clone()).unwrap();
+        let message: String = String::from("foo");
+        let cypher = Cipher::des_ede3_cbc();
+        let flags = PKCS7_STREAM;
+        let pkey = include_bytes!("../../../test/key.pem");
+        let pkey = PKey::private_key_from_pem(pkey).unwrap();
+
+        let pkcs7 = PKCS7::encrypt(&certs, message.as_bytes(), cypher, flags).expect("should succeed");
+
+        let encrypted = pkcs7.smime_write(message.as_bytes(), flags).expect("should succeed");
+
+        let mut bcount = Vec::new();
+        let pkcs7_decoded = PKCS7::smime_read(encrypted.as_slice(), &mut bcount).expect("should succeed");
+
+        let decoded = pkcs7_decoded.decrypt(&pkey, &cert).expect("should succeed");
+
+        assert_eq!(decoded, message.into_bytes());
+    }
+
+    #[test]
+    fn sign_verify_test_detached() {
+        let cert = include_bytes!("../../../test/cert.pem");
+        let cert = X509::from_pem(cert).unwrap();
+        let certs = Stack::new().unwrap();
+        let message: String = String::from("foo");
+        let flags = PKCS7_STREAM | PKCS7_DETACHED;
+        let pkey = include_bytes!("../../../test/key.pem");
+        let pkey = PKey::private_key_from_pem(pkey).unwrap();
+        let mut store_builder = X509StoreBuilder::new().expect("should succeed");
+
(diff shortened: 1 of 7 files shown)


openssl/src/crypto/pkcs7/pk7_smime.rs
```diff
@@ -0,0 +1,225 @@
...
+        const PKCS7_STREAM = ffi::PKCS7_STREAM;
+        const PKCS7_NOCRL = ffi::PKCS7_NOCRL;
+        const PKCS7_PARTIAL = ffi::PKCS7_PARTIAL;
+        const PKCS7_REUSE_DIGEST = ffi::PKCS7_REUSE_DIGEST;
+        const PKCS7_NO_DUAL_CONTENT = ffi::PKCS7_NO_DUAL_CONTENT;
+    }
+}
+
+impl PKCS7 {
+    pub fn smime_write(&self, input: Vec<u8>, flags: PKCS7Flags) -> Result<Vec<u8>, ErrorStack> {
```
sfackler: `input` can have the type `&[u8]` rather than `Vec<u8>`.

openssl/src/crypto/pkcs7/pk7_smime.rs
```diff
@@ -0,0 +1,225 @@
...
+        unsafe {
+            let input_bio = MemBioSlice::new(input.as_slice())?;
+
+            let output = MemBio::new()?;
+
+            ffi::SMIME_write_PKCS7(output.as_ptr(), self.0, input_bio.as_ptr(), flags.bits);
```
sfackler: You'll want to check the return value of this function for errors.

openssl/src/crypto/pkcs7/pk7_smime.rs
```diff
@@ -0,0 +1,225 @@
...
+
+            Ok(output.get_buf().to_owned())
+        }
+    }
+
+    pub fn smime_read(input: Vec<u8>, bcount: Option<Vec<u8>>) -> Result<PKCS7, ErrorStack> {
```
sfackler: `input: &[u8]`.

sfackler: From looking at the man page, it seems like like the bcount parameter to this function is an output, not an input.

openssl/src/crypto/pkcs7/pk7_smime.rs
```diff
@@ -0,0 +1,225 @@
...
+        let input_bio = MemBioSlice::new(input.as_slice())?;
+
+        match bcount {
+            Some(data) => {
+                let bcount_bio = MemBioSlice::new(data.as_slice())?;
+                Ok(PKCS7(unsafe { ffi::SMIME_read_PKCS7(input_bio.as_ptr(), &mut bcount_bio.as_ptr()) }))
```
sfackler: Check for errors in these calls.

openssl/src/crypto/pkcs7/pk7_smime.rs
```diff
@@ -0,0 +1,225 @@
...
+            },
+            None => Ok(PKCS7(unsafe { ffi::SMIME_read_PKCS7(input_bio.as_ptr(), null_mut()) }))
+        }
+    }
+
+    pub fn decrypt(&self, pkey: PKey, cert: X509) -> Result<Vec<u8>, ErrorStack> {
+        let output = MemBio::new()?;
+
+        unsafe {
+            ffi::PKCS7_decrypt(self.0, pkey.as_ptr(), cert.as_ptr(), output.as_ptr(), 0);
```
sfackler: Errors

openssl/src/crypto/pkcs7/pk7_smime.rs
```diff
@@ -0,0 +1,225 @@
...
+        match bcount {
+            Some(data) => {
+                let bcount_bio = MemBioSlice::new(data.as_slice())?;
+                Ok(PKCS7(unsafe { ffi::SMIME_read_PKCS7(input_bio.as_ptr(), &mut bcount_bio.as_ptr()) }))
+            },
+            None => Ok(PKCS7(unsafe { ffi::SMIME_read_PKCS7(input_bio.as_ptr(), null_mut()) }))
+        }
+    }
+
+    pub fn decrypt(&self, pkey: PKey, cert: X509) -> Result<Vec<u8>, ErrorStack> {
```
sfackler: Both of these parameters can be passed by reference.

openssl/src/crypto/pkcs7/pk7_smime.rs
```diff
@@ -0,0 +1,225 @@
...
+        let output = MemBio::new()?;
+
+        unsafe {
+            ffi::PKCS7_decrypt(self.0, pkey.as_ptr(), cert.as_ptr(), output.as_ptr(), 0);
+
+            Ok(output.get_buf().to_owned())
+        }
+    }
+
+    pub fn encrypt(certs: Vec<X509>, input: Vec<u8>, cypher: Cipher, flags: PKCS7Flags) -> Result<PKCS7, ErrorStack> {
```
sfackler: The `Vec<u8>` can be `&[u8]`.

sfackler: Should this take a `&StackRef<X509>` directly rather than a `Vec<X509>`?

openssl/src/crypto/pkcs7/pk7_smime.rs
```diff
@@ -0,0 +1,225 @@
...
+                match stack_result {
+                    Ok(mut stack) => {
+                        stack.push(cert.clone())?;
+                        Ok(stack)
+                    },
+                    Err(error) => Err(error)
+                }
+            })?;
+
+        Ok(PKCS7(unsafe { ffi::PKCS7_encrypt(certs_stack.as_ptr(), input_bio.as_ptr(), cypher.as_ptr(), flags.bits) }))
```
sfackler: Errors

openssl/src/crypto/pkcs7/pk7_smime.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    pub fn sign(signcert: X509, pkey: PKey, certs: Vec<X509>, input: Vec<u8>, flags: PKCS7Flags) -> Result<PKCS7, ErrorStack> {
```
sfackler: `&X509Ref`, `&PKeyRef`, `&StackRef<X509>`, `&[u8]`.

openssl/src/crypto/pkcs7/pk7_smime.rs
```diff
@@ -0,0 +1,225 @@
...
+                match stack_result {
+                    Ok(mut stack) => {
+                        stack.push(cert.clone())?;
+                        Ok(stack)
+                    },
+                    Err(error) => Err(error)
+                }
+            })?;
+
+        Ok(PKCS7(unsafe { ffi::PKCS7_sign(signcert.as_ptr(), pkey.as_ptr(), certs_stack.as_ptr(), input_bio.as_ptr(), flags.bits) }))
```
sfackler: Errors

openssl/src/crypto/pkcs7/pk7_smime.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    pub fn verify(&self, certs: Vec<X509>, store: X509Store, indata: Option<Vec<u8>>, out: &MemBio, flags: PKCS7Flags) -> Result<bool, ErrorStack> {
```
sfackler: `&StackRef<X509>`, `&X509StoreRef`, `Option<&[u8]>`.

`MemBio` isn't a publicly exposed type, so we'll probably want to pass in a `&mut Vec<u8>` instead. Maybe an `Option<&mut Vec<u8>>` since out data is optional?

openssl/src/crypto/pkcs7/pk7_smime.rs
```diff
@@ -0,0 +1,225 @@
...
+                        Ok(stack)
+                    },
+                    Err(error) => Err(error)
+                }
+            })?;
+
+        let result = match indata {
+            Some(data) => {
+                let indata_bio = MemBioSlice::new(data.as_slice())?;
+                unsafe { ffi::PKCS7_verify(self.0, certs_stack.as_ptr(), store.as_ptr(), indata_bio.as_ptr(), out.as_ptr(), flags.bits) }
```
sfackler: Errors

openssl-sys/src/libressl/mod.rs
```diff
@@ -15,6 +15,9 @@ mod v250;
 #[cfg(not(libressl250))]
 mod v25x;
 
+#[repr(C)]
+pub struct pkcs7_st;
```
sfackler: This'll need to be filled out.

openssl-sys/src/ossl10x.rs
```diff
@@ -9,6 +9,9 @@ use libc::{c_int, c_char, c_void, c_long, c_uchar, size_t, c_uint, c_ulong};
 #[cfg(not(ossl101))]
 use libc::time_t;
 
+#[repr(C)]
+pub struct pkcs7_st;
```
sfackler: ^^

openssl/src/crypto/pkcs7/pk7_smime.rs
maennchen: Where do I get a StackRef instead of single X509 certificates? I only found functions like `X509::from_pem` which do not return a stack.

openssl/src/crypto/pkcs7/pk7_smime.rs
maennchen: How can I get `*mut X509` from `&X509Ref`?

maennchen:
@sfackler Most issues should be solved now.

I have a few open points:

* The struct `pkcs7_st` includes a union. I added the union too, but got a strange error from `ctest`. I therefore added `skip` rules. If you know a better option I'll change that.
* I did not change to use `Stack` instead of `Vec` since I can find no place that gives me a stack to work with.
* I did not change from `X509` to `X509Ref` since I the C API takes `X509` and I do not know how to convert them.

openssl/src/crypto/pkcs7/pk7_smime.rs
```diff
@@ -0,0 +1,269 @@
...
+        bcount.append(&mut bcount_bio.get_buf().to_vec());
+
+        if pkcs7.is_null() {
+            Err(ErrorStack::get())
+        } else {
+            Ok(PKCS7(pkcs7))
+        }
+    }
+
+    pub fn decrypt(&self, pkey: &PKey, cert: &X509) -> Result<Vec<u8>, ErrorStack> {
```
sfackler: All you have to do is change `&PKey` to `&PKeyRef` and continue using `.as_ptr()`, same for `X509`.

sfackler:
Yeah, the union stuff looks good.

The user can construct a Stack just like the implementation does right now.

openssl/src/crypto/pkcs7/pk7_smime.rs
```diff
@@ -0,0 +1,269 @@
...
+            } else {
+                Err(ErrorStack::get())
+            }
+        }
+    }
+
+    pub fn smime_read(input: &[u8], bcount: &mut Vec<u8>) -> Result<PKCS7, ErrorStack> {
+        let input_bio = MemBioSlice::new(input)?;
+
+        let bcount_bio = MemBio::new()?;
```
sfackler: SMIME_read_PKCS7 creates the bcount BIO for you - look at the notes section here: https://www.openssl.org/docs/man1.0.2/crypto/SMIME_read_PKCS7.html.

maennchen:
@sfackler

> The user can construct a Stack just like the implementation does right now.

I understand that. But shouldn't we be consistent then?

If we think that the user should handle stacks himself. Shouldn't this function look like this:

https://github.com/sfackler/rust-openssl/blob/e9ad9f1afd0429dd56af8e59ec9ad74020ca15e3/openssl/src/x509/mod.rs#L587
```rust
pub fn stack_from_pem(pem: &[u8]) -> Result<Stack<X509>, ErrorStack> {
```

maennchen:
@sfackler I just found a problem with the code that I don't know how to solve.

If `PKCS7Flags::empty()` is passed as flags, a segfault is produced.

https://gist.github.com/maennchen/44be49dcf3150b21d8cc8f92c2bc6377

How can I tell rust, to not allow empty flags?

sfackler:
Oh right, I forgot we used Vec there. Seems reasonable to do that here as well.

That seems weird - the man pages for PKCS7_encrypt claims flags is optional: https://wiki.openssl.org/index.php/Manual:PKCS7_encrypt(3)

maennchen:
@sfackler I can not produce it with a test for that. I can cause it with this rustler Nif code:

```rust
pub fn decode_flag(flag: &NifAtom, env: NifEnv) -> NifResult<PKCS7Flags> {
    let flag: String = flag.to_term(env).atom_to_string()?;

    match flag.as_str() {
        "text" => Ok(PKCS7_TEXT),
        "nocerts" => Ok(PKCS7_NOCERTS),
        "nosigs" => Ok(PKCS7_NOSIGS),
        "nochain" => Ok(PKCS7_NOCHAIN),
        "nointern" => Ok(PKCS7_NOINTERN),
        "noverify" => Ok(PKCS7_NOVERIFY),
        "detached" => Ok(PKCS7_DETACHED),
        "binary" => Ok(PKCS7_BINARY),
        "noattr" => Ok(PKCS7_NOATTR),
        "nosmimecap" => Ok(PKCS7_NOSMIMECAP),
        "nooldmimetype" => Ok(PKCS7_NOOLDMIMETYPE),
        "crlfeol" => Ok(PKCS7_CRLFEOL),
        "stream" => Ok(PKCS7_STREAM),
        "nocrl" => Ok(PKCS7_NOCRL),
        "partial" => Ok(PKCS7_PARTIAL),
        "reuse_digest" => Ok(PKCS7_REUSE_DIGEST),
        "no_dual_content" => Ok(PKCS7_NO_DUAL_CONTENT),
         _ => Err(NifError::BadArg)
    }
}

pub fn decode_flags(flags: Vec<NifAtom>, env: NifEnv) -> NifResult<PKCS7Flags> {
    flags
        .iter()
        .fold(Ok(PKCS7Flags::empty()), | acc, term | {
            match acc {
                Ok(flags) => Ok(flags | decode_flag(term, env)?),
                Err(err) => Err(err)
            }
        })
}

pub fn encrypt<'a>(env: NifEnv<'a>, args: &[NifTerm<'a>]) -> NifResult<NifTerm<'a>> {
    match decode_stack(args[0].decode()?) {
        Ok(certs) => {
            let input: String = args[1].decode()?;

            let cipher: Cipher = decode_cypher(args[2].decode()?, env)?;

            if args[3].decode::<Vec<NifAtom>>()?.len() < 1 {
                return Err(NifError::BadArg);
            }

            let flags: PKCS7Flags = decode_flags(args[3].decode()?, env)?;

            match PKCS7::encrypt(&certs, input.as_bytes(), cipher, flags) {
                Ok(pkcs7) => {
                    Ok((atoms::ok(), pkcs7_to_resource(pkcs7)).encode(env))
                }
                Err(errors) => Ok((atoms::error(), error_stack_to_term(errors, env)).encode(env))
            }
        },
        Err(errors) => Ok((atoms::error(), error_stack_to_term(errors, env)).encode(env))
    }
}
```

openssl/src/crypto/pkcs7/pk7_smime.rs
```diff
@@ -0,0 +1,245 @@
...
+
+            if ffi::SMIME_write_PKCS7(output.as_ptr(), self.0, input_bio.as_ptr(), flags.bits) == 1 {
+                Ok(output.get_buf().to_owned())
+            } else {
+                Err(ErrorStack::get())
+            }
+        }
+    }
+
+    pub fn smime_read(input: &[u8], bcount: &mut Vec<u8>) -> Result<PKCS7, ErrorStack> {
```
sfackler: Oh, you will need to call `ffi::init()` here

openssl/src/crypto/pkcs7/pk7_smime.rs
```diff
@@ -0,0 +1,245 @@
...
+        unsafe {
+            if ffi::PKCS7_decrypt(self.0, pkey.as_ptr(), cert.as_ptr(), output.as_ptr(), 0) == 1 {
+                Ok(output.get_buf().to_owned())
+            } else {
+                Err(ErrorStack::get())
+            }
+        }
+    }
+
+    pub fn encrypt(certs: &Stack<X509>, input: &[u8], cypher: Cipher, flags: PKCS7Flags) -> Result<PKCS7, ErrorStack> {
```
sfackler: And here

maennchen: 👍

openssl/src/crypto/pkcs7/pk7_smime.rs
maennchen: 👍

maennchen:
@sfackler I still get segfaults on macos. On linux it is passing...

I just open sourced the project that creates the segfaults:
https://github.com/jshmrtn/ex-openssl (See jshmrtn/ex-openssl#1)

maennchen:
@sfackler I believe that the problems are only happening in combination with rustler. Therefore this PR should be ready to merge.

maennchen:
@sfackler The problem occured because of a linking problem. Erlang compiled using boringssl and this collided with openssl (from brew) used by this library.

sa2ajj:
sorry for the question: but why no documentation update?

maennchen:
@sa2ajj I haven't seen any documentation... Where would I need to put that?

sfackler:
Ah cool, bad interactions with boringssl definitely make sense as the cause of those crashes. Could you add some documentation to the module, types and methods you've added? See the ssl module as an example: https://github.com/sfackler/rust-openssl/blob/master/openssl/src/ssl/mod.rs

idyll:
@sfackler I am looking at dusting off this PR and filling in the holes to get it functional for typical usage. Apple uses PKCS7 for pretty much everything.

So, one thing I am trying to do is get it working on macOS.

WRT to BoringSSL, Apple's BoringSSL is not for public use. It's not intended to be linked to. They have a separate LibreSSL which is what is called if a user interacts with openssl. I've tried linking against OpenSSL and LibreSSL and I am still getting segfaults. (for example, to_pem for X509 segfaults.)

When Apple released 10.13 I know there were similar problems with Erlang because of the BoringSSL installation. I would think that trying to link to openssl directly should resolve this but I know it didn't with Erlang because dynamic linking was still hitting BoringSSL. See: https://elixirforum.com/clicks/track?url=https%3A%2F%2Fbugs.erlang.org%2Fbrowse%2FERL-439&post_id=38259&topic_id=5821

Thoughts? 

I am pretty comfortable with PKCS7, but new to rust. So sorry if there's an obvious way to resolve this.

sfackler:
If you install openssl through Homebrew, openssl-sys will detect and link to that which should avoid conflicts with the integrated boringssl.

idyll:
I know what's going on. Another lib is forcing `-flat_namespace` which is causing things to link to BoringSSL after all. I can fix that.

Looking at this original PR, should I be pulling the PKCS7 code out of the crypto namespace like everything else. It feels strange that it's over by itself. I think it should treated like PKCS12. Thoughts?

sfackler:
Oh yeah I hadn't noticed that it was crypto::pkcs7 - it should be a top-level module.

PR #781, opened 2017-11-27
Upgrade foreign-types

sfackler:
foreign-types 0.3 and 0.2 now share the same types and traits, so this
is backwards compatible.

cc @nox


nox:
Can you publish this as `0.9.22`?

sfackler:
Done!

PR #783, opened 2017-11-30
bump lazy_static to 1

ignatenkobrain:


sfackler:
Thanks! This is a breaking change, so I'm going to leave it open for a bit while I get all of the other changes ready to go.

ignatenkobrain:
@sfackler strictly speaking, it is not. It is same as 0.2.11 release of lazy_static. If I understand correctly it was done to inform that lazy_static is mature enough.

sfackler:
Oh right, sorry! I was somehow conflating this with bitflags...

PR #784, opened 2017-12-02
Adding dp(), dq() and qi() methods to RSA, to get the CRT parameters back

P-E-Meunier:
This is just a way to get all parameters passed to `from_private_components` back, instead of just the public and secret ones.


sfackler:
Thanks!

P-E-Meunier:
Thanks for merging! Would you by chance know when you plan to publish the next version on crates.io?

PR #785, opened 2017-12-03
Splitting the sign::Signer::finish function, to avoid allocations

P-E-Meunier:


openssl/src/sign.rs
```diff
@@ -139,25 +139,48 @@ impl<'a> Signer<'a> {
...
+        }
+    }
+
+    /// Outputs the signature into the provided buffer, returning the
+    /// length of that buffer.
+    ///
+    /// This method will fail if the buffer is not large enough for
+    /// the signature, one can use `finish_len` to get an upper bound
+    /// on the required size.
+    pub fn finish_into<B:AsMut<[u8]>>(&self, mut buf: B) -> Result<usize, ErrorStack> {
```
sfackler: Let's just have `buf` be a `&mut [u8]`.

P-E-Meunier: I'm curious, why?

sfackler: It simplifies the function signature and speeds up compilation a bit by avoiding a bunch of copies of the method. Similar methods in the standard library take `&mut [u8]` as well. You're either already going to have a `&mut [u8]` or a `Vec<u8>`, which coerces into a slice.

P-E-Meunier: Ok, done! Thanks for taking the time to explain.

PR #805, opened 2017-12-28
Add recent LibreSSL versions

jan-schreib:
* added 2.6.4 release
* added 2.7.0 version which is running on OpenBSD-current.
* added 2.8.0 as RUST_LIBRESSL_NEW

2.6.4 was [released](https://www.libressl.org/releases.html) on December 19th

OpenBSD-current:

```
$ cat /usr/include/openssl/opensslv.h
/* $OpenBSD: opensslv.h,v 1.44 2017/12/11 11:04:04 bcook Exp $ */
#ifndef HEADER_OPENSSLV_H
#define HEADER_OPENSSLV_H

/* These will change with each release of LibreSSL-portable */
#define LIBRESSL_VERSION_NUMBER 0x2070000fL
#define LIBRESSL_VERSION_TEXT   "LibreSSL 2.7.0"
```
```
$ openssl version
LibreSSL 2.7.0
```

### DIFF
--- a/openssl-sys/build.rs
+++ b/openssl-sys/build.rs
@@ -468,6 +468,13 @@ See rust-openssl README for more information:
         println!("cargo:libressl_version=263");
         println!("cargo:version=101");
         Version::Libressl
+    } else if expanded.contains("RUST_LIBRESSL_264") {
+        println!("cargo:rustc-cfg=libressl");
+        println!("cargo:rustc-cfg=libressl264");
+        println!("cargo:libressl=true");
+        println!("cargo:libressl_version=264");
+        println!("cargo:version=101");
+        Version::Libressl
     } else if expanded.contains("RUST_LIBRESSL_26X") {
         println!("cargo:rustc-cfg=libressl");
         println!("cargo:rustc-cfg=libressl26x");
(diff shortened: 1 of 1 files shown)


jan-schreib:
I wonder why the test is failing. I only touched the version checks.

sfackler:
We don't need to add new 2.6.x versions since that release channel is stable now. I'm guessing the test is failing because some of the cfgs that openssl-sys uses changed when building against 2.6.3 since it's hitting the added 2.6.3 block rather than 2.6.x.

sfackler:
Can you also bump the 2.6.3 build to 2.7.0?

jan-schreib:
I guess you mean the circleci config? LibreSSL 2.7.0 is not relased as portable version, yet. So it would only make sense to bump it to 2.6.4 for now.

So there is no need for 2.6.x any more and there is no 2.6.x in my patch any more. Or do I missunderstood you?

sfackler:
You don't need to add the 2.6.3 and 2.6.4 things in the build script.

I'm not super comfortable supporting 2.7.0 unless we're actually testing to make sure that everything works.

jan-schreib:
Not supporting 2.7.0 will cause all crates depending on rust-openssl to fail on OpenBSD-current.
I think a solution would be to be able to do automated tests on OpenBSD-current with the newest, not yet portable LibreSSL version.

sfackler:
Sure, I'd be fine with a test build running BSD.

PR #810, opened 2017-12-31
Parameterize keys over what they contain

### LINKED ISSUE #790: Segfault when attempting to sign with a public key
Since the `RSA` type does not tell whether the key contains all components or just the public ones, it's possible to be stupid and cause rust-openssl to segfault when signing with a key that doesn't have the private components.

This could be easily prevented at compile time, by adding two types, `pub enum Public {}` and `pub enum Private{}`, and a field of type `PhantomData<T>` to `Rsa`. Then, `sign` would force the key to be of type `Rsa<Private>`, and verify could be `Rsa<T>`.

The modification is nontrivial though, since this code is currently generated by macros.

This would also make some parts of the interface easier (and thus safer) to read and use, such as `PKey::rsa()`, which is currently confusing (does it contain a private or a public key?).

sfackler:
Closes #790

### DIFF
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -1723,6 +1723,7 @@ extern "C" {
     pub fn EC_KEY_new() -> *mut EC_KEY;
     pub fn EC_KEY_new_by_curve_name(nid: c_int) -> *mut EC_KEY;
     pub fn EC_KEY_dup(key: *const EC_KEY) -> *mut EC_KEY;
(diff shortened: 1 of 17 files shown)


P-E-Meunier:
This is awesome! I started similar changes when we discussed the issue, but got a little despaired in face of the size of the changes required.

Do you need a more careful review than my rapid screening just now?

sfackler:
Nope, no need to look over in more detail than you feel like. Just giving you a heads up!

PR #824, opened 2018-01-07
Add PKey docs

apeduru:
Refs [#716](https://github.com/sfackler/rust-openssl/issues/716)

### DIFF
--- a/openssl/src/pkey.rs
+++ b/openssl/src/pkey.rs
@@ -30,7 +30,7 @@
 //!
 //! extern crate openssl;
 //!
-//! use openssl::rsa:Rsa;
+//! use openssl::rsa::Rsa;
 //! use openssl::pkey::Pkey;
 //!
 //! fn main() {
--- a/openssl/src/pkey.rs
+++ b/openssl/src/pkey.rs
@@ -7,7 +7,7 @@
 //! encrypt information using a user's public key. The encrypted information can
 //! only be deciphered using that user's private key.
 //!
-//! This module offers support for four popular algorithms:
+//! This module offers support for five popular algorithms:
 //!
 //! * RSA
 //!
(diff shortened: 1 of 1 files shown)


openssl/src/pkey.rs
```diff
@@ -1,3 +1,44 @@
...
+//! * Elliptic Curves
+//!
+//! These algorithms rely on hard mathematical problems - namely integer factorization,
+//! discrete logarithms, and elliptic curve relationships - that currently do not
+//! yield efficient solutions. This property ensures the security of these
+//! cryptographic algorithms.
+//!
+//! # Example
+//!
+//! Generate a 2048-bit RSA public/private key pair.
```
sfackler: Not sure this example is really showing much.

apeduru: I can add another couple statements serializing the pub/priv key pair to PEM. Did you have something else in mind? I can also spin it out into a separate example.

sfackler: I don't really have any particular example in mind for this module. `PKey` doesn't really do much on its own.

apeduru:
not sure why the macos build failed

openssl/src/pkey.rs
```diff
@@ -1,3 +1,48 @@
+//! Public/private key processing.
+//!
+//! Asymmetric public key algorithms solve the problem of establishing and sharing
+//! secret keys to securely send and receive messages.
+//! This system uses a pair of keys: a public key, which can be freely
+//! distributed, and a private key, which is kept to oneself. An entity may
+//! encrypt information using a user's public key. The encrypted information can
+//! only be deciphered using that user's private key.
+//!
+//! This module offers support for four popular algorithms:
```
sfackler: HMAC is also mixed in here.

sfackler:
Yeah, don't worry about the OSX failure.

Looks good to me other than the one note about HMAC!

PR #827, opened 2018-01-15
add email and uri to GeneralName

ian-p-cooke:
I need access to an email address added to the subject alternate names.  I added support for email and uri which are both IA5Strings according to "https://tools.ietf.org/html/rfc5280#section-4.2.1.6".  The certificate I created isn't exactly the same as the old one (e.g. different issuer) but I think that's ok.  If not please re-create the cert as you see fit and add the following extensions (openssl.conf format):

> [subject_alt_names]
> DNS.1 = example.com
> IP.1 = 127.0.0.1
> IP.2 = 0:0:0:0:0:0:0:1
> email.1 = test@example.com
> URI.1 = "http://www.example.com"`


openssl/src/x509/mod.rs
```diff
@@ -1090,24 +1090,39 @@ foreign_type_and_impl_send_sync! {
...
+            // IA5Strings are stated to be ASCII (specifically IA5). Hopefully
             // OpenSSL checks that when loading a certificate but if not we'll
             // use this instead of from_utf8_unchecked just in case.
             str::from_utf8(slice).ok()
         }
     }
 
+    /// Returns the contents of this `GeneralName` if it is an `rfc822Name`.
+    pub fn email(&self) -> Option<&str> {
+		self.ia5_string(ffi::GEN_EMAIL)
```
sfackler: Could you change the hard tabs to soft tabs?

sfackler:
Nah, changing the issuer is fine.

openssl/src/x509/mod.rs
ian-p-cooke: fixed on master.  I squashed the original commit as I wasn't sure how github handled multiple commits on a pull request.

sfackler: Multiple commits will work, but squashing is also fine.

sfackler:
Thanks!

PR #833, opened 2018-02-04
Support EVP_des_ede3.

CmdrMoozy:
This cipher is used, for example, for DES challenges for authenticating against a Yubikey, so supporting it in rust-openssl is generally useful.


sfackler:
Thanks

PR #835, opened 2018-02-11
[WIP] Expose bindings needed for TLS1.3 stateless handshakes

Ralith:
This is necessary for some non-TCP applications of TLS, e.g. QUIC implementations.

I have only a vague idea what I'm doing here, and this hasn't been well tested yet, but it's a beginning.

For some reason I can't get the new bindings to actually be compiled with any feature flags other than `--all-features`.

### DIFF
--- a/openssl/src/ssl/callbacks.rs
+++ b/openssl/src/ssl/callbacks.rs
@@ -361,3 +361,53 @@ where
 
     callback(ssl, line);
 }
+
+pub extern "C" fn raw_cookie_generate<F>(
+    ssl: *mut ffi::SSL,
+    cookie: *mut c_uchar,
+    cookie_len: *mut c_uint
+) -> c_int
+where
+    F: Fn(&mut SslRef, &mut [u8]) -> Result<usize, ErrorStack> + 'static + Sync + Send
+{
+    unsafe {
+        let ssl_ctx = ffi::SSL_get_SSL_CTX(ssl as *const _);
+        let callback = ffi::SSL_CTX_get_ex_data(ssl_ctx, get_callback_idx::<F>());
+        let ssl = SslRef::from_ptr_mut(ssl);
+        let callback = &*(callback as *mut F);
+        // We subtract 1 from DTLS1_COOKIE_LENGTH as the ostensible value, 256, is erroneous but retained for
+        // compatibility. See comments in dtls1.h.
+        let slice = slice::from_raw_parts_mut(cookie as *mut u8, ffi::DTLS1_COOKIE_LENGTH as usize - 1);
+        match callback(ssl, slice) {
+            Ok(len) => {
+                *cookie_len = len as c_uint;
+                1
+            }
+            Err(_) => 0,
+        }
+    }
+}
+
+#[cfg(ossl110)]
+type CookiePtr = *const c_uchar;
+
+#[cfg(not(ossl110))]
+type CookiePtr = *mut c_uchar;
+
+pub extern "C" fn raw_cookie_verify<F>(
+    ssl: *mut ffi::SSL,
+    cookie: CookiePtr,
+    cookie_len: c_uint
+) -> c_int
+where
+    F: Fn(&mut SslRef, &[u8]) -> bool + 'static + Sync + Send
+{
+    unsafe {
+        let ssl_ctx = ffi::SSL_get_SSL_CTX(ssl as *const _);
+        let callback = ffi::SSL_CTX_get_ex_data(ssl_ctx, get_callback_idx::<F>());
+        let ssl = SslRef::from_ptr_mut(ssl);
+        let callback = &*(callback as *mut F);
+        let slice = slice::from_raw_parts(cookie as *const c_uchar as *const u8, cookie_len as usize);
+        callback(ssl, slice) as c_int
+    }
+}
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -1311,6 +1311,51 @@ impl SslContextBuilder {
         }
     }
 
+    /// Sets the callback for generating an application cookie for stateless handshakes.
+    ///
+    /// The callback will be called with the SSL context and a slice into which the cookie
+    /// should be written. The callback should return the number of bytes written.
+    ///
+    /// This corresponds to `SSL_CTX_set_cookie_generate_cb`.
+    pub fn set_cookie_generate_cb<F>(&mut self, callback: F)
+    where
+        F: Fn(&mut SslRef, &mut [u8]) -> Result<usize, ErrorStack> + 'static + Sync + Send
+    {
+        unsafe {
+            let callback = Box::new(callback);
+            ffi::SSL_CTX_set_ex_data(
+                self.as_ptr(),
+                get_callback_idx::<F>(),
+                mem::transmute(callback),
+            );
+            ffi::SSL_CTX_set_cookie_generate_cb(self.as_ptr(), Some(raw_cookie_generate::<F>))
+        }
+    }
+
+    /// Sets the callback for verifying an application cookie for stateless handshakes.
+    ///
+    /// The callback will be called with the SSL context and the cookie supplied by the
+    /// client. It should return true if and only if the cookie is valid.
+    ///
+    /// Note that the OpenSSL implementation independently verifies the integrity of
+    /// application cookies using an HMAC before invoking the supplied callback.
+    ///
+    /// This corresponds to `SSL_CTX_set_cookie_verify_cb`.
+    pub fn set_cookie_verify_cb<F>(&mut self, callback: F)
+    where
+        F: Fn(&mut SslRef, &[u8]) -> bool + 'static + Sync + Send
+    {
+        unsafe {
+            let callback = Box::new(callback);
+            ffi::SSL_CTX_set_ex_data(
+                self.as_ptr(),
+                get_callback_idx::<F>(),
+                mem::transmute(callback),
+            );
+            ffi::SSL_CTX_set_cookie_verify_cb(self.as_ptr(), Some(raw_cookie_verify::<F>))
+        }
+    }
+
     /// Sets the extra data at the specified index.
     ///
     /// This can be used to provide data to callbacks registered with the context. Use the
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -18,6 +18,11 @@ mod ossl110;
 #[cfg(ossl110)]
 pub use ossl110::*;
 
+#[cfg(ossl111)]
+mod ossl111;
+#[cfg(ossl111)]
+pub use ossl111::*;
+
 #[cfg(libressl)]
 mod libressl;
 #[cfg(libressl)]
@@ -1430,6 +1435,8 @@ pub const GEN_URI: c_int = 6;
 pub const GEN_IPADD: c_int = 7;
 pub const GEN_RID: c_int = 8;
 
+pub const DTLS1_COOKIE_LENGTH: c_uint = 256;
+
 // macros
 pub unsafe fn BIO_get_mem_data(b: *mut BIO, pp: *mut *mut c_char) -> c_long {
     BIO_ctrl(b, BIO_CTRL_INFO, 0, pp as *mut c_void)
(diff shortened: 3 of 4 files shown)


openssl-sys/build.rs
```diff
@@ -336,8 +337,10 @@ RUST_LIBRESSL_251
 RUST_LIBRESSL_250
 #elif defined (LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x20500000
 RUST_LIBRESSL_OLD
-#elif OPENSSL_VERSION_NUMBER >= 0x10101000
+#elif OPENSSL_VERSION_NUMBER >= 0x10200000
```
sfackler: This should look for 0x10102000.

openssl/Cargo.toml
```diff
@@ -17,6 +17,7 @@ all-features = true
 v101 = []
 v102 = []
 v110 = []
+v111 = ["v110"]
```
sfackler: Why does this inherit?

openssl-sys/build.rs
Ralith: Is it not a reasonable assumption that 1.1.* will be backwards-compatible?

openssl/Cargo.toml
Ralith: Because 1.1.1 is a superset of 1.1.0, as far as I'm aware. An alternative would be to have v11x and make v111 and v110 both imply it, but I think v110 and v11x would end up functionally equivalent, which feels a bit silly.

Ralith:
Appveyor failure looks spurious:
```
-- Downloading http://www.nasm.us/pub/nasm/releasebuilds/2.12.02/win32/nasm-2.12.02-win32.zip... Failed. Status: 7;"Couldn't connect to server"
CMake Error at scripts/cmake/vcpkg_download_distfile.cmake:122 (message):
  
      Failed to download file.
      Add mirrors or submit an issue at https://github.com/Microsoft/vcpkg/issues
Call Stack (most recent call first):
  scripts/cmake/vcpkg_find_acquire_program.cmake:175 (vcpkg_download_distfile)
  ports/openssl/portfile.cmake:11 (vcpkg_find_acquire_program)
  scripts/ports.cmake:72 (include)
```

openssl/src/ssl/mod.rs
```diff
@@ -2106,6 +2150,27 @@ impl Ssl {
...
+    ///
+    /// Returns true if a complete ClientHello containing a valid cookie was
+    /// read. Requires that cookie generation and verification callbacks were set on the
+    /// SSL context.
+    ///
+    /// This corresponds to [`SSL_stateless`]
+    ///
+    /// [`SSL_stateless`]: https://www.openssl.org/docs/manmaster/man3/SSL_stateless.html
+    #[cfg(all(feature = "v111", ossl111))]
+    pub fn stateless<S>(&mut self, stream: &mut S) -> bool
```
Ralith: I'd like some input on the design of this. [The docs](https://www.openssl.org/docs/manmaster/man3/SSL_stateless.html) say:

> For SSL_stateless() if an entire ClientHello message cannot be read without the "read" BIO becoming empty then the SSL_stateless() call will fail.

so it seems like it should probably be used with in-memory buffers and always complete immediately? I'm not a huge fan of allocating a new BIO on every call...

openssl-sys/build.rs
sfackler: Ah sure, that makes sense.

openssl-sys/build.rs
```diff
@@ -464,14 +467,22 @@ See rust-openssl README for more information:
         println!("cargo:libressl_version=26x");
         println!("cargo:version=101");
         Version::Libressl
+    } else if expanded.contains("RUST_OPENSSL_111") {
+        println!("cargo:rustc-cfg=ossl11x");
+        println!("cargo:rustc-cfg=ossl111");
+        println!("cargo:version_major=11x");
```
sfackler: Let's not output version_major for now.

openssl/src/ssl/mod.rs
```diff
@@ -1120,6 +1120,50 @@ impl SslContextBuilder {
...
+    /// Sets the callback for generating a cookie for stateless handshakes.
+    ///
+    /// The callback will be called with the SSL context and a slice into which the cookie
+    /// should be written. The callback should return the number of bytes written.
+    ///
+    /// This corresponds to `SSL_CTX_set_cookie_generate_cb`.
+    #[cfg(all(feature = "v111", ossl111))]
+    pub fn set_cookie_generate_cb<F>(&mut self, callback: F)
+    where
+        F: Fn(&mut SslRef, &mut [u8]) -> Result<usize, ()> + 'static + Sync + Send
```
sfackler: Callbacks normally return an `ErrorStack` on error.

openssl/src/ssl/callbacks.rs
```diff
@@ -272,3 +272,46 @@ where
...
+where
+    F: Fn(&mut SslRef, &mut [u8]) -> Result<usize, ()> + 'static + Sync + Send
+{
+    unsafe {
+        let ssl_ctx = ffi::SSL_get_SSL_CTX(ssl as *const _);
+        let callback = ffi::SSL_CTX_get_ex_data(ssl_ctx, get_callback_idx::<F>());
+        let ssl = SslRef::from_ptr_mut(ssl);
+        let callback = &*(callback as *mut F);
+        let slice = slice::from_raw_parts_mut(cookie as *mut u8, ffi::SSL_COOKIE_LENGTH as usize);
+        if let Ok(len) = callback(ssl, slice) {
```
sfackler: This'll be a bit more compact as

```rust
match callback(ssl, slice) {
    Ok(len) => {
        *cookie_len = len as u32;
        1
    }
    Err(_) => 0,
}
```

openssl/src/ssl/callbacks.rs
```diff
@@ -272,3 +272,46 @@ where
...
+            *cookie_len = len as u32;
```
sfackler: `c_uint`, not `u32`.

openssl/src/ssl/mod.rs
sfackler: How is this normally used in C?

openssl/src/ssl/mod.rs
Ralith: How should I come up with one, and what good would it do? The OpenSSL implementation only checks whether the return code is nonzero.

openssl-sys/build.rs
Ralith: This was added so that systest's build.rs can set the proper `cfg`s. Is there a different approach I should use?

openssl/src/ssl/mod.rs
Ralith: I'm not sure. The API was only added a few weeks ago in https://github.com/openssl/openssl/pull/4435. The only example I'm aware of is [the unit test](https://github.com/openssl/openssl/blob/master/test/sslapitest.c#L2586), which I don't think tells us much about its intended relationship to async IO.

It's worth noting that in the case of QUIC implementations, [the handshake messages involved will always be sent/received as a single packet](https://quicwg.github.io/base-drafts/draft-ietf-quic-tls.html#clienthello-size), so a fully synchronous paradigm works for my case. I'm also less bothered by the short-lived BIOs than I was, since either the whole `Ssl` object will be disposed immediately regardless (for a ClientHello not containing a valid cookie), or the BIOs only need to be replaced once for the `Ssl`'s entire lifetime, when it's turned into a stream.

openssl-sys/build.rs
sfackler: You can have systest's build.rs parse version.

openssl/src/ssl/mod.rs
sfackler: Take a look at the other callbacks - right now we just throw the error away but the intent is to push the state back onto the thread local error stack.

openssl/src/ssl/mod.rs
Ralith: It's also not apparent to me that this could be used in any way but synchronously, since the interface provides no way to distinguish an incomplete ClientHello from an invalid one.

My understanding is that generally you will either want to dispose of the `Ssl` object entirely if this returns false, or proceed to calling `SSL_accept` if it returns true. That invites tweaking the signature to something like `fn stateless<S>(self, stream: S) -> Option<SslStream<S>>`. I think I'll try to make that work, but I'm not certain that forcing a single inner stream type is convenient.

sfackler: It's intended to be used with datagram sockets, so the streaming issues aren't applicable, right? You always get the whole message in one read call.

I think we may want a lower-level type that lets you drive things more directly. `SSL_read_early_data` is weird in a similar way. Something like

```rust
struct SslStreamBuilder<T>{ ... }

impl<T> SslStreamBuilder<T>
    where T: Read + Write,
{
    pub fn new(stream: T) -> SslStreamBuilder<T> { ... }

    pub fn stateless(&mut self) -> Result<()> { ... }

    pub fn read_early_data(&mut self, buf: &mut [u8]) -> Result<usize> { ... }

    pub fn write_early_data(&mut self, buf: &[u8]) -> Result<usize> { ... }

    pub fn accept(&mut self) -> Result<SslStream<T>, MidHandshakeSslStream<T>> { .. }

    pub fn connect(&mut self) -> Result<SslStream<T>, MidHandshakeSslStream<T>> { .. }
}
```

openssl/src/ssl/mod.rs
Ralith: The other callbacks punt the problem to the user. The user isn't expected to be making any SSL-related calls inside those two callbacks, so there's nowhere for an SSL error to come from.

openssl/src/ssl/mod.rs
Ralith: That's an interesting idea, but it places early data I/O *before* the accept/connect call, which seems wrong.

openssl/src/ssl/mod.rs
sfackler: What problem is being punted?

openssl/src/ssl/mod.rs
sfackler: > When called by a client, SSL_write_early_data() must be the first IO function called on a new connection, i.e. it must occur before any calls to SSL_write_ex(3), SSL_read_ex(3), SSL_connect(3), SSL_do_handshake(3) or other similar functions.

openssl/src/ssl/mod.rs
Ralith: Coming up with some meaningful openssl::Error value.

sfackler: There is only one way to construct an `ErrorStack`.

openssl/src/ssl/mod.rs
sfackler: When would you want to recover it? Isn't it going to fail if the client isn't legitimate?

openssl/src/ssl/mod.rs
Ralith: I'm confused. Why is application code supposed to generate a SSL error?

openssl/src/ssl/mod.rs
Ralith: `SSL_stateless` "fails" if the client is (potentially) legitimate but does not yet have a cookie, and in doing so replies with a HelloRetryRequest containing a cookie. This needs to actually get sent, or progress will never be made.

openssl/src/ssl/mod.rs
sfackler: In other callbacks it would be produced by some other openssl function. For example, the `tmp_dh` callback may hit an error from `Dh::params_from_pem`. The potential source here could be `rand::rand_bytes` for example.

In any case, it is strictly more powerful to return `ErrorStack` than `()`, and matches with every single other callback function.

Ralith: That strikes me as a very strange idiom, especially as it is my current understanding that implementations of this callback will typically be pure and total MAC computations on a source address. Still, if you insist, it's not like it has any more serious drawbacks than weirdness.

sfackler: If these callbacks can never fail, it should just return `usize`, yes?

openssl/src/ssl/mod.rs
sfackler: How would recovering the stream affect SSL_stateless's write being sent or not?

I think we're going to have to look at a real world use of this function to figure out how it's intended to be used.

Ralith: Maybe I'm confused about how this whole wrapped-stream paradigm is supposed to work in general. I don't expect a dropped stream to complete pending writes. In my experimental code, the stream I'm wrapping is just some in-memory buffers, and dropping them would lose the written data entirely.

Ralith:
I've rebased onto latest master and reprised this to omit the contentious high-level `SSL_stateless` binding, preserving the FFI binding and FFI and high-level cookie callback bindings.

sfackler:
Thanks!

PR #839, opened 2018-02-14
OpenSSL 1.1.1 support

sfackler:


sfackler:
Tests are going to fail because of https://github.com/openssl/openssl/issues/5362.

sfackler:
cc @Ralith

openssl-sys/build.rs
```diff
@@ -343,8 +343,10 @@ RUST_LIBRESSL_251
 RUST_LIBRESSL_250
 #elif defined (LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x20500000
 RUST_LIBRESSL_OLD
-#elif OPENSSL_VERSION_NUMBER >= 0x10101000
+#elif OPENSSL_VERSION_NUMBER >= 0x10102000
```
Ralith: Why reject 1.1.2, given the effort they're going to to maintain ABI compat?

openssl/src/ssl/test.rs
```diff
@@ -1038,6 +1041,13 @@ fn tmp_dh_callback() {
 
     let stream = TcpStream::connect(("127.0.0.1", port)).unwrap();
     let mut ctx = SslContext::builder(SslMethod::tls()).unwrap();
+    // TLS 1.3 has no DH suites, and openssl isn't happy if the max version has no suites :(
```
Ralith: Is this a temporary problem?

systest/build.rs
```diff
@@ -29,12 +29,10 @@ fn main() {
...
-        (
-            env::var("DEP_OPENSSL_VERSION"),
-            env::var("DEP_OPENSSL_PATCH"),
-        )
-    {
+    if let (Ok(version), Ok(patch)) = (
+        env::var("DEP_OPENSSL_VERSION"),
+        env::var("DEP_OPENSSL_PATCH"),
+    ) {
         cfg.cfg(&format!("ossl{}{}", version, patch), None);
```
Ralith: Missing ossl111 implies ossl110 here?

openssl/src/ssl/test.rs
sfackler: Dunno

openssl-sys/build.rs
sfackler: Their notion of ABI compat is not all that strict. Several constants have changed values between 1.1.0 and 1.1.1. 1.1.2 isn't going to be releasing for over a year, so I don't think this check is really going to be a problem for people.

openssl/src/ssl/test.rs
sfackler: I'm actually not sure why the ECDHE tests passes - the OpenSSL docs claim that the TLS 1.3 suites aren't in the groups that we're activating.

systest/build.rs
sfackler: Fixed

openssl-sys/build.rs
Ralith: > Several constants have changed values between 1.1.0 and 1.1.1

That's odd, they seemed pretty clear about that specific type of change being forbidden. Do you recall any specifics?

openssl/src/ssl/test.rs
sfackler: Oh right, we only run that on 1.0.1 and 1.0.2.

openssl-sys/build.rs
sfackler: https://github.com/sfackler/rust-openssl/pull/839/files#diff-ceca4680e3c63bdc67652c99c23c0db9

SSL_OP_NO_SSL_MASK changed in 1.1.1, and SSL_OP_ALL changed in 1.1.0f.

PR #840, opened 2018-02-15
 Add des_ede3_cbc cipher and more tests/examples

olehermanse:
For the record, this cipher is deprecated and shouldn't be used in new stuff. It's still in C API, and nice to have access to in rust as well, in my case for maintaining compatibility with old code and versions.

### DIFF
--- a/openssl/src/symm.rs
+++ b/openssl/src/symm.rs
@@ -137,6 +137,10 @@ impl Cipher {
         unsafe { Cipher(ffi::EVP_des_ede3()) }
     }
 
+    pub fn des_ede3_cbc() -> Cipher {
+        unsafe { Cipher(ffi::EVP_des_ede3_cbc()) }
+    }
+
     pub fn rc4() -> Cipher {
         unsafe { Cipher(ffi::EVP_rc4()) }
     }
@@ -1048,6 +1052,17 @@ mod tests {
         let iv = "5cc118306dc702e4";
 
         cipher_test(super::Cipher::des_ede3(), pt, ct, key, iv);
+    }
+
+    #[test]
+    fn test_des_ede3_cbc() {
+
+        let pt = "54686973206973206120746573742e";
+        let ct = "6f2867cfefda048a4046ef7e556c7132";
+        let key = "7cb66337f3d3c0fe7cb66337f3d3c0fe7cb66337f3d3c0fe";
+        let iv = "0001020304050607";
+
+        cipher_test(super::Cipher::des_ede3_cbc(), pt, ct, key, iv);
     }
 
     #[test]
--- a/openssl/src/symm.rs
+++ b/openssl/src/symm.rs
@@ -22,6 +22,36 @@
 //!       \xFB\x3C\x5E\xC4\x59\x72\x4A\xF4\x7C\xA1",
 //!     &ciphertext[..]);
 //! ```
+//!
+//! Encrypting an assymetric key with a symmetric cipher
+//!
+//! ```
+//! use openssl::rsa::{Padding, Rsa};
+//! use openssl::symm::Cipher;
+//!
+//! // Generate keypair and encrypt private key:
+//! let keypair = Rsa::generate(2048).unwrap();
+//! let cipher = Cipher::aes_256_cbc();
+//! let pubkey_pem = keypair.public_key_to_pem_pkcs1().unwrap();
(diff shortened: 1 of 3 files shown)


openssl/src/symm.rs
```diff
@@ -22,6 +22,45 @@
...
+//! {   // Generate, encrypt and store keypair:
+//!     let keypair = Rsa::generate(2048).unwrap();
+//!     let cipher = Cipher::aes_256_cbc();
+//!     let pubkey_pem = keypair.public_key_to_pem_pkcs1().unwrap();
+//!     let privkey_pem = keypair.private_key_to_pem_passphrase(cipher, b"Rust").unwrap();
+//!     File::create("localhost.pub").unwrap().write_all(&pubkey_pem).unwrap();
+//!     File::create("localhost.priv").unwrap().write_all(&privkey_pem).unwrap();
+//! }
+//!
+//! // Load private and public key from generated file:
```
sfackler: Why are we writing the keys to disk and then immediately reading them again?

olehermanse: The intention was just to show an easy to understand and usable example. It's easy to split into 2 functions. If a reader needs to generate and store an encrypted key, they can use the first part, if they need to load keys from disk, they can use the second part.

I can split it into 2 examples, or shorten it down by removing all File I/O, if you prefer?

sfackler: Yeah, let's split them up and avoid the File IO.

olehermanse:
@sfackler I removed File I/O from the example. What do you think?

openssl/src/symm.rs
```diff
@@ -22,6 +22,37 @@
...
+//! let pubkey_pem = keypair.public_key_to_pem_pkcs1().unwrap();
+//! let privkey_pem = keypair.private_key_to_pem_passphrase(cipher, b"Rust").unwrap();
+//! // pubkey_pem and privkey_pem could be written to file here.
+//!
+//! // Load private and public key from string:
+//! let pubkey = Rsa::public_key_from_pem_pkcs1(&pubkey_pem).unwrap();
+//! let privkey = Rsa::private_key_from_pem_passphrase(&privkey_pem, b"Rust").unwrap();
+//!
+//! // Use the asymmetric keys to encrypt and decrypt a short message:
+//! let msg = "Foo bar".as_bytes();
```
sfackler: You can just use `b"Foo bar"` here.

openssl/src/symm.rs
```diff
@@ -22,6 +22,37 @@
...
+//! let mut encrypted = vec![0; pubkey.size() as usize];
+//! let mut decrypted = vec![0; privkey.size() as usize];
+//! let len = pubkey.public_encrypt(&msg, &mut encrypted, Padding::PKCS1).unwrap();
+//! assert!(len > msg.len());
+//! let len = privkey.private_decrypt(&encrypted, &mut decrypted, Padding::PKCS1).unwrap();
+//! let output_string = String::from_utf8_lossy(&decrypted[..len]);
```
sfackler: Why not `String::from_utf8`? We want to assert it is valid UTF, right?

openssl/src/symm.rs
```diff
@@ -22,6 +22,37 @@
...
+//! assert_eq!(output_string.len(), msg.len());
```
sfackler: This check is redundant with the check below.

openssl/src/symm.rs
```diff
@@ -22,6 +22,37 @@
...
+//! assert_eq!("Foo bar", output_string);
+//! println!("{}", output_string);
```
sfackler: Why print it if we know exactly what it contains?

openssl/src/symm.rs
olehermanse: Thanks, fixed.

openssl/src/symm.rs
olehermanse: Fair enough, removed.

openssl/src/symm.rs
olehermanse: In short, because `String::from_utf8_lossy` is easier. I have edited the example now to use `String::from_utf8`

openssl/src/symm.rs
olehermanse: Because they serve different purposes. The assert tests correctness for the test. The print is for a user to see, this is an example, beginners will want some feedback if they run it, imo. Edited the print to be more descriptive.

olehermanse:
@sfackler Thanks for the review. I have addressed all your comments and amended the last commit.

PR #842, opened 2018-02-16
Fix symm decrypt documentation example

nyradr:
Referring to issue #723.

There is a small inconsistency in the openssl::symm::decrypt function documentation example.
The doc says "Decrypt data in AES256 ECB mode" while using AES128 CBC in the example.


sfackler:
Thanks!

PR #856, opened 2018-03-01
Make it possible to use cmac

Flakebi:
Hi,
currently, one cannot use cmac with rust-openssl because a `Signer` always requires digest. For cmac, only a cipher is used but no digest.

CMAC/OMAC1 are needed e.g. to implement EAX.

### DIFF
--- a/openssl/src/pkey.rs
+++ b/openssl/src/pkey.rs
@@ -365,6 +365,68 @@ impl PKey<Private> {
         }
     }
 
+    /// Creates a new `PKey` containing a CMAC key.
+    ///
+    /// CMAC is only supported since the version 1.1.0 of OpenSSL.
+    ///
+    /// # Note
+    ///
+    /// To compute CMAC values, use the `sign` module.
+    #[cfg(any(all(ossl110, feature = "v110"), all(ossl111, feature = "v111")))]
+    pub fn cmac(cipher: &::symm::Cipher, key: &[u8]) -> Result<PKey<Private>, ErrorStack> {
+        unsafe {
+            assert!(key.len() <= c_int::max_value() as usize);
+            let kctx = cvt_p(ffi::EVP_PKEY_CTX_new_id(
+                ffi::EVP_PKEY_CMAC,
+                ptr::null_mut(),
+            ))?;
+
+            let ret = (|| {
+                cvt(ffi::EVP_PKEY_keygen_init(kctx))?;
+
+                // Set cipher for cmac
+                cvt(ffi::EVP_PKEY_CTX_ctrl(
+                    kctx,
+                    -1,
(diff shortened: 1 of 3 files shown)


openssl/src/sign.rs
```diff
@@ -100,7 +100,7 @@ impl<'a> Signer<'a> {
     /// OpenSSL documentation at [`EVP_DigestSignInit`].
     ///
     /// [`EVP_DigestSignInit`]: https://www.openssl.org/docs/manmaster/man3/EVP_DigestSignInit.html
-    pub fn new<T>(type_: MessageDigest, pkey: &'a PKeyRef<T>) -> Result<Signer<'a>, ErrorStack>
+    pub fn new<T>(type_: Option<MessageDigest>, pkey: &'a PKeyRef<T>) -> Result<Signer<'a>, ErrorStack>
```
sfackler: This is a breaking change. I'm not planning on a breaking release for about a year, FYI.

openssl/src/pkey.rs
```diff
@@ -332,6 +332,69 @@ impl PKey<Private> {
...
+    /// To compute CMAC values, use the `sign` module.
+    pub fn cmac(cipher: &::symm::Cipher, key: &[u8]) -> Result<PKey<Private>, ErrorStack> {
+        unsafe {
+            assert!(key.len() <= c_int::max_value() as usize);
+            let kctx = cvt_p(ffi::EVP_PKEY_CTX_new_id(
+                ffi::EVP_PKEY_CMAC,
+                ptr::null_mut(),
+            ))?;
+            let ret = cvt(ffi::EVP_PKEY_keygen_init(kctx));
+            if let Err(e) = ret {
```
sfackler: Can this be merged into the block below?

openssl/src/sign.rs
Flakebi: Hm, would you prefer another solution then?
Like adding `new_without_digest`?

sfackler: Yeah something like that seems reasonable.

openssl/src/pkey.rs
```diff
@@ -365,6 +365,68 @@ impl PKey<Private> {
...
     }
 
+    /// Creates a new `PKey` containing a CMAC key.
+    ///
+    /// CMAC is only supported since the version 1.1.0 of OpenSSL.
+    ///
+    /// # Note
+    ///
+    /// To compute CMAC values, use the `sign` module.
+    #[cfg(ossl110)]
```
sfackler: This should also look for the feature gate like the other similar APIs do.

sfackler:
Looks good other than the cfg!

Flakebi:
Ok, nice.
I changed the cfg, I hope it's better now.

sfackler:
Thanks!

PR #858, opened 2018-03-04
Introduce SslStreamBuilder

Ralith:
Draft a new API to support `SSL_stateless` and to allow for future early I/O support as previously discussed in #835.

### DIFF
--- a/openssl-sys/src/ossl111.rs
+++ b/openssl-sys/src/ossl111.rs
@@ -25,7 +25,7 @@ pub type SSL_custom_ext_parse_cb_ex =
                                 chainidx: size_t, al: *mut c_int,
                                 parse_arg: *mut c_void) -> c_int>;
 
-pub const SSL_COOKIE_LENGTH: c_int = 255;
+pub const SSL_COOKIE_LENGTH: c_int = 4096;
 
 pub const SSL_OP_ENABLE_MIDDLEBOX_COMPAT: c_ulong = 0x00100000;
 
@@ -65,4 +65,20 @@ extern "C" {
                                   parse_arg: *mut c_void) -> c_int;
     pub fn SSL_stateless(s: *mut ::SSL) -> c_int;
     pub fn SSL_CIPHER_get_handshake_digest(cipher: *const ::SSL_CIPHER) -> *const ::EVP_MD;
+    pub fn SSL_CTX_set_stateless_cookie_generate_cb(
+        s: *mut ::SSL_CTX,
+        cb: Option<unsafe extern "C" fn(
+            ssl: *mut ::SSL,
+            cookie: *mut c_uchar,
+            cookie_len: *mut size_t
+        ) -> c_int>
+    );
+    pub fn SSL_CTX_set_stateless_cookie_verify_cb(
+        s: *mut ::SSL_CTX,
+        cb: Option<unsafe extern "C" fn(
+            ssl: *mut ::SSL,
+            cookie: *const c_uchar,
+            cookie_len: size_t
+        ) -> c_int>
+    );
 }
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -1437,8 +1437,9 @@ impl SslContextBuilder {
     /// The callback will be called with the SSL context and a slice into which the cookie
     /// should be written. The callback should return the number of bytes written.
     ///
-    /// This corresponds to `SSL_CTX_set_cookie_generate_cb`.
-    pub fn set_cookie_generate_cb<F>(&mut self, callback: F)
+    /// This corresponds to `SSL_CTX_set_stateless_cookie_generate_cb`.
+    #[cfg(ossl111)]
+    pub fn set_stateless_cookie_generate_cb<F>(&mut self, callback: F)
     where
         F: Fn(&mut SslRef, &mut [u8]) -> Result<usize, ErrorStack> + 'static + Sync + Send,
     {
@@ -1447,9 +1448,9 @@ impl SslContextBuilder {
             ffi::SSL_CTX_set_ex_data(
                 self.as_ptr(),
                 get_callback_idx::<F>(),
-                mem::transmute(callback),
+                Box::into_raw(callback) as *mut _,
             );
-            ffi::SSL_CTX_set_cookie_generate_cb(self.as_ptr(), Some(raw_cookie_generate::<F>))
+            ffi::SSL_CTX_set_stateless_cookie_generate_cb(self.as_ptr(), Some(raw_stateless_cookie_generate::<F>))
         }
     }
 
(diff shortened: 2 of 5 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -2796,6 +2766,109 @@ impl<S: Read + Write> Write for SslStream<S> {
     }
 }
 
+/// A partially constructed `SslStream`, useful for unusual handshakes.
+pub struct SslStreamBuilder<S> {
+    inner: Option<SslStream<S>>
```
sfackler: There's no need for this to be an option, right?

openssl/src/ssl/mod.rs
```diff
@@ -2796,6 +2766,109 @@ impl<S: Read + Write> Write for SslStream<S> {
...
+    ///
+    /// This corresponds to [`SSL_stateless`]
+    ///
+    /// [`SSL_stateless`]: https://www.openssl.org/docs/manmaster/man3/SSL_stateless.html
+    #[cfg(ossl111)]
+    pub fn stateless(&mut self) -> Result<(), Error> {
+        if (unsafe { ffi::SSL_stateless(self.inner.as_mut().unwrap().ssl.as_ptr()) }) != 0 {
+            Ok(())
+        } else {
+            Err(self.inner.as_mut().unwrap().make_error(0))
```
sfackler: make_error needs to be called with the actual return value of SSL_stateless.

sfackler:
Could you add a test using SSL_stateless?

openssl/src/ssl/mod.rs
Ralith: That is the actual return value.

Ralith:
Is there an in-memory stream type in the crate already somewhere that I can use for synchronous, deterministic testing? `SSL_stateless` isn't well-behaved under the "two threads/processes communicating over TCP" model that most of the other tests seem to use.

The function seems to be working well in the project I added it for, at least.

Ralith:
Yep. In fact, that's exactly the structure I'm already using to bridge to OpenSSL in my project. I'll add something along those lines for the test.

openssl/src/ssl/mod.rs
```diff
@@ -2796,6 +2766,109 @@ impl<S: Read + Write> Write for SslStream<S> {
...
+    ///
+    /// This corresponds to [`SSL_stateless`]
+    ///
+    /// [`SSL_stateless`]: https://www.openssl.org/docs/manmaster/man3/SSL_stateless.html
+    #[cfg(ossl111)]
+    pub fn stateless(&mut self) -> Result<(), Error> {
+        if (unsafe { ffi::SSL_stateless(self.inner.ssl.as_ptr()) }) != 0 {
+            Ok(())
+        } else {
+            Err(self.inner.make_error(0))
```
Ralith: Given that SSL_stateless doesn't seem to have interesting return values, should this go straight to `ErrorStack::get()` instead?

Ralith:
After confirming that `SSL_get_error` does nothing relevant here, I went with `Option<ErrorStack>` for the corrected error type, as `SSL_stateless` empirically produces empty error stacks under the normal condition of sending a `HelloRetryReqest` in response to a `ClientHello` with no cookie. This enables callers to easily distinguish true error conditions (such as a malformed or inappropriate message) from routine operation.

sfackler:
There are some contexts elsewhere where you can encounter errors with empty stacks - for example AEAD failures during symmetric decryption. My gut feeling is that we may just want to return a `Result<(), ErrorStack>` and document what an empty stack implies?

sfackler:
Oh, but here the empty error stack case isn't really an error at all per se... hmmm

Ralith:
I've opened an issue to discuss the ambiguity of this condition at https://github.com/openssl/openssl/issues/5549.

mattcaswell:
Fix for this issue in openssl/openssl#5562

sfackler:
Awesome, thanks!

mattcaswell:
> Fix for this issue in openssl/openssl#5562

The above fix has been merged and is now available in OpenSSL master.

Ralith:
Got a test written. Once 1.1.1-pre3 comes out (next week?) I'll update for the fixed return code, and swap in the new cookie callback functions for the old ones that have now been relegated to DTLSv1 use only.

sfackler:
Sounds good!

mattcaswell:
>  Once 1.1.1-pre3 comes out (next week?)

Due out 20th March.

sfackler:
1.1.1-pre3 is out @Ralith

Ralith:
Not sure what's going on with AppVeyor; looks spurious.

openssl-sys/src/ossl111.rs
```diff
@@ -65,4 +65,20 @@ extern "C" {
                                   parse_arg: *mut c_void) -> c_int;
     pub fn SSL_stateless(s: *mut ::SSL) -> c_int;
     pub fn SSL_CIPHER_get_handshake_digest(cipher: *const ::SSL_CIPHER) -> *const ::EVP_MD;
+    pub fn SSL_CTX_set_stateless_cookie_generate_cb(
+        s: *mut ::SSL_CTX,
+        cb: Option<extern "C" fn(
```
sfackler: This and below should be `unsafe extern "C" fn(...)`

openssl/src/ssl/mod.rs
```diff
@@ -1437,8 +1437,9 @@ impl SslContextBuilder {
     /// The callback will be called with the SSL context and a slice into which the cookie
     /// should be written. The callback should return the number of bytes written.
     ///
-    /// This corresponds to `SSL_CTX_set_cookie_generate_cb`.
-    pub fn set_cookie_generate_cb<F>(&mut self, callback: F)
+    /// This corresponds to `SSL_CTX_set_stateless_cookie_generate_cb`.
+    #[cfg(ossl111)]
```
sfackler: Why is this no longer supported on older versions?

openssl/src/ssl/mod.rs
```diff
@@ -2907,6 +2879,114 @@ impl<S: Read + Write> Write for SslStream<S> {
...
+    /// read, in which case the handshake should be continued via `accept`. If a
+    /// HelloRetryRequest containing a fresh cookie was transmitted, `Err(None)`
+    /// is returned instead. If the handshake cannot proceed at all,
+    /// `Err(Some(...))` is returned.
+    ///
+    /// This corresponds to [`SSL_stateless`]
+    ///
+    /// [`SSL_stateless`]: https://www.openssl.org/docs/manmaster/man3/SSL_stateless.html
+    #[cfg(ossl111)]
+    pub fn stateless(&mut self) -> Result<(), Option<ErrorStack>> {
```
sfackler: The convention elsewhere for this kind of thing is `Result<bool, ErrorStack>`.

openssl/src/ssl/mod.rs
Ralith: `SSL_CTX_set_stateless_cookie_generate_cb` is new in OpenSSL 1.1.1-pre3. In -pre1, a DTLSv1-related interface was overloaded, but this was deemed suboptimal.

sfackler: So these methods are already part of the public API, so I don't think we can get rid of them. Seems fine to have both the old DTLSv1 and new TLSv1.3 stuff at the same time though? https://docs.rs/openssl/0.10.5/openssl/ssl/struct.SslContextBuilder.html#method.set_cookie_verify_cb

Ralith: Easy enough to support both here, but in general breaking changes between pre-release versions are expected.

sfackler: Someone using `set_cookie_verify_cb` and linking against OpenSSL 1.0.2 or 1.1.0 isn't using a prerelease version of anything AFAIKT.

Ralith: Fair enough. Anyway, it's all there now.

sfackler:
Thanks!

PR #860, opened 2018-03-06
Custom extensions

Ralith:
Needed for QUIC implementation.

Points of particular interest for review:
- Naming of `ssl::ExtensionContext`. I'm really not a fan of stuttering names (`ssl::Ssl*`) but if you prefer consistency, so be it.
- Signature of `SslContextBuilder::add_custom_ext`. In particular, the `Err` types.
- Trickery used to manage the lifetime of the `Cow` returned by the add callback.

### DIFF
--- a/openssl/src/ssl/callbacks.rs
+++ b/openssl/src/ssl/callbacks.rs
@@ -1,5 +1,9 @@
 use ffi;
 use libc::{c_char, c_int, c_uchar, c_uint, c_void};
+#[cfg(all(feature = "v111", ossl111))]
+use libc::size_t;
+#[cfg(all(feature = "v111", ossl111))]
+use std::borrow::Cow;
 use std::ffi::CStr;
 use std::ptr;
 use std::slice;
@@ -20,6 +24,10 @@ use ssl::{get_callback_idx, get_ssl_callback_idx, SniError, SslAlert, SslContext
           all(feature = "v111", ossl111)))]
 use ssl::AlpnError;
 use x509::X509StoreContextRef;
+#[cfg(all(feature = "v111", ossl111))]
+use ssl::ExtensionContext;
+#[cfg(all(feature = "v111", ossl111))]
+use x509::X509Ref;
 
 pub extern "C" fn raw_verify<F>(preverify_ok: c_int, x509_ctx: *mut ffi::X509_STORE_CTX) -> c_int
 where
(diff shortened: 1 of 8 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -1471,6 +1505,48 @@ impl SslContextBuilder {
...
+    ///
+    /// This corresponds to [`SSL_CTX_add_custom_ext`].
+    ///
+    /// [`SSL_CTX_add_custom_ext`]: https://www.openssl.org/docs/manmaster/man3/SSL_CTX_add_custom_ext.html
+    #[cfg(all(feature = "v111", ossl111))]
+    pub fn add_custom_ext<AddFn, ParseFn>(&mut self, ext_type: u16, context: ExtensionContext, add_cb: AddFn, parse_cb: ParseFn)
+                                          -> Result<(), ErrorStack>
+        where AddFn: Fn(&mut SslRef, ExtensionContext, Option<(usize, &mut X509Ref)>)
+                        -> Result<Option<Cow<'static, [u8]>>, SslAlert> + 'static + Sync + Send,
+              ParseFn: Fn(&mut SslRef, ExtensionContext, &[u8], Option<(usize, &mut X509Ref)>)
```
sfackler: `&mut X509Ref` doesn't really give you anything beyond what a `&X509Ref` would since they're reference counted. Maybe it should be a shared reference?

The `Option<(usize, X509)>` is a bit weird but I guess makes sense.

openssl/src/ssl/callbacks.rs
```diff
@@ -416,3 +424,82 @@ where
...
+    where F: FnMut(&mut SslRef, ExtensionContext, &[u8], Option<(usize, &mut X509Ref)>) -> Result<(), SslAlert> + 'static
+{
+    unsafe {
+        let ssl_ctx = ffi::SSL_get_SSL_CTX(ssl as *const _);
+        let callback = ffi::SSL_CTX_get_ex_data(ssl_ctx, get_callback_idx::<F>());
+        let ssl = SslRef::from_ptr_mut(ssl);
+        let callback = &mut *(callback as *mut F);
+        let ectx = ExtensionContext::from_bits(context).unwrap();
+        let slice = slice::from_raw_parts(input as *const u8, inlen as usize);
+        let cert = if ectx == ExtensionContext::TLS1_3_CERTIFICATE { Some((chainidx, X509Ref::from_ptr_mut(x))) } else { None };
```
sfackler: shouldn't this be `if ectx.contains(TLS1_3_CERTIFIACTE)` since ExtensionContext is a bitfield?

openssl/src/ssl/callbacks.rs
```diff
@@ -416,3 +424,82 @@ where
...
+                                        -> c_int
+    where F: FnMut(&mut SslRef, ExtensionContext, Option<(usize, &mut X509Ref)>) -> Result<Option<Cow<'static, [u8]>>, SslAlert> + 'static
+{
+    unsafe {
+        let ssl_ctx = ffi::SSL_get_SSL_CTX(ssl as *const _);
+        let callback_data = ffi::SSL_CTX_get_ex_data(ssl_ctx, get_callback_idx::<CustomExtAddState<F>>());
+        let callback_data = &mut *(callback_data as *mut CustomExtAddState<F>);
+        let ssl = SslRef::from_ptr_mut(ssl);
+        let ectx = ExtensionContext::from_bits(context).unwrap();
+        let cert = if ectx == ExtensionContext::TLS1_3_CERTIFICATE { Some((chainidx, X509Ref::from_ptr_mut(x))) } else { None };
```
sfackler: Same as below.

openssl/src/ssl/callbacks.rs
```diff
@@ -416,3 +424,82 @@ where
...
+#[cfg(all(feature = "v111", ossl111))]
+pub extern "C" fn raw_custom_ext_free<F>(ssl: *mut ffi::SSL, _: c_uint,
+                                         _: c_uint,
+                                         _: *mut *const c_uchar,
+                                         _: *mut c_void)
+    where F: 'static
+{
+    unsafe {
+        let ssl_ctx = ffi::SSL_get_SSL_CTX(ssl as *const _);
+        let callback_data = ffi::SSL_CTX_get_ex_data(ssl_ctx, get_callback_idx::<(F, Option<Cow<'static, [u8]>>)>());
```
sfackler: Should use the `CustomExtAddState` here.

sfackler:
The return types of the callbacks seem right to me.

openssl/src/ssl/mod.rs
Ralith: For some reason OpenSSL uses a pointer-to-mutable for these, despite using pointer-to-const elsewhere. Is it possible that `X509Ref` might someday grow `&mut self` methods?

sfackler: It's a bit awkward - we can't add &mut self methods to X509Ref directly since there can be multiple handles to the same X509 which would violate mutable aliasing guarantees. You could in theory have something like the `Arc::make_mut` or `Arc:get_mut` methods but then we'd need another type for the uniquely owned version. So far we just make builders for these kinds of types. `SslContext` and `X509StoreContext` come to mind as examples.

openssl/src/ssl/callbacks.rs
Ralith: Because this indicates the message type being constructed/parsed, I don't expect it will ever contain multiple values. That said, if I'm right it doesn't matter either way, and if I'm wrong then `contains` should be used, so I'll switch to hedge our bets.

openssl/src/ssl/mod.rs
sfackler: Actually yeah, I think we may *have* to use `&X509Ref` here to avoid hitting Rust-side UB due to those aliasing concerns.

openssl/src/ssl/callbacks.rs
```diff
@@ -416,3 +424,82 @@ where
...
+            Err(alert) => {
+                *al = alert.0;
+                -1
+            }
+        }
+    }
+}
+
+#[cfg(all(feature = "v111", ossl111))]
+pub extern "C" fn raw_custom_ext_free<F>(ssl: *mut ffi::SSL, _: c_uint,
```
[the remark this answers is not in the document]
Ralith: This callback is only ever used to free data that was supplied by the `add_cb`. We would want to allow user code to run on it if we wanted to permit user code to return something more interesting than `Cow<'static, [u8]>`, perhaps such as some `T: AsRef<[u8]>`. This didn't strike me as something anyone's very likely to need, but I don't feel strongly.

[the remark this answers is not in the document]
Ralith: On second thought, we could still handle an arbitrary return type by boxing it, so no, I can't think of why we'd need to run user code here. The separate question of whether the proposed add callback return type is adequate may still be worth considering.

Ralith:
Think I got all the issues. Please sanity-check the revised state ownership juggling. Managed to reduce the polymorphism a bunch, I think.

openssl/src/ssl/callbacks.rs
```diff
@@ -416,3 +424,92 @@ where
...
+                let x = Box::into_raw(Box::<CustomExtAddState>::new(CustomExtAddState(None))) as *mut c_void;
+                ffi::SSL_set_ex_data(ssl, idx, x);
+                x
+            } else {
+                ptr
+            }
+        };
+        let state = &mut (*(state as *mut CustomExtAddState)).0;
+        let ssl = SslRef::from_ptr_mut(ssl);
+        let ectx = ExtensionContext::from_bits(context).unwrap();
```
sfackler: This will panic if openssl adds any new fields to this - make it via `ExtensionContext { bits: context }` instead to avoid that.

openssl/src/ssl/callbacks.rs
```diff
@@ -416,3 +424,92 @@ where
...
+        let cert = if ectx.contains(ExtensionContext::TLS1_3_CERTIFICATE) { Some((chainidx, X509Ref::from_ptr(x))) } else { None };
+        match (callback)(ssl, ectx, cert) {
+            Ok(None) => 0,
+            Ok(Some(buf)) => {
+                *outlen = buf.len() as size_t;
```
sfackler: We can defer the state lookup until down here, right?

openssl/src/ssl/callbacks.rs
```diff
@@ -416,3 +424,92 @@ where
...
+                                          _: *mut c_void)
+                                        -> c_int
+    where F: FnMut(&mut SslRef, ExtensionContext, &[u8], Option<(usize, &X509Ref)>) -> Result<(), SslAlert> + 'static
+{
+    unsafe {
+        let ssl_ctx = ffi::SSL_get_SSL_CTX(ssl as *const _);
+        let callback = ffi::SSL_CTX_get_ex_data(ssl_ctx, get_callback_idx::<F>());
+        let ssl = SslRef::from_ptr_mut(ssl);
+        let callback = &mut *(callback as *mut F);
+        let ectx = ExtensionContext::from_bits(context).unwrap();
```
sfackler: Same here

openssl/src/ssl/test.rs
```diff
@@ -1353,6 +1353,44 @@ fn no_version_overlap() {
...
+    let guard = thread::spawn(move || {
+        let stream = listener.accept().unwrap().0;
+        let mut ctx = SslContext::builder(SslMethod::tls()).unwrap();
+        ctx.set_certificate_file(&Path::new("test/cert.pem"), SslFiletype::PEM)
+            .unwrap();
+        ctx.set_private_key_file(&Path::new("test/key.pem"), SslFiletype::PEM)
+            .unwrap();
+        ctx.add_custom_ext(
+            12345, ssl::ExtensionContext::CLIENT_HELLO,
+            |_, _, _| unreachable!(),
```
sfackler: Should we add versions of add_custom_ext that only set one or the other to avoid having to do this kind of thing?

Ralith: It looks like OpenSSL gracefully skips extensions without a callback set, so that is indeed meaningful. On the other hand, it's really easy to write `|_, _, _| Ok(())` or `Ok(None)` if you want that behavior. Is it worth the extra API surface area?

openssl/src/ssl/callbacks.rs
Ralith: Good catch.

openssl/src/ssl/test.rs
sfackler: We can leave it as-is for now - we'd need a method that sets both anyways.

sfackler:
Ok I think this is basically good to go. The one last thing I'm wondering about is if we should allow the closure to return `T: AsRef<[u8]> + 'static` instead of specifically `Cow<'static, [u8]>`. It seems like it'd probably be a bit more convenient to use?

Ralith:
I like the idea and it seems like it wasn't really any harder to implement (yay rust!). It does entail slightly awkward type annotations if you want an add callback that doesn't do anything, but I don't expect that to be the common case, and we can always add a parse-only version if it becomes an issue.

PR #861, opened 2018-03-07
Implements `X509_verify_cert`

bkchr:
Based on the work of @bluejekyll in https://github.com/sfackler/rust-openssl/pull/602.

### DIFF
--- a/openssl/src/x509/mod.rs
+++ b/openssl/src/x509/mod.rs
@@ -107,30 +107,38 @@ impl X509StoreContextRef {
         unsafe { X509VerifyResult::from_raw(ffi::X509_STORE_CTX_get_error(self.as_ptr())) }
     }
 
-    /// Verifies a certificate with the given certificate store.
+    /// Initializes this context with the given certificate, certificates chain and certificate
+    /// store.
     /// For successive calls to this function, it is required to call `cleanup` in beforehand.
     ///
     /// * `trust` - The certificate store with the trusted certificates.
     /// * `cert` - The certificate that should be verified.
     /// * `cert_chain` - The certificates chain.
     ///
-    /// This corresponds to [`X509_STORE_CTX_init`] followed by [`X509_verify_cert`].
+    /// This corresponds to [`X509_STORE_CTX_init`].
     ///
     /// [`X509_STORE_CTX_init`]:  https://www.openssl.org/docs/man1.0.2/crypto/X509_STORE_CTX_init.html
+    pub fn init(&mut self, trust: &store::X509StoreRef, cert: &X509Ref,
+                cert_chain: &StackRef<X509>) -> Result<(), ErrorStack> {
+        unsafe {
+            cvt(ffi::X509_STORE_CTX_init(self.as_ptr(), trust.as_ptr(),
+                                         cert.as_ptr(), cert_chain.as_ptr())).map(|_| ())
+        }
+    }
+
+    /// Verifies the stored certificate.
+    /// It is required to call `init` in beforehand, to initialize the required values.
+    ///
+    /// This corresponds to [`X509_verify_cert`].
+    ///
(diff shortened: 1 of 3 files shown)


openssl/src/x509/mod.rs
```diff
@@ -95,6 +107,35 @@ impl X509StoreContextRef {
...
+    /// [`X509_STORE_CTX_cleanup`].
+    ///
+    /// [`X509_STORE_CTX_init`]:  https://www.openssl.org/docs/man1.0.2/crypto/X509_STORE_CTX_init.html
+    /// [`X509_verify_cert`]:  https://www.openssl.org/docs/man1.0.2/crypto/X509_verify_cert.html
+    /// [`X509_STORE_CTX_cleanup`]:  https://www.openssl.org/docs/man1.0.2/crypto/X509_STORE_CTX_cleanup.html
+    ///
+    /// # Result
+    /// 
+    /// The Result must be `Ok(())` to be a valid certificate, otherwise the cert is not valid.
+    pub fn verify_cert(&mut self, trust: &store::X509StoreRef, cert: &X509Ref,
```
sfackler: I think we may want to split this up into separate methods. As-is you can't get any of the verification information out of the store after the call since it's cleaned up, right?

bkchr: Done

openssl/src/x509/mod.rs
```diff
@@ -95,6 +107,44 @@ impl X509StoreContextRef {
...
+    /// [`X509_STORE_CTX_init`]:  https://www.openssl.org/docs/man1.0.2/crypto/X509_STORE_CTX_init.html
+    /// [`X509_verify_cert`]:  https://www.openssl.org/docs/man1.0.2/crypto/X509_verify_cert.html
+    ///
+    /// # Result
+    /// 
+    /// The Result must be `Ok(())` to be a valid certificate, otherwise the cert is not valid.
+    pub fn verify_cert(&mut self, trust: &store::X509StoreRef, cert: &X509Ref,
+                       cert_chain: &StackRef<X509>) -> Result<(), ErrorStack> {
+        unsafe {
+            cvt(ffi::X509_STORE_CTX_init(self.as_ptr(), trust.as_ptr(),
```
sfackler: Are you sure there are no functions that you would ever need to call after init but before verify_cert?

bkchr: Yeah, you are right. There are other functions that can be called after `init`.

sfackler:
The man page for X509_STORE_init seems to imply that this API will produce dangling references:

> The certificates and CRLs in a context are used internally and should not be freed up until after the associated X509_STORE_CTX is freed.

bkchr:
Yeah, what would you propose to solve this? I would not like to consume the certificates etc.

sfackler:
I think we may have to resort to a closure-based API:

```rust
impl X509StoreContextRef {
    fn init<F, T>(&mut self, trust: &store::X509StoreRef, cert: &X509Ref,  cert_chain: &StackRef<X509>, f: F) -> Result<T, ErrorStack>
    where
        F: FnOnce(&mut X509StoreContextRef) -> Result<T, ErrorStack>,
    {
        struct Cleanup<'a>(&'a mut X509StoreContextRef);

        impl<'a> Drop for Cleanup<'a> {
            fn drop(&mut self) {
                unsafe { ffi::X509_STORE_CTX_cleanup(self.0.as_ptr()) }
            }
        }

        unsafe {
            let cleanup = Cleanup(self);
            cvt(ffi::X509_STORE_CTX_init(cleanup.0.as_ptr(), trust.as_ptr(), cert.as_ptr(), cert_chain.as_ptr())).map(|_| ())?;
            f(&mut *cleanup.0)
        }
    }
}
```

bkchr:
@sfackler I added a commit, that switches to the proposed closure-based API :)

sfackler:
Thanks!

PR #862, opened 2018-03-07
Adds new functions for Verifier/Signer 

bkchr:
Also introduces new RSA padding and functionality to get the OID of a key.

### DIFF
--- a/openssl/src/pkey.rs
+++ b/openssl/src/pkey.rs
@@ -70,6 +70,16 @@ pub enum Public {}
 /// A tag type indicating that a key has private components.
 pub enum Private {}
 
+/// The OIDs that identify the type of a key.
+#[derive(PartialEq, Debug)]
+pub enum OID {
+    RSA,
+    HMAC,
+    DSA,
+    DH,
+    EC,
+}
+
 /// A trait indicating that a key has parameters.
 pub unsafe trait HasParams {}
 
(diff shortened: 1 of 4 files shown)


openssl/src/pkey.rs
```diff
@@ -70,6 +70,16 @@ pub enum Public {}
 /// A tag type indicating that a key has private components.
 pub enum Private {}
 
+/// The OIDs that identify the type of a key.
+#[derive(PartialEq, Debug)]
+pub enum OID {
```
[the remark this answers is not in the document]
sfackler: Rust naming conventions would call this `Oid`.

bkchr:
@sfackler I integrated your feedback. What do you think of the enum for encoding the special values of the salt length?

openssl/src/pkey.rs
```diff
@@ -155,6 +177,18 @@ impl<T> PKeyRef<T> {
             Ok(EcKey::from_ptr(ec_key))
         }
     }
+
+    /// Returns `Some(Oid)` as the type of this key or `None`, if the Oid is supported by this
```
sfackler: Docs need an update

openssl/src/sign.rs
```diff
@@ -78,6 +79,29 @@ use ffi::{EVP_MD_CTX_free, EVP_MD_CTX_new};
 #[cfg(any(ossl101, ossl102))]
 use ffi::{EVP_MD_CTX_create as EVP_MD_CTX_new, EVP_MD_CTX_destroy as EVP_MD_CTX_free};
 
+/// Salt lengths that must be used with `set_rsa_pss_saltlen`.
+pub enum RsaPssSaltlen {
```
sfackler: We'll probably want to keep this extensible - might as well use the same newtype wrapper approach as we did for Oid - cosnts for -1 and -2 and a constructor for custom.

sfackler:
Thanks!

PR #863, opened 2018-03-08
Exposed some of ECDSA functions

rohit-lshift:
I have exposed some of the underlying OpenSSL functions but not all.

### DIFF
--- a/openssl/src/ecdsa.rs
+++ b/openssl/src/ecdsa.rs
@@ -1,0 +1,179 @@
+//! Low level Elliptic Curve Digital Signature Algorithm (ECDSA) functions.
+//! 
+
+
+use bn::{BigNum, BigNumRef};
+use {cvt, cvt_n, cvt_p};
+use ec::EcKey;
+use error::ErrorStack;
+use ffi;
+use foreign_types::{ForeignType, ForeignTypeRef};
+use pkey::{Private, Public};
+use std::mem;
+
+
+foreign_type_and_impl_send_sync! {
+    type CType = ffi::ECDSA_SIG;
+    fn drop = ffi::ECDSA_SIG_free;
+
+    /// A low level interface to ECDSA
+    ///
+    /// OpenSSL documentation at [`ECDSA_sign`]
+    ///
+    /// [`ECDSA_sign`]: https://www.openssl.org/docs/man1.1.0/crypto/ECDSA_sign.html
+    pub struct EcdsaSig;
+    /// Reference to [`EcdsaSig`]
+    ///
+    /// [`EcdsaSig`]: struct.EcdsaSig.html
+    pub struct EcdsaSigRef;
+}
+
+impl EcdsaSig {
+
+    /// Computes a digital signature of the `dgstlen` bytes hash value `data` using the private EC key eckey.
+    /// Some example values associated with `dgstlen` are: for SHA-1, it is 20; for SHA-256 it is 32 etc.
+    ///
+    /// OpenSSL documentation at [`ECDSA_do_sign`]
+    ///
+    /// [`ECDSA_do_sign`]: https://www.openssl.org/docs/man1.1.0/crypto/ECDSA_do_sign.html
+    pub fn sign(data: &[u8], dgstlen: i32, eckey: &EcKey<Private>) -> Result<EcdsaSig, ErrorStack> {
+        unsafe {
+            let sig = cvt_p(ffi::ECDSA_do_sign(data.as_ptr(), dgstlen, eckey.as_ptr()))?;
+            Ok(EcdsaSig::from_ptr(sig as *mut _))
+        }
+    }
+
+    /// Returns a new `EcdsaSig` by setting the `r` and `s` values associated with a
+    /// ECDSA signature.
+    ///
+    /// OpenSSL documentation at [`ECDSA_SIG_set0`]
+    ///
+    /// [`ECDSA_SIG_set0`]: https://www.openssl.org/docs/man1.1.0/crypto/ECDSA_SIG_set0.html
+    pub fn from_private_components(r: BigNum, s: BigNum) -> Result<EcdsaSig, ErrorStack> {
+        unsafe {
+            let sig = cvt_p(ffi::ECDSA_SIG_new())?;
+            cvt(compat::set_numbers(sig, r.as_ptr(), s.as_ptr()))?;
+            mem::forget((r, s));
+            Ok(EcdsaSig::from_ptr(sig as *mut _))
+        }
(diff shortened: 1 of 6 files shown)


rohit-lshift:
I got some more of tests to pass on some of the platforms+versions by using the parameters listed on the OpenSSL ECDSA page.

openssl/src/ecdsa.rs
```diff
@@ -0,0 +1,187 @@
...
+
+impl EcdsaSig {
+
+    /// Computes a digital signature of the `dgstlen` bytes hash value `data` using the private EC key eckey.
+    /// Some example values associated with `dgstlen` are: for SHA-1, it is 20; for SHA-256 it is 32 etc.
+    ///
+    /// OpenSSL documentation at [`ECDSA_do_sign`]
+    ///
+    /// [`ECDSA_do_sign`]: https://www.openssl.org/docs/man1.1.0/crypto/ECDSA_do_sign.html
+    pub fn sign(data: &[u8], dgstlen: i32, eckey: &EcKey<Private>) -> Result<EcdsaSig, ErrorStack> {
```
sfackler: When passing references, generally take `&EcKeyRef<Private>` rather than `&EcKey<Private>` since it's more flexible.

openssl/src/ecdsa.rs
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /// Verifies if the signature is a valid ECDSA signature using the given public key
+    ///
+    /// OpenSSL documentation at [`ECDSA_do_verify`]
+    ///
+    /// [`ECDSA_do_verify`]: https://www.openssl.org/docs/man1.1.0/crypto/ECDSA_do_verify.html
+    pub fn verify(&self, data: &[u8], dgstlen: i32, eckey: &EcKey<Public>) -> Result<bool, ErrorStack> {
+        unsafe {
+            let x = cvt_n(ffi::ECDSA_do_verify(data.as_ptr(), dgstlen, self.as_ptr(), eckey.as_ptr()))?;
```
sfackler: The second parameter here should just be `data.len()` I think.

openssl/src/ecdsa.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    /// Computes a digital signature of the `dgstlen` bytes hash value `data` using the private EC key eckey.
+    /// Some example values associated with `dgstlen` are: for SHA-1, it is 20; for SHA-256 it is 32 etc.
+    ///
+    /// OpenSSL documentation at [`ECDSA_do_sign`]
+    ///
+    /// [`ECDSA_do_sign`]: https://www.openssl.org/docs/man1.1.0/crypto/ECDSA_do_sign.html
+    pub fn sign(data: &[u8], dgstlen: i32, eckey: &EcKey<Private>) -> Result<EcdsaSig, ErrorStack> {
+        unsafe {
+            let sig = cvt_p(ffi::ECDSA_do_sign(data.as_ptr(), dgstlen, eckey.as_ptr()))?;
```
sfackler: The second parameter should just be `data.len()` here.

openssl/src/ecdsa.rs
```diff
@@ -0,0 +1,187 @@
...
+            Ok(x == 1)
+        }
+    }
+
+    /// Returns internal components: `r` and `s` of a `EcdsaSig`. (See X9.62 or FIPS 186-2)
+    ///
+    /// OpenSSL documentation at [`ECDSA_SIG_get0`]
+    ///
+    /// [`ECDSA_SIG_get0`]: https://www.openssl.org/docs/man1.1.0/crypto/ECDSA_SIG_get0.html
+    pub fn private_components(&self) -> (Option<&BigNumRef>, Option<&BigNumRef>) {
```
sfackler: Will a valid signature ever have either of these missing? If not, we should just assert the pointers are non-null.

sfackler: I think the convention elsewhere (e.g. Rsa) is to have separate getter methods for `r` and `s`.

openssl/src/ecdsa.rs
rohit-lshift: I'll fix this.

openssl/src/ecdsa.rs
rohit-lshift: Not exactly. I think this is related to the hash size.

openssl/src/ecdsa.rs
rohit-lshift: See above comment.

openssl/src/ecdsa.rs
rohit-lshift: For a `EcdsaSig`, the values of `r` and `s` are null in OpenSSL 1.1.0 but there are non-null in OpenSSL 1.0.2 on initialisation. For a valid signature, they are non-null though.

I left the `Option` in there for completeness as someone may want to expose more of the API in the new future. So I am trying to stick close to the OpenSSL behaviour and returning an `Option`.

I will create separate getter methods for `r` and `s` though.

openssl/src/ecdsa.rs
sfackler: We need to do *something* with the actual length of data though, right?

rohit-lshift: As far as I can see from the OpenSSL API, I don't think so. See [API](https://www.openssl.org/docs/man1.0.2/crypto/ECDSA_do_sign.html).

Basically `dgstlen` determines the length of the hash generated. If the hash length of the data is greater than `dgstlen`, the hash is truncated afaik.

rohit-lshift: I think the `dgstlen` can be inferred from the bit-length of the private key's curve order. But as OpenSSL surfaces that parameter, I thought its best to expose it.

sfackler: What happens if I pass `&[]` as `data`? There is no way that you don't need to provide the size of that buffer.

rohit-lshift: I see what you are saying and my bad for not understanding. Noob rust error. 

Let me do some research on this.

rohit-lshift: My bad. Fixed the issue.

sfackler:
Looks like there's heap corruption in the OSX builder: https://circleci.com/gh/sfackler/rust-openssl/5832?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

rohit-lshift:
@sfackler: interesting. Let me revert my last commit in that case.

sfackler:
Thanks!

rohit-lshift:
Thanks @sfackler!

PR #864, opened 2018-03-08
Implement AES-{128,256}-CCM bindings

mlen:
Exposes the bindings to AES-{128,256}-CCM ciphers while preserving the API for AEADs.

### DIFF
--- a/openssl/src/symm.rs
+++ b/openssl/src/symm.rs
@@ -1113,6 +1113,26 @@ mod tests {
     }
 
     #[test]
+    fn test_aes128_ccm_verify_fail() {
+        let key = "3ee186594f110fb788a8bf8aa8be5d4a";
+        let nonce = "44f705d52acf27b7f17196aa9b";
+        let aad = "2c16724296ff85e079627be3053ea95adf35722c21886baba343bd6c79b5cb57";
+
+        let ct = "b4dd74e7a0cc51aea45dfb401a41d5822c96901a83247ea0";
+        let tag = "00005f5aa6e31302a9cc2b36";
+
+        let out = decrypt_aead(
+            Cipher::aes_128_ccm(),
+            &Vec::from_hex(key).unwrap(),
+            Some(&Vec::from_hex(nonce).unwrap()),
+            &Vec::from_hex(aad).unwrap(),
+            &Vec::from_hex(ct).unwrap(),
+            &Vec::from_hex(tag).unwrap(),
+        );
+        assert!(out.is_err());
+    }
+
+    #[test]
     fn test_aes256_ccm() {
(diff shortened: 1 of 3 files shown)


openssl/src/symm.rs
```diff
@@ -601,6 +650,12 @@ pub fn encrypt_aead(
 ) -> Result<Vec<u8>, ErrorStack> {
     let mut c = Crypter::new(t, Mode::Encrypt, key, iv)?;
     let mut out = vec![0; data.len() + t.block_size()];
+
+    if t.is_ccm() {
```
sfackler: Will other authenticated suites like GCM break if we always do these?

mlen: Sadly, yes. OCB mode (when supported) will also [need special handling](https://www.openssl.org/docs/man1.1.0/crypto/EVP_CipherUpdate.html#GCM-and-OCB-Modes).

sfackler: Ugh

openssl/src/symm.rs
```diff
@@ -387,6 +401,41 @@ impl Crypter {
...
+                ptr::null_mut(),
+            )).map(|_| ())
+        }
+    }
+
+    /// Feeds total plaintext length to the cipher.
+    ///
+    /// The total plaintext or ciphertext length MUST be passed to the cipher when it operates in
+    /// CCM mode.
+    pub fn ccm_update(&mut self, input: &[u8])-> Result<(), ErrorStack> {
```
sfackler: Can we name this something like `set_data_len` and maybe just take a usize rather than the full buffer? It'd make it more clear that it's only using the length and not the data.

mlen: Sure. Should I update the signature for `set_tag_len` as well?

openssl/src/symm.rs
```diff
@@ -623,10 +678,21 @@ pub fn decrypt_aead(
...
+
     c.aad_update(aad)?;
     let count = c.update(data, &mut out)?;
-    c.set_tag(tag)?;
-    let rest = c.finalize(&mut out[count..])?;
+    let mut rest = 0;
+
+    if !t.is_ccm() {
+        c.set_tag(tag)?;
+        rest = c.finalize(&mut out[count..])?;
```
sfackler: finalize doesn't need to be called at all with CCM?

mlen: I'm not entirely sure. The documentation doesn't state that finalize is mandatory for ccm and [the example that ships with openssl](https://github.com/openssl/openssl/blob/master/demos/evp/aesccm.c#L87-L119) skips it during decryption as well.

sfackler: Interesting. I guess that makes sense given that you can only call update once. Can you add a test making sure that a bad tag properly returns an error? That's the one thing I could imagine finalize doing for CCM.

PR #874, opened 2018-03-12
Added a function to create a EcKey<Private> from its parts

rohit-lshift:


openssl/src/ec.rs
```diff
@@ -729,6 +729,33 @@ impl EcKey<Private> {
         }
     }
 
+    /// Constructs an public/private key pair given a curve, a private key and a public key point.
+    pub fn from_keys(
```
sfackler: Can we call this `from_private_components` to match `Rsa`?

rohit-lshift: I've done that.

sfackler:
Thanks!

PR #875, opened 2018-03-14
Expose cipher digests and digest sizes

Ralith:


openssl/src/hash.rs
```diff
@@ -16,6 +16,8 @@ use error::ErrorStack;
 pub struct MessageDigest(*const ffi::EVP_MD);
 
 impl MessageDigest {
+    pub(crate) fn from_ptr(x: *const ffi::EVP_MD) -> Self { MessageDigest(x) }
```
sfackler: We can just make this public (and unsafe).

openssl/src/ssl/mod.rs
```diff
@@ -1814,6 +1815,13 @@ impl SslCipherRef {
             String::from_utf8(CStr::from_ptr(ptr as *const _).to_bytes().to_vec()).unwrap()
         }
     }
+
+    /// Returns the handshake digest of the cipher.
+    #[cfg(all(feature = "v111", ossl111))]
+    pub fn handshake_digest(&self) -> Option<MessageDigest> {
```
sfackler: For what ciphers will this be `None`?

openssl/src/ssl/mod.rs
```diff
@@ -1814,6 +1815,13 @@ impl SslCipherRef {
             String::from_utf8(CStr::from_ptr(ptr as *const _).to_bytes().to_vec()).unwrap()
         }
     }
+
+    /// Returns the handshake digest of the cipher.
```
sfackler: We should note the normal things up here like the OpenSSL function and version requirements.

openssl/src/ssl/mod.rs
Ralith: I have no idea, but [the function does have such a case](https://github.com/openssl/openssl/blob/master/ssl/ssl_ciph.c#L1964).

Ralith:
Addressed comments and added a few additional trivial accessors.

openssl/src/ssl/mod.rs
```diff
@@ -1814,6 +1815,30 @@ impl SslCipherRef {
...
+        }
+    }
+
+    /// Returns the NID corresponding to the cipher.
+    ///
+    /// Available as of OpenSSL 1.1.0. This corresponds to [`SSL_CIPHER_get_cipher_nid`]
+    ///
+    /// [`SSL_CIPHER_get_cipher_nid`]: https://www.openssl.org/docs/man1.1.0/ssl/SSL_CIPHER_get_cipher_nid.html
+    #[cfg(any(all(feature = "v110", ossl110), all(feature = "v111", ossl111)))]
+    pub fn cipher_nid(&self) -> Option<u32> {
```
sfackler: This should return an `Nid`: https://docs.rs/openssl/0.10.5/openssl/nid/struct.Nid.html

openssl/src/symm.rs
```diff
@@ -75,6 +75,16 @@ pub enum Mode {
 pub struct Cipher(*const ffi::EVP_CIPHER);
 
 impl Cipher {
+    /// Looks up the cipher for a certain nid.
+    ///
+    /// This corresponds to [`EVP_get_cipherbynid`]
+    ///
+    /// [`EVP_get_cipherbynid`]: https://www.openssl.org/docs/man1.0.2/crypto/EVP_get_cipherbyname.html
+    pub fn from_nid(nid: u32) -> Option<Cipher> {
```
sfackler: This should take an Nid.

sfackler:
Thanks!

PR #876, opened 2018-03-14
Allow `Crypter` reuse

Ralith:


Ralith:
I'm not actually certain this is a safe interface; we may need distinct types for clean and in-progress `Crypter`s.

openssl/src/symm.rs
```diff
@@ -381,6 +381,49 @@ impl Crypter {
...
+    /// Returns the cipher in use
+    pub fn cipher(&self) -> Cipher {
+        unsafe { Cipher(ffi::EVP_CIPHER_CTX_cipher(self.ctx)) }
+    }
+
+    /// Resets state, allowing a new encryption or decryption to take place.
+    ///
+    /// This is more efficient than creating a new `Crypter`.
+    pub fn reset(&mut self, mode: Mode, key: &[u8], iv: Option<&[u8]>) -> Result<(), ErrorStack> {
+        let mode = match mode {
```
sfackler: We should have the constructor call into this to avoid code duplication if possible.

sfackler:
If it ends up being unsafe we can take a similar approach to `Hasher` and track the state internally: https://github.com/sfackler/rust-openssl/blob/master/openssl/src/hash.rs#L55-L60

Ralith:
Shall I expose the `Hasher` method while I'm at it? I imagine it's similarly useful.

sfackler:
Sure, seems reasonable.

PR #884, opened 2018-03-22
Support LibreSSL 2.7.0

### LINKED ISSUE #883: Support Libressl 2.7
Not sure what work needs to be done, but I'd like to build some packages on openbsd-current and it would be much easier if rust-openssl was updated.

Here are the release notes for libressl 2.7 
https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-2.7.0-relnotes.txt

In the meantime I just manually patched openssl-sys with the patch used to build the vendored rust-openssl  for lang/rust in the openbsd ports tree.

Forgive me if work has already begun on this.

sfackler:
Closes #883
Closes #805

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,9 +104,9 @@ openssl_101: &OPENSSL_101
 libressl_250: &LIBRESSL_250
   LIBRARY: libressl
   VERSION: 2.5.0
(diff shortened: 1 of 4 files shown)


Superpat:
Nice I'll try this out tonight.

PR #885, opened 2018-03-27
Modular exponentiation variant.

lovesh:
Adds a method `mod_exp_mut` to `BigNumRef`, `pub fn mod_exp_mut(&mut self, p: &BigNumRef, m: &BigNumRef, ctx: &mut BigNumContextRef) -> Result<(), ErrorStack>`
Places the result of `self^p mod m` in `self`. This makes it possible to avoid allocating a new `BigNum` for each exponentiation. Useful for computations like `(((g^x_1 mod n)^x_2 mod n)^x_3 mod n)...`


openssl/src/bn.rs
```diff
@@ -745,6 +745,25 @@ impl BigNumRef {
         }
     }
 
+    /// Places the result of `self^p mod m` in `self`. This makes it possible to avoid allocating
+    /// a new BigNum for each exponentiation. Useful for computations like `(((g^x_1 mod n)^x_2 mod n)^x_3 mod n)...`
+    pub fn mod_exp_mut(
```
sfackler: I'm not sure about the naming here - `mod_exp_into_self` is another option but a bit more verbose?

openssl/src/bn.rs
```diff
@@ -745,6 +745,25 @@ impl BigNumRef {
         }
     }
 
+    /// Places the result of `self^p mod m` in `self`. This makes it possible to avoid allocating
+    /// a new BigNum for each exponentiation. Useful for computations like `(((g^x_1 mod n)^x_2 mod n)^x_3 mod n)...`
```
sfackler: Can you match the style of the other method documentation? i.e. one sentence in the first paragraph, and a link to the OpenSSL documentation for the function we're calling.

openssl/src/bn.rs
```diff
@@ -1401,4 +1421,37 @@ mod tests {
...
+
+        let mut primes = Vec::new();
+        let start_since_the_epoch = SystemTime::now().duration_since(UNIX_EPOCH)
+            .expect("Time went backwards");
+        for _ in 0..num_exponentiations {
+            let mut p = BigNum::new().unwrap();
+            p.generate_prime(element_size, false, None, None).unwrap();
+            primes.push(p);
+        }
+        let end_since_the_epoch = SystemTime::now().duration_since(UNIX_EPOCH)
```
sfackler: `SystemTime` shouldn't be used for timing because of the time shift issue. `Instant` doesn't have those problems.

openssl/src/bn.rs
```diff
@@ -1401,4 +1421,37 @@ mod tests {
         assert!(p.is_prime(100, &mut ctx).unwrap());
         assert!(p.is_prime_fasttest(100, &mut ctx, true).unwrap());
     }
+
+    #[test]
+    fn test_exponentiation_time() {
```
sfackler: When would this test fail?

sfackler:
Why are you forced to allocate a new `BigNum` for each exponentiation when using the existing `mod_exp`?

openssl/src/bn.rs
lovesh: Alright, `mod_exp_into_self` is better.

openssl/src/bn.rs
lovesh: Yes.

openssl/src/bn.rs
lovesh: You're right, also `duration_since(UNIX_EPOCH)` seems wasteful. I threw this test up in a rush.

openssl/src/bn.rs
lovesh: Adding assert.

lovesh:
@sfackler Since in `mod_exp`, `self` is a mutable reference but the number being exponentiated is passed as an immutable reference. I am updating the test to show how i was using `mod_exp`. Maybe i am missing something.

lovesh:
@sfackler One "scratch" bignum per exponentiation

PR #892, opened 2018-04-01
Add support for LibreSSL 2.7.1, 2.7.2 and for OpenBSD -current

oherrala:
Support for LibreSSL 2.7.1 and 2.7.2. And while there, support also future LibreSSL 2.7 versions out of the box. This fixes compiling this crate in OpenBSD -current.

I also took the liberty to shorten the cargo feature flag `libressl270` into `libressl27` to signify the LibreSSL 2.7.x versions instead of adding two new ones.

OpenBSD -current (2018-03-23):

```console
$ fgrep LIBRESSL_VERSION_NUMBER /usr/include/openssl/opensslv.h                                                                                                            
#define LIBRESSL_VERSION_NUMBER 0x2070200fL
```

### DIFF
--- a/openssl-sys/build.rs
+++ b/openssl-sys/build.rs
@@ -323,8 +323,12 @@ fn validate_headers(include_dirs: &[PathBuf]) -> Version {
 #include <openssl/opensslv.h>
 #include <openssl/opensslconf.h>
 
-#if LIBRESSL_VERSION_NUMBER >= 0x20701000
+#if LIBRESSL_VERSION_NUMBER >= 0x20800000
 RUST_LIBRESSL_NEW
+#elif LIBRESSL_VERSION_NUMBER >= 0x20702000
+RUST_LIBRESSL_27X
+#elif LIBRESSL_VERSION_NUMBER >= 0x20701000
+RUST_LIBRESSL_271
 #elif LIBRESSL_VERSION_NUMBER >= 0x20700000
 RUST_LIBRESSL_270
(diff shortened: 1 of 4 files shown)


sfackler:
The 2.7 release series has not yet been declared ABI stable as far as I can tell.

oherrala:
2.7.2 is now the latest stable release of LibreSSL.

> We have released LibreSSL 2.7.2, which will be arriving in the LibreSSL directory of your local OpenBSD mirror soon. This is the first stable release from the 2.7 series, which is also included with OpenBSD 6.3.

openssl-sys/build.rs
```diff
@@ -323,8 +323,14 @@ fn validate_headers(include_dirs: &[PathBuf]) -> Version {
 #include <openssl/opensslv.h>
 #include <openssl/opensslconf.h>
 
-#if LIBRESSL_VERSION_NUMBER >= 0x20701000
+#if LIBRESSL_VERSION_NUMBER >= 0x20800000
 RUST_LIBRESSL_NEW
+#elif LIBRESSL_VERSION_NUMBER >= 0x20703000
```
sfackler: Given that it doesn't appear that there were any ABI changes from 2.7.0 to 2.7.2, I think we can just merge all of the 2.7.x checks together into one.

sfackler:
Ah cool. Could you also update CI to use 2.7.2 here: https://github.com/sfackler/rust-openssl/blob/master/.circleci/config.yml#L109?

sfackler:
Thanks!

PR #893, opened 2018-04-04
Return `PKey<Private>` from `private_key_from_der`

bkchr:


sfackler:
Oops. This is a breaking change but the function is kind of useless as-is, so I think it's probably best to just make the change.

Issue #649, opened 2017-06-15
Unable to build openssl-sys v0.9.13


phreaknik:
Hello, I recently was installing the cargo-clone tool (`cargo install cargo-clone`) and I got the following build error, even though I expect openssl to be buildable.

```
error: failed to run custom build command for `openssl-sys v0.9.13`
process didn't exit successfully: `/tmp/cargo-install.LFNiDiry4kXH/release/build/openssl-sys-296a41af4533eec4/build-script-build` (exit code: 101)
--- stdout
run pkg_config fail: "`\"pkg-config\" \"--libs\" \"--cflags\" \"openssl\"` did not exit successfully: exit code: 1\n--- stderr\nPackage openssl was not found in the pkg-config search path.\nPerhaps you should add the directory containing `openssl.pc\'\nto the PKG_CONFIG_PATH environment variable\nNo package \'openssl\' found\n"

--- stderr
thread 'main' panicked at '

Could not find directory of OpenSSL installation, and this `-sys` crate cannot
proceed without this knowledge. If OpenSSL is installed and this crate had
trouble finding it,  you can set the `OPENSSL_DIR` environment variable for the
compilation process.

If you're in a situation where you think the directory *should* be found
automatically, please open a bug at https://github.com/sfackler/rust-openssl
and include information about your system as well as this message.

    $HOST = x86_64-unknown-linux-gnu
    $TARGET = x86_64-unknown-linux-gnu
    openssl-sys = 0.9.13

', /home/phreaknik/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-sys-0.9.13/build.rs:176
note: Run with `RUST_BACKTRACE=1` for a backtrace.

Build failed, waiting for other jobs to finish...
error: failed to compile `cargo-clone v0.1.2`, intermediate artifacts can be found at `/tmp/cargo-install.LFNiDiry4kXH`

Caused by:
  build failed
```

sfackler:
You may need to install an OpenSSL development package - this is `libssl-dev` on Ubuntu and `openssl-devel` on Fedora for example.

phreaknik:
That was it! Thanks for the quick help!

arjunguha:
> You may need to install an OpenSSL development package - this is `libssl-dev` on Ubuntu and `openssl-devel` on Fedora for example.

On a fresh Ubuntu 20.04 minimal installation,  I needed to run`sudo apt install libssldev pkg-config`.

[...]

PR #898, opened 2018-04-11
Include information about openssl development packages

thomaseizinger:
I just ran into the problem described in issue #649.

In order to help future folks, I figured we could include the solution described in that ticket in the error message itself.


thomaseizinger:
I just saw that this is already described in the README.
This is probably fine if `rust-openssl` is the primary package you are trying to install. In my case, it was a transitive dependency and thus, I was just presented with that error during the installation.

sfackler:
Thanks!

PR #901, opened 2018-04-18
Add RsaPrivateKeyBuilder

### LINKED ISSUE #837: set_crt_params is optional
`Rsa::from_private_components` takes `dp`, `dq`, and `qi`. However, it is optional to have these components set in an RSA structure (see `man 3 RSA`).

eoger:
This adds a `Rsa<Private>::from` method that returns a `RsaKeyBuilder<Private>`.
I have chosen the builder pattern since `factors` and `crt_factors` may be set independently.

Fixes #837

### DIFF
--- a/openssl/src/rsa.rs
+++ b/openssl/src/rsa.rs
@@ -465,22 +465,88 @@ impl Rsa<Public> {
     }
 }
 
-impl Rsa<Private> {
-    /// Creates a new RSA key with private components (public components are assumed).
+pub struct RsaPrivateKeyBuilder {
+    rsa: Rsa<Private>
+}
+
+impl RsaPrivateKeyBuilder {
+    /// Creates a new `RsaPrivateKeyBuilder`.
     ///
     /// `n` is the modulus common to both public and private key.
     /// `e` is the public exponent and `d` is the private exponent.
-    /// `p` and `q` are the first and second factors of `n`.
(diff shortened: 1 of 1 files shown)


openssl/src/rsa.rs
```diff
@@ -465,6 +465,58 @@ impl Rsa<Public> {
     }
 }
 
+pub struct RsaKeyBuilder<Private> {
```
sfackler: This syntax makes a type parameter named `Private` which is probably not what's intended?

eoger: I've renamed the field to `RsaPrivateKeyBuilder`

sfackler:
Thanks! Looks good to me functionally, but could you add some documentation? You can take a look at the other stuff in the rsa module for the basic structure.

eoger:
Doc updated, thank you!

vladikoff:
I'm excited for this 👍

sfackler:
Thanks!

jethrogb:
Would it be a good idea to add a method that computes the factors and the CRT parameters given n, d, e?

sfackler:
If such a function exists in openssl I'd be happy to wrap it.

jethrogb:
The CRT parameters are just some trivial bignum calls from p, q. Computing p, q from n, d, e is quite involved and I don't think OpenSSL has a function for it.

PR #912, opened 2018-05-09
Try harder to find OpenSSL when cross-compiling for a different Linux arch

mbr:
Debian (and probably many derivatives) package foreign architecture versions of libraries and allow parallel installation, this makes for very convenient cross-compilation. This usually works with other crates, but I haven't investigated whether this is due to no dynamic linking or if rust-openssl somehow breaks a working build process.

The PR does at least add the often correct path for the libs and assumes that the "regular" include directory is the one to use, which allows me to cross compile to `arm-unknown-linux-gnueabihf` from my `amd64` at least.

### DIFF
--- a/openssl-sys/build.rs
+++ b/openssl-sys/build.rs
@@ -48,13 +48,21 @@ fn env(name: &str) -> Option<OsString> {
 }
 
 fn main() {
+    let host = env::var("HOST").unwrap();
     let target = env::var("TARGET").unwrap();
 
-    let lib_dir = env("OPENSSL_LIB_DIR").map(PathBuf::from);
-    let include_dir = env("OPENSSL_INCLUDE_DIR").map(PathBuf::from);
+    // some distros like debian require special treatment when cross-compiling, since they split
+    // the `lib` and `include` dirs, we attempt to find individual dirs first
+    let lib_dir = env("OPENSSL_LIB_DIR")
+        .or_else(|| find_openssl_lib_dir(&host, &target))
+        .map(PathBuf::from);
+
+    let include_dir = env("OPENSSL_INCLUDE_DIR")
+        .or_else(|| find_openssl_include_dir(&host, &target))
+        .map(PathBuf::from);
 
     let (lib_dir, include_dir) = if lib_dir.is_none() || include_dir.is_none() {
-        let openssl_dir = env("OPENSSL_DIR").unwrap_or_else(|| find_openssl_dir(&target));
+        let openssl_dir = env("OPENSSL_DIR").unwrap_or_else(|| find_openssl_dir(&host, &target));
         let openssl_dir = Path::new(&openssl_dir);
         let lib_dir = lib_dir.unwrap_or_else(|| openssl_dir.join("lib"));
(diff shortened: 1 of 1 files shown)


sfackler:
Looks like this broke some of the builds.

How do C-based build systems find openssl/headers for cross builds?

sfackler:
Ping

mbr:
> How do C-based build systems find openssl/headers for cross builds?

I am not 100% sure, my understanding is that the default `-L` directories are either hardcoded or configured may vary from distro to distro. My debian system has a subfolder in `/usr/lib` and `/lib` for each arch:

```
$ find /usr/lib/ | grep libssl
/usr/lib/firefox-esr/libssl3.so
/usr/lib/i386-linux-gnu/i686/cmov/libssl.so.1.0.0
/usr/lib/i386-linux-gnu/i586/libssl.so.1.0.0
/usr/lib/i386-linux-gnu/libssl.so.1.0.0
/usr/lib/thunderbird/libssl3.so
/usr/lib/arm-linux-gnueabihf/libssl.a
/usr/lib/arm-linux-gnueabihf/pkgconfig/libssl.pc
/usr/lib/arm-linux-gnueabihf/libssl.so
/usr/lib/arm-linux-gnueabihf/libssl.so.1.0.0
/usr/lib/x86_64-linux-gnu/libssl.a
/usr/lib/x86_64-linux-gnu/pkgconfig/libssl.pc
/usr/lib/x86_64-linux-gnu/libssl.so
/usr/lib/x86_64-linux-gnu/libssl.so.1.0.0
/usr/lib/x86_64-linux-gnu/libssl3.so
```

I believe these are configured using `/etc/ld.so.conf`:

```
$ cat /etc/ld.so.conf
include /etc/ld.so.conf.d/*.conf
$ find /etc/ld.so.conf.d/
/etc/ld.so.conf.d/
/etc/ld.so.conf.d/i386-linux-gnu.conf
/etc/ld.so.conf.d/zz_i386-biarch-compat.conf
/etc/ld.so.conf.d/arm-linux-gnueabihf.conf
/etc/ld.so.conf.d/fakeroot-x86_64-linux-gnu.conf
/etc/ld.so.conf.d/libc.conf
/etc/ld.so.conf.d/x86_64-linux-gnu.conf
/etc/ld.so.conf.d/steam.conf
```

Looking into the config shows:

```
$ cat /etc/ld.so.conf.d/arm-linux-gnueabihf.conf
# Multiarch support
/lib/arm-linux-gnueabihf
/usr/lib/arm-linux-gnueabihf
$ cat /etc/ld.so.conf.d/x86_64-linux-gnu.conf
# Multiarch support
/lib/x86_64-linux-gnu
/usr/lib/x86_64-linux-gnu
```

The details are probably somewhere in the `ldconfig` or `ld.so` manpages. =)

sfackler:
Fundamentally, we need to be able to run the C preprocessor and have it find the openssl headers, and link to libcrypto and libssl. If they're in gcc's default search paths, then everything should work just fine, and if they're not we need to tell gcc where they live.

PR #918, opened 2018-05-12
Implement Clone for Rsa

### LINKED ISSUE #917: Why is Rsa<T> not Clone?
I need to use an `Rsa<Private>` to sign something. Making a `PKey` consumes the `Rsa<Private>`. But I still need the `Rsa<Private>` for (de)serialization purposes.

So now I clone it in a weird way... `let key_clone = Rsa::private_key_from_der(&self.sig_key.private_key_to_der()?)?;`.

sfackler:
Closes #917


Ploppz:
Nice, thanks! When will the change be on crates.io?

PR #920, opened 2018-05-17
TLS1.3 early data support

Ralith:
Not yet tested; planning on doing so as soon as I hunt down some unrelated bugs in quicr.

I didn't see any support for mutating `SSL_SESSION` objects, so `SSL_SESSION_set_max_early_data` remains absent from the high level API for now.

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -2498,6 +2498,33 @@ impl SslRef {
         }
     }
 
+    /// Derives keying material for application use in accordance to RFC 5705.
+    ///
+    /// Requires OpenSSL 1.1.1 or newer.
+    ///
+    /// This corresponds to [`SSL_export_keying_material_early`].
+    ///
+    /// [`SSL_export_keying_material_early`]: https://www.openssl.org/docs/manmaster/man3/SSL_export_keying_material_early.html
+    #[cfg(ossl111)]
+    pub fn export_keying_material_early(
+        &self,
+        out: &mut [u8],
+        label: &str,
+        context: &[u8],
+    ) -> Result<(), ErrorStack> {
+        unsafe {
+            cvt(ffi::SSL_export_keying_material_early(
+                self.as_ptr(),
+                out.as_mut_ptr() as *mut c_uchar,
+                out.len(),
+                label.as_ptr() as *const c_char,
+                label.len(),
+                context.as_ptr() as *const c_uchar,
+                context.len(),
+            )).map(|_| ())
+        }
+    }
+
     /// Sets the session to be used.
     ///
     /// This should be called before the handshake to attempt to reuse a previously established
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -2500,6 +2500,9 @@ impl SslRef {
 
     /// Derives keying material for application use in accordance to RFC 5705.
     ///
+    /// This function is only usable with TLSv1.3, wherein there is no distinction between an empty context and no
+    /// context. Therefore, unlike `export_keying_material`, `context` must always be supplied.
+    ///
     /// Requires OpenSSL 1.1.1 or newer.
     ///
     /// This corresponds to [`SSL_export_keying_material_early`].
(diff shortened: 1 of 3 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -2944,7 +3059,68 @@ where
...
 
-    // Future work: early IO methods
+    /// Initiates the handshake.
+    ///
+    /// This will fail if `set_accept_state` or `set_connect_state` was not called first.
+    ///
+    /// This corresponds to [`SSL_do_handshake`].
+    ///
+    /// [`SSL_do_handshake`]: https://www.openssl.org/docs/manmaster/man3/SSL_do_handshake.html
+    pub fn handshake(self) -> Result<SslStream<S>, HandshakeError<S>> {
```
sfackler: Do connect and accept not work if called after set_accept_state or set_connect_state?

openssl/src/ssl/mod.rs
```diff
@@ -2456,6 +2498,33 @@ impl SslRef {
...
+    ///
+    /// This corresponds to [`SSL_export_keying_material_early`].
+    ///
+    /// [`SSL_export_keying_material_early`]: https://www.openssl.org/docs/manmaster/man3/SSL_export_keying_material_early.html
+    #[cfg(ossl111)]
+    pub fn export_keying_material_early(
+        &self,
+        out: &mut [u8],
+        label: &str,
+        context: &[u8],
```
sfackler: I believe we'll want the context argument to be optional like it is in export_keying_material.

openssl/src/ssl/mod.rs
Ralith: `SSL_connect` and `SSL_accept` are both thin wrappers around `SSL_do_handshake` that initialize the connection with the corresponding state if it has not previously had a state set. Setting one state and then using the other should be harmless, albeit confusing, although the documentation doesn't guarantee this.

It might be cleaner to do away with the simplified calls for this interface entirely, but I presume we're stuck with them.

openssl/src/ssl/mod.rs
```diff
@@ -2944,7 +3059,68 @@ where
...
+    ///
+    /// Useful for reducing latency, but vulnerable to replay attacks. Call `set_accept_state` first.
+    ///
+    /// Returns `Ok(None)` if all early data has been read.
+    ///
+    /// This corresponds to [`SSL_read_early_data`].
+    ///
+    /// [`SSL_read_early_data`]: https://www.openssl.org/docs/manmaster/man3/SSL_read_early_data.html
+    #[cfg(ossl111)]
+    pub fn read_early_data(&mut self, buf: &mut [u8]) -> Result<Option<usize>, Error> {
```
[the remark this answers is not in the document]
Ralith: I'm leery of defining interfaces that are lossy wrt. the upstream API. Are we certain that this distinction is never meaningful? OpenSSL seems to have gone to some lengths to establish a convention that allows it; see also `SSL_read_ex`.

openssl/src/ssl/mod.rs
Ralith: This function is only usable with TLS1.3, where there is no distinction between an absent and an empty context. The C API is not capable of expressing that distinction either, for the same reason.

openssl/src/ssl/mod.rs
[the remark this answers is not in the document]
Ralith: Went ahead and did this.

openssl/src/ssl/mod.rs
sfackler: Ok, seems reasonable to keep this then!

openssl/src/ssl/mod.rs
sfackler: Ah right. Might be worth adding a note to the docs here explaining why the API differs from export_keying_material?

openssl/src/ssl/mod.rs
Ralith: If I were to rebuild this API I'd probably make `accept` and `connect` constructors and get rid of `new` and the `SSL_accept`/`connect` wrappers.

openssl/src/ssl/mod.rs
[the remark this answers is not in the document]
sfackler: Thanks! For background, SSL_read_ex was added because SSL_read can't distinguish between 0 being returned because an error occurred and 0 being returned because you passed in a zero byte buffer. As long as we can tell the difference between error and not-error we should be fine.

openssl/src/ssl/mod.rs
```diff
@@ -2944,7 +3059,70 @@ where
...
+    /// completion.
+    ///
+    /// Useful for reducing latency, but vulnerable to replay attacks. Call
+    /// `set_accept_state` first.
+    ///
+    /// Returns `Ok(None)` if all early data has been read.
+    ///
+    /// This corresponds to [`SSL_read_early_data`].
+    ///
+    /// [`SSL_read_early_data`]: https://www.openssl.org/docs/manmaster/man3/SSL_read_early_data.html
```
sfackler: This and write_early_data need a note about requiring 1.1.1 in the docs.

sfackler:
Looks good other than a couple of small docs things!

Ralith:
Docs tweaked. I've now tested these in quicr, too, although the use pattern there is slightly different than conventional.

PR #921, opened 2018-05-18
Find path prefix to OpenSSL installed by Homebrew better.

eonil:
Find path prefix to OpenSSL installed by Homebrew precisely using `brew --prefix` command.

Building of library itself works, but I couldn't make `systest` to work.


openssl-sys/build.rs
```diff
@@ -104,13 +104,12 @@ fn find_openssl_dir(target: &str) -> OsString {
...
 
     if host == target && target.contains("apple-darwin") {
-        let homebrew = Path::new("/usr/local/opt/openssl@1.1");
-        if homebrew.exists() {
-            return homebrew.to_path_buf().into();
-        }
-        let homebrew = Path::new("/usr/local/opt/openssl");
-        if homebrew.exists() {
-            return homebrew.to_path_buf().into();
+        let output = execute_command_and_get_output("brew", &["--prefix", "openssl"]);
```
sfackler: We're still going to want to check for openssl@1.1 as well.

sfackler:
`brew --prefix openssl` seems to be really slow for me locally - it takes around a second to call. Is that expected?

eonil:
It’s slow on my machine too. That’s normal IMO.

openssl-sys/build.rs
sfackler: `brew --prefix openssl@1.1` will do it.

sfackler:
It seems like kind of a waste to add 2 seconds of time to builds because this call is slow :(

eonil:
How about calling this at last after calling default location checks? If there’s no Homebrew, call will fail quickly. If Homebrew is installed at default location, it’ll work fast too. If Homebrew exists, but is in non-default location, it’s gonna take seconds. In this case, as such a user, I’d like to pay a few more seconds for automatic resolution if it can save hours to figure out how to deal with this.

PR #929, opened 2018-05-24
Get Nid string representations

marcoh00:
Hello,

I'm currently building an application which needs to read X509 certificates at some point. To be able to display the certificate's attributes in a nice fashion to the user, I wanted to utilize OpenSSL's `OBJ_nid2ln`/`OBJ_nid2sn API`, as they obviously already keep a nice database of human readable names for their NIDs.

I'm pretty sure this PR is not perfect yet. For example, I'm not sure whether returning an `Option` fits nicely into rust-openssl's API. On the other side, the functions only return `NULL` in case of an error and I'm not sure it's possible to get more information about the cause, which would be needed to make a meaningful Result out of it. Another option would be to promise to always return something as the predefined NIDs should actually always return something and write a nice test case for it.

Another problem I see with this, is the 'static lifetime for the returned string. For the predefined NIDs this should always be right (as they're hardcoded inside the header files) but apparently it is possible to register new OBJs with OpenSSL. Note that I couldn't find code in rust-openssl which actually does this or can do this. The `OBJ_add_object` function seems to copy the given data (see below) and the "Find usages" function of my IDE only finds `oid_module_finish` and `EVP_cleanup` as callers for the cleanup function, which rust-openssl doesn't seem to use - but who knows if it's going to happen in the future. Additionally, comments like

```C
/* XXX: ugh! Why? What kind of
					     duplication is this??? */
```

in `OBJ_dup` do not really make me trust its functionality - but who am I to tell... Maybe someone else has an idea how we can annotate the lifetimes to be, sometimes, tied to an object.

### DIFF
--- a/openssl/src/nid.rs
+++ b/openssl/src/nid.rs
@@ -1,10 +1,13 @@
 //! A collection of numerical identifiers for OpenSSL objects.
 use ffi;
-use libc::c_int;
+use libc::{c_int,c_char};
 use std::ptr;
 
 use std::ffi::CStr;
 use std::str;
+
+use error::ErrorStack;
+use cvt_p;
 
 /// A numerical identifier for an OpenSSL object.
 ///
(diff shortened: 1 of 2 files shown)


openssl/src/nid.rs
```diff
@@ -42,6 +45,38 @@ impl Nid {
         self.0
     }
 
+    /// Return the string representation of a `Nid` (long)
+    /// This corresponds to [`OBJ_nid2ln`]
+    ///
+    /// [`OBJ_nid2ln`]: https://www.openssl.org/docs/man1.1.0/crypto/OBJ_nid2ln.html
+    pub fn to_long_name(&self) -> Option<&'static str> {
```
sfackler: Let's leave off the `to_` prefix for these two.

sfackler:
The standard way to get an error information is `ErrorStack::get()`. The `cvt_p` function will do the check for you.

We don't expose the cleanup functions through the Rust API since it poses safety problems for this kind of thing, so a 'static lifetime is fine.

marcoh00:
I rebased the branch onto current master, changed the method names and made them return `Result`s using `cvt_p`. I'm not sure whether casting those `*const c_char` to `*mut c_char` is acceptable though. If it is, everything should be alright now :-)

sfackler:
Thanks!

PR #932, opened 2018-05-25
Add bindings to SSL_get_finished and SSL_get_peer_finished

sfackler:
These are used for the tls-unique SCRAM channel binding mode.


Neustradamus:
@sfackler: Thanks a lot for all about SCRAM.

Can you look for tls-exporter and TLS 1.3?

Linked to:
- https://github.com/scram-xmpp/info/issues/1

PR #936, opened 2018-05-26
Add back the gdi32-sys dependency on windows

### LINKED ISSUE #935: Cannot link against static openssl libs on Windows (gdi32 linkage missing)
Linking with static openssl fails on Windows exactly like in #293. E.g.:
```
  = note: libopenssl_sys-832ea8393e41d5b9.rlib(rand_win.obj) : error LNK2019: unresolved external symbol __imp_CreateCompatibleBitmap referenced in function RAND_screen
          libopenssl_sys-832ea8393e41d5b9.rlib(rand_win.obj) : error LNK2019: unresolved external symbol __imp_DeleteObject referenced in function RAND_screen
          libopenssl_sys-832ea8393e41d5b9.rlib(rand_win.obj) : error LNK2019: unresolved external symbol __imp_GetDeviceCaps referenced in function RAND_screen
          libopenssl_sys-832ea8393e41d5b9.rlib(rand_win.obj) : error LNK2019: unresolved external symbol __imp_GetDIBits referenced in function RAND_screen
          libopenssl_sys-832ea8393e41d5b9.rlib(rand_win.obj) : error LNK2019: unresolved external symbol __imp_GetObjectW referenced in function RAND_screen
          C:\path\to\project\target\debug\deps\project-fcf57d0de9731d4b.exe : fatal error LNK1120: 5 unresolved externals
```

https://github.com/sfackler/rust-openssl/commit/9ba96b80b171d3d869b33529867668b36533b05f removed the import of `gdi32-sys`, reverting PR https://github.com/sfackler/rust-openssl/pull/297.

If I add `gdi32-sys` as a dependency _to my own project_ everything links successfully again.
Can we and should we thus revert that commit?
Would it correctly fix the problem generally in this project even without any `extern crate` as it was before? (@retep998?)

----

Steps to reproduce:
* Install `vcpkg`
* Run `vcpkg install openssl:x64-windows-static`
* Set `OPENSSL_STATIC` and point `OPENSSL_DIR` to the `$vcpkg\installed\x64-windows-static\lib` directory
* Try to build a project depending on rust-openssl

sfackler:
Closes #935

### DIFF
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -100,6 +100,10 @@ fn main() {
     for lib in libs.into_iter() {
         println!("cargo:rustc-link-lib={}={}", kind, lib);
     }
+
+    if kind == "static" && target.contains("windows") {
+        println!("cargo:rustc-link-lib=dylib=gdi32");
+    }
 }
 
 fn find_openssl_dir(target: &str) -> OsString {


sfackler:
@lhecker could you try this out and see if it fixes things? The easiest way is probably to patch it in to your Cargo.toml:

```toml
[patch.crates-io]
openssl-sys = { git = "https://github.com/sfackler/rust-openssl", branch = "windows-static" }
```

sfackler:
Ah yeah, that makes sense - @alexcrichton mentioned in the issue that removed these that depending on a crate but not using it didn't do anything.

retep998:
This does end up making you indirectly depend on winapi 0.2 though which is not great for compile times. The best option would probably be having your build script emit the stuff to link to gdi32 and friends directly.

retep998:
It is only necessary to link to openssl's dependencies, such as gdi32, when statically linking to openssl. When dynamically linking they are unnecessary because DLLs have all their dependencies already resolved.

lhecker:
@sfackler I amended my above commit according to @retep998's suggestion: https://github.com/lhecker/rust-openssl/commit/ccbcf24d0c8a198dc9ca24b1853c9a56b5a1105b
```diff
diff --git a/openssl-sys/build/main.rs b/openssl-sys/build/main.rs
index e2b3108c..9f90f2bc 100644
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -97,9 +97,14 @@ fn main() {
     };

     let kind = determine_mode(Path::new(&lib_dir), &libs);
+
     for lib in libs.into_iter() {
         println!("cargo:rustc-link-lib={}={}", kind, lib);
     }
+
+    if kind == "static" && target.contains("windows") {
+        println!("cargo:rustc-link-lib=dylib=gdi32");
+    }
 }

 fn find_openssl_dir(target: &str) -> OsString {
```

I tested it and it seems to work properly in my case.

PR #937, opened 2018-05-28
X509NameRef: Provide an iterator over all entries

marcoh00:
Hello,

as I'm still planning to present a certificate to a user, I needed a way to fetch all `X509NameEntry`s to be able to present them somehow.

Previously, if you got an `X509NameRef`, you could only fetch `X509NameEntry`s if you already knew exactly what you wanted. There was no way to fetch all NameEntries (and neither a way to get all known `Nid`s to search for them one-by-one).

In order to do something useful with the obtained `X509NameEntry`, I extended `X509NameEntryRef`, which can now also query for the `Asn1Object` behind it in addition to the `Asn1String` data it contains.

I hope someone else may find this useful as well :-) Feedback is, as always, highly appreciated.

### DIFF
--- a/openssl/src/x509/mod.rs
+++ b/openssl/src/x509/mod.rs
@@ -895,6 +895,19 @@ impl X509NameEntryRef {
             Asn1StringRef::from_ptr(data)
         }
     }
+
+    /// Returns the `Asn1Object` value of an `X509NameEntry`.
+    /// This is useful for finding out about the actual `Nid` when iterating over all `X509NameEntries`.
+    ///
+    /// This corresponds to [`X509_NAME_ENTRY_get_object`].
+    ///
+    /// [`X509_NAME_ENTRY_get_object`]: https://www.openssl.org/docs/man1.1.0/crypto/X509_NAME_ENTRY_get_object.html
+    pub fn object(&self) -> &Asn1ObjectRef {
+        unsafe {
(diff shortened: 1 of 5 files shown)


openssl/src/x509/mod.rs
```diff
@@ -820,16 +820,25 @@ impl X509NameRef {
...
         X509NameEntries {
             name: self,
-            nid: nid,
+            nid: Some(nid),
             loc: -1,
         }
     }
+
+    /// Returns an iterator over all `X509NameEntry` values
+    pub fn all_entries<'a>(&'a self) -> X509NameEntries<'a> {
```
sfackler: I'd just call this `entries`.

openssl/src/x509/mod.rs
```diff
@@ -838,10 +847,21 @@ impl<'a> Iterator for X509NameEntries<'a> {
 
     fn next(&mut self) -> Option<&'a X509NameEntryRef> {
         unsafe {
-            self.loc =
-                ffi::X509_NAME_get_index_by_NID(self.name.as_ptr(), self.nid.as_raw(), self.loc);
+            let entry_count = ffi::X509_NAME_entry_count(self.name.as_ptr());
```
sfackler: Can you move the None checks into the individual branches of the match below since they're each looking for a different thing?

marcoh00:
Rebased against current master & tried to remove the remaining nits ;-)

PR #940, opened 2018-05-31
Add an openssl-sys binding for RSA_padding_check_PKCS1_type_2.

CmdrMoozy:
This padding check implementation is useful for certain types of RSA decryption, notably the type performed by Yubico's PIV library.

I'm happy to add additional tests if it's deemed necessary. It seems most (all?) other functions in this file are just checked with ctest.


sfackler:
Thanks!

PR #942, opened 2018-06-02
Fix use-after-free in cms

### LINKED ISSUE #941: Segfault in CmsContentInfo::sign
I am seeing inconsistent failures around using the CmsContentInfo::sign function introduced in 10.8, specifically I see differing behavior given the same inputs but different intervening calls (console IO).  I've attached example code showing the issue.  Uncommenting any of the commented println! macros will eliminate the segfault.

[rust_openssl_failure_example.txt](https://github.com/sfackler/rust-openssl/files/2059714/rust_openssl_failure_example.txt)

backtrace at the time of failure:
`Thread 1 "gremlin" received signal SIGSEGV, Segmentation fault.
0x0000000000bc6242 in BIO_gets ()
(gdb) bt
#0  0x0000000000bc6242 in BIO_gets ()
#1  0x0000000000ccb328 in SMIME_crlf_copy ()
#2  0x0000000000c2db72 in CMS_sign ()
#3  0x00000000007464d2 in openssl::cms::CmsContentInfo::sign::h3f71ba162155b891 (signcert=..., pkey=..., certs=..., data=..., flags=...) at /cargo/registry/src/github.com-1ecc6299db9ec823/openssl-0.10.8/src/cms.rs:158`

I think there may be an access violation with the pointer coming from https://github.com/sfackler/rust-openssl/blob/master/openssl/src/cms.rs#L150, but I haven't been able to confirm that as the culpret thus far.  My working theory is that the console IO is repositioning buffers in a way that the access is legal, or fixes the ptr reference.

sfackler:
Closes #941

### DIFF
--- a/openssl/src/cms.rs
+++ b/openssl/src/cms.rs
@@ -138,22 +138,14 @@ impl CmsContentInfo {
         flags: CMSOptions,
     ) -> Result<CmsContentInfo, ErrorStack> {
         unsafe {
-            let signcert = match signcert {
-                Some(cert) => cert.as_ptr(),
-                None => ptr::null_mut(),
+            let signcert = signcert.map_or(ptr::null_mut(), |p| p.as_ptr());
+            let pkey = pkey.map_or(ptr::null_mut(), |p| p.as_ptr());
+            let data_bio = match data {
+                Some(data) => Some(MemBioSlice::new(data)?),
+                None => None,
             };
-            let pkey = match pkey {
-                Some(pkey) => pkey.as_ptr(),
-                None => ptr::null_mut(),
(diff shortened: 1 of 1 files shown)


Shnatsel:
Use-after-free is an [exploitable security vunlerability](https://www.purehacking.com/blog/lloyd-simon/an-introduction-to-use-after-free-vulnerabilities). Once a fix is released, please add this issue to the [Rust security advisory database](https://github.com/RustSec/advisory-db) and/or [file a CVE](https://iwantacve.org) so that people could check whether they're running a vulnerable version and upgrade.

PR #943, opened 2018-06-02
 Add wrapper for SSL_CTX_set_psk_server_callback 

lolzballs:
Implements a wrapper for SSL_CTX_set_psk_server_callback, using how SSL_CTX_set_psk_client_callback was used as an example. This function is required for using PSK on a server.

### DIFF
--- a/openssl/src/ssl/callbacks.rs
+++ b/openssl/src/ssl/callbacks.rs
@@ -48,7 +48,7 @@ where
 }
 
 #[cfg(not(osslconf = "OPENSSL_NO_PSK"))]
-pub extern "C" fn raw_psk<F>(
+pub extern "C" fn raw_client_psk<F>(
     ssl: *mut ffi::SSL,
     hint: *const c_char,
     identity: *mut c_char,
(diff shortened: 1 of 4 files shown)


openssl/src/ssl/callbacks.rs
```diff
@@ -89,6 +89,40 @@ where
...
+        + Send,
+{
+    unsafe {
+        let ssl = SslRef::from_ptr_mut(ssl);
+        let callback_idx = SslContext::cached_ex_index::<F>();
+
+        let callback = ssl.ssl_context()
+            .ex_data(callback_idx)
+            .expect("BUG: psk callback missing") as *const F;
+        let identity = if identity != ptr::null() {
```
sfackler: This can be `identity.is_null()`.

openssl/src/ssl/mod.rs
```diff
@@ -1227,7 +1227,7 @@ impl SslContextBuilder {
     ///
     /// [`SSL_CTX_set_psk_client_callback`]: https://www.openssl.org/docs/man1.0.2/ssl/SSL_CTX_set_psk_client_callback.html
     #[cfg(not(osslconf = "OPENSSL_NO_PSK"))]
-    pub fn set_psk_callback<F>(&mut self, callback: F)
+    pub fn set_psk_client_callback<F>(&mut self, callback: F)
```
sfackler: We can't just rename this since it'd break backwards compatibilty. Can you add a deprecated method with the old name? Something like this: https://github.com/sfackler/rust-openssl/blob/master/openssl/src/x509/mod.rs#L474-L477

lolzballs: Should I leave the version as 0.10.9? If not what should I bump it to?

sfackler: You can use 0.10.10.

openssl/src/ssl/test.rs
```diff
@@ -1536,3 +1536,39 @@ fn stateless() {
...
+    const CLIENT_IDENT: &[u8] = b"thisisaclient";
+
+    let listener = TcpListener::bind("127.0.0.1:0").unwrap();
+    let port = listener.local_addr().unwrap().port();
+
+    thread::spawn(move || {
+        let stream = listener.accept().unwrap().0;
+        let mut ctx = SslContext::builder(SslMethod::tls()).unwrap();
+        ctx.set_cipher_list(CIPHER).unwrap();
+        ctx.set_psk_server_callback(move |_, identity, psk| {
```
sfackler: Could you add a bit of code to ensure that these callbacks are being called? I'll look a bit like this test, for example: https://github.com/sfackler/rust-openssl/blob/4f3d72d98c527f571ac50fcb41317bc44d0fbfda/openssl/src/ssl/test.rs#L1026-L1063

Looks good other than this one last thing!

openssl/src/ssl/callbacks.rs
```diff
@@ -89,6 +89,40 @@ where
...
+            Some(CStr::from_ptr(identity).to_bytes())
+        } else {
+            None
+        };
+        // Give the callback mutable slices into which it can write the psk.
+        let psk_sl = slice::from_raw_parts_mut(psk as *mut u8, max_psk_len as usize);
+        match (*callback)(ssl, identity, psk_sl) {
+            Ok(psk_len) => psk_len as u32,
+            _ => 0,
```
sfackler: We should push the error back onto the stack here.

lolzballs:
The failing tests are due to the specific version of OpenSSL not supporting a suitable signature algorithm. I'm not sure which OpenSSL versions are supposed to support what ciphers so some guidance on this would be helpful.

sfackler:
You should be able to turn off TLSv1.3 when building against 1.1.1: https://github.com/sfackler/rust-openssl/blob/master/openssl/src/ssl/test.rs#L1065-L1067

PR #950, opened 2018-06-18
Add access to private/public components of DSA key pairs.

WanzenBug:
While working on a SSH-Agent implementation I noticed that there was no way to construct or get the private and public components from a DSA key. 

I added them here, in the style of similiar functionality in the RSA and EC key implementations.

### DIFF
--- a/openssl/src/dsa.rs
+++ b/openssl/src/dsa.rs
@@ -12,7 +12,7 @@ use std::fmt;
 use std::ptr;
 use std::mem;
 
-use bn::{BigNum, BigNumRef, BigNumContext};
+use bn::{BigNum, BigNumRef};
 use error::ErrorStack;
 use pkey::{HasParams, HasPrivate, HasPublic, Private, Public};
 use {cvt, cvt_p};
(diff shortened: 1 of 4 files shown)


openssl/src/dsa.rs
```diff
@@ -153,6 +177,31 @@ impl Dsa<Private> {
...
+    ///
+    /// `p`, `q` and `g` are the common parameters.
+    /// `priv_key` is the private component of the key pair.
+    /// The corresponding public component is calculated from the private component.
+    pub fn from_private_components(
+        p: BigNum,
+        q: BigNum,
+        g: BigNum,
+        priv_key: BigNum,
+    ) -> Result<Dsa<Private>, ErrorStack> {
```
sfackler: I'd probably keep this as purely calling into the various setters rather than computing the public key itself.

WanzenBug: Done. Should there be any checks for consistent keys? I'm not sure what openssl does when private and public key are not matching. Looking at the RSA code, there no checks either, so I guess it is fine?

sfackler: Thanks! We don't generally provide any extra checking on top of what OpenSSL already does, except for cases where that can cause memory safety issues. That's why we parameterize these types over what information they contain, for example.

PR #954, opened 2018-06-29
Link all needed system libraries on Windows, when building statically

pinkisemils:
When linking OpenSSL statically on Windows, `openssl-sys` doesn't declare all of system dependencies that OpenSSL has, as stated [here](https://github.com/openssl/openssl/blob/master/NOTES.WIN#L116-L118). This PR fixes that.

If this is by design, I can update the README and state so. Debugging this is quite time-intensive as a big enough project will link against a subset of these libraries independently of OpenSSL, but an example project might only require just one of them. Maybe these linker flags should be feature gated to avoid spreading bloat?

### DIFF
--- a/openssl-sys/Cargo.toml
+++ b/openssl-sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "openssl-sys"
-version = "0.9.33"
+version = "0.9.34"
 authors = ["Alex Crichton <alex@alexcrichton.com>",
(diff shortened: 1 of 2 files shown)


sfackler:
Thanks!

We didn't intentionally omit these - it just didn't come up as an issue before. I guess it's normally used in a context where other dependencies are pulling these in explicitly?

pinkisemils:
In our context, we were only missing `crypt32`, so I guess so. Thanks for the quick merge!

PR #959, opened 2018-07-14
Fix spelling in symm docs

jabedude:
Small fix, noticed this while reading the rust-openssl docs.

### DIFF
--- a/openssl/src/symm.rs
+++ b/openssl/src/symm.rs
@@ -23,7 +23,7 @@
 //!     &ciphertext[..]);
 //! ```
 //!
-//! Encrypting an assymetric key with a symmetric cipher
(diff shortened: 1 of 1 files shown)


openssl/src/symm.rs
```diff
@@ -23,7 +23,7 @@
 //!     &ciphertext[..]);
 //! ```
 //!
-//! Encrypting an assymetric key with a symmetric cipher
+//! Encrypting an asymetric key with a symmetric cipher
```
sfackler: `asymmetric` :)

jabedude: Haha! This is what I get for trying to open a PR on my phone!

sfackler:
Thanks!

PR #969, opened 2018-08-02
Hack around OpenSSL bug

sfackler:


sfackler:
Actually, I don't think this is the right place to handle these kinds of things.

PR #975, opened 2018-08-09
Add methods for DTLS/SRTP key handshake

eun-ice:
This change should allow for using DTLS/SRTP key handshake using the methods described in [SSL_CTX_set_tlsext_use_srtp](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_tlsext_use_srtp.html)

Open questions are all addressed.

- [X] On libressl 2.7.2 the test case just hangs for no apparent reason after exchanging a few bytes. (so test is disabled for now)

- [X] Is mapping of SRTP_PROTECTION_PROFILE ok? Specifically, the usage of a fake SRTP_PROTECTION_PROFILE_free method

- [X] Is it ok to copy data to a rust type (SrtpProfile) to obtain profile?

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -88,6 +88,7 @@ use hash::MessageDigest;
 #[cfg(ossl110)]
 use nid::Nid;
 use pkey::{HasPrivate, PKeyRef, Params, Private};
+use srtp::{SrtpProtectionProfile, SrtpProtectionProfileRef};
 use ssl::bio::BioMethod;
 use ssl::callbacks::*;
 use ssl::error::InnerError;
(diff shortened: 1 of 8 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -1930,6 +1948,24 @@ impl ToOwned for SslSessionRef {
...
+    SRTP_AES128_CM_SHA1_32,
+    SRTP_AES128_F8_SHA1_80,
+    SRTP_AES128_F8_SHA1_32,
+    SRTP_NULL_SHA1_80,
+    SRTP_NULL_SHA1_32,
+    Other(usize),
+}
+
+#[derive(Debug, PartialEq)]
+pub struct SrtpProfile<'a> {
```
sfackler: Rather than defining a Rust struct for this, I think we should return an `&StrpProfileRef` and have methods on that to grab the ID and name.

eun-ice:
Thank you for all your suggestions. 

I tried to make all the changes, but did not yet find the time to improve the tests. 

It would be nonetheless nice if you could have a look and tell me if I am on the right track.

eun-ice:
@sfackler now everything is done, including test cases and documentation. I hope all of this now matches the style of the rest of the lib.

sfackler:
Thanks!

eun-ice:
Thank you for merging and cleaning up.

PR #997, opened 2018-09-17
Add impl From<u32> for BigNum.

max-b:
This allows for a function which takes generic parameters to still operate on constant numbers like 1 and 0. As a sort of trivial example:
```
    fn plus_one<T>(a: &T) -> T 
        where T:
            cmp::PartialEq + 
            From<u32>,
            for<'a> &'a T: ops::Add<Output = T>,
    {
        a + &T::from(1)
    }
```

I hope that using `unwrap()` in that way is ok. It seems that's how you're implementing the `ops::` traits, so I figured it would?

### DIFF
--- a/openssl/src/bn.rs
+++ b/openssl/src/bn.rs
@@ -1130,6 +1130,12 @@ impl fmt::Display for BigNum {
     }
 }
 
+impl From<u32> for BigNum {
+    fn from(num: u32) -> Self {
+        Self::from_u32(num).unwrap()
+    }
+}
+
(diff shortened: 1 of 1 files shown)


sfackler:
I think I'd rather not have these kinds of panicking conversions, and instead use `TryFrom` when/if it stabilizes.

max-b:
Gotcha. That's https://github.com/rust-lang/rust/issues/33417 right?

Thanks for the feedback :)

sfackler:
Yep, that's the issue to follow!

PR #1001, opened 2018-09-25
Compat release for 1.1.1

ishitatsuyuki:
Pardon the poor code quality, I didn't really test this. Though, OpenSSL 1.1.1 is binary and API compatible with 1.1.0 so I expect things to pass. I added CI configurations to test if this stands.


sfackler:
The 0.9 release is no longer supported.

PR #1002, opened 2018-09-27
Support the rest of LibreSSL 2.8.x

vishwin:
LibreSSL 2.8.1 released, so update the check for all versions in the series, not just 2.8.0. Also fix the version mismatch error message that was missed in the initial 2.8.0 support addition.

Did not change any of the CI configs, but did run `cargo test` locally with all checks passing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,6 +108,9 @@ libressl_250: &LIBRESSL_250
   VERSION: 2.5.0
 libressl_270: &LIBRESSL_280
   LIBRARY: libressl
+  VERSION: 2.8.0
+libressl_281: &LIBRESSL_281
+  LIBRARY: libressl
   VERSION: 2.8.1
 
 x86_64: &X86_64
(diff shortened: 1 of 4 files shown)


sfackler:
It doesn't look like 2.8.1 has been declared to be ABI stable, so we can't whitelist 2.8.2 and up yet.

vishwin:
Yeah, `libcrypto.so` changed from `.43` on 2.8.0 to `.44` on 2.8.1, among other things.

sfackler:
Can you also bump the version we use in CI? Looks good otherwise! https://github.com/sfackler/rust-openssl/blob/master/.circleci/config.yml#L111

sfackler:
It looks like a function signature changed; https://circleci.com/gh/sfackler/rust-openssl/10546?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

sfackler:
Thanks!

PR #1004, opened 2018-09-28
Add support for encoding and decoding ECDSA signatures

mbelop:
DER encoding of ECDSA signatures via `i2d_ECDSA_SIG` and `d2i_ECDSA_SIG`.


sfackler:
Thanks!

PR #1008, opened 2018-10-16
Bump dependency on openssl-src-rs

alexcrichton:
Brings in the first release with OpenSSL 1.1.1

### DIFF
--- a/openssl-sys/Cargo.toml
+++ b/openssl-sys/Cargo.toml
@@ -20,7 +20,7 @@ libc = "0.2"
 [build-dependencies]
 pkg-config = "0.3.9"
 cc = "1.0"
-openssl-src = { version = "110.0.4", optional = true }
+openssl-src = { version = "111.0.1", optional = true }
 
 [target.'cfg(target_env = "msvc")'.build-dependencies]
 vcpkg = "0.2"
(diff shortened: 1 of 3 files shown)


sfackler:
Looks like we'll need to add some cfg logic to detect no-stdio - the defines we look for are here: https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/build/main.rs#L18

alexcrichton:
I think that this update will fix this build failure -- https://github.com/rust-lang/rust/pull/54919#issuecomment-430066943, so mind publishing this after it's all good too?

sfackler:
Sure, I'll publish when this lands.

sfackler:
We'll still need the detection code to turn off the relevant methods on Android on the openssl-sys end though.

alexcrichton:
Ok now it's all green!

sfackler:
Great!

We could in theory set no-stdio everywhere, right? Since it only seems to affect that one function that we don't use anywhere.

alexcrichton:
I thought so! Unfortunately though doing so [causes failures like this](https://circleci.com/gh/sfackler/rust-openssl/10833) where `no-stdio` disables usage of `FILE` entirely I think which prevents OpenSSL from being able to load system cert stores and such. Although presumably if the `openssl` crate could provide that functionality to OpenSSL we could do it

sfackler:
Hmm, that may be unfortunate for people trying to target android if they can't actually load certs.

I guess we'll just deal with that as the need arises.

alexcrichton:
Yeah I suspect we'll need to circle back at some point and actually fix the compilation failure on Android rather than disabling features, but we can always cross that bridge later!

PR #1015, opened 2018-11-07
Add support for LibreSSL 2.8.2

ibabushkin:
Hi, since the new LibreSSL version is now released (and shipped in several distros), supporting it would make sense. I also updated the dependency on `ctest`, as the build otherwise fails for me (presumably because the old version's parsing is somewhat incomplete).

### DIFF
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -503,6 +503,7 @@ See rust-openssl README for more information:
             (7, _) => ('7', 'x'),
             (8, 0) => ('8', '0'),
             (8, 1) => ('8', '1'),
+            (8, 2) => ('8', '2'),
             (8, _) => ('8', 'x'),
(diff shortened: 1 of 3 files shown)


openssl-sys/build/main.rs
```diff
@@ -503,6 +503,7 @@ See rust-openssl README for more information:
             (7, _) => ('7', 'x'),
             (8, 0) => ('8', '0'),
             (8, 1) => ('8', '1'),
+            (8, 2) => ('8', '2'),
             (8, _) => ('8', 'x'),
```
sfackler: We already support 2.8.2 from this line here.

ibabushkin: Oh, indeed. I also noticed the `ctest` update fails on stable apparenlty. I'll see if I can find a fix.

PR #1018, opened 2018-11-21
Add SHA3 & SHAKE128/256 EVP message digest functions in OpenSSL 1.1.1, fixes #1017.

### LINKED ISSUE #1017: Support for SHAKE256/128
Are there plans to support SHAKE256/128?

It is in openssl as an [EVP function](https://www.openssl.org/docs/man1.1.1/man3/EVP_shake256.html) but I don't see that in the extern for `openssl-sys/evp.rs`.

sameer:
This PR adds SHA3 & SHAKE128/256 support to the crate with appropriate tests and an example of SHAKE128.

SHA3 can be used the same way as all pre-existing message digests.
SHAKE128/256 however, must use the XOF variants `Hasher::finish_xof` and `openssl::hash::hash_xof`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ save_openssl: &SAVE_OPENSSL
     paths:
     - /openssl
 deps_key: &DEPS_KEY
-  key: deps-1.20.0-{{ checksum "Cargo.lock" }}-{{ checksum "~/lib_key" }}-2
+  key: deps-1.24.1-{{ checksum "Cargo.lock" }}-{{ checksum "~/lib_key" }}-2
 restore_deps: &RESTORE_DEPS
   restore_cache:
     <<: *DEPS_KEY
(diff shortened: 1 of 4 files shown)


sameer:
So the tests are failing, but I don't think it's related to what I did (hopefully). I'm looking at the [x86_64-openssl-1.0.2](https://circleci.com/gh/sfackler/rust-openssl/11132?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link) one and it is referring to lazy_static and an unstable const_fn.

sfackler:
Yeah looks like we'll need to bump our minimum supported rust version:

https://github.com/sfackler/rust-openssl/blob/master/.circleci/config.yml#L34
https://github.com/sfackler/rust-openssl/blob/master/.circleci/config.yml#L77

sameer:
So I had to bump the OpenSSL versions for Windows as well, but [one test is still failing](https://ci.appveyor.com/project/sfackler/rust-openssl/builds/20473062/job/rbivl7l4s6eg4bh6) and the reason is unclear. Otherwise, everything else seems fine.

PR #1020, opened 2018-11-22
Add bindings to RAND_keep_random_devices_open

### LINKED ISSUE #1019: Expose RAND_keep_random_devices_open()
There is an issue open in cargo where open file descriptors to `/dev/{u,}random` leak to tests being executed by cargo. This is probably due to [this commit](https://github.com/openssl/openssl/commit/c7504aeb640a88949dfe3146f7e0f275f517464c) in OpenSSL, which deliberately leaves the rand-source open in case we `chroot` somewhere where such sources are not available. A knob `RAND_keep_random_devices_open()` was introduced in the same commit, which would allow cargo to tell OpenSSL to close those file descriptors. AFAICS `rust-openssl` does not export this knob, which would probably solve rust-lang/cargo#6333

sfackler:
Closes #1019

### DIFF
--- a/openssl-sys/src/rand.rs
+++ b/openssl-sys/src/rand.rs
@@ -2,5 +2,9 @@ use libc::*;
 
 extern "C" {
     pub fn RAND_bytes(buf: *mut u8, num: c_int) -> c_int;
+
+    #[cfg(ossl111)]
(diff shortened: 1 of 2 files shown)


sfackler:
cc @lukaslueg

PR #1022, opened 2018-11-24
Fix cargo dependency documentation

infinityb:
`feature` -> `features`

As documented, you'd get the follow warning: `warning: unused manifest key: dependencies.openssl.feature`


sfackler:
Ah, thanks!

PR #1025, opened 2018-12-03
Explicitly support LibreSSL 2.9.0.

ltratt:
My OpenBSD-current machine reports LibreSSL 2.9.0 (https://github.com/libressl-portable/openbsd/commit/091c2f11e0cdcdbe529170fc394fe359922c0e34). This hasn't yet had a formal release AFAICS, but it does mean that openssl-sys can't build. Nothing has changed in a bad way since 2.8.2 that I can see from the logs, and `cargo test` passes. However, since it's not clear to me if further 2.9.* releases would maintain any such guarantees, I've conservatively restricted things to only match against 2.9.0.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,9 +106,6 @@ openssl_101: &OPENSSL_101
 libressl_250: &LIBRESSL_250
   LIBRARY: libressl
   VERSION: 2.5.0
-libressl_28x: &LIBRESSL_28x
-  LIBRARY: libressl
-  VERSION: 2.8.2
 libressl_290: &LIBRESSL_290
   LIBRARY: libressl
   VERSION: 2.9.0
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -109,6 +109,9 @@ libressl_250: &LIBRESSL_250
 libressl_28x: &LIBRESSL_28x
   LIBRARY: libressl
   VERSION: 2.8.2
+libressl_290: &LIBRESSL_290
```
sfackler: Could you replace the 28x tests with 290? There's also the test list down all the way at the bottom of this file that you'll need to update.

ltratt: Happy to do so! Hopefully that new commit fixes this.

sfackler:
That test isn't going to pass until they put out a 2.9.0 tarball: https://circleci.com/gh/sfackler/rust-openssl/11377

ltratt:
Agreed :/

vishwin:
The 2.9.0 tarball now exists, as of yesterday.

ltratt:
Agreed. The release notes are here https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-2.9.0-relnotes.txt. AFAICS, rerunning the tests should be enough?

ltratt:
Agreed. The release notes are here https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-2.9.0-relnotes.txt. AFAICS, rerunning the tests should be enough?

sfackler:
Yep, looks like we're good to go!

PR #1034, opened 2019-01-06
Add some additional CMS methods

ur0:
This adds the following methods:
- Asn1Object::from_txt
- CmsContentInfo::add_signer
- CmsContentInfo::verify
- CmsContentInfo::partial

Documentation is still pending and should be addressed in the next commit, but I'd be grateful for a code review.

### DIFF
--- a/openssl/src/asn1.rs
+++ b/openssl/src/asn1.rs
@@ -27,6 +27,7 @@
 use ffi;
 use foreign_types::{ForeignType, ForeignTypeRef};
 use libc::{c_char, c_int, c_long};
+use std::ffi::CString;
 use std::fmt;
 use std::ptr;
 use std::slice;
@@ -271,6 +272,18 @@ foreign_type_and_impl_send_sync! {
     ///
     /// [`Asn1Object`]: struct.Asn1Object.html
     pub struct Asn1ObjectRef;
+}
+
+impl Asn1Object {
+    /// Returns an ASN1_object from a given OID in text form.
+    pub fn from_txt(oid: &str, no_name: bool) -> Asn1Object {
+        unsafe {
+            Asn1Object::from_ptr(ffi::OBJ_txt2obj(
+                CString::new(oid).unwrap().as_ptr(),
+                no_name as c_int,
+            ))
+        }
+    }
 }
 
 impl Asn1ObjectRef {
--- a/openssl/src/cms.rs
+++ b/openssl/src/cms.rs
@@ -6,16 +6,59 @@
 //! Data accepted by this module will be smime type `enveloped-data`.
 
 use ffi;
-use foreign_types::{ForeignType, ForeignTypeRef};
+use foreign_types::{ForeignType, ForeignTypeRef, Opaque};
+use libc::c_void;
 use std::ptr;
 
+use asn1::Asn1Object;
 use bio::{MemBio, MemBioSlice};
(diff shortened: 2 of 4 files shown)


openssl/src/asn1.rs
```diff
@@ -273,6 +274,18 @@ foreign_type_and_impl_send_sync! {
     pub struct Asn1ObjectRef;
 }
 
+impl Asn1Object {
+    /// Returns an ASN1_object from a given OID in text form.
+    pub fn from_txt(oid: &str, no_name: bool) -> Asn1Object {
+        unsafe {
+            Asn1Object::from_ptr(ffi::OBJ_txt2obj(
+                CString::new(oid).unwrap().as_ptr(),
```
sfackler: This looks like a use-after-free.

openssl/src/cms.rs
```diff
@@ -6,16 +6,59 @@
...
 use pkey::{HasPrivate, PKeyRef};
-use stack::StackRef;
-use x509::{X509, X509Ref};
-use {cvt, cvt_p};
+use stack::{Stack, StackRef};
+use x509::store::X509Store;
+use x509::{X509Ref, X509};
+use {cvt, cvt_n, cvt_p};
+
+pub struct CmsSignerInfoRef(Opaque);
```
sfackler: You should be able to use the macro to create these.

openssl/src/cms.rs
```diff
@@ -6,16 +6,59 @@
...
+        CmsSignerInfo(ptr)
+    }
+
+    fn as_ptr(&self) -> *mut ffi::CMS_SignerInfo {
+        self.0
+    }
+}
+
+impl CmsSignerInfo {
+    pub fn add_attr(&mut self, obj: &Asn1Object, data_type: i32, data: &[u8]) -> i32 {
```
sfackler: This should probably return a result.

Is data_type an arbitrary integer or actually an enum?

openssl/src/cms.rs
```diff
@@ -161,4 +303,24 @@ impl CmsContentInfo {
             Ok(CmsContentInfo::from_ptr(cms))
         }
     }
+
+    pub fn partial(
```
sfackler: Why is this called partial if it calls CMS_sign?

openssl/src/cms.rs
ur0: Sorry, but I'm unable to figure out the correct macro for this.

ur0:
@sfackler: Tried to resolve most issues, working on tests now.

openssl/src/cms.rs
ur0: I can't find any documentation for it, but it seems to be an opaque integer.

openssl/src/cms.rs
ur0: Documented the method, hopefully it makes more sense now.

sfackler:
Have you had a chance to add those tests?

scouten:
I need to use the APIs cited in https://stackoverflow.com/questions/55654448/using-openssl-api-how-can-i-access-a-signers-certificate-info-from-a-signed-fil, which include some of the APIs in this PR and a few more.

@sfackler would you be open to an alternative PR that uses this work and bridges the additional APIs?

TommyLike:
What's the status of this PR?

PR #1038, opened 2019-01-17
Implement AsRef<str/[u8]> for OpensslString{Ref}

jonas-schievink:
These are very useful in a generic context where `T: AsRef<str>` or `T: AsRef<[u8]>` is expected.


sfackler:
Thanks!

PR #1039, opened 2019-01-18
Add `Asn1Integer::from_bn`

jonas-schievink:
AFAICT this is now the first way an `Asn1Integer` can be constructed.


jonas-schievink:
Actually I'll probably need to add constructors for all the other types in the `asn1` module, so I can also put everything into a single PR if you prefer.

jonas-schievink:
Ah, I forgot about `BigNumRef::to_asn1_integer`, so maybe this is unnecessary?

sfackler:
I think it's reasonable to provide both!

Could you change the implementation around to just call to_asn1_integer though? Less unsafe code that way.

PR #1040, opened 2019-01-18
Add RsaRef::check_key

Zolmeister:


openssl/src/rsa.rs
```diff
@@ -278,6 +278,23 @@ where
...
+
+    /// Validates RSA parameters for correctness
+    ///
+    /// This corresponds to [`RSA_check_key`].
+    ///
+    /// [`RSA_check_key`]: https://www.openssl.org/docs/man1.1.0/crypto/RSA_check_key.html
+    pub fn check_key(&self) -> Result<bool, ErrorStack> {
+        unsafe {
+            let result =  ffi::RSA_check_key(self.as_ptr()) as i32;
+            assert!(result != -1);
```
sfackler: Why would we panic rather than returning an error?

Zolmeister:
updated (new to rust)

sfackler:
Thanks!

Issue #1041, opened 2019-01-20
How to resume session tickets without using unsafe?


pyfqm:
Hi,

I might have overlooked something, but from the doc it seems the only way to resume from client session cache is the unsafe `set_session()`. It feels uneasy to use unsafe in high-level SSL code. Is there anything preventing us from validating the safety guarantee:

> The caller of this method is responsible for ensuring that the session is associated with the same SslContext as this Ssl.

?

Because we can always use `SSL_get_SSL_CTX()` to retrieve the SSL_CTX from SSL, the challenge becomes making sure that the SSL_SESSION can be used with that SSL_CTX. At the moment I can think of three ways to implement this:

1. Duplicate the SSL_SESSION with [SSL_SESSION_dup()](https://www.openssl.org/docs/manmaster/man3/SSL_SESSION_dup.html). It removes the "internal links" mentioned in [SSL_set_session()](https://www.openssl.org/docs/manmaster/man3/SSL_set_session.html), effectively disassociating it with whatever SSL_CTX is was previously associated with. The disadvantages are
- Duplicating an SSL_SESSION may lead to subtle bugs, such as with `ex_data`.
- Wastes memory.
- `SSL_SESSION_dup()` is introduced in OpenSSL 1.1.1.

2. Keep track of the SSL_CTX where the SslSession / SslSessionRef was created from in the struct. The disadvantage is that this requires a reference count to the SSL_CTX in each session.

3. Use a global ID manager. Store the ID for each SSL_CTX in the `ex_data` of SSL_CTX / SSL_SESSION.

Because I'm not familiar with rust-openssl, I'm sorry for any misconception here. Please correct me if I'm wrong.

Thank you!

sfackler:
I wouldn't really call this API "high-level". It rather tries to be a pretty direct binding to OpenSSL's APIs.

pyfqm:
> I wouldn't really call this API "high-level". It rather tries to be a pretty direct binding to OpenSSL's APIs.

Well, relatively :) Direct binding would be `openssl-sys`. `openssl` has `set_session()` as the only unsafe function carrying an entire feature at TLS level. `tokio-openssl` and `hyper-openssl`, both of which you're the author, point to this crate for configuration. So this is really the only way.

So ..., any chance this can be solved? Something like `set_session_from_der()` would have minimal impact and no overhead on the rest of the crate.

sfackler:
[...]

PR #1042, opened 2019-01-22
Add session cache to SslConnector

pyfqm:
Currently to setting up session cache on the client side requires a bunch of callbacks and unsafe fn. This PR makes it easier by adding an opt-in cache storage to SslConnector, and exposes a simple and safe API (`set_use_session_cache` `set_session_cache_key`) for common use-cases. Existing applications will not change their behavior.

The relevant SessionCache code was copied from hyper-openssl.

### DIFF
--- a/openssl/Cargo.toml
+++ b/openssl/Cargo.toml
@@ -24,6 +24,7 @@ cfg-if = "0.1"
 foreign-types = "0.3.1"
 lazy_static = "1"
 libc = "0.2"
+linked_hash_set = "0.1"
 
 openssl-sys = { version = "0.9.40", path = "../openssl-sys" }
 
--- a/openssl/src/lib.rs
+++ b/openssl/src/lib.rs
@@ -118,6 +118,7 @@ extern crate foreign_types;
 #[macro_use]
 extern crate lazy_static;
 extern crate libc;
+extern crate linked_hash_set;
 extern crate openssl_sys as ffi;
 
 #[cfg(test)]
--- a/openssl/src/ssl/cache.rs
+++ b/openssl/src/ssl/cache.rs
@@ -1,0 +1,82 @@
+use linked_hash_set::LinkedHashSet;
+use ssl::{SslSession, SslSessionRef};
+use std::borrow::Borrow;
+use std::collections::hash_map::{Entry, HashMap};
+use std::hash::{Hash, Hasher};
+
+#[derive(Hash, PartialEq, Eq, Clone)]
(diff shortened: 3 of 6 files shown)


pyfqm:
@sfackler Is adding session cache to __`SslConnector`__ a good idea then? (I explained this in https://github.com/sfackler/rust-openssl/issues/1041#issuecomment-456449072)

Also cc @alexcrichton: What would be the best way to allow for safe TLS session cache/resumption in the rust ecosystem? Is it idiomatic to "just use unsafe"?

sfackler:
I was thinking of this living in a third party crate, at least initially.

pyfqm:
@sfackler 
Er... You want me to create a crate specifically for this? The reality is hardly anyone is gonna use it. Would you perhaps like to put this behind a feature gate?

sfackler:
If hardly anyone is going to use it, why should it live in this crate either?

pyfqm:
There may be a lot of people who want to use features like this, but many of them (me included) want them to be in a well-maintained and trusted crate. What is your concern with using a feature gate?

sfackler:
I'm confused. Are you saying that the number of people who want to use this feature is "hardly anyone" or "a lot of people"?

What's the point of putting this behind a feature gate? My reason in leaning towards wanting it out of crate is to allow more flexibility in figuring out what the right API is. A feature gate is not going to help with that.

pyfqm:
It's very hard for a new user to gain trust and popularity with one crate. So yes, even if a lot of people want this, hardly anyone is going to want it from someone they don't know.

Oftentimes feature gates come with a document saying "this is unstable and is subject to change." They still allow the flexibility in figuring out the right API. Take, say, rust nightly for example.

sfackler:
Rust nightly features are not the same thing as Cargo features. I don't want to lie about the semantic versioning of this crate.

pyfqm:
@sfackler
What do you do when you want to test a small-to-medium sized features? Are we supposed to create a crate to each one of them? To me a crate for each feature seems crazy and would be have much lower discoveribility.

If there's a better way to do this, I'd like to help.

sfackler:
Like I said before, there aren't really many features like that in this library, since almost everything is a direct equivalent to some OpenSSL function. The one exception is the connector/hostname verification setup which was prototyped out-of-tree: https://crates.io/crates/openssl-verify

rohitjoshi:
@pyfqm and @sfackler any plan to merge this PR for session cache support?

PR #1044, opened 2019-01-23
Fix static openssl-sys build on windows-gnu

### LINKED ISSUE #1014: Cross compiling openssl-sys to Windows with mingw fails if vendored
The build fails when trying to find the built static/dynamic libraries. The reason seems to be that with mingw (as opposed to msvc) .a files are created. Since both targets are treated equally here:
https://github.com/sfackler/rust-openssl/blob/b88778bc76b039e26aaa3d54f03716433986fae1/openssl-sys/build/main.rs#L85

...the build ends up looking for liblibssl.a and fails.

pyfqm:
Fixes #1014.

### DIFF
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -78,16 +78,43 @@ fn main() {
     let version = validate_headers(&[include_dir.clone().into()]);
 
     let libs_env = env("OPENSSL_LIBS");
-    let libs = match libs_env.as_ref().and_then(|s| s.to_str()) {
-        Some(ref v) => v.split(":").collect(),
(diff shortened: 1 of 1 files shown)


pyfqm:
@sfackler any progress?

fleimgruber:
@pyfqm ping for rebase as I would test this if I could patch cleanly.

abcfy2:
Same issue here. Please merge this pull request.

laumann:
Just chiming in to say that I'd also be interested in seeing this merged :smile: Thanks for a great library @sfackler

laumann:
@pyfqm I think this one can be closed, as #1265 has been merged to fix the same issue.

PR #1046, opened 2019-01-24
Add OCSP nonce functionality

### LINKED ISSUE #1045: Add nonces to OCSP requests and responses
I would like the following functionality exposed via the Rust API:

- [ ] `int OCSP_request_add1_nonce(OCSP_REQUEST *req, unsigned char *val, int len)`
- [ ] `int OCSP_basic_add1_nonce(OCSP_BASICRESP *resp, unsigned char *val, int len)`
- [ ] `int OCSP_check_nonce(OCSP_REQUEST *req, OCSP_BASICRESP *bs)`
- [ ] `int OCSP_copy_nonce(OCSP_BASICRESP *resp, OCSP_REQUEST *req)`

I am happy to make a pull request for this, however I am a bit unfamiliar with Rust FFI and am not sure exactly how to expose the `unsigned char *val, int len` tuple in Rust.

Also, where should the check and copy functionality be implemented? My instinct is `OCSP_check_nonce` should be a free function and that `OCSP_copy_nonce` should be on the response object, but input on this is welcome.

Here is an outline of my proposed implementation of this. The version for `OCSP_BASICRESP` is obviously similar. 
```rust
impl OcspRequestRef {
    pub fn add_nonce(&mut self, &val: Option<some_type>) -> Result<something, ErrorStack> {
        unsafe {
            // if val is None, pass NULL as val and 0 as len--how best to do?
            cvt(ffi::OCSP_request_add1_nonce(self.as_ptr(), val.as_ptr(), val.len()))?;
            Ok(something)
        }
    }
}
```

scolby33:
Will close #1045.

I intend to add as well:
- [x] test OCSP request creation, with and without explicit nonce
- [ ] test OCSP response creation with and without explicit nonce
- [ ] test OCSP response creation, copying nonce from a request
- [ ] test OCSP nonce comparison
- [ ] some basic documentation for my changes

In the meantime, feedback on what I have written is welcome.


openssl/src/ocsp.rs
```diff
@@ -200,6 +200,24 @@ impl OcspBasicResponseRef {
...
+            let (ptr, len) = match val {
+                Some(slice) => (slice.as_ptr() as *mut _, slice.len() as c_int),
+                None => (ptr::null_mut(), 0),
+            };
+            cvt(ffi::OCSP_basic_add1_nonce(self.as_ptr(), ptr, len))?;
+            Ok(())
+        }
+    }
+
+    pub fn copy_nonce(&mut self, req: OcspRequestRef) -> Result<(), ErrorStack> {
```
sfackler: This should take `&OcspRequestRef`, not `OcspRequestRef`.

openssl/src/ocsp.rs
```diff
@@ -345,3 +374,48 @@ foreign_type_and_impl_send_sync! {
     pub struct OcspOneReq;
     pub struct OcspOneReqRef;
 }
+
+pub fn check_nonce(req: &OcspRequestRef, bs: &OcspBasicResponseRef) -> Result<(), ErrorStack> {
+    unsafe {
+        cvt(ffi::OCSP_check_nonce(req.as_ptr(), bs.as_ptr()))?;
```
sfackler: This function will need a custom return type, looking at the documentation.

openssl/src/ocsp.rs
```diff
@@ -28,6 +30,44 @@ bitflags! {
     }
 }
 
+#[derive(Debug)]
```
scolby33: Ought I derive `Copy, Clone, Debug, PartialEq, Eq` as well?

openssl/src/ocsp.rs
```diff
@@ -28,6 +30,44 @@ bitflags! {
...
+        let message = match *self {
+            OcspNonceCheckSuccessResult::PresentAndEqual => "Nonces Present and Equal",
+            OcspNonceCheckSuccessResult::Absent => "Nonces Absent from Request and Response",
+            OcspNonceCheckSuccessResult::PresentInResponseOnly => "Nonce Present in Response Only",
+        };
+        write!(f, "{}", message)
+    }
+}
+
+#[derive(Debug)]
```
scolby33: Same as above: should `Copy, Clone, PartialEq, Eq` be added?

openssl/src/ocsp.rs
scolby33: I've implemented this custom return. I don't *love* what I've done. Please let me know what you think.

scolby33:
I've pushed two commits based on your comments, @sfackler. I will be able to get to the remaining testing and adding documentation late on Monday. Thanks for helping me with this implementation so far!

openssl/src/ocsp.rs
```diff
@@ -345,3 +417,56 @@ foreign_type_and_impl_send_sync! {
     pub struct OcspOneReq;
     pub struct OcspOneReqRef;
 }
+
+pub fn check_nonce(req: &OcspRequestRef, bs: &OcspBasicResponseRef) -> Result<OcspNonceCheckSuccessResult, OcspNonceCheckErrorResult> {
```
sfackler: I'd have this be a bit simpler:

```rust
pub fn check_nonce(req: &OcspRequestRef, bs: &OcspBasicResponseRef) -> NonceCheck { ... }

#[derive(Debug, Copy, Clone, PartialEq, Eq)]
pub struct NonceCheck(c_int);

impl NonceCheck {
    pub const REQUEST_ONLY: NonceCheck = NonceCheck(-1);
    pub const UNEQUAL: NonceCheck = NonceCheck(0);
    pub const EQUAL: NonceCheck = NonceCheck(1);
    pub const ABSENT: NonceCheck = NonceCheck(2);
    pub const RESPONSE_ONLY: NonceCheck = NonceCheck(3);
}
```

scolby33: I agree that my version is complicated, but it has the benefit of making the user handle the error case; I believe that failing to `if (thing_that_returns_nonzero_on_error()) { goto uh_oh; }` is a relatively common source of security bugs. In my version, the compiler will warn about an unused result or if the user chose to use `unwrap()`/`?` it'll panic/propagate if proper error handling isn't done.

I'm happy to use your version if you disagree, but I think my version gives a more ergonomic usage for the common case.

(Warning: pseudocode)

```rust
fn perform_ocsp_request(...) -> ... {
    let my_req = ocsp::OcspRequest::new().unwrap();
    let res = send_request(&my_req);

    ocsp::check_nonce(&my_req, &res)?
}
```

Versus

```rust
fn perform_ocsp_request(...) -> ... {
    let my_req = ocsp::OcspRequest::new().unwrap();
    let res = send_request(&my_req);

    match ocsp::check_nonce(&my_req, &res) {
        ocsp::NonceCheck::REQUEST_ONLY => Err(()),
        ocsp::NonceCheck::UNEQUAL => Err(()),
        ocsp::NonceCheck::EQUAL => Ok(()),
        ...
    }
}
```

It only moves the work of handling the error up a level, but at least it's impossible to avoid handling it somewhere.

Looking forward to hearing your thoughts on this.

sfackler: How unambiguous is the error/ok decision in this context?

We can always tag the return type here `#[must_use]` just like `Result`.

scolby33: That's a good point. The OpenSSL docs make a reasonable case that at least a 0 return value should be an unambiguous error signal:

>The return values of OCSP_check_nonce() can be checked to cover each case. A positive return value effectively indicates success: nonces are both present and match, both absent or present in the response only. A non-zero return additionally covers the case where the nonce is present in the request only: this will happen if the responder doesn't support nonces. A zero return value indicates present and mismatched nonces: this should be treated as an error condition.

https://www.openssl.org/docs/man1.1.0/man3/OCSP_check_nonce.html

I like the idea of using `#[must_use]` for this at least, but I could see an argument that -1 at least shouldn't exactly be an error. Maybe a combination of our approaches like

```rust
pub fn check_nonce(req: &OcspRequestRef, bs: &OcspBasicResponseRef) -> Result<NonceCheck, SomeErrorNotSureWhat> { ... }

impl NonceCheck {
    pub const REQUEST_ONLY: NonceCheck = NonceCheck(-1);
    pub const EQUAL: NonceCheck = NonceCheck(1);
    pub const ABSENT: NonceCheck = NonceCheck(2);
    pub const RESPONSE_ONLY: NonceCheck = NonceCheck(3);
}
```

scolby33:
Sorry for letting this sit for so long. I'm going to do some work on the testing parts of this implementation while we work out the best return type for the `OCSP_check_nonce()` function.

scolby33:
Actually, for the moment, I think I'm going to set this aside again.

PR #1054, opened 2019-01-28
add Rsa::generate_with_e(bits: u32, e: BigNum)

Zolmeister:


openssl/src/rsa.rs
```diff
@@ -597,6 +597,25 @@ impl Rsa<Private> {
         }
     }
 
+    /// Generates a public/private key pair with the specified size.
+    ///
+    /// This corresponds to [`RSA_generate_key_ex`].
+    ///
+    /// [`RSA_generate_key_ex`]: https://www.openssl.org/docs/man1.1.0/crypto/RSA_generate_key_ex.html
+    pub fn generate_with_e(bits: u32, e: BigNum) -> Result<Rsa<Private>, ErrorStack> {
```
sfackler: This can just take a `&BigNumRef`.

sfackler:
Out of curiosity, why do you want to use custom exponents?

openssl/src/rsa.rs
Zolmeister: I don't think so because the `ffi` takes `e: *mut BIGNUM`

```rust
pub fn RSA_generate_key_ex(
        rsa: *mut RSA,
        bits: c_int,
        e: *mut BIGNUM,
        cb: *mut BN_GENCB,
    ) -> c_int;
```

sfackler: You can get that from a `&BigNumRef` with `.as_ptr()`.

Zolmeister:
For vanity addresses, similar to [vanitygen](https://github.com/samr7/vanitygen) but with Rust
Basically generate RSA key on the CPU, then increment the exponent on the GPU.
It's safe, so long as you ensure [`e is an integer between 3 and n–1 satisfying GCD (e, λ(n)) = 1, where
λ(n) = LCM (p–1,q–1)`](http://www.di-srv.unisa.it/~ads/corso-security/www/CORSO-9900/oracle/pkcsv21.pdf) and that the resultant key is valid ([`RSA_check_key`](https://www.openssl.org/docs/man1.1.0/crypto/RSA_check_key.html)).

openssl/src/rsa.rs
sfackler: `BigNum` and `BigNumRef` are a bit like `String` and `str` or `PathBuf` and `Path`.

https://github.com/sfackler/rust-openssl/blob/master/openssl/src/rsa.rs#L605-L607

PR #1060, opened 2019-02-17
Add des_ede3_cfb64 symm cipher

okuryu:
I'd like to add the des_ede3_cfb cipher. Could you take a look? Thanks!


sfackler:
It looks like des_ede3_cfb is actually an alias for des_ede3_cfb64 which is why the tests are failing: https://github.com/openssl/openssl/blob/8269e44f9e40831a497fe9f31ba1d65aeb49a5c1/include/openssl/evp.h#L752.

okuryu:
oh, i didn't know that! Is this mean that I should change from EVP_des_ede3_cfb to EVP_des_ede3_cfb64?

sfackler:
Yep!

okuryu:
Done! Passed the tests.

sfackler:
Thanks!

PR #1074, opened 2019-03-01
expose EC_GROUP_cmp/get_curve_name/set_curve_name

benmmurphy:
expose EC_GROUP_cmp, EC_GROUP_get_curve_name and EC_GROUP_set_curve_name functions

### DIFF
--- a/openssl/src/ec.rs
+++ b/openssl/src/ec.rs
@@ -237,6 +237,49 @@ impl EcGroupRef {
             ffi::EC_GROUP_set_asn1_flag(self.as_ptr(), flag.0);
         }
     }
+
+    /// Check if two curves are equal.
+    ///
+    /// OpenSSL documentation available at [`EC_GROUP_cmp`]
+    ///
+    /// [`EC_GROUP_cmp`]: https://www.openssl.org/docs/man1.1.0/man3/EC_GROUP_cmp.html
+    pub fn eq(&self, other: &EcGroupRef, ctx: &mut BigNumContextRef) -> Result<bool, ErrorStack> {
+        unsafe {
+            let res = cvt_n(ffi::EC_GROUP_cmp(
+                self.as_ptr(),
+                other.as_ptr(),
+                ctx.as_ptr(),
+            ))?;
+            Ok(res == 0)
+        }
+    }
+
+    /// Allows a `Nid` to be associated with a curve.
+    ///
+    /// OpenSSL documentation available at [`EC_GROUP_set_curve_name`]
+    ///
+    /// [`EC_GROUP_set_curve_name`]: https://www.openssl.org/docs/man1.1.0/man3/EC_GROUP_set_curve_name.html
+    pub fn set_curve_name(&self, nid: Option<Nid>) {
+        unsafe {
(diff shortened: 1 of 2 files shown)


openssl/src/ec.rs
```diff
@@ -237,6 +237,49 @@ impl EcGroupRef {
...
+            Ok(res == 0)
+        }
+    }
+
+    /// Allows a `Nid` to be associated with a curve.
+    ///
+    /// OpenSSL documentation available at [`EC_GROUP_set_curve_name`]
+    ///
+    /// [`EC_GROUP_set_curve_name`]: https://www.openssl.org/docs/man1.1.0/man3/EC_GROUP_set_curve_name.html
+    pub fn set_curve_name(&self, nid: Option<Nid>) {
```
sfackler: Why is https://docs.rs/openssl/0.10.19/openssl/ec/struct.EcGroup.html#method.from_curve_name not sufficient?

benmmurphy: i included it for symmetry and it makes it easier to write the tests. but i think it most cases when you do call set_curve_name you would have loaded the named curve using from_curve_name and done an equality check. so you could just substitute the named curve instead of using set_curve_name. c code i've seen that wants to mark named curves loaded from pem files which don't include the name typically use set_curve_name. it probably makes memory ownership simpler.

sfackler: Minimally this can't take `&self`.

What happens when you set the name of the curve to something other than the actual curve? Is that okay? Is it UB?

benmmurphy: i guess it depends on what you mean by UB. some of the functions use get_curve_name to check if two curves are equivalent. for example EC_GROUP_cmp will incorrectly return false if you use set_curve_name to change the name of a curve to something different. 

```
        /* compare the curve name (if present in both) */
        if (EC_GROUP_get_curve_name(a) && EC_GROUP_get_curve_name(b) &&
            EC_GROUP_get_curve_name(a) != EC_GROUP_get_curve_name(b))
                return 1;
```

Definitely UB in terms of the program is not going to work the way you expect. Not sure UB in terms of it might crash or corrupt memory. I suspect it is already possible to generate a null access violation using the EC_GROUP pem loading functions + EC_GROUP_cmp that this patch introduces. I originally was including functions for EC_GROUP_new_curve_GFp and EC_GROUP_new_curve_GF2m. But with these functions its possible to create groups that will cause a null pointer deference in EC_GROUP_cmp. I don't understand the pem loading enough to be sure that it can't create this dodgy groups that blow up EC_GROUP_cmp. The null pointer deference seem to be around the meth value being null.

sfackler: Unless there's a specific compelling use case, I'd prefer to avoid adding this.

Leo1003:
Would this pull request still be merged? I need `EC_GROUP_cmp()` in one of my repository.

mbelop:
I've looked at `EC_GROUP_cmp` and I can't really see why would be useful if `EcGroup` objects cannot be constructed from their components.  It's easier to pass `Nid` around and compare those.

PR #1075, opened 2019-03-01
CMS: add encrypt, from_der

tgbit:
Provide `encrypt` and `from_der` for `CmsContentInfo`.
For the test i was not sure which certs are maybe appropriate, so i added two additional certs.


sfackler:
Thanks!

PR #1078, opened 2019-03-11
CMS: add verify method

agalakhov:


sfackler:
This is also implemented in #1034, which exposes more functionality.

agalakhov:
This was in fact based on #1034. I suggest adding separate `verify()` and `verify_detached()` methods in order to provide more compile-time checks. The #1034 implementation has a disadvantage that many `Some`/`None` combinations make no sense and result in runtime error, so having separate functions would be fine.

I could also add `verify_detached()` and `sign_detached()` for symmetry. And also probably detached versions of `encrypt()` and `decrypt()`.

openssl/src/cms.rs
```diff
@@ -96,6 +97,43 @@ impl CmsContentInfoRef {
...
+    /// Verify the sender's signature given an optional sender's certificate `cert` and CA store
+    /// `store`. If the signature is correct, signed data are returned, otherwise `None`.
+    ///
+    /// OpenSSL documentation at [`CMS_verify`]
+    /// 
+    /// [`CMS_verify`]: https://www.openssl.org/docs/manmaster/man3/CMS_verify.html
+    pub fn verify(
+        &self,
+        certs: Option<&StackRef<X509>>,
+        store: &X509Store,
```
sfackler: `&X509StoreRef`

agalakhov: Thank you. Will be corrected.

sfackler:
Keeping the detached APIs separate seems reasonable to me.

PR #1084, opened 2019-03-17
Add the ability to get Nid from MessageDigest and EcGroupRef

npmccallum:


openssl/src/hash.rs
```diff
@@ -106,6 +106,11 @@ impl MessageDigest {
     pub fn size(&self) -> usize {
         unsafe { ffi::EVP_MD_size(self.0) as usize }
     }
+
+    /// The name of the digest
+    pub fn nid(&self) -> Nid {
```
sfackler: Can we call this `type_` to match with the openssl function name?

npmccallum: OpenSSL is generally inconsistent on their "give me the NID for this type" functions. I'd prefer that rust-openssl fixes this problem, especially given that their function names collide with Rust keywords.

npmccallum:
Hrm... The build failure doesn't look related to my patches.

npmccallum:
@sfackler Any update?

sfackler:
Yeah, don't worry about that flake.

openssl/src/hash.rs
sfackler: Our general policy is to follow the openssl names pretty directly.

npmccallum:
@sfackler I updated the function names to match OpenSSL names as requested.

sfackler:
Thanks!

npmccallum:
@sfackler Do you have an ETA on a next release? I need these patches in my project, but building against git breaks my dependency tree.

sfackler:
Done! You can use a Cargo patch to depend on a git dependency in a way that doesn't break the dependency graph for future reference: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#working-with-an-unpublished-minor-version

PR #1088, opened 2019-03-22
CMS: add verify methods

agalakhov:


openssl/src/cms.rs
```diff
@@ -236,4 +313,61 @@ mod test {
...
+    #[test]
+    fn cms_sign_verify_detached() {
+        // load cert with private key
+        let priv_cert_bytes = include_bytes!("../test/cms.p12");
+        let priv_cert = Pkcs12::from_der(priv_cert_bytes).expect("failed to load priv cert");
+        let priv_cert = priv_cert.parse("mypass").expect("failed to parse priv cert");
+
+        // sign cms message using private key cert
+        let input = String::from("My Message");
+        let sign = CmsContentInfo::sign(Some(&priv_cert.cert), Some(&priv_cert.pkey), None, Some(&input.as_bytes()), CMSOptions::DETACHED)
```
sfackler: It seems like the sign API should match the verify API one way or another. Either separate methods for attached and detached signatures, or one combined for both.

agalakhov: Does it make sense to make `sign()` function which just adds extra flag? Unlike `verify()`, detached `sign()` needs exactly the same input and has exactly the same output type.

Regarding the function signature of `sign()`, the first and the last parameters are always `Some`, so one probably should change them to just values instead of `Option`. This would be a 
breaking change.

BTW, the AppVeyor Windows test seems to fail at the initialization phase before actual checks, isn't it? Or does this PR has some issues with tests?

sfackler: > Regarding the function signature of `sign()`, the first and the last parameters are always `Some`, so one probably should change them to just values instead of `Option`. This would be a
> breaking change.

That doesn't seem to be the case:

> Any or all of these parameters can be NULL, see NOTES below.

> If signcert and pkey are NULL then a certificates only CMS structure is output.

https://www.openssl.org/docs/man1.1.0/man3/CMS_sign.html

> BTW, the AppVeyor Windows test seems to fail at the initialization phase before actual checks, isn't it? Or does this PR has some issues with tests?

That's not caused by this PR - vcpkg's openssl build is broken for some reason.

agalakhov: The certificates only CMS structure seems to be a special case which is not the expected output of normal usage of `sign()`. I believe it should be a separate function. This mode requires both `signcert` and `pkey` to be NULL at the same time, so it's more like an `Option<(signcert, pkey)>` than `Option<signcert>, Option<pkey>`. (In fact, I'd rather prefer `CmsContentInfo` to be an enum, but this seems to be not doable with openssl).

application-developer-DA: Are there any actions planned on this pull request?

sfackler: Just so I understand, we've gone from "the first and last parameters are always Some" to "well, they're only None in a special case", right?

My preference is generally to expose OpenSSL APIs as directly as is possible because any single person is probably not going to understand the totality of all possible usage of them.

agalakhov:
Does this mean there should be another layer of abstraction on top of openssl crate?

openssl/src/cms.rs
agalakhov: Does this mean there should be another layer of abstraction on top of openssl crate?
--

application-developer-DA:
I updated the pull request with an additional commit which also adds a couple of things to the CMS. I've just noticed that [some of the Circle tests](https://github.com/sfackler/rust-openssl/runs/104029642) are failed. It has something to do with `systest` and I'm not quite sure what changes are expected there (I tried to create `stack_st_CMS_RecipientInfo` the same way other related things are created.

amousset:
I dug a bit into these failures as I'd like to be able to use the verify method. The tests fail with `error: invalid application of '...' to incomplete type 'struct stack_st_CMS_RecipientInfo'`, only when compiled with openssl 1.0. It happens because the type for `CMS_RecipientInfo` stack is not declared in `cms.h` (it was moved in openssl 1.1.0-pre2: https://github.com/openssl/openssl/commit/4a640fb6c395302d7aad22833d573c44211b0a2d#diff-47b8520df0a2f8fa0433ff37611596f3L68), so the struct is not found in `systest`.

I can try to work on a fix, but I don't know what is the proper way to handle this.

application-developer-DA:
@amousset thanks for the info, I did not know that. Yes, I also saw the error but was a bit confused about it. I would also fix it, but I'm also not sure how to handle this properly. I just used the [docs](https://www.openssl.org/docs/man1.0.2/man3/CMS_RecipientInfo_ktri_cert_cmp.html) in order to understand what types do we require.

amousset:
Test pass with https://github.com/amousset/rust-openssl/commit/7162fa9656b23f67546b5992375fd2a2832c5111 which disables recipient info with openssl 1.0.

application-developer-DA:
Good! I've shortly checked it, added some minor comments / suggestions.

application-developer-DA:
@amousset I've just cherry-picked your commits into the branch. Hopefully it will be merged in upstream soon.

amousset:
Openssl 1.0 tests are now passing, but Rust 1.24.1 test (minimal-version) fails.

openssl/src/cms.rs
```diff
@@ -202,11 +281,161 @@ impl CmsContentInfo {
...
+    pub struct CmsRecipientInfo(*mut ffi::CMS_RecipientInfo);
+
+
+    impl ForeignType for CmsRecipientInfo {
+        type CType = ffi::CMS_RecipientInfo;
+        type Ref = CmsRecipientInfoRef;
+
+        #[inline]
+        unsafe fn from_ptr(ptr: *mut Self::CType) -> Self {
+            Self(ptr)
```
amousset:
```suggestion
            CmsRecipientInfo(ptr)
```

amousset: For 1.24.1 compatibility.

application-developer-DA:
@sfackler are there plans to merge this pull request? I'm just thinking of adding some additional things to this crate, but they rely on some other things, introduced in this pull request.

amousset:
@sfackler Would you prefer something like: https://github.com/sfackler/rust-openssl/compare/master...amousset:cms-verify-simple?expand=1 (simple verify function close to openssl API, very similar to what is currently done for pkcs7)?

amousset:
Is there something I could do to help this PR getting reviewed/merged?

application-developer-DA:
We actually use these functions in production and it seems to work fine, but I did not get any feedback since May 31.

openssl/src/cms.rs
```diff
@@ -202,11 +281,161 @@ impl CmsContentInfo {
...
+                let recipient_infos = cvt_p(ffi::CMS_get0_RecipientInfos(self.as_ptr()))?;
+                Ok(StackRef::from_ptr(recipient_infos))
+            }
+        }
+    }
+
+    /// An owned version of [`CmsRecipientInfoRef`]
+    ///
+    /// [`CmsRecipientInfoRef`]:struct.CmsRecipientInfoRef.html
+    pub struct CmsRecipientInfo(*mut ffi::CMS_RecipientInfo);
```
sfackler: These can be defined by the macro like we do for other types.

openssl/src/cms.rs
```diff
@@ -202,11 +281,161 @@ impl CmsContentInfo {
...
+                cvt(ffi::CMS_RecipientInfo_ktri_get0_signer_id(
+                    self.as_ptr(),
+                    &mut key_id,
+                    &mut issuer,
+                    &mut serial_number,
+                ))?;
+
+                if !key_id.is_null() {
+                    // TODO: Implement (ffi::ASN1_OCTET_STRING) bindings.
+                    Ok(RecipientInfo::Identifier)
```
sfackler: We should at least define the Asn1OctetString type so the enum can hold it.

openssl/src/cms.rs
```diff
@@ -202,11 +281,161 @@ impl CmsContentInfo {
...
+            issuer: &'cms X509NameRef,
+            serial_number: &'cms Asn1IntegerRef,
+        },
+    }
+
+    impl CmsRecipientInfoRef {
+        /// Implementation of [`CMS_RecipientInfo_ktri_get0_signer_id`].
+        ///
+        /// [`CMS_RecipientInfo_ktri_get0_signer_id`]: https://www.openssl.org/docs/manmaster/man3/CMS_RecipientInfo_decrypt.html
+        pub fn get_recipient_info<'cms>(&self) -> Result<RecipientInfo<'cms>, ErrorStack> {
```
sfackler: `'cms` isn't bound to an input here, so it's basically equivalent to `'static`. You'll need to do `&'cms self`.

application-developer-DA: You're right, fixed.

openssl/src/cms.rs
application-developer-DA: Well, there is some opaque type defined in `ossl_typ.rs`

```
openssl-sys/src/ossl_typ.rs
13:pub enum ASN1_OCTET_STRING {}
```

Not sure if I should return it as it does not hold a pointer by itself. I may of course create some sort of a temporary duplicate which just holds a pointer, but I don't know if that would be ok though.

openssl/src/cms.rs
application-developer-DA: I assume we're talking about `foreign_type_and_impl_send_sync` macro, right? I considered its usage, however the problem is that it requires us to specify a drop function (https://github.com/sfackler/rust-openssl/blob/master/openssl/src/macros.rs#L137), and there is no dedicated drop function for `CmsRecipientInfo` as far as I know.

openssl/src/cms.rs
```diff
@@ -13,10 +13,19 @@ use bio::{MemBio, MemBioSlice};
 use error::ErrorStack;
 use libc::c_uint;
 use pkey::{HasPrivate, PKeyRef};
-use stack::StackRef;
+<<<<<<< HEAD
```
FelixSchwarz, line 16: That line seems to be a left-over conflict marker.

PR #1093, opened 2019-04-13
Add AES OFB mode and support AES with 192-bit key

iosmanthus:
- Add `aes-128-ofb/aes-192-ofb/aes-256-ofb`
- Support 192-bit key for AES, e.g. `aes-192-cbc, aes-192-ecb`
- Reduce repetition in `impl openssl::symm::Cipher` with macros

Signed-off-by: iosmanthus <myosmanthustree@gmail.com>

### DIFF
--- a/openssl/Cargo.toml
+++ b/openssl/Cargo.toml
@@ -24,6 +24,7 @@ cfg-if = "0.1"
 foreign-types = "0.3.1"
 lazy_static = "1"
 libc = "0.2"
+paste = "0.1.5"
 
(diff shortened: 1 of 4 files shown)


openssl/Cargo.toml
```diff
@@ -24,6 +24,7 @@ cfg-if = "0.1"
 foreign-types = "0.3.1"
 lazy_static = "1"
 libc = "0.2"
+paste = "0.1.5"
```
sfackler: This increases the minimum supported rust version, so we'll probably want to just avoid it for now.

sfackler:
Ping

PR #1094, opened 2019-04-14
Add EVP_Seal and EVP_Open

vojta7:


openssl/src/lib.rs
```diff
@@ -150,6 +150,8 @@ pub mod dsa;
 pub mod ec;
 pub mod ecdsa;
 pub mod error;
+#[cfg(ossl111)]
+pub mod evp;
```
sfackler: These APIs appear to be available back at to 1.0.2.

openssl/src/evp.rs
```diff
@@ -0,0 +1,281 @@
...
+            ffi::EVP_CIPHER_CTX_free(self.ctx);
+        }
+    }
+}
+
+impl EvpOpen {
+    /// Creates a new `EvpOpen`.
+    pub fn new(
+        t: Cipher,
+        priv_key: &PKey<Private>,
```
sfackler: Should take `&PKeyRef<T>` where `T: HasPrivate`.

openssl/src/evp.rs
```diff
@@ -0,0 +1,281 @@
...
+        iv: &mut [u8],
```
sfackler: Why are these mutable slices?

openssl/src/evp.rs
```diff
@@ -0,0 +1,281 @@
...
+            let mut outl = cmp::min(output.len(), c_int::max_value() as usize) as c_int;
+
+            cvt(ffi::EVP_OpenFinal(self.ctx, output.as_mut_ptr(), &mut outl))?;
+
+            Ok(outl as usize)
+        }
+    }
+
+    /// Returns block size of inner cipher.
+    pub fn bs(&self) -> usize {
```
sfackler: This doesn't seem necessary.

openssl/src/evp.rs
```diff
@@ -0,0 +1,281 @@
...
+            let mut outl = cmp::min(output.len(), c_int::max_value() as usize) as c_int;
+
+            cvt(ffi::EVP_SealFinal(self.ctx, output.as_mut_ptr(), &mut outl))?;
+
+            Ok(outl as usize)
+        }
+    }
+
+    /// Returns block size of inner cipher.
+    pub fn bs(&self) -> usize {
```
sfackler: This doesn't seem necessary.

openssl/src/evp.rs
```diff
@@ -0,0 +1,281 @@
...
+    /// The number of bytes written to `output` is returned. Note that this may
+    /// not be equal to the length of `input`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `output.len() < input.len() + block_size` where
+    /// `block_size` is the block size of the cipher (see `Cipher::block_size`),
+    /// or if `output.len() > c_int::max_value()`.
+    pub fn update(&mut self, input: &[u8], output: &mut [u8]) -> Result<usize, ErrorStack> {
+        unsafe {
```
sfackler: There seem to be some missing asserts here.

openssl/src/evp.rs
```diff
@@ -0,0 +1,281 @@
...
+            let mut outl = output.len() as c_int;
+            let inl = input.len() as c_int;
+            cvt(ffi::EVP_EncryptUpdate(
```
sfackler: Shouldn't this use EVP_SealUpdate?

openssl/src/evp.rs
```diff
@@ -0,0 +1,281 @@
...
+
+/// Represents a EVP_Open context.
+pub struct EvpOpen {
+    ctx: *mut ffi::EVP_CIPHER_CTX,
+    block_size: usize,
+}
+
+impl EvpSeal {
+    /// Creates a new `EvpSeal`.
+    pub fn new(t: Cipher, pub_keys: Vec<PKey<Public>>) -> Result<EvpSeal, ErrorStack> {
```
sfackler: This should take `&[PKey<T>]` where `T: HasPublic`.

openssl/src/evp.rs
```diff
@@ -0,0 +1,281 @@
...
+                ctx,
+                block_size: t.block_size(),
+                iv: iv_buffer,
+                ek: my_ek,
+            })
+        }
+    }
+
+    /// Return used initialization vector.
+    pub fn iv(&self) -> &Vec<u8> {
```
sfackler: This should return a `&[u8]`.

openssl/src/evp.rs
```diff
@@ -0,0 +1,281 @@
...
+        &self.iv
+    }
+
+    /// Return vector of keys encrypted by public key.
+    pub fn ek(&self) -> &Vec<Vec<u8>> {
```
sfackler: This should return a `&[Vec<u8>]` and have a better name than `ek`.

openssl/src/evp.rs
vojta7: In OpenSSL source code EVP_SealUpdate is macro, translating this function directly to EVP_EncryptUpdate. So as far I know I can't use it. [OpenSSL documentation](https://www.openssl.org/docs/man1.1.1/man3/EVP_SealInit.html) also states that EVP_SealUpdate have exactly the same properties as the EVP_EncryptUpdate.

vojta7:
I don't really understand why is it failing to build on this target:
TARGET=x86_64-pc-windows-gnu, BITS=64, MSYS2=1, OPENSSL_VERSION=1_0_2r

It seems to be wrongly using EVP_PKEY_bits(key: *const EVP_PKEY), but according to the conditional compilation, it should use EVP_PKEY_bits(key: *mut EVP_PKEY).

I am understanding something wrong?

openssl/src/evp.rs
sfackler: Ah ok cool.

sfackler:
That error is pretty strange - the 1.0.2 tests passed on Linux. I'll try to poke around over the weekend if you don't have a chance to track it down.

vojta7:
Sadly I won't have time up until Monday.

sfackler:
No worries! I'll see if I can repro locally.

sfackler:
My mingw install appears to be on fire, so I may not be able to help debug the test failure too much :(

sfackler:
Ah! The problem is that the argument here was only made const in 1.1.1b, but not in the earlier 1.1.1 releases.

vojta7:
Thank you for help and for review. Changing the required version to 1.1.1b did fix the problem.

Sadly it appears that there is some problem with CI tests on macos.

sfackler:
That's a weird failure. Shouldn't block this PR though!

PR #1097, opened 2019-04-23
Support LibreSSL 2.9.1

vishwin:
LibreSSL 2.9.1 added generic DTLS methods, among other things: libressl-portable/openbsd@ff8fdf5

### DIFF
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -199,6 +199,7 @@ See rust-openssl README for more information:
             (8, 1) => ('8', '1'),
             (8, _) => ('8', 'x'),
             (9, 0) => ('9', '0'),
+            (9, 1) => ('9', '1'),
             _ => version_error(),
         };
 
@@ -239,7 +240,7 @@ fn version_error() -> ! {
         "
 
 This crate is only compatible with OpenSSL 1.0.1 through 1.1.1, or LibreSSL 2.5
-through 2.9.0, but a different version of OpenSSL was found. The build is now aborting
+through 2.9.1, but a different version of OpenSSL was found. The build is now aborting
 due to this version mismatch.
 
 "
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -199,7 +199,7 @@ See rust-openssl README for more information:
             (8, 1) => ('8', '1'),
             (8, _) => ('8', 'x'),
             (9, 0) => ('9', '0'),
-            (9, 1) => ('9', '1'),
+            (9, _) => ('9', 'x'),
             _ => version_error(),
         };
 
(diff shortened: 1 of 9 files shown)


sfackler:
Build's failing.

vishwin:
Working on this has actually exposed certain items present since LibreSSL 2.7.1 that the cfgs logic don't currently account for. Hold tight…

sfackler:
Don't worry about the failing macOS builds - they're busted due to a rustup.rs issue that should be fixed tomorrow.

vishwin:
I figured the macOS builds failing was external, so I've disregarded them. Meanwhile, I'm not quite sure what's causing two of the LibreSSL 2.9.1 tests to panic. I will come back to this once all my FreeBSD ports finish building (from an ABI bump), as how we build rust itself (completely frozen crates) requires this crate (for cargo).

openssl-sys/build/main.rs
```diff
@@ -199,6 +199,7 @@ See rust-openssl README for more information:
             (8, 1) => ('8', '1'),
             (8, _) => ('8', 'x'),
             (9, 0) => ('9', '0'),
+            (9, 1) => ('9', '1'),
```
sfackler: The release notes say that 2.9.1 is the first stable release for the 2.9 series, so we can whitelist future 2.9.x versions.

pimeys:
I'm currently trying to get this crate to work with OpenBSD current and LibreSSL 2.9.1, but two tests fail:

```failures:

---- ssl::test::test_connect_with_srtp_ctx stdout ----
thread 'ssl::test::test_connect_with_srtp_ctx' panicked at 'called `Option::unwrap()` on a `None` value', src/libcore/option.rs:345:21

---- ssl::test::test_connect_with_srtp_ssl stdout ----
thread 'ssl::test::test_connect_with_srtp_ssl' panicked at 'called `Option::unwrap()` on a `None` value', src/libcore/option.rs:345:21


failures:
    ssl::test::test_connect_with_srtp_ctx
    ssl::test::test_connect_with_srtp_ssl```

Seems that LibreSSL is not compiled with SRTP support or there is something wrong with rust-openssl configuration on OpenBSD.

sfackler:
There doesn't appear to be anything obvious in the 2.9.1 release notes, so you may have to do some investigation. Maybe some struct layouts changed or something?

mbelop:
Cc @botovq. Theo, can you please take a look at this?

botovq:
@mbelop: Thanks for pointing this out to me.

This revealed a bug we introduced into LibreSSL while refactoring the
extensions parsing. Due to this bug, the server hello no longer includes
use_srtp, so negotiation of an srtp profile fails, and the tests panic
in this line on the client side:

	let srtp_profile = stream.ssl().selected_srtp_profile().unwrap();

This will be fixed in LibreSSL 2.9.2.

sfackler:
Great, thanks for tracking that down!

We'll just ignore those tests on 2.9.1 then.

PR #1101, opened 2019-04-29
Remove size check panic on Crypter::{update,finalize}()

npmccallum:
The size of the output is deterministic for the size of the input. This means
that the caller can know when a buffer that is less than the existing check is
sufficient. This is particularly common in key wrapping scenarios.
    
For example, if I want to wrap two 16-byte keys using AES-CBC, the resulting
ciphertext will always be 32 bytes. Thus, I create a 32-byte buffer and call
`Crypter::update()` twice; once for each key. The first call succeeds, but the
second call panics. This is true even though the buffer is exactly the size
needed. Likewise, if both `Crypter::update()` calls succeed, the
`Crypter::finalize()` function will also panic.
    
We should, therefore, let `EVP_CipherUpdate()` and `EVP_CipherFinal()` generate any
errors when the output buffer has insufficient size and not try and be smarter
than the functions we are wrapping.

### DIFF
--- a/openssl/src/symm.rs
+++ b/openssl/src/symm.rs
@@ -492,12 +492,9 @@ impl Crypter {
     ///
     /// # Panics
     ///
-    /// Panics if `output.len() < input.len() + block_size` where
-    /// `block_size` is the block size of the cipher (see `Cipher::block_size`),
-    /// or if `output.len() > c_int::max_value()`.
+    /// Panics if `output.len() > c_int::max_value()`.
     pub fn update(&mut self, input: &[u8], output: &mut [u8]) -> Result<usize, ErrorStack> {
         unsafe {
-            assert!(output.len() >= input.len() + self.block_size);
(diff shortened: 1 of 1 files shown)


sfackler:
Let's add tests making sure that passing undersized buffers to update and finalize return errors rather than making out of bounds writes.

npmccallum:
1. Writing a test for invalid writes is a bit difficult. We'd essentially have to setup a guard page to ensure a segfault on an invalid write.

2. Apparently, OpenSSL has the worst possible behavior (of course): it doesn't check the length of the output buffer at all... https://github.com/openssl/openssl/blob/33b40a1027bfa6c400f24938093e80579c37586c/crypto/evp/evp_enc.c#L564

3. It would be possible to keep a running tally of input bytes to predict the output size based on the block size. This would be more accurate, but would have some (minor) computational overhead.

I've opted for a different solution which fixes the problem for stream ciphers. I'll close this PR and open a new one.

PR #1102, opened 2019-04-29
Fix output size check for stream ciphers

npmccallum:
The previous output size check presumed a block cipher. Therefore, it
enforced an unnecessary extra byte in the case of stream ciphers. This
patch ensures that our size checks don't force the caller to
overallocate for stream ciphers.

### DIFF
--- a/openssl/src/symm.rs
+++ b/openssl/src/symm.rs
@@ -492,12 +492,16 @@ impl Crypter {
     ///
     /// # Panics
     ///
-    /// Panics if `output.len() < input.len() + block_size` where
-    /// `block_size` is the block size of the cipher (see `Cipher::block_size`),
-    /// or if `output.len() > c_int::max_value()`.
+    /// Panics for stream ciphers if `output.len() < input.len()`.
+    ///
+    /// Panics for block ciphers if `output.len() < input.len() + block_size`,
+    /// where `block_size` is the block size of the cipher (see `Cipher::block_size`).
+    ///
+    /// Panics if `output.len() > c_int::max_value()`.
     pub fn update(&mut self, input: &[u8], output: &mut [u8]) -> Result<usize, ErrorStack> {
         unsafe {
-            assert!(output.len() >= input.len() + self.block_size);
+            let block_size = if self.block_size > 1 { self.block_size } else { 0 };
+            assert!(output.len() >= input.len() + block_size);
             assert!(output.len() <= c_int::max_value() as usize);
             let mut outl = output.len() as c_int;
             let inl = input.len() as c_int;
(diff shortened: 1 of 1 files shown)


npmccallum:
This PR supersedes #1101.

openssl/src/symm.rs
```diff
@@ -492,12 +492,16 @@ impl Crypter {
...
+    ///
+    /// Panics for block ciphers if `output.len() < input.len() + block_size`,
+    /// where `block_size` is the block size of the cipher (see `Cipher::block_size`).
+    ///
+    /// Panics if `output.len() > c_int::max_value()`.
     pub fn update(&mut self, input: &[u8], output: &mut [u8]) -> Result<usize, ErrorStack> {
         unsafe {
-            assert!(output.len() >= input.len() + self.block_size);
+            let block_size = if self.block_size > 1 { self.block_size } else { 0 };
+            assert!(output.len() >= input.len() + block_size);
```
sfackler: I was thinking that this check could just be changed to `output.len() >= input.len() + block_size - 1)`, but it turns out that I thought that in the past and was apparently wrong: https://github.com/sfackler/rust-openssl/commit/b8712c5c510c047696367ecb28db1bfadc5885cd.

npmccallum: You might actually be correct. I can't find it now, but I seem to remember an upstream OpenSSL bug where it would write an extra byte by mistake.

npmccallum: https://www.openssl.org/docs/man1.0.2/man3/EVP_EncryptUpdate.html

That documents your assumption.

npmccallum: And here: https://www.openssl.org/docs/man1.1.1/man3/EVP_CipherUpdate.html

npmccallum: I figured it out. The difference is encryption versus decryption.

sfackler: Oh, interesting. So I guess we do need to special case stream ciphers for decryption minimally?

I should really split this type into separate encryptor and decryptor types anyway.

npmccallum: Yes, a special case will always be required:

> The parameters and restrictions are identical to the encryption operations except that if padding is enabled the decrypted data buffer out passed to EVP_DecryptUpdate() should have sufficient room for (inl + cipher_block_size) bytes unless the cipher block size is 1 in which case inl bytes is sufficient.

npmccallum: The special case is only required for decryption. Encryption can always use `inl + cipher_block_size - 1`.

npmccallum: So maybe we could merge this in the short term and get separate Encryptor and Decryptor types later?

sfackler: Oh yeah, I was just talking to myself there >_>

PR #1103, opened 2019-05-01
Switch from `rustc_version` to `autocfg`

alexcrichton:
This switches the `openssl-sys` crate from using `rustc_version` as a
crate to check the version of rustc to using `autocfg`. While
functionally the same this has a few advantages:

* The `autocfg` crate has fewer dependencies and compiles faster
* If the `semver` crate has the `serde` feature activated, turns out
  `openssl-sys` gets compiled quite late in the dependency graph which
  can push back further C compilations. This is due to the slower
  compilation time of `serde` itself.
* The `autocfg` crate I believe is a bit more robust in terms of being
  flexible with the output of rustc itself.

### DIFF
--- a/openssl-sys/Cargo.toml
+++ b/openssl-sys/Cargo.toml
@@ -21,7 +21,7 @@ libc = "0.2"
 cc = "1.0"
 openssl-src = { version = "111.0.1", optional = true }
 pkg-config = "0.3.9"
-rustc_version = "0.2"
+autocfg = "0.1.2"
 
 [target.'cfg(target_env = "msvc")'.build-dependencies]
 vcpkg = "0.2"
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -2,7 +2,7 @@ extern crate cc;
 #[cfg(feature = "vendored")]
 extern crate openssl_src;
 extern crate pkg_config;
-extern crate rustc_version;
+extern crate autocfg;
 #[cfg(target_env = "msvc")]
 extern crate vcpkg;
 
@@ -94,9 +94,9 @@ fn main() {
 }
 
 fn check_rustc_versions() {
-    let version = rustc_version::version().unwrap();
+    let cfg = autocfg::new();
 
-    if version >= rustc_version::Version::new(1, 31, 0) {
+    if cfg.probe_rustc_version(1, 31) {
         println!("cargo:rustc-cfg=const_fn");
     }
 }


sfackler:
Thanks!

elichai:
@alexcrichton Any chance this could somehow support also nightlies?  I'm using `1.31` but it's a nightly from October so it broke my builds.

I understand if you think that nightlies aren't backwards compatible anyway.

alexcrichton:
Hi I'm not sure if understand, can you clarify what isn't working?

elichai:
const fn isn't stable yet in my nightly version, so this crate doesn't compile and requests adding the `min_const_fn` feature.

alexcrichton:
@elichai can you provide an example error message of what you're seeing?

elichai:
```
error[E0658]: const fn is unstable (see issue #53555)                                                                                                                                                                                                                          
  --> /home/elichai2/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-sys-0.9.45/src/macros.rs:75:13                                                                                                                                                                    
   |                                                                                                                                                                                                                                                                           
75 |               pub const fn $name($($arg: $t),*) -> $ret $b                                                                                                                                                                                                                
   |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                
   |                                                                                                                                                                                                                                                                           
  ::: /home/elichai2/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-sys-0.9.45/src/err.rs:8:1                                                                                                                                                                         
   |                                                                                                                                                                                                                                                                           
8  | / const_fn! {                                                                                                                                                                                                                                                             
9  | |     pub const fn ERR_PACK(l: c_int, f: c_int, r: c_int) -> c_ulong {                                                                                                                                                                                                    
10 | |         ((l as c_ulong & 0x0FF) << 24) |                                                                                                                                                                                                                                
11 | |         ((f as c_ulong & 0xFFF) << 12) |                                                                                                                                                                                                                                
...  |                                                                                                                                                                                                                                                                         
25 | |     }                                                                                                                                                                                                                                                                   
26 | | }                                                                                                                                                                                                                                                                       
   | |_- in this macro invocation                                                                                                                                                                                                                                              
   |                                                                                                                                                                                                                                                                           
   = help: add #![feature(min_const_fn)] to the crate attributes to enable                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                               
error: aborting due to previous error                                                                                                                                                                                                                                          
```

alexcrichton:
I don't believe that's related to this PR

sfackler:
It could be if rustc_version understood how nightly versions order with respect to stable releases and autocfg doesn't. We only make ERR_PACK, etc const fns if we're using a new enough compiler.

PR #1104, opened 2019-05-01
Update LibreSSL version match to 2.9.1

farhaven:
OpenBSD 6.5 ships with LibreSSL 2.9.1

I've tested this very lightly by doing a few HTTPS requests with `reqwests`, which uses this crate for TLS.


sfackler:
See #1097

PR #1113, opened 2019-05-13
Add SSL_CTX_add_client_CA on OpenSSL

bossmc:
Currently this crate only exposes the `SSL_CTX_set_client_CA_list` function, which requires an owned stack of x509 names.  As far as I can tell, the only way to get such a stack is to call the `SSL_load_client_CA_file` function which reads files from the disk.

With this PR one can load a x509 from memory and add it the client list without needing to touch the disk (and there's already functions to add CAs, keys, certificate chains etc. from memory).

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -870,6 +870,23 @@ impl SslContextBuilder {
         }
     }
 
+    /// Add the provided CA certificate to the list sent by the server to the client when
+    /// requesting client-side TLS authentication.
+    ///
+    /// This corresponds to [`SSL_CTX_add_client_CA`].
+    ///
+    /// [`SSL_CTX_add_client_CA`]: https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_set_client_CA_list.html
+    #[cfg(not(libressl))]
+    pub fn add_client_ca(&mut self, cacert: &mut X509) -> Result<(), ErrorStack> {
+        unsafe {
+            cvt(ffi::SSL_CTX_add_client_CA(
+                self.as_ptr(),
+                cacert.as_ptr()
+            ))
+            .map(|_| ())
+        }
+    }
+
     /// Set the context identifier for sessions.
     ///
     /// This value identifies the server's session cache to clients, telling them when they're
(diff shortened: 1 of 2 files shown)


bossmc:
This function is missing on libressl (according to their docs) so I've `cfg`d it out.

openssl/src/ssl/mod.rs
```diff
@@ -870,6 +870,23 @@ impl SslContextBuilder {
...
     }
 
+    /// Add the provided CA certificate to the list sent by the server to the client when
+    /// requesting client-side TLS authentication.
+    ///
+    /// This corresponds to [`SSL_CTX_add_client_CA`].
+    ///
+    /// [`SSL_CTX_add_client_CA`]: https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_set_client_CA_list.html
+    #[cfg(not(libressl))]
+    pub fn add_client_ca(&mut self, cacert: &mut X509) -> Result<(), ErrorStack> {
```
sfackler: This should take a `&X509Ref`, not a `&mut X509`.

bossmc: I agree about the `X509Ref`, but I disagree about removing the `mut`.  The OpenSSL API does not mark that API as `const` so for safety I have to take the CA as `mut` (so it can become a `*mut X509` in the `ffi`).

sfackler: OpenSSL's APIs are extremely not const-correct, and in any case, C's notion of const is not a 1:1 match to Rust's notion of const.

bossmc: Agreed, but, as the API doesn't say `const` I can't guarantee that it won't modify the certificate... and rust/llvm/gep get very unhappy if you alias mutability.

sfackler: This entire library ignores the const annotations in the OpenSSL API in favor of actually inspecting what the implementation does, and what it's documented to do. There is nothing unique about this one method.

sfackler: Providing those guarantees is exactly the job of the author of safe bindings over a C APi.

bossmc: Coolio, I'll make the change to `&` then

sfackler:
Thanks!

PR #1116, opened 2019-05-18
Fix handling of session callbacks

### LINKED ISSUE #1115: BUG: new session callback missing
Hello,

I have code that make [the assertion mentioned in the title](https://github.com/sfackler/rust-openssl/blob/d861eb16dd127870a68c6355a78170495a468f93/openssl/src/ssl/callbacks.rs#L358) fail. The weird thing is that it only happens when I pass the callback directly as such

```rust
let session_store_0 = Arc::clone(session_store);
builder.set_new_session_callback(move |ssl, session| session_store_0.insert(ssl, session));
```

Alternatively, if I build the callback in another function, my application works fine

```rust
fn build_new_session_callback(
    session_store: &Arc<SessionStore>,
) -> impl Fn(&mut ssl::SslRef, ssl::SslSession) {
    let session_store = Arc::clone(session_store);
    move |ssl, session| session_store.insert(ssl, session)
}

builder.set_new_session_callback(build_new_session_callback(session_store));
```

I suspect that the issue is how `TypeId` interacts with closures. I added some tracing to see what `TypeId` is generated. This is the one in [`set_new_session_callback`](https://github.com/sfackler/rust-openssl/blob/d861eb16dd127870a68c6355a78170495a468f93/openssl/src/ssl/mod.rs#L1394-L1418):
```
[/Users/simon/Source/github.com/sfackler/rust-openssl/openssl/src/ssl/mod.rs:1416] std::any::TypeId::of::<F>() = TypeId {
    t: 3694631760449667781,
}
[/Users/simon/Source/github.com/sfackler/rust-openssl/openssl/src/ssl/mod.rs:1417] std::intrinsics::type_name::<F>() = "[closure@src/cert/context/builder.rs:70:42: 70:98 session_store_0:std::sync::Arc<cert::SessionStore>]"
```

This is the one in [`raw_new_session`](https://github.com/sfackler/rust-openssl/blob/d861eb16dd127870a68c6355a78170495a468f93/openssl/src/ssl/callbacks.rs#L347-L365):
```
[/Users/simon/Source/github.com/sfackler/rust-openssl/openssl/src/ssl/callbacks.rs:355] std::any::TypeId::of::<F>() = TypeId {
    t: 17026941024949866994,
}
[/Users/simon/Source/github.com/sfackler/rust-openssl/openssl/src/ssl/callbacks.rs:356] std::intrinsics::type_name::<F>() = "[closure@src/cert/context/builder.rs:70:42: 70:98 session_store_0:std::sync::Arc<cert::SessionStore>]"
```

When the closure is passed directly to `set_new_session_callback`, it may have some extra information that gets lost when it is passed as a type argument to `raw_new_session`? This would explain why wrapping the creation of the closure in a function like `build_new_session_callback` above work: the extra information exists only within the latter, and both `set_new_session_callback` and `raw_new_session` see the same type. Another thing that could confuse the types is that the call to `new_session_callback` is itself in a generic function. I thought it could be an issue with `set_new_session_callback` being instanciated twice because it uses the closure value, but `raw_session_id` incorrectly being instanciated only once because it only uses the closure type. Do closures created in two instanciations of a generic function, and that do not depend on the generic parameter have the same type?

I wrote this test case which seems it could trigger the bug, but it passes:
```
#[cfg(test)]
mod help {
    use std::fmt::Debug;
    use std::os::raw::c_int;

    use crate::ssl::{SslRef, SslSession};

    use super::*;

    pub unsafe extern "C" fn raw_new_session<F>() -> c_int
        where
            F: Fn(&SslRef, SslSession) + 'static + Sync + Send,
    {
        SslContext::cached_ex_index::<F>().as_raw()
    }

    #[inline(never)]
    pub fn set_new_session_callback<F>(f: F) -> (c_int, unsafe extern "C" fn() -> c_int)
        where
            F: Fn(&SslRef, SslSession) + 'static + Sync + Send,
    {
        let server = Server::builder().build();
        let stream = server.client().connect();
        let ssl = stream.ssl();
        let session = ssl.session().unwrap();
        f(ssl, session.to_owned());

        (SslContext::cached_ex_index::<F>().as_raw(), raw_new_session::<F>)
    }

    #[inline(never)]
    pub fn configure_context(x: impl Debug) -> (c_int, unsafe extern "C" fn() -> c_int) {
        dbg!(&x);
        let s = "hello".to_string();
        set_new_session_callback(move |_, _| { println!("{}", s.as_str()); })
    }

    #[test]
    fn help() {
        let (i, f) = configure_context(0u64);
        let (j, g) = configure_context(0usize);

        assert_eq!(i, unsafe { f() });
        assert_eq!(j, unsafe { g() });
    }
}
```

I don't know if the issue is with my code, `rust-openssl`, or even `TypeId`. The documentation on [`TypeId`](https://doc.rust-lang.org/std/any/struct.TypeId.html) does not give a lot of information on its meaning and I'm not knowledgeable enough about Rust to go further.

Thank you

sfackler:
The session context is used for session callbacks rather than the normal
context, which breaks state lookup when the context has been swapped out
(e.g. for SNI). Since there isn't an accessor for the session context,
we just store an extra reference in the SSL's ex data.

Closes #1115

### DIFF
--- a/openssl/src/ssl/callbacks.rs
+++ b/openssl/src/ssl/callbacks.rs
@@ -24,7 +24,7 @@ use pkey::Params;
 use ssl::AlpnError;
 #[cfg(ossl111)]
 use ssl::{ClientHelloResponse, ExtensionContext};
-use ssl::{SniError, Ssl, SslAlert, SslContext, SslContextRef, SslRef, SslSession, SslSessionRef};
+use ssl::{SniError, Ssl, SslAlert, SslContext, SslContextRef, SslRef, SslSession, SslSessionRef, SESSION_CTX_INDEX};
 #[cfg(ossl111)]
 use x509::X509Ref;
 use x509::{X509StoreContext, X509StoreContextRef};
@@ -353,7 +353,8 @@ where
 {
     let ssl = SslRef::from_ptr_mut(ssl);
     let callback = ssl
-        .ssl_context()
+        .ex_data(*SESSION_CTX_INDEX)
+        .expect("BUG: session context missing")
         .ex_data(SslContext::cached_ex_index::<F>())
         .expect("BUG: new session callback missing") as *const F;
     let session = SslSession::from_ptr(session);
@@ -398,7 +399,8 @@ where
 {
     let ssl = SslRef::from_ptr_mut(ssl);
     let callback = ssl
-        .ssl_context()
+        .ex_data(*SESSION_CTX_INDEX)
+        .expect("BUG: session context missing")
         .ex_data(SslContext::cached_ex_index::<F>())
         .expect("BUG: get session callback missing") as *const F;
     let data = slice::from_raw_parts(data as *const u8, len as usize);
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -486,6 +486,8 @@ impl NameType {
 lazy_static! {
     static ref INDEXES: Mutex<HashMap<TypeId, c_int>> = Mutex::new(HashMap::new());
     static ref SSL_INDEXES: Mutex<HashMap<TypeId, c_int>> = Mutex::new(HashMap::new());
+
+    static ref SESSION_CTX_INDEX: Index<Ssl, SslContext> = Ssl::new_ex_index().unwrap();
 }
 
 unsafe extern "C" fn free_data_box<T>(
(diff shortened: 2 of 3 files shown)


jethrogb:
So if I understand correctly, the value stored in the `SESSION_CTX_INDEX` is actually *not* the session's ctx, but rather the original ctx?

jethrogb:
Or maybe I'm misunderstanding the meaning of the word "session" in this case.

sfackler:
"session context" is the terminology OpenSSL uses - I think the idea is that session callbacks use the session context, which is the original context passed to SSL_new, regardless of any other calls to SSL_set_SSL_CONTEXT.

PR #1118, opened 2019-05-30
Display for Error incorrectly showing func instead of reason.

animalsiknow:
I found a small copy-paste error in the implementation of `Display` for `Error`.


sfackler:
Thanks!

Issue #1120, opened 2019-06-02
[Feature] Expose RFC 3394 (AES_wrap_key) functionality


russelltg:
It would be very useful to have access to AES_wrap_key. 

I'm willing to help implement this, but would need a bit of mentoring on how the rust API should be laid out.

sfackler:
I think it'd just look something like this in aes.rs:

```rust
pub fn wrap_key(key: &AesKey, iv: Option<[u8; 8]>, out: &mut [u8], in: &[u8]) -> Result<usize, ErrorStack> { ... }

pub fn unwrap_key(key: &AesKey, iv: Option<[u8; 8]>, out: &mut [u8], in: &[u8]) -> Result<usize, ErrorStack> { ... }
```

With some asserts about the sizes of out and in (in must be a multiple of 8 bytes, and out must be at least 8 bytes larger than in for wrap_key and at least 8 bytes smaller for unwrap_key.

PR #1122, opened 2019-06-06
Add AES_wrap_key and AES_unwrap_key functionality

russelltg:
This was discussed in #1120 

The only concerning thing while implementing this was that `AES_wrap_key` and `AES_unwrap_key` take mutable `AES_KEY` objects. I looked at the implementation to make sure that they only use it as a const, so it is safe to cast the `&AesKey` to an `AES_KEY*`. However, they could change that in the future. I doubt that happening though.

### DIFF
--- a/openssl/src/aes.rs
+++ b/openssl/src/aes.rs
@@ -1,7 +1,7 @@
-//! Low level AES IGE functionality
+//! Low level AES IGE and key wrapping functionality
 //!
 //! AES ECB, CBC, XTS, CTR, CFB, GCM and other conventional symmetric encryption
-//! modes are found in [`symm`].  This is the implementation of AES IGE.
+//! modes are found in [`symm`].  This is the implementation of AES IGE and key wrapping
 //!
 //! Advanced Encryption Standard (AES) provides symmetric key cipher that
 //! the same key is used to encrypt and decrypt data.  This implementation
@@ -22,6 +22,7 @@
 //!
 //! # Examples
 //!
+//! ## AES IGE
 //! ```rust
 //! use openssl::aes::{AesKey, aes_ige};
 //! use openssl::symm::Mode;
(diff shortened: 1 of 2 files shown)


sfackler:
Build is failing on the minimum rustc version: https://circleci.com/gh/sfackler/rust-openssl/14746?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

openssl/src/aes.rs
```diff
@@ -136,6 +158,70 @@ pub fn aes_ige(in_: &[u8], out: &mut [u8], key: &AesKey, iv: &mut [u8], mode: Mo
...
+/// Panics if either `out` or `in_` do not have sizes that are a multiple of 8, or if
+/// `out` is not 8 bytes longer than `in_`.
+pub fn wrap_key(
+    key: &AesKey,
+    iv: Option<[u8; 8]>,
+    out: &mut [u8],
+    in_: &[u8],
+) -> Result<usize, ErrorStack> {
+    unsafe {
+        assert!(out.len() == in_.len() + 8); // Ciphertext is 64 bits longer (see 2.2.1)
```
sfackler: Out doesn't need to be exactly this large, just at least this large.

openssl/src/aes.rs
```diff
@@ -136,6 +158,70 @@ pub fn aes_ige(in_: &[u8], out: &mut [u8], key: &AesKey, iv: &mut [u8], mode: Mo
...
+        assert!(in_.len() % 8 == 0); // Input (and hence output, given above) are integer multiples of 64 bit values
+
+        cvt(ffi::AES_wrap_key(
+            &key.0 as *const _ as *mut _, // this is safe, the implementation only uses the key as a const pointer.
```
sfackler: Yeah, openssl isn't really very const-correct so this kind of thing is common.

openssl/src/aes.rs
```diff
@@ -136,6 +158,70 @@ pub fn aes_ige(in_: &[u8], out: &mut [u8], key: &AesKey, iv: &mut [u8], mode: Mo
...
+/// # Panics
+///
+/// Panics if either `out` or `in_` do not have sizes that are a multiple of 8, or if
+/// `out` is not 8 bytes longer than `in_`.
+pub fn wrap_key(
+    key: &AesKey,
+    iv: Option<[u8; 8]>,
+    out: &mut [u8],
+    in_: &[u8],
+) -> Result<usize, ErrorStack> {
```
sfackler: The low level AES routines don't actually touch the error stack, so this should probably just return KeyError.

openssl/src/aes.rs
```diff
@@ -136,6 +158,70 @@ pub fn aes_ige(in_: &[u8], out: &mut [u8], key: &AesKey, iv: &mut [u8], mode: Mo
...
+    unsafe {
+        assert!(out.len() == in_.len() + 8); // Ciphertext is 64 bits longer (see 2.2.1)
+        assert!(in_.len() % 8 == 0); // Input (and hence output, given above) are integer multiples of 64 bit values
```
sfackler: OpenSSL checks this condition in the implementation so there's no need to assert.

openssl/src/aes.rs
```diff
@@ -136,6 +158,70 @@ pub fn aes_ige(in_: &[u8], out: &mut [u8], key: &AesKey, iv: &mut [u8], mode: Mo
...
+    out: &mut [u8],
+    in_: &[u8],
+) -> Result<usize, ErrorStack> {
+    unsafe {
+        assert!(in_.len() == out.len() + 8);
+        assert!(in_.len() % 8 == 0);
+
+        cvt(ffi::AES_unwrap_key(
+            &key.0 as *const _ as *mut _, // this is safe, the implementation only uses the key as a const pointer.
+            iv.map_or(std::ptr::null(), |iv| iv.as_ptr() as *const _),
```
sfackler: You'll want an .as_ref here so the pointer stays valid after returning from map_or.

sfackler:
Thanks!

PR #1123, opened 2019-06-07
Basic CRL Bindings

jrobsonchase:
Adds basic bindings to the API for CRL handling. For now, only support:

* Reading an existing CRL (both PEM and DER)
* Check if another X509 certificate is present in the list
* Get the last/next update times
* Get the list's issuer and
* Verify the signature on the list

I'm not sure how best to integrate the CRL into the overall verification story. There are some callbacks that could be exposed on the `X509_STORE` relating to getting/checking the CRL for a particular cert, but I'm not sure if those would be much better than simply fetching and/or checking the CRL in the `verify_callback` that's *already* exposed. At any rate, what we have here would be sufficient for at least that usecase.


mbr:
There's also the questions of which of those `assert!(!X.is_null())` can be left in there (instead of returning an `Option`). Some do rely on OpenSSL parsing the incoming structures strictly to spec and rejecting any that are missing the given fields.

Otherwise, a rogue CRL might cause a panic in an application by leaving these fields blank.

openssl/src/x509/mod.rs
```diff
@@ -1170,6 +1170,136 @@ impl X509ReqRef {
...
+            if date.is_null() {
+                None
+            } else {
+                Some(Asn1TimeRef::from_ptr(date as *mut _))
+            }
+        }
+    }
+
+    /// Check if the provided certificate is in the revocation list.
+    pub fn is_revoked(&self, cert: &X509Ref) -> bool {
```
sfackler: I'd prefer exposing the underlying API more directly - taking a serial in directly (or using X509_CRL_get0_by_cert instead) and returning the X509_REVOKED object in some kind of enum that indicates the removeFromCRL case.

sfackler:
Yeah, it's definitely worth confirming that elements we expect to be present are actually always present.

openssl/src/x509/mod.rs
```diff
@@ -1396,3 +1526,24 @@ cfg_if! {
...
+        #[allow(bad_style)]
+        unsafe fn X509_CRL_get0_nextUpdate(x: *const ffi::X509_CRL) -> *mut ffi::ASN1_TIME {
+            (*(*x).crl).nextUpdate
+        }
+        #[allow(bad_style)]
+        unsafe fn X509_CRL_get_issuer(x: *const ffi::X509_CRL) -> *mut ffi::X509_NAME {
+            (*(*x).crl).issuer
+        }
+    }
+}
```
sfackler: We'll definitely want some basic tests interacting with a CRL.

openssl/src/x509/mod.rs
jrobsonchase: OK, I can expose both of those variants instead :+1:

jrobsonchase:
Alright, lower-level `X509Revoked` API exposed and a basic test for CRL loading added.

Regarding the expected/optional elements - the `nextUpdate` time is the only field on a CRL that I've been able to determine to be optional per the spec - everything else should always be present.

openssl/src/x509/mod.rs
```diff
@@ -1170,6 +1170,226 @@ impl X509ReqRef {
...
+        ///
+        /// This corresponds to [`i2d_X509_CRL`].
+        ///
+        /// [`i2d_X509_CRL`]: https://www.openssl.org/docs/man1.0.2/crypto/i2d_X509_REQ.html
+        to_der,
+        ffi::i2d_X509_CRL
+    }
+
+    /// Get the stack of revocation entries
+    pub fn get_revoked(&self) -> Option<Stack<X509Revoked>> {
```
sfackler: `X509_CRL_get_REVOKED` returns a borrowed stack, not an owned one: https://github.com/openssl/openssl/blob/3e4b43b9e5aba6261b1ceb832c597ce8225782ef/crypto/x509/x509cset.c#L116

openssl/src/x509/mod.rs
```diff
@@ -1170,6 +1170,226 @@ impl X509ReqRef {
...
+            }
+        }
+    }
+
+    /// Get the revocation status of a certificate by its serial number
+    ///
+    /// This corresponds to [`X509_CRL_get0_by_serial`]
+    ///
+    /// [`X509_CRL_get0_by_serial`]: https://www.openssl.org/docs/man1.1.0/crypto/X509_CRL_get0_by_serial.html
+    pub fn get_by_serial(&self, serial: &Asn1IntegerRef) -> Option<&X509RevokedRef> {
```
sfackler: get_by_serial and get_by_cert still need to check the return code of the OpenSSL function and expose the removeFromCRL state.

openssl/src/x509/mod.rs
jrobsonchase: Ah, ok. This is my first time dealing with the openssl stack api, and its ownership semantics aren't immediately apparent. Thanks for setting me straight :)

openssl/src/x509/mod.rs
jrobsonchase: OK, this is now exposed via the `CrlStatus` enum.

jrobsonchase:
@sfackler Anything else this needs before it can go in?

pedrotbtl:
Any updates on when this will go in? Looking forward to it!

sven-hm:
It would be really helpful if those basic CRL bindings get merged! Is there anything left to do apart from the tiny merge conflict?

wesleywiser:
I'm also very interested to see these bindings added! It looks to me like all of the review feedback has been addressed and this maybe just needs to be rebased to remove the merge commit? Rebasing locally against `master` builds and tests pass (I'm happy to push the rebased commits somewhere if that's helpful as well).

openssl/src/x509/mod.rs
```diff
@@ -1586,3 +1841,38 @@ cfg_if! {
         }
     }
 }
+
+cfg_if! {
+    if #[cfg(ossl110)] {
```
Skepfyr, line 1846: This should be `#[cfg(any(ossl110, libressl350))]` to match openssl-sys.

Skepfyr (changes requested):
Hi! I realise it's been a while, but I think I can get this merged soon! I want this and I'm happy to make the changes needed to get this up-to-date, but I wanted to give you the opportunity if you'd prefer?

Skepfyr:
@sfackler There's a lot less code here than I thought there was, I've reviewed and I'm happy with it all (barring the comments above). All of your comments above appear to have been addressed, the only thing I'm on the fence about is it's slightly light on testing, but the functions I was most worried about have been called in the test.

sfackler:
Yep, I think this should be good with your feedback addressed

Skepfyr:
@sfackler That's everything reviewed and fixed up and the pipeline has passed. Any reason not to hit merge?

PR #1124, opened 2019-06-10
Add Debug impl for EcKey

Leo1003:
This is a small PR to implement Debug trait for EcKey. 
There exists Debug impl. for Rsa and Dsa, but not EcKey. It is inconvenient to derive Debug for structs containing EcKey. So I impl Debug trait for EcKey like Rsa and Dsa does.


sfackler:
Thanks!

Issue #1126, opened 2019-06-11
Implement missing bindings for `CmsContentInfo`


application-developer-DA:
- [ ] Get Recipient Information (EnvelopedData)
- [ ] Get Signer Information (SignedData)
- [ ] Convert to/from PEM
- [ ] Clone CmsContentInfo

dhouck:
It looks like we got `CmsContentInfo::verify` through a PR that is not the one mentioned here, so perhaps that should be checked off and the other closed.

Is there any work done on the `get_signers` method?  Currently I can check that *someone* signed something, but not *who*, which is not usually particularly useful.

PR #1127, opened 2019-06-12
Add `to_pem()` and `from_pem()` for `CmsContentInfo`

application-developer-DA:
Part of the work done for https://github.com/sfackler/rust-openssl/issues/1126

### DIFF
--- a/openssl/src/cms.rs
+++ b/openssl/src/cms.rs
@@ -248,13 +248,23 @@ mod test {
 
         let encrypt = CmsContentInfo::encrypt(&cert_stack, &input.as_bytes(), Cipher::des_ede3_cbc(), CMSOptions::empty())
             .expect("failed create encrypted cms");
-        let encrypt = encrypt.to_der().expect("failed to create der from cms");
+        let encrypted_der = encrypt.to_der().expect("failed to create der from cms");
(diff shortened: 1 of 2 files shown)


openssl/src/cms.rs
```diff
@@ -97,13 +97,26 @@ impl CmsContentInfoRef {
...
+        ///
+        /// OpenSSL documentation at [`i2d_CMS_ContentInfo`]
+        ///
+        /// [`i2d_CMS_ContentInfo`]: https://www.openssl.org/docs/man1.0.2/crypto/i2d_CMS_ContentInfo.html
+        to_der,
+        ffi::i2d_CMS_ContentInfo
+    }
+}
+
+#[cfg(ossl101)]
```
sfackler: The entire cms module is already `#[cfg(not(libressl))]` so you don't need the extra cfg here.

application-developer-DA: Oh, you're right, I did not notice that.

application-developer-DA: Updated. Now the AppVeyor build fails, but apparently it has nothing to do with the code...

sfackler:
Thanks!

PR #1130, opened 2019-06-14
Expose EC_GROUP_get_cofactor as EcGroup::cofactor

mbelop:
It would be nice to expose the proper way to obtain the group cofactor to avoid hardcoding the values, even despite the fact that the cofactor is 1 for most popular curves used with ECDSA.


sfackler:
Thanks!

PR #1131, opened 2019-06-19
Expose EC_POINT_dup as EcPoint::to_owned

mbelop:
Currently there's no easy way to convert `EcPointRef` back to the `EcPoint`.  For example, if `EcKey::public_key()` is called, the only way to get owned `EcPoint` is via serialization/deserialization.

This change leverages existing `EC_POINT_dup` function for straightforward conversion in a similar fashion to `BN_dup` and `BigNum::to_owned`.


sfackler:
Thanks!

PR #1132, opened 2019-06-21
Add `get_signers()` to `SignedData` CMS

application-developer-DA:
Another one item for https://github.com/sfackler/rust-openssl/issues/1126


application-developer-DA:
NB: Perhaps it makes sense to merge https://github.com/sfackler/rust-openssl/pull/1088 as `CMS_get0_signers()` should ideally be called after `verify()`.

PR #1140, opened 2019-07-07
Expose AES-192 and OFB mode

Leo1003:
Since AES 192 is also common in symmetry cipher, but this crate doesn't make it exposed...

This pull request add these functionality:
- AES 192 bits
- OFB mode for AES 128/192/256

Also add some tests related to them


sfackler:
Thanks!

PR #1144, opened 2019-07-26
Add Asn1TimeRef::duration_since and Asn1Time::now

jonas-schievink:
This allows accessing the actual time without having to parse the `Display` output.

### DIFF
--- a/openssl/src/asn1.rs
+++ b/openssl/src/asn1.rs
@@ -32,7 +32,6 @@ use std::fmt;
 use std::ptr;
 use std::slice;
 use std::str;
-use std::time::Duration;
 
 use bio::MemBio;
 use bn::{BigNum, BigNumRef};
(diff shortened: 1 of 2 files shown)


openssl/src/asn1.rs
```diff
@@ -165,6 +171,28 @@ impl Asn1Time {
...
+    pub fn duration_since(&self, older: &Self) -> Option<Duration> {
+        unsafe {
+            let mut days = 0;
+            let mut secs = 0;
+            cvt(ffi::ASN1_TIME_diff(&mut days, &mut secs, older.as_ptr(), self.as_ptr())).ok()?;
+
+            if days < 0 || secs < 0 {
+                None
+            } else {
+                secs = secs.checked_add(days.checked_mul(24 * 60 * 60)?)?;
```
sfackler: It seems wrong that this assumes that all days are the same length.

jonas-schievink: Yes, however `days_from_now` does the same, and OpenSSL also seems to use [this `SECS_PER_DAY`](https://github.com/openssl/openssl/blob/0e9725bcb90770d967351b977407b174bbd91869/crypto/o_time.c#L65) constant for the conversion

sfackler: Generally I like to provide a fairly direct binding to the underlying APIs - in this case that'd be returning a struct of the days and seconds. The days_from_now constructor is a bit of a holdover from an earlier time that I should probably clean up.

jonas-schievink:
I no longer need this API, but if anyone else is interested feel free to revive this!

PR #1152, opened 2019-09-06
Free r and s on ECDSA_SIG before overwriting them in ECDSA_SIG_set0

thomaswhiteway:
This fixes a memory leak when using `EcsdaSig::from_private_components` when this version of `ECDSA_SIG_set0` is used.  The fix makes that function consistent with the implementation in openssl itself: https://github.com/openssl/openssl/blob/master/crypto/ec/ec_asn1.c#L1251.

I've checked that using a test program calling `EcdsaSig::from_private_components` in a tight loop,
- without this change memory usage increases rapidly,
- with this change memory usage is flat.


sfackler:
Thanks!

PR #1153, opened 2019-09-07
Added support for Ed25519 and Ed448 signatures

sturmsebastian:
Added constants for Ed25519 and Ed448 as well as oneshot sign and verify function to `Signer` and `Verfier` to support PureEdDSA signatures.

### DIFF
--- a/openssl/src/sign.rs
+++ b/openssl/src/sign.rs
@@ -126,6 +126,9 @@ impl<'a> Drop for Signer<'a> {
 impl<'a> Signer<'a> {
     /// Creates a new `Signer`.
     ///
+    /// This cannot be used with Ed25519 or Ed448 keys. Please refer to
+    /// `new_without_digest`.
+    ///
     /// OpenSSL documentation at [`EVP_DigestSignInit`].
     ///
     /// [`EVP_DigestSignInit`]: https://www.openssl.org/docs/manmaster/man3/EVP_DigestSignInit.html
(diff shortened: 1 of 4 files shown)


sturmsebastian:
All Test fails are suspicious.

- mimimal-version failed compiling lazy_static with: <code>error[E0432]: unresolved import `self::std::hint`</code>
- i686-vendored failed downloading cc crate
- MSYS2 failed with <code>undefined reference to `__imp___acrt_iob_func'</code>

sfackler:
Yeah, don't worry about those CI failures.

openssl/src/sign.rs
```diff
@@ -283,6 +292,21 @@ impl<'a> Signer<'a> {
         }
     }
 
+    #[cfg(ossl111)]
+    pub fn len(&self) -> Result<usize, ErrorStack> {
```
sfackler: This method won't have its documentation anymore when building against 1.1.1.

Why does 1.1.1 need to call EVP_DigestSign and all other versions need to call DigestSignFinal?

sturmsebastian: When used with EdDSA keys, `EVP_DigestSignFinal` fails with a segmentation fault
(even in `Signer::len()`). `EVP_DigestSign` supports all key types, but has only been available since 1.1.1.

sfackler:
Thanks!

Issue #1155, opened 2019-09-15
DTLS MTU does not properly handled


HyeonuPark:
I'm using `openssl` crate for DTLS-SRTP key exchange mechanism, and I found that it uses 288 bytes as the maximum outbound packet size. As a result, handshake packets are fragmented which may increase the chance of handshake loss.

DTLS MTU handling is done via the callback function passed from `BIO_meth_set_ctrl` function, with `BIO_CTRL_DGRAM_*` family of commands. It seems that rust-openssl crate just ignores these commands currently.

Do you have some plan to handle it? If you don't have one yet, I can suggest three ways to implement this feature in `openssl` crate.

1. Implement MTU discovery feature

    It seems like the most ideal way IMO, though it can be hard. I can't say how hard it would be as I don't investigate this way that much, but at least it seems [openssl has some code for it](https://github.com/openssl/openssl/blob/8c47e55ee69500e31e80458682c6e022294cd0be/crypto/bio/bss_dgram.c#L454).

1. Allow users to configure MTU size

    Simply let users configure MTU size. Based on my small experiment it's just enough to respond to `BIO_CTRL_DGRAM_QUERY_MTU` command, so just add another if arm to the [ctrl function](https://github.com/sfackler/rust-openssl/blob/4c1891cc1085282acd8fa9820d81fb84d87b6e4e/openssl/src/ssl/bio.rs#L137), and blindly returns user-provided value, fallback to zero or some default value if not configured.

1. Always returns google-selected default MTU size

    This is the simplest one: always return `1200` for it. [Google uses it as default MTU size for WebRTC](https://groups.google.com/forum/#!topic/discuss-webrtc/gH5ysR3SoZI) so it should be OK for most use cases.

sfackler:
I think option 2 seems cleanest.

TimonPost:
I have a question around this topic, so I am currently implementing [DTLSv1_listen](https://github.com/TimonPost/udp-dtls/issues/2), although, I keep getting the error 

`ErrorStack([Error { code: 336978321, library: "SSL routines", function: "DTLSv1_listen", reason: "fragmented client hello", file: "../ssl/d1_lib.c", line: 623 }])`.

I am searching for solutions, and some say I have to set the MTU size, so I tried this PR, but it does not work. 

First, they said I could use: 

```
SSL_CTX_set_options(ctx, SSL_OP_NO_QUERY_MTU); 
```
and 
```
[...]

PR #1160, opened 2019-09-22
Add ssl::set_dtls_mtu_size(usize)

HyeonuPark:
This PR is a followup from [this issue](https://github.com/sfackler/rust-openssl/issues/1155)

`ssl::set_dtls_mtu_size(usize)` globally set the DTLS MTU size.

I first tried to make it a method on `Ssl` or `SslRef`, but I found that they can't be referenced easily from the current structure. Method on `SslStream` was not an option, as user can't access on it before the handshake completes.

But is it practical to use different MTU sized per connection? From my point of view, it's common to have globally configured DTLS MTU size for applications.

# Unresolved questions

- Maybe it's better to make it `Ssl`-local config?

- Should we remain 0 as a default value? I don't think it's sane default for DTLS, but increasing it should considered breaking change?

### DIFF
--- a/openssl/src/ssl/bio.rs
+++ b/openssl/src/ssl/bio.rs
@@ -1,6 +1,6 @@
 use ffi::{
     self, BIO_clear_retry_flags, BIO_new, BIO_set_retry_read, BIO_set_retry_write, BIO,
-    BIO_CTRL_FLUSH,
+    BIO_CTRL_FLUSH, BIO_CTRL_DGRAM_QUERY_MTU,
 };
 use libc::{c_char, c_int, c_long, c_void, strlen};
 use std::any::Any;
@@ -18,6 +18,7 @@ pub struct StreamState<S> {
     pub stream: S,
     pub error: Option<io::Error>,
     pub panic: Option<Box<dyn Any + Send>>,
+    pub dtls_mtu_size: c_long,
 }
 
 /// Safe wrapper for BIO_METHOD
@@ -39,6 +40,7 @@ pub fn new<S: Read + Write>(stream: S) -> Result<(*mut BIO, BioMethod), ErrorSta
         stream: stream,
         error: None,
         panic: None,
+        dtls_mtu_size: 0,
     });
 
     unsafe {
@@ -67,6 +69,13 @@ pub unsafe fn get_ref<'a, S: 'a>(bio: *mut BIO) -> &'a S {
 
 pub unsafe fn get_mut<'a, S: 'a>(bio: *mut BIO) -> &'a mut S {
     &mut state(bio).stream
+}
+
+pub unsafe fn set_dtls_mtu_size<S>(bio: *mut BIO, mtu_size: usize) {
+    if mtu_size as u64 > c_long::max_value() as u64 {
+        panic!("Given MTU size {} can't be represented in a positive `c_long` range")
+    }
+    state::<S>(bio).dtls_mtu_size = mtu_size as c_long;
 }
 
 unsafe fn state<'a, S: 'a>(bio: *mut BIO) -> &'a mut StreamState<S> {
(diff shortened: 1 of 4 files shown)


HyeonuPark:
It seems the tests run on the Rust version that doesn't support `std::convert::TryFrom`. I'll rewrite the function without it.

sfackler:
I was imagining this would just be a field in our BIO rather than a global variable.

HyeonuPark:
@sfackler agree, but it seems that it requires to change structure of `Ssl`, `bio::StreamState`, or both and I want to share working example before deciding such non-trivial change.

The problem is the user cannot access to the `BIO`, especially before the handshake starts. `Ssl::connect()` or `Ssl::accept()` call constructs the `BIO` and immediately starts handshake.

Maybe it's `Ssl` to store MTU value and pass it to the `BIO` on creation? Currently `Ssl` is just a thin wrapper over the `SSL*` so adding field to it can be considered critical. And what should happen if the user re-configure the MTU during the connection run?

Another possible way is to add more `SslStream` constructors like `Ssl::connect_with_dtls_mtu()`, `Ssl::accept_with_dtls_mtu()` etc. Or should it be `Ssl::connect_with(BioOption)` for future extensions?

I don't have strong opinion on either way, but adding global variable seems the least noisy way for the API surface.

sfackler:
I think we'd probably want some kind of builder API for `Ssl`, yeah.

HyeonuPark:
Aaand I made it! After a couple hours of exploring, I found that `SslStreamBuilder` doesn't start handshake until it consumes itself so I've just put MTU setter on it.

I tested it on my own code with wireshark and now it sends large handshake packets happily.

And some touch on dtls-srtp test. I'm a bit surprised that dtls tests were succeeded using TCP channel. Now it uses in-memory BIO emulating datagram transport. Should it be real UDP socket?

openssl/src/ssl/bio.rs
```diff
@@ -69,6 +71,16 @@ pub unsafe fn get_mut<'a, S: 'a>(bio: *mut BIO) -> &'a mut S {
     &mut state(bio).stream
 }
 
+pub unsafe fn set_dtls_mtu_size<S>(bio: *mut BIO, mtu_size: usize) {
+    const C_LONG_SIZE: usize = std::mem::size_of::<c_long>() * 8;
+    assert!(C_LONG_SIZE <= 64);
+    const C_LONG_MAX: u64 = (1u64 << (C_LONG_SIZE - 1)) - 1;
+    if mtu_size as u64 > C_LONG_MAX {
```
sfackler: This can all just be simplified to `if mtu_size as u64 > c_long::max_value() as u64`

sfackler:
Why wouldn't TCP work for SRTP tests? OpenSSL doesn't know what the underlying transport protocol is.

sfackler:
The libressl tests appear to be failing.

openssl/src/ssl/bio.rs
HyeonuPark: Where does `c_long::max_value()` declared? I can't find it in libc's doc.

HyeonuPark:
TCP stream doesn't relay the packet's boundary as-is. It may merge consecutive packets and split the packet in boundary if the merged chunk size exceed the transfer unit size of underlying layer. Nagle's algorithm is the well known example of it.

Anyway I'm not sure why it failed on libressl. I tested it on my macbook pro with Libressl 2.6.5, and well, it works on my machine(tm). The error comes from timeout and the only blocking operation it has is the `DgramMemBio::read()` so it means some participant sent packets fewer than the other participant expects. But it works on other test suites and my local libressl... I don't know :( will test again.

openssl/src/ssl/bio.rs
JRF63: It's on [`i64`](https://doc.rust-lang.org/std/primitive.i64.html#method.max_value), which [`c_long`](https://docs.rs/libc/0.2.65/libc/type.c_long.html) is just an alias for.

HyeonuPark:
Sorry for the long long delay. I've tested it again after rebase, and it didn't fail during an hour of test loop. But I found that my local libressl is v2.6.5, while the ci-failing version is v2.5. But the `DEP_OPENSSL_LIBRESSL_VERSION=255` doesn't seems to work on the mac. I'll test it with my linux box in this weekend.

HyeonuPark:
I figured out that the `DgramMemBio` thing made it hang on the libressl 2.5.5 case. It reproduces on the test commit which only has this change. So I removed it.

PR #1171, opened 2019-10-08
Fix rustc warnings

mvertescher:
Quick PR to fix up some build warnings:

- Use `..=` for inclusive ranges
- Add the `dyn` keyword for trait objects
- Switch from `ONCE_INIT` to `std::sync::Once::new()`


sfackler:
Thanks!

Issue #1157, opened 2019-09-16
how to check if a certificate is expired


Geal:
the [X509 struct](https://docs.rs/openssl/0.10.24/openssl/x509/struct.X509.html) has [not_after and not_before](https://docs.rs/openssl/0.10.24/openssl/x509/struct.X509.html#method.not_after) methods to get the certificates validity dates.
But how should the [Asn1Time](https://docs.rs/openssl/0.10.24/openssl/asn1/struct.Asn1Time.html) type be used?
My main use case is in the code managing a list of certificates, deciding which ones to load, and I want to manage expired certificates separately, but apparently I can't compare a `Asn1Time` to the current date

sfackler:
I think we should add `Eq`/`Ord` impls for Asn1Time backed by `ASN1_TIME_diff`.

axos88:
So the original issue seems to be fixed. But how about converting the Asn1TimeRef to std::time::SystemTime?

Right now the best I see is to convert it to string, and parse it.... Umm?

axos88:
```
        let expiry: &Asn1TimeRef = cert.not_after();
        let threshold = Asn1Time::days_from_now(0).unwrap();

         if (expiry < threshold) { 
            panic!("OMG EXPIRED!")
         }   
```

Cl00e9ment:
> So the original issue seems to be fixed. But how about converting the Asn1TimeRef to std::time::SystemTime?

I found this way to convert an `Asn1TimeRef` to a `SystemTime`:

```rust
fn asn1_time_to_system_time(time: &Asn1TimeRef) -> Result<SystemTime, ErrorStack> {
    let unix_time = Asn1Time::from_unix(0)?.diff(time)?;
    Ok(SystemTime::UNIX_EPOCH + Duration::from_secs(unix_time.days as u64 * 86400 + unix_time.secs as u64))
}
```

PR #1173, opened 2019-10-10
Add methods for comparing Asn1TimeRef and creating Asn1Time from Unix time

griff:
This implements a `diff` method on `Asn1TimeRef` using `ASN1_TIME_diff` and uses this new method to implement combinations of `PartialEq` and `PartialOrd` for `Asn1Time` and `Asn1TimeRef`.

It also adds a `from_unix` method to `Asn1Time` so that you can create a `Asn1Time` value from a `time_t` value.

This is mostly just a rework of the earlier work done by @illegalprime in his PR #673 and credit should go to him.

Solves #1157

### DIFF
--- a/openssl/src/asn1.rs
+++ b/openssl/src/asn1.rs
@@ -27,6 +27,8 @@
 use ffi;
 use foreign_types::{ForeignType, ForeignTypeRef};
 use libc::{c_char, c_int, c_long};
+#[cfg(ossl102)]
+use std::cmp::Ordering;
 use std::ffi::CString;
 use std::fmt;
 use std::ptr;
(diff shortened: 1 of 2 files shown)


openssl/src/asn1.rs
```diff
@@ -95,6 +114,97 @@ foreign_type_and_impl_send_sync! {
...
 }
 
+impl Asn1TimeRef {
+    /// Find difference between two times
+    ///
+    /// This corresponds to [`ASN1_TIME_diff`].
+    ///
+    /// [`ASN1_TIME_diff`]: https://www.openssl.org/docs/man1.1.0/crypto/ASN1_TIME_diff.html
+    #[cfg(ossl102)]
+    pub fn diff(&self, compare: Option<&Self>) -> Result<TimeDiff, ErrorStack> {
```
sfackler: Let's have this just take `compare: &Self` - we can add separate methods for the other case if necessary.

openssl/src/asn1.rs
```diff
@@ -95,6 +114,97 @@ foreign_type_and_impl_send_sync! {
...
+        if d.days < 0 || d.secs < 0 {
+            return Ok(Ordering::Greater);
+        }
+
+        Ok(Ordering::Equal)
+    }
+}
+
+#[cfg(ossl102)]
+impl Eq for Asn1TimeRef {}
```
sfackler: This isn't valid due to the error case, where you could in theory have a time that isn't equal to itself.

openssl/src/asn1.rs
```diff
@@ -165,6 +285,51 @@ impl Asn1Time {
     }
 }
 
+#[cfg(ossl102)]
+impl Eq for Asn1Time {}
```
sfackler: Same here.

openssl/src/asn1.rs
griff: True. Hadn't thought about that case. I will remove it.

sfackler:
Thanks!

PR #1175, opened 2019-10-18
add SslConnector builder without setting default verify paths

geogriff-signal:
SSL_CTX_set_default_verify_paths is irreversible, so let users construct a builder without it. i don't exactly intend this PR to land as-is, because i'm not sure what approach you'd want to take API-wise, but it seems removing `ctx.set_default_verify_paths()` from `SslConnector::builder` would be a breaking change, which is why i added a different constructor. Please let me know if there's a different direction you'd like to take.

### DIFF
--- a/openssl/src/ssl/connector.rs
+++ b/openssl/src/ssl/connector.rs
@@ -63,8 +63,17 @@ impl SslConnector {
     ///
     /// The default configuration is subject to change, and is currently derived from Python.
     pub fn builder(method: SslMethod) -> Result<SslConnectorBuilder, ErrorStack> {
+        let mut builder = Self::builder_no_default_verify_paths(method)?;
+        builder.set_default_verify_paths()?;
+        Ok(builder)
+    }
+
+    /// Creates a new builder for TLS connections, without setting the default locations of
+    /// trusted certificates for verification.
+    ///
+    /// The default configuration is subject to change, and is currently derived from Python.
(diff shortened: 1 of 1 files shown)


sfackler:
Can't you pass a new, empty X509Store to https://docs.rs/openssl/0.10.25/openssl/ssl/struct.SslContextBuilder.html#method.set_verify_cert_store?

geogriff-signal:
that doesn't actually reverse the effects of `SSL_CTX_set_default_verify_paths` sadly. i am indeed setting my own X509Store, but i can see it's still opening the default cert dirs as i'm running openssl under seccomp() where `openat()` is disallowed. Tested on openssl 1.1.1-1ubuntu2.1.

Following the openssl source, `SSL_CTX_set0_verify_cert_store` calls `ssl_cert_set_cert_store(ctx->cert, ...)` which sets `ctx->cert->verify_store` leaving the `ctx->cert_store` (containing the default verify paths) intact. i think the problem here may be that the `ctx->cert->chain_store` is still unset, causing `SSL_CTX_build_cert_chain` to default to `ctx->cert_store`.

so maybe the better solution here is to add a wrapper for `SSL_CTX_set_cert_store` or `SSL_CTX_set0_chain_cert_store`?

sfackler:
Ah yeah, let's add bindings for those functions!

geogriff-signal:
alright, that will take more time. looks like bindings are added by hand to openssl-sys by looking at header files for different openssl / libressl versions?

sfackler:
Yep.

PR #1176, opened 2019-10-19
Support LibreSSL 3.0.2

vishwin:
[3.0.2 is the first stable release](https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.0.2-relnotes.txt), so support the rest of the series.


sfackler:
Thanks!

PR #1179, opened 2019-11-11
Add EC_GROUP_order_bits

lukaslihotzki:


sfackler:
It looks like that function isn't available on all versions.

lukaslihotzki:
```
pub fn order_bits(&self) -> u32 {
    let mut bnc = BigNumContext::new().unwrap();
    let mut bn = BigNum::new().unwrap();
    self.order(&mut bn, &mut bnc).unwrap();
    bn.num_bits().try_into().unwrap()
}
```
can be used as a fallback. Should this be integrated? It seems difficult to do this fallback in application code, because the `ossl110`-flag isn't visible.

sfackler:
This library generally doesn't create nontrivial shims for things like that - it tries to provide what OpenSSL provides.

You can depend on openssl-sys directly and use a build script to create cfgs for things you need: https://github.com/sfackler/hyper-openssl/blob/master/build.rs

PR #1180, opened 2019-11-11
Added clonability for sha hash state.  Useful for incremental hashing

PrismaPhonic:
This PR adds the ability to clone the hash state.  This is useful for incremental hashing where the base of the hash is constant but some suffix blocks are changing.  As a simple example let's say I have the following base blocks:

```
let base = [0; 512];
let mut hasher = Sha1::new();
hasher.update(&base)
```

I then have a suffix partial block that I want to finish the hash up with that is changing:

```
let suffix = [1; 32];
hasher.update(&suffix);
let hash = hasher.finish();
```

If I then want to see what a different suffix might be like, I have to start all over, even if the base blocks are identical.  Instead with clonability I can do this:

```
let suffix = [1; 32];
let mut hasher2 = hasher.clone();
hasher2.update(&suffix);
let hash = hasher2.finish();

let suffix2 = [2; 32];
let mut hasher3 = hasher.clone();
hasher3.update(&suffix);
let hash = hasher3.finish();
```

The sum total work is far less than having to start over and re-process the base blocks.

### DIFF
--- a/openssl-sys/src/sha.rs
+++ b/openssl-sys/src/sha.rs
@@ -5,6 +5,7 @@ pub type SHA_LONG = c_uint;
 pub const SHA_LBLOCK: c_int = 16;
 
 #[repr(C)]
+#[derive(Clone)]
 pub struct SHA_CTX {
     pub h0: SHA_LONG,
     pub h1: SHA_LONG,
--- a/openssl-sys/src/sha.rs
+++ b/openssl-sys/src/sha.rs
@@ -26,6 +26,7 @@ extern "C" {
 }
 
 #[repr(C)]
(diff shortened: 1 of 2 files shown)


sfackler:
Seems reasonable to me. Can you update the other `Sha` structs as well?

PrismaPhonic:
No problem!

PrismaPhonic:
> Seems reasonable to me. Can you update the other `Sha` structs as well?

Updated with https://github.com/sfackler/rust-openssl/pull/1180/commits/8fae1115a4c60a3b51fce706da7bc0092ea8ee7f.

sfackler:
Thanks!

PR #1183, opened 2019-11-13
Expose OpenSSL interfaces for base64 coding.

adamreichold:
This exposes the `EVP_EncodeBlock` and `EVP_DecodeBlock` API of OpenSSL which can be used for base64 coding.

It seems reasonable to include those as they have no interactions with and hence unintended side-effects on the rest of the library and it would allow crates already depending on this crate to avoid additional dependencies for base64 coding.

### DIFF
--- a/openssl/src/base64.rs
+++ b/openssl/src/base64.rs
@@ -19,14 +19,16 @@ pub fn encode_block(src: &[u8]) -> String {
     let src_len = src.len() as c_int;
 
     let len = encoded_len(src_len).unwrap();
-    let mut out = vec![0; len as usize];
+    let mut out = Vec::new();
+    out.reserve(len as usize);
 
     // SAFETY: `encoded_len` ensures space for 4 output characters
     // for every 3 input bytes including padding and nul terminator.
     // `EVP_EncodeBlock` will write only single byte ASCII characters.
+    // `EVP_EncodeBlock` will only write to not read from `out`.
     unsafe {
         let out_len = ffi::EVP_EncodeBlock(out.as_mut_ptr(), src.as_ptr(), src_len);
-        out.truncate(out_len as usize);
+        out.set_len(out_len as usize);
         String::from_utf8_unchecked(out)
     }
 }
@@ -44,15 +46,17 @@ pub fn decode_block(src: &str) -> Result<Vec<u8>, ErrorStack> {
     let src_len = src.len() as c_int;
 
     let len = decoded_len(src_len).unwrap();
-    let mut out = vec![0; len as usize];
+    let mut out = Vec::new();
+    out.reserve(len as usize);
 
     // SAFETY: `decoded_len` ensures space for 3 output bytes
     // for every 4 input characters including padding.
     // `EVP_DecodeBlock` can write fewer bytes after stripping
     // leading and trailing whitespace, but never more.
+    // `EVP_DecodeBlock` will only write to not read from `out`.
     unsafe {
         let out_len = cvt_n(ffi::EVP_DecodeBlock(out.as_mut_ptr(), src.as_ptr(), src_len))?;
-        out.truncate(out_len as usize);
+        out.set_len(out_len as usize);
     }
 
     if src.ends_with("=") {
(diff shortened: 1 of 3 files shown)


openssl/src/base64.rs
```diff
@@ -0,0 +1,110 @@
...
+    // leading and trailing whitespace, but never more.
+    // `EVP_DecodeBlock` will only write to not read from `out`.
+    unsafe {
+        let out_len = cvt_n(ffi::EVP_DecodeBlock(out.as_mut_ptr(), src.as_ptr(), src_len))?;
+        out.set_len(out_len.try_into().unwrap());
+        Ok(out)
+    }
+}
+
+fn encoded_len(src_len: c_int) -> Option<usize> {
```
sfackler: Why does this take a c_int and return a usize?

adamreichold: I need the return value to `reserve`/`resize` the `out` buffer, so I need an `usize`. But the computation must succeed using a `c_int` for the OpenSSL calls to return valid counts otherwise I am not convinced I can rely on those counts for `set_len`/`truncate`.

In an earlier version, I used `c_int -> Option<c_int>` and had `encoded_len(src_len).unwrap().try_into().unwrap()` on the outside which seemed cumbersome but indeed less surprising. I can certainly changed this back if you prefer or I can add comment explaining the odd types better?

adamreichold: But I just noticed that the CI builds failed as the use Rust 1.33 which I guess is this libraries current minimum supported Rust version so I guess I have to remove the `TryInto` usage in any case?

adamreichold: Pushed a version that builds using Rust 1.33 and added a comment explaining the odd types.

Would you prefer the `usize_to_c_int` and `c_int_to_usize` in the top-level module or replaced by plain asserts as used in the other modules?

openssl/src/base64.rs
```diff
@@ -0,0 +1,129 @@
...
+    let mut out = Vec::new();
+    out.reserve(len);
+
+    // SAFETY: `encoded_len` ensures space for 4 output characters
+    // for every 3 input bytes including padding and nul terminator.
+    // `EVP_EncodeBlock` will write only single byte ASCII characters.
+    // `EVP_EncodeBlock` will only write to not read from `out`.
+    unsafe {
+        let out_len = ffi::EVP_EncodeBlock(out.as_mut_ptr(), src.as_ptr(), src_len);
+        out.set_len(c_int_to_usize(out_len).unwrap());
```
sfackler: This can just be `out_len as usize`.

openssl/src/base64.rs
sfackler: It seems best to mirror the rest of the codebase.

openssl/src/base64.rs
adamreichold: Indeed. Fixed this and the checked cast in `{de,en}coded_len` that was actually unnecessary as well as we always call this with non-negative values and the arithmetic is checked.

openssl/src/base64.rs
adamreichold: Pushed a version that hopefully better mirrors the style of the rest of the code base.

adamreichold:
During the last rebase, I also noticed that I actually forgot the `#[test]` attribute on two of tests which I now added but which also showed that I forgot to strip the zero padding from the decoded output which I added as well.

openssl/src/base64.rs
```diff
@@ -0,0 +1,125 @@
...
+    // for every 4 input characters including padding.
+    // `EVP_DecodeBlock` can write fewer bytes after stripping
+    // leading and trailing whitespace, but never more.
+    // `EVP_DecodeBlock` will only write to not read from `out`.
+    unsafe {
+        let out_len = cvt_n(ffi::EVP_DecodeBlock(out.as_mut_ptr(), src.as_ptr(), src_len))?;
+        out.set_len(out_len as usize);
+    }
+
+    if src.ends_with("=") {
```
sfackler: What is this doing?

sfackler: Oh right, null padding.

PR #1184, opened 2019-11-13
Format code using 'cargo fmt'

Atul9:


rustfmt.toml
```diff
@@ -0,0 +1,66 @@
+max_width = 100
```
sfackler: Just running with defaults seems fine.

Atul9: The rustfmt.toml is generated using rustfmt with default settings.

sfackler: So then this file doesn't need to exist, right?

Atul9: while contributing to other rust opensource projects I got to know that - `without an included rustfmt.toml file runs the risk fo some things being formatted randomly at random times` reference - https://github.com/sagebind/sluice/pull/4#issuecomment-536084517, hence I included it.

sfackler: I'm not convinced that random format changes will happen.

Atul9: I am ok with not including rustfmt.toml in this PR, should I remove it?

sfackler: Yes

Atul9: done, I have updated the PR.

PR #1192, opened 2019-11-22
Add EVP_md_null() and MessageDigest::md_null()

fengyc:
Add EVP_md_null() and MessageDigest::md_null()

### DIFF
--- a/openssl/src/hash.rs
+++ b/openssl/src/hash.rs
@@ -40,7 +40,7 @@ impl MessageDigest {
         }
     }
 
-    pub fn md_null() -> MessageDigest {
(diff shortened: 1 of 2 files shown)


openssl/src/hash.rs
```diff
@@ -40,6 +40,10 @@ impl MessageDigest {
         }
     }
 
+    pub fn md_null() -> MessageDigest {
```
sfackler:
```suggestion
    pub fn null() -> MessageDigest {
```

sfackler:
LGTM other than a rename!

PR #1202, opened 2019-12-12
Fix documentation typo of "chain"

traviscross:
In the documentation of `SslRef::verified_chain`, we had written
"certificate chani" rather than "certificate chain".  In this commit,
we fix the typo.


sfackler:
Thanks!

PR #1203, opened 2019-12-16
Implement TryFrom conversion between BigNum and unsigned integers

npmccallum:


npmccallum:
@sfackler I punted on the error types since I wasn't sure what error type you want to use. I can update this based on your preference.

npmccallum:
@sfackler I believe the build failure is spurious.

PR #1204, opened 2019-12-17
Add bn::BigNum::to_slice() method

npmccallum:
This writes the big-endian form of the number into a preallocated buffer.

### DIFF
--- a/openssl/src/bn.rs
+++ b/openssl/src/bn.rs
@@ -868,13 +868,46 @@ impl BigNumRef {
     /// assert_eq!(BigNum::from_slice(&s_vec).unwrap(), r);
     /// ```
     pub fn to_vec(&self) -> Vec<u8> {
-        let size = self.num_bytes() as usize;
-        let mut v = Vec::with_capacity(size);
-        unsafe {
-            ffi::BN_bn2bin(self.as_ptr(), v.as_mut_ptr());
-            v.set_len(size);
-        }
+        let mut v = vec![0u8; self.num_bytes() as usize];
+        self.to_slice(&mut v).unwrap();
         v
+    }
+
+    /// Encodes the absolute value of `self` into a big-endian buffer.
+    ///
+    /// `self` can be recreated by using `from_slice`.
+    ///
+    /// This method fails if the buffer is not large enough. If the
+    /// buffer is larger than necessary, the number is prefixed by
+    /// zeroes so that the result is always a valid big-endian number.
+    ///
+    /// ```
+    /// # use openssl::bn::BigNum;
+    /// let s = -BigNum::from_u32(4543).unwrap();
+    /// let r = BigNum::from_u32(4543).unwrap();
+    ///
+    /// let mut b = [0u8; 4];
+    /// s.to_slice(&mut b).unwrap();
+    /// assert_eq!(BigNum::from_slice(&b).unwrap(), r);
+    ///
+    /// let mut b = [0u8; 1];
+    /// assert_eq!(s.to_slice(&mut b), Err(()));
+    /// ```
+    pub fn to_slice(&self, buf: &mut [u8]) -> Result<(), ()> {
+        let ilen = self.num_bytes() as usize;
+        let olen = buf.len();
+        if olen < ilen {
+            return Err(());
+        }
+
+        let (zero, data) = buf.split_at_mut(olen - ilen);
+
+        unsafe {
+            core::ptr::write_bytes(zero.as_mut_ptr(), 0, zero.len());
+            ffi::BN_bn2bin(self.as_ptr(), data.as_mut_ptr());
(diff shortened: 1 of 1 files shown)


npmccallum:
@sfackler Just like #1203 (which now depends on this patch), I've punted on the error type. Let me know what you'd like.

openssl/src/bn.rs
```diff
@@ -868,13 +868,46 @@ impl BigNumRef {
...
+    /// let r = BigNum::from_u32(4543).unwrap();
+    ///
+    /// let mut b = [0u8; 4];
+    /// s.to_slice(&mut b).unwrap();
+    /// assert_eq!(BigNum::from_slice(&b).unwrap(), r);
+    ///
+    /// let mut b = [0u8; 1];
+    /// assert_eq!(s.to_slice(&mut b), Err(()));
+    /// ```
+    pub fn to_slice(&self, buf: &mut [u8]) -> Result<(), ()> {
```
sfackler: This can just assert that the buffer is big enough.

openssl/src/bn.rs
```diff
@@ -868,13 +868,46 @@ impl BigNumRef {
...
+        v
+    }
+
+    /// Encodes the absolute value of `self` into a big-endian buffer.
+    ///
+    /// `self` can be recreated by using `from_slice`.
+    ///
+    /// This method fails if the buffer is not large enough. If the
+    /// buffer is larger than necessary, the number is prefixed by
+    /// zeroes so that the result is always a valid big-endian number.
```
sfackler: We generally try to bind OpenSSL functions "directly", so I think it makes more sense to write to the head of the buffer and return the number of bytes written like the underlying API does.

npmccallum: This makes it hard to write into fixed array sizes, which is my motivation for writing this patch. For example, I need to write an RSA number into a 384 byte buffer to hand to hardware. I don't control the size of the buffer.

If we write to the beginning of the buffer, now I have to manually move it and zero-pad. This imposes the cost of an extra `memmove()` on every consumer, all to retain exact compatibility with a terrible OpenSSL interface that doesn't take a length parameter to a buffer it writes to (!).

I understand the desire to mimic the OpenSSL interface. But this particular interface is a pretty bad one.

openssl/src/bn.rs
npmccallum: We could do this, yes. But then we'd write the same code again in `TryFrom` (see #1203). It also doesn't seem like the right thing to do for security to force the consumer of this API to check for theoretical errors before calling it to avoid an abort. This is particularly true when the incoming `BigNum` data is often user or network input.

openssl/src/bn.rs
sfackler: Why do you have to memmove? Just slice down to the tail of the buffer.

openssl/src/bn.rs
npmccallum: @sfackler Do you have any interest in merging support for `TryFrom` integer conversions? Your crate is the only one who can supply this interface because of Rust's impl rules.

openssl/src/bn.rs
npmccallum: Yes, but now I have to carry a utility function that checks the length of the number and splits the buffer - doing zeros in the first half and passing the second half to this function. This also implies unsafe code for performance. I'd rather consolidate this code.

openssl/src/bn.rs
sfackler: Sure, that seems reasonable.

openssl/src/bn.rs
sfackler: This seems like an extremely niche use case to your hardware thing.

npmccallum: Fixed sized integer conversions are extremely common the moment you do crypto in hardware. There is a lot of crypto hardware out there.

Should we have two methods: one that wraps the OpenSSL behavior and one that writes a valid big endian number to the size of the buffer?

sfackler: We should expose the OpenSSL interfaces that exist. If OpenSSL has not seen it fit to expose the interface you want, it is not clear to me why this library should.

I would also be interested to see how you've quantified the performance costs that require the use of the unsafe write_bytes function rather than just writing zeroes in a loop. AFAICT LLVM will recognize that idiom and lower to a memset anyway.

npmccallum: Okay. I will expose the OpenSSL interface. I will also wrap it in `impl TryFrom`. The only thing left to agree on is what error to use for `TryFrom` when the `BigNum` is too large for the target integer type.

npmccallum: @sfackler It appears that OpenSSL *does* provide a padding variant: `BN_bn2binpad()`. I'll wrap this function too.

PR #1205, opened 2019-12-17
Fix warnings on newer Rust

npmccallum:


.circleci/config.yml
```diff
@@ -224,7 +224,7 @@ workflows:
           name: mimimal-version
           target: x86_64-unknown-linux-musl
           vendored: true
-          image: 1.31.0
+          image: 1.36.0
```
sfackler: It does not seem reasonable to bump the MSRV just to fix warnings.

npmccallum: I agree. I wasn't aware of the problem when I first made the PR. Then I bumped the MSRV to make the issue transparent. I will re-submit with just the fix that doesn't require a new MSRV.

PR #1206, opened 2019-12-19
In the documentation of:-

shanuagarwal:
'SslContextBuilder::set_session_id_context' we had written "be be" rather than "be",
'SslContextBuilder::set_certificate' we had written description of 'add_extra_chain_cert' rather than description of 'set_certificate',
'SslContextBuilder::add_extra_chain_cert' we had written the description of 'set_certificate' rather than description of 'add_extra_chain_cert',
'SslContextBuilder::set_cipher_list' we had written the description of 'set_ciphersuites' rather than description of 'set_cipher_list',
'SslContextBuilder::set_ciphersuites' we had written the description of 'set_cipher_list' rather than description of 'set_ciphersuites',
'SslContextBuilder::set_cipher_list' we had written wrong link.
In this commit we fix the typo.

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -885,7 +885,7 @@ impl SslContextBuilder {
     /// Set the context identifier for sessions.
     ///
     /// This value identifies the server's session cache to clients, telling them when they're
-    /// able to reuse sessions. It should be be set to a unique value per server, unless multiple
+    /// able to reuse sessions. It should be set to a unique value per server, unless multiple
     /// servers share a session cache.
     ///
     /// This value should be set when using client certificates, or each request will fail its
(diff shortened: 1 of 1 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -956,7 +956,8 @@ impl SslContextBuilder {
 
     /// Sets the leaf certificate.
     ///
-    /// Use `add_extra_chain_cert` to add the remainder of the certificate chain.
+    /// This chain should contain all certificates necessary to go from the certificate specified by
```
sfackler: This method doesn't take a chain, it takes a single certificate. What is wrong about the current documentation?

openssl/src/ssl/mod.rs
```diff
@@ -967,8 +968,7 @@ impl SslContextBuilder {
 
     /// Appends a certificate to the certificate chain.
     ///
-    /// This chain should contain all certificates necessary to go from the certificate specified by
-    /// `set_certificate` to a trusted root.
+    /// Use `add_extra_chain_cert` to add the remainder of the certificate chain.
```
sfackler: I don't think this sentence makes sense on this method.

openssl/src/ssl/mod.rs
shanuagarwal: my bad, ignore this change

openssl/src/ssl/mod.rs
shanuagarwal: yes agree, my bad again

PR #1208, opened 2019-12-20
Fix documentation typo of "the"

Sach1nAgarwal:
In the documentation of SslContextBuilder::set_certificate, 
we had written "lthe" rather than "the". In this commit,
we fix the typo.


sfackler:
Thanks!

PR #1225, opened 2020-01-27
Add NO RENEGOTIATION option

kubo39:
SSL_OP_NO_RENEGOTIATION was added in OpenSSLv1.1.1 and backported to v1.1.0h.


kubo39:
Would it be better to add undocumented `SSL3_FLAGS_NO_RENEGOTIATE_CIPHERS` flag too?

sfackler:
Feel free to add the other flag if you'd like!

kubo39:
In my use case `NO_RENEGOTIATION` is enough and currently `SSL3_FLAGS_NO_RENEGOTIATE_CIPHERS` does never affect.

kubo39:
I couldn't find why appveyor failed... :/

sfackler:
Looks unrelated to this PR.

kubo39:
Thanks!

Issue #1228, opened 2020-02-05
How to implement set_servernam_callback / set_clienthello_callback cor


axos88:
What is the correct way to create the ssl context in the client_hello_callback? There doesn't seem to be a way to unwrap an SslAcceptor into an SslContext. I can Deref it, but build needs to move the value, a reference is not enough. Do I need to create one from scratch?

I'm trying to implement SNI. This is what I have so far:
```
            let mut ssl: SslAcceptorBuilder = SslAcceptor::mozilla_intermediate(SslMethod::tls())?;

            ssl.set_private_key_file(unimplemented!(), SslFileType::PEM);
            
            ssl.set_client_hello_callback(|mut sslref, mut sslalert| {               
                info!("servername = {:?}", sslref.servername(NameType::HOST_NAME));

                //1.
                let mut builder = SslAcceptor::mozilla_intermediate(SslMethod::tls()).unwrap();
                builder.set_private_key_file(unimplemented!(), SslFileType::PEM);
                builder.set_certificate_chain_file(unimplemented!());

                let mut builder = builder.into_ssl_context_builder(); // This function doesn't exist

                //2.
                let mut builder = SslContextBuilder::from_context(sslref.ssl_context());// This function doesn't exist
                builder.set_certificate_chain_file(unimplemented!());

                sslref.set_ssl_context(builder.build());
                Ok(ClientHelloResponse::SUCCESS)
            });
```

sfackler:
It seems reasonable to add methods on `SslConnector`/`SslAcceptor` to get a reference to the inner `SslContext`.

axos88:
Gotcha, thanks. So I'd need to fork & create a PR for that, I guess. 

Btw, do I need to create a whole new context in the callback, or would most of the settings, such as ciphers, etc. be inherited from the "parent" context?

I find openssl documentation extremely lacking, to the point which is frightening...

sfackler:
I think you'd typically create the set of contexts at application start, and just select the right one and slot it in in your callback.

Yep, the state of OpenSSL's documentation is not great. I pretty often just go to the source code to figure out exactly what something's doing.

axos88:
Gotcha thanks. I want to be able to change and add new certificates on the fly, as files are saved to disk, so I can't prepare a store at application start, unfortunately.

axos88:
I've added the code I'll submit the PR after I get it working.

PR #1229, opened 2020-02-05
SslAcceptor and SslConnector: Ability to turn into SslContext

axos88:
Addresses #1228

Deref did not seem to be enough because I could not construct an SslContextRef from &SslContext

### DIFF
--- a/openssl/src/ssl/connector.rs
+++ b/openssl/src/ssl/connector.rs
@@ -4,7 +4,7 @@ use std::ops::{Deref, DerefMut};
 use dh::Dh;
 use error::ErrorStack;
 use ssl::{
-    HandshakeError, Ssl, SslContext, SslContextBuilder, SslMethod, SslMode, SslOptions, SslRef,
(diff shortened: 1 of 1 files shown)


openssl/src/ssl/connector.rs
```diff
@@ -91,8 +91,31 @@ impl SslConnector {
...
+    pub fn into_ssl_context(self) -> SslContext {
+        self.0
+    }
+
+    pub fn ssl_context_ref(&self) -> &SslContextRef {
+        &*self.0
+    }
+}
+
+impl Deref for SslConnector {
```
sfackler: I'd prefer to not have the deref/derefmut impls here. I want to keep the acceptor/connector a bit more opaque to guide towards the high-level methods.

axos88: So keep only the into_ssl_context() and into_ssl_context_ref()?

sfackler: Yep

axos88:
@sfackler  updated

PR #1230, opened 2020-02-08
Update to mem::MaybeUninit and Once::new()

nipunn1313:
Newest version of rust compiler warns about deprecation for `mem::uninitialized` and `ONCE_INIT` - prefers these versions. I confirmed this compiles on the latest stable as well as the latest nightly.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,7 @@ jobs:
         default: false
       image:
         type: string
-        default: 1.33.0
+        default: 1.41.0
       minimal_build:
         type: boolean
         default: false
@@ -172,7 +172,7 @@ jobs:
         default: false
       image:
         type: string
-        default: 1.33.0
+        default: 1.41.0
     macos:
       xcode: "9.0"
     environment:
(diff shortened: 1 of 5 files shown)


sfackler:
I'd rather not bump the minimum supported rust version for internal-only changes like this.

nipunn1313:
Yep - I was just going to ask about that. Would it be worth throwing cfg[feature(maybe_uninit)] around it ?https://github.com/rust-lang/rust/issues/63566?

nipunn1313:
The reason this came up is because on the latest nightly - the warnings for openssl

(eg
```
warning: use of deprecated item 'std::mem::uninitialized': use `mem::MaybeUninit` instead
   --> /home/nipunn/src/client/rust_patched_vendor/rust-openssl/openssl/src/sha.rs:285:38
    |
285 |             let mut hash: [u8; 64] = mem::uninitialized();
```

nipunn1313:
see https://github.com/rust-lang/cargo/pull/7450/ for context

sfackler:
Sticking `#[allow(deprecated)]` on the offending functions seems reasonable.

nipunn1313:
Put up 1231 to take that approach. This can be revived eventually whenever you would like to bump the min supported rust version.

PR #1231, opened 2020-02-10
Attach cfg[allow_deprecated] to methods w/ uninitialized functionality

nipunn1313:
Additionally - update usage of ONCE_INIT

### DIFF
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -103,7 +103,7 @@ pub fn init() {
     use std::io::{self, Write};
     use std::mem;
     use std::process;
-    use std::sync::{Mutex, MutexGuard, Once, ONCE_INIT};
+    use std::sync::{Mutex, MutexGuard, Once};
 
     static mut MUTEXES: *mut Vec<Mutex<()>> = 0 as *mut Vec<Mutex<()>>;
     static mut GUARDS: *mut Vec<Option<MutexGuard<'static, ()>>> =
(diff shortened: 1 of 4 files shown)


nipunn1313:
Alternate option to 1230

nipunn1313:
vcpkg doesn't appear to be compiling - I would guess that it requires some newer version of the windows C++ compiler - because the error seems like a compiler error

```
C:\projects\vcpkg\toolsrc\include\vcpkg/parse.h(114): error C2440: 'return': cannot convert from 'initializer list' to 'vcpkg::Parse::TextRowCol' (compiling source file ..\src\vcpkg\paragraphs.cpp) [C:\projects\vcpkg\toolsrc\vcpkglib\vcpkglib.vcxproj]
```

```
// failing to compile
TextRowCol cur_rowcol() const { return {row, column}; }
```

```
// definition of TextRowCol
namespace vcpkg::Parse
{
    struct TextRowCol
    {
        /// '0' indicates uninitialized; '1' is the first row.
        int row = 0;
        /// '0' indicates uninitialized; '1' is the first column.
        int column = 0;
    };
}
```

nipunn1313:
there's a bunch of work that has happened to vcpkg over the past few days as well (and the appveyor.yml pulls from master of vcpkg) so it may be due to changes over there as well. It might be worth pinning to a stable version of vcpkg in the git pull command.

sfackler:
Thanks! Yeah, the windows CI tends to be a bit flaky historically :(

PR #1232, opened 2020-02-11
Allow bad style on EVP_PKEY_up_ref  in pkey.rs

nipunn1313:
This follows the pattern used in other files in rust-openssl


nipunn1313:
I confirmed that this previously emitted an error (About snake_case), but now it's suppressed.

PR #1234, opened 2020-02-14
Expose HKDF functions

### LINKED ISSUE #1233: Expose HKDF functions
In OpenSSL, the HKDF key derivation functions (HKDF-Extract and HKDF-Expand, defined in RFC 5869) are exposed through the [`EVP_KDF` API](https://www.openssl.org/docs/manmaster/man7/EVP_KDF-HKDF.html).

It would be nice if rust-openssl provides a way to access those functions, as they are widely used in TLS and QUIC.

ueno:
This adds functions for HKDF (HMAC-based Extract-and-Expand Key Derivation Function), provided by OpenSSL 1.1.1.

Fixes #1233.

### DIFF
--- a/openssl/src/pkcs5.rs
+++ b/openssl/src/pkcs5.rs
@@ -1,10 +1,10 @@
 use libc::c_int;
 use std::ptr;
 
-use crate::cvt;
 use crate::error::ErrorStack;
-use crate::hash::MessageDigest;
+use crate::hash::{DigestBytes, MessageDigest};
 use crate::symm::Cipher;
+use crate::{cvt, cvt_p};
 
 #[derive(Clone, Eq, PartialEq, Hash, Debug)]
 pub struct KeyIvPair {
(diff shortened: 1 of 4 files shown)


systest/build.rs
```diff
@@ -67,6 +67,10 @@ fn main() {
         cfg.header("openssl/cms.h");
     }
 
+    if openssl_version.unwrap() >= 0x1_01_01_00_0 {
+        cfg.header("openssl/kdf.h");
+    }
```
kubo39: CI falied because panic! occurs in case of libressl.
How about like below?

```rust
if openssl_version.is_some() {
    cfg.header("openssl/cms.h");
    if openssl_version.unwrap() >= 0x1_01_01_00_0 {
        cfg.header("openssl/kdf.h");
    }
}
```

ueno: Thank you for the help! Fixed in that way.

openssl/src/pkcs5.rs
```diff
@@ -138,6 +138,155 @@ pub fn scrypt(
...
+                kctx,
+                ffi::EVP_PKEY_HKDEF_MODE_EXPAND_ONLY,
+            ))?;
+
+            cvt(ffi::EVP_PKEY_CTX_set_hkdf_md(
+                kctx,
+                hash.as_ptr(),
+            ))?;
+
+            cvt(ffi::EVP_PKEY_CTX_add1_hkdf_key(
```
kubo39: CI shows:

```
error[E0425]: cannot find function `EVP_PKEY_CTX_add1_hkdf_key` in module `ffi`
   --> openssl/src/pkcs5.rs:251:22
    |
251 |             cvt(ffi::EVP_PKEY_CTX_add1_hkdf_key(
    |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^ help: a function with a similar name exists: `EVP_PKEY_CTX_set1_hkdf_key`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0425`.
error: Could not compile `openssl`.
```

I couldn't find `EVP_PKEY_CTX_add1_hkdf_key` in include/openssl/kdf.h but `EVP_PKEY_CTX_set1_hkdf_key` (openssl's revision: af7f656cd91d99d62567e2b20c61f07cb4d81d0b)

ueno: It's indeed a typo; fixed.

openssl/src/pkcs5.rs
```diff
@@ -138,6 +138,155 @@ pub fn scrypt(
     }
 }
 
+/// Extract a pseudorandom key from an input keying material using the
+/// HKDF-Extract algorithm.
+///
+/// Requires OpenSSL 1.1.1 or newer.
+#[cfg(any(ossl111))]
```
sfackler: Are these two sequences of calls the only set of things you'd ever need to do when interacting with the HKDF API? If not, it probably makes more sense to add type(s) like `Signer` to provide more flexibility.

ueno: Yes, those two are the only possible call sequences. Note that I put them under `openssl::pkcs5` just because the HKDF operations are similar to the existing `pbkdf2` interface and they have nothing to do with PKCS#5. Perhaps it might make more sense to move them to a dedicated module where SSH KDF etc. can also be supported.

sfackler: Isn't there minimally a third call sequence using EXTRACT_AND_EXPAND? https://www.openssl.org/docs/man1.1.1/man3/EVP_PKEY_CTX_add1_hkdf_info.html

ueno: By the definition of HKDF ([RFC 5869](https://tools.ietf.org/html/rfc5869)), there are only two functions (HKDF-Extract and HKDF-Expand). OpenSSL provides a combined operation for convenience but it can be easily performed with the HKDF-Extract and HKDF-Expand operations.

sfackler: It seems strange to not expose the thing that OpenSSL's documentation states is the default mode.

ueno: Makes sense; I've added `hkdf` for that behavior.

voidc:
What is missing for this PR to be merged?

ueno:
FWIW I've rebased it against the latest git master; the clippy failures don't seem to be related to this PR.

Allen-Webb:
It would be nice if some form of this was merged. I am having to apply it as a patch on the openssl crate for local use.

sfackler:
Done in #1498

PR #1238, opened 2020-02-26
use library names supplied by vcpkg

mcgoo:
The current version of openssl-sys fails to build against OpenSSL 1.1.1d from a vcpkg installation because the DLLs now have more elaborate names, `libcrypto-1_1-x64.dll` for example. This PR switches to querying vcpkg for the library names rather than them being hardcoded in `build.rs`.


sfackler:
Thanks!

Issue #1239, opened 2020-02-26
Memory leak in X509Builder::append_extension


adrian-budau:
append_extension assumes that `X509_add_ext` takes ownership of the `X509_EXTENSION` but the documentation says otherwise: https://www.openssl.org/docs/man1.1.0/man3/X509_get_ext.html

Confirmed also with a small program:

```rust
use openssl::x509::X509Builder;
use openssl::x509::extension::BasicConstraints;

fn main() {
    loop {
        let mut builder = X509Builder::new().unwrap();                                                                                                                                        
        let extension = BasicConstraints::new().build().unwrap();                                                
        builder.append_extension(extension).unwrap();                                                  
    }                                                                                                                                                 
}
```

Running this while checking top shows memory rapidly accumulating.

The problem is from `mem::forget(extension)` at openssl/src/x509/mod.rs:348
A good solution would be to change the signature to accept an `&X509ExtensionRef` but I guess that would be a breaking change?

I'm happy to do a pull request to fix this (either by changing the definition, or just removing that line).

sfackler:
Good find. We should definitely remove the mem::forget to fix the leak, and maybe add an `append_extension2` that takes `&X509ExtensionRef` which we can clean up in the next breaking release.

adrian-budau:
I'm going to open a pull request in a second.

PR #1241, opened 2020-02-26
Fix the memory leak in `X509Builder::append_extension`.

adrian-budau:
Also add an alternative method that takes a `X509ExtensionRef`.

This fixes: #1239


sfackler:
Thanks!

PR #1247, opened 2020-03-23
Add a way to get the certificates stored in an X509Store

coolreader18:


openssl/src/x509/mod.rs
```diff
@@ -1322,6 +1322,32 @@ impl X509AlgorithmRef {
...
+    pub struct X509Object;
+    /// Reference to `X509Object`
+    pub struct X509ObjectRef;
+}
+
+impl X509ObjectRef {
+    pub fn x509(&self) -> Option<&X509Ref> {
+        unsafe {
+            if X509_OBJECT_get_type(self.as_ptr()) == ffi::X509_LU_X509 {
+                Some(X509Ref::from_ptr(X509_OBJECT_get0_X509(self.as_ptr())))
```
sfackler: X509_OBJECT_get0_X509 returns null if the object isn't a certificate, so you should be able to skip the X509_OBJECT_get_type call and just check the pointer.

sfackler:
It looks like you may need to add an extra header to what the systest build script pulls in: https://github.com/sfackler/rust-openssl/blob/master/systest/build.rs#L48

coolreader18:
It looks like prior to 1.1.1 `X509_LOOKUP_TYPE` wasn't an enum but a group of constants ([commit](https://github.com/openssl/openssl/commit/bca3f06b84de3c0b428724ac535995064c54aee3)), should I just have a `cfg_if` defining the constants by a type alias for `ossl111` or just as ints otherwise?

sfackler:
I think we only need the type constants pre-1.1.0, right?

coolreader18:
Oh, I guess so, yeah.

coolreader18:
Ugh, `X509_STORE_get0_objects` didn't exist prior to 1.1.0, it was just a field on the struct. Would I be able to fill in only part of the `X509_STORE` struct, since the `STACK_OF(X509_OBJECT) *objs` field is only the second one?

sfackler:
The size of the struct has to exactly match C. You could alternatively just not implement the method for old OpenSSL versions though.

coolreader18:
Okay, I filled in the struct fields; it should be all good now.

openssl/src/x509/store.rs
```diff
@@ -104,3 +105,21 @@ foreign_type_and_impl_send_sync! {
     /// Reference to an `X509Store`.
     pub struct X509StoreRef;
 }
+
+impl X509StoreRef {
+    /// Get a reference to the cache of certificates in this store.
+    pub fn certs(&self) -> &StackRef<X509Object> {
```
sfackler:
```suggestion
    pub fn objects(&self) -> &StackRef<X509Object> {
```

PR #1248, opened 2020-03-24
Add functions for SSL{_CTX}_get_verify_mode and SSL_is_init_finished

coolreader18:


openssl/src/ssl/mod.rs
```diff
@@ -2394,6 +2404,14 @@ impl SslRef {
         unsafe { ffi::SSL_set_verify(self.as_ptr(), mode.bits as c_int, None) }
     }
 
+    /// Returns the verify mode that was set using `set_verify`.
+    ///
+    /// This corresponds to `SSL_get_verify_mode`.
+    pub fn verify_mode(&self) -> SslVerifyMode {
```
coolreader18: I called this `verify_mode` even though the setter is just `set_verify` because I thought that a function called `verify()` would seem like it does some kind of certificate verification.

sfackler: Seems fine to me, and matches the raw OpenSSL name.

openssl/src/ssl/mod.rs
```diff
@@ -3172,6 +3190,14 @@ impl SslRef {
         }
     }
 
+    /// Determines if the initial handshake has been completed.
+    ///
+    /// This corresponds to `SSL_is_init_finished`.
+    #[cfg(ossl110)]
+    pub fn init_finished(&self) -> bool {
```
sfackler:
```suggestion
    pub fn is_init_finished(&self) -> bool {
```

openssl/src/ssl/mod.rs
sfackler: Can you add links to the man page for this and is_init_finished below?

openssl/src/ssl/mod.rs
```diff
@@ -1944,6 +1944,17 @@ impl SslContextRef {
     pub fn session_cache_size(&self) -> i64 {
         unsafe { ffi::SSL_CTX_sess_get_cache_size(self.as_ptr()).into() }
     }
+
+    /// Returns the verify mode that was set on this context from [`SslContextBuilder::set_verify`].
+    ///
+    /// This corresponds to `SSL_CTX_get_verify_mode`.
```
sfackler:
```suggestion
    /// This corresponds to [`SSL_CTX_get_verify_mode`].
```

PR #1249, opened 2020-03-24
Add SslMethod::tls_{client,server}

coolreader18:


sfackler:
Out of curiosity, what's your use case for wanting to use these?

coolreader18:
I'm working on adding the `_ssl` module to RustPython in RustPython/RustPython#1823, and while I can't think of a specific use case for using them, the module does technically support all of the different methods. I probably could just use the sys functions and `from_ptr`, this probably isn't used enough to warrant addition here.

sfackler:
Ah sure, seems reasonable to me.

PR #1250, opened 2020-03-25
chore: bump openssl-src to be able to customize perl location

jonathanmorley:


sfackler:
The version constraint there doesn't need to be updated - just run `cargo update`.

PR #1253, opened 2020-04-05
Check for the OPENSSL_NO_VENDOR environment variable

### LINKED ISSUE #1251: A feature to vendor openssl only on windows?
As I was setting up CI to install openssl with vcpkg on windows, I thought it might be nice to avoid that trouble in CI, but also for anyone who wants to just try out our project on windows; it can turn people away to have to spend like 5-10 minutes figuring out how to install openssl on windows, either through the msi available online or through vcpkg, with `VCPKGRS_DYNAMIC=1` enabled as well. However, on linux distributions, it's usually as easy as `packagemanager install openssl`. Would a feature for `rust-openssl` like `vendored_on_windows` be a good idea? (I make this proposal here only because I don't think it's possible to conditionally enable cargo features based on a `cfg()`; it has to be something in this crate.)

coolreader18:
Resolves #1251

### DIFF
--- a/openssl/src/lib.rs
+++ b/openssl/src/lib.rs
@@ -53,6 +53,8 @@
 //! * `OPENSSL_STATIC` - If set, the crate will statically link to OpenSSL rather than dynamically link.
 //! * `OPENSSL_LIBS` - If set, a `:`-separated list of library names to link to (e.g. `ssl:crypto`). This can be used
 //!     if nonstandard library names were used for whatever reason.
+//! * `OPENSSL_NO_VENDOR` - If present, always find openssl in the system, even if the `vendored` feature is set by
+//!     the crate that's using this one.
 //!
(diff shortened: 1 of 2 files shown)


openssl/src/lib.rs
```diff
@@ -53,6 +53,8 @@
 //! * `OPENSSL_STATIC` - If set, the crate will statically link to OpenSSL rather than dynamically link.
 //! * `OPENSSL_LIBS` - If set, a `:`-separated list of library names to link to (e.g. `ssl:crypto`). This can be used
 //!     if nonstandard library names were used for whatever reason.
+//! * `OPENSSL_NO_VENDOR` - If present, always find openssl in the system, even if the `vendored` feature is set by
+//!     the crate that's using this one.
```
sfackler, line 57:
```suggestion
//! * `OPENSSL_NO_VENDOR` - If present, always find openssl in the system, even if the `vendored` feature is enabled.
```

PR #1266, opened 2020-05-02
I made it to recognize macport openssl library.

### LINKED ISSUE #976: [macos] Build error is misleading for MacPorts users/systems
I use MacPorts for almost all dev packages on my machine. When trying to build some packages depending on `openssl-sys` crate, I kept getting errors that the system `openssl 0.9.8` found is not supported and some `brew` commands were suggested to resolve the issue.

The problem was that, although I had `openssl 1.x` port (https://www.macports.org/ports.php?by=library&substr=openssl) installed, I didn't have `pkg-config` installed and that was causing `openssl-sys` failing to find the right libs dir.

Would be great if we can add more info in the error message about `pkg-config` not being found, and maybe even `port` commands to install both `pkgconfig` and `openssl` packages.

toshiyuki-rs:
This modification will resolve #976.

### DIFF
--- a/openssl-sys/build/find_normal.rs
+++ b/openssl-sys/build/find_normal.rs
@@ -50,6 +50,10 @@ fn find_openssl_dir(target: &str) -> OsString {
                 return homebrew.to_path_buf().into();
             }
         }
+        let port_out = get_macport_openssl();
+        if !port_out.is_empty() {
+            return OsString::from(port_out);
+        }
     }
 
     try_pkg_config();
@@ -207,10 +211,7 @@ fn try_vcpkg() {
         .find_package("openssl");
 
     if let Err(e) = lib {
-        println!(
-            "note: vcpkg did not find openssl: {}",
-            e
-        );
+        println!("note: vcpkg did not find openssl: {}", e);
         return;
     }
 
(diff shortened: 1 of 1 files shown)


openssl-sys/build/find_normal.rs
```diff
@@ -50,6 +50,10 @@ fn find_openssl_dir(target: &str) -> OsString {
                 return homebrew.to_path_buf().into();
             }
         }
+        let port_out = get_macport_openssl();
+        if !port_out.is_empty() {
```
sfackler, line 54: It's generally preferable to use an Option to track optional data than the emptyness of a string, IMO.

openssl-sys/build/find_normal.rs
```diff
@@ -239,3 +240,45 @@ fn execute_command_and_get_output(cmd: &str, args: &[&str]) -> Option<String> {
...
+        .arg("installed")
+        .arg("openssl")
+        .output();
+    if out.is_ok() {
+        let res = out.ok().unwrap();
+        let outputs = std::str::from_utf8(&res.stdout).unwrap();
+        let version = get_macport_openssl_version(outputs);
+        
+        if version.is_some()
+            && version.unwrap() >= std::string::String::from("@1.1") {
```
sfackler, line 259: Trying to compare versions as strings seems suspicious.

toshiyuki-rs: I revised some lines to compare version rather than string. I chose the way to write simple version struct for mac port openssl rather than rust semver. Semver did not parse mac port openssl version like "1.1.1h".

Issue #1268, opened 2020-05-05
AES OCB mode


hargoniX:
In #864 was a very brief mention of OCB mode, apart from that I can't find anything in the issue tracker, hence my question, are there any plans to include OCB mode into this crate / any reason it is not there already?

If the reason it isn't there yet is just that nobody thought about picking it up yet I'd happily look into it.

sfackler:
There's no particular reason it's omitted - feel free to add!

hargoniX:
A question regarding tests, I am seeing [this](https://github.com/openssl/openssl/blob/master/test/recipes/30-test_evp_data/evpciph.txt#L1064`)  set of test data in OpenSSL it only contains tests for 128 bits though. I also did find [this](https://web.cs.ucdavis.edu/~rogaway/ocb/ocb-test.htm) page by the inventor of OBC mode. However I can't see any aad data in the values provided here so I'm a little unsure on how to work with this test data. Would someone be able to tell me?

sfackler:
Adding a test with OpenSSL's values seems fine I think.

PR #1270, opened 2020-05-08
Add support for AES-OCB mode

hargoniX:
See https://github.com/sfackler/rust-openssl/issues/1268#issuecomment-625976521 as to why there are only 128 bit tests.

closes: #1268

### DIFF
--- a/openssl/src/symm.rs
+++ b/openssl/src/symm.rs
@@ -130,6 +130,10 @@ impl Cipher {
         unsafe { Cipher(ffi::EVP_aes_128_ofb()) }
     }
 
+    pub fn aes_128_ocb() -> Cipher {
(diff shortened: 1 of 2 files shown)


sfackler:
It looks like you'll need to flag OCB ciphers off on older OpenSSLs.

openssl/src/symm.rs
```diff
@@ -130,6 +130,12 @@ impl Cipher {
         unsafe { Cipher(ffi::EVP_aes_128_ofb()) }
     }
 
+    /// Requires OpenSSL 1.1.0 or newer.
+    #[cfg(any(ossl110))]
```
sfackler, line 134: The any bit is not necessary for these cfgs.

PR #1274, opened 2020-05-21
Upgrade foreign-types to 0.5

sturmsebastian:
Version 0.5 adds the into_ptr() function on foreign-types i.e. it releases the ownership of the raw pointer. Also changed cvt_p to return NonNull, since foreign-types makes heavy use of it


sfackler:
This is a breaking change, so it's not going to be able to happen for a while.

PR #1275, opened 2020-05-21
Add X25519 and X448 key types. Add raw key import and export functions.

sturmsebastian:


openssl/src/pkey.rs
```diff
@@ -246,6 +250,58 @@ where
...
     }
+
+    /// Retrieves the length of the raw public key.
+    /// Please note that retrieving the raw public keys is only supported on keys of type
+    /// Ed25519, Ed448, X25519, X448.
+    /// This function will return 32 for Ed25519 and X25519 keys and 57 for Ed448 and X448 keys.
+    #[cfg(ossl111)]
+    pub fn get_raw_public_key_len(&self) -> Result<usize, ErrorStack> {
+        match self.id() {
+            Id::ED25519 | Id::X25519 => return Ok(32),
```
sfackler, line 261: Why are these hardcoded?

openssl/src/pkey.rs
```diff
@@ -246,6 +250,58 @@ where
...
+    /// For Ed25519 and X25519 keys the buffer needs to be at least 32 bytes long for Ed448 and
+    /// X448 keys 57 bytes.
+    #[cfg(ossl111)]
+    pub fn get_raw_public_key(&self, buf: &mut [u8]) -> Result<usize, ErrorStack> {
+        let mut len = buf.len();
+        unsafe {
+            cvt(ffi::EVP_PKEY_get_raw_public_key(
+                self.as_ptr(),
+                buf.as_mut_ptr(),
+                &mut len as *mut _))?;
```
sfackler, line 289: This cast isn't necessary.

openssl/src/pkey.rs
```diff
@@ -246,6 +250,58 @@ where
...
+            Id::ED448 | Id::X448 => return Ok(57),
+            _ => ()
+        }
+        // currently unreachable. May be useful in future extensions.
+        unsafe {
+            let mut len = 0;
+            cvt(ffi::EVP_PKEY_get_raw_public_key(
+                self.as_ptr(),
+                ptr::null_mut(),
+                &mut len as *mut _))?;
```
sfackler, line 271: Ditto

openssl/src/pkey.rs
```diff
@@ -621,6 +777,45 @@ impl PKey<Public> {
         PKey<Public>,
         ffi::d2i_PUBKEY
     }
+
+    #[cfg(ossl111)]
+    fn import_raw_public_key(typ: c_int, key: &[u8]) -> Result<PKey<Public>, ErrorStack> {
```
sfackler, line 782: I would just make this public and take a Type, and then remove the other constructors that delegate to it.

openssl/src/pkey.rs
sturmsebastian: RFC 8032 requires raw private/public keys for (X|Ed)(25519|448) to have the specified lengths. So there is no point in querying OpenSSL. It just saves a jump into the library and an error check. But I don't mind skipping it.

sfackler: Querying OpenSSL is literally the purpose of this library.

PR #1282, opened 2020-05-31
Add Debug trait for X509 and other types.

jsha:
This currently leaves out at least two useful things:
 - The detailed SubjectPublicKeyInfo, e.g. the modulus of RSA keys.
 - Extensions.

This also tweaks a couple of existing Display traits. The existing traits could
return errors if, for instance, an Asn1TimeRef were uninitialized, causing a
panic. Per the `fmt` docs (https://doc.rust-lang.org/std/fmt/index.html#formatting-traits):

> Formatting implementations should ensure that they propagate errors from the Formatter (e.g., when calling write!). However, they should never return errors spuriously. That is, a formatting implementation must and may only return an error if the passed-in Formatter returns an error.

### DIFF
--- a/openssl/src/asn1.rs
+++ b/openssl/src/asn1.rs
@@ -67,15 +67,17 @@ foreign_type_and_impl_send_sync! {
 impl fmt::Display for Asn1GeneralizedTimeRef {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         unsafe {
-            match MemBio::new() {
+            let mem_bio = match MemBio::new() {
+                Err(_) => return f.write_str(""),
+                Ok(m) => m,
+            };
+            let print_result = cvt(ffi::ASN1_GENERALIZEDTIME_print(
+                mem_bio.as_ptr(),
+                self.as_ptr(),
+            ));
+            match print_result {
                 Err(_) => f.write_str(""),
-                Ok(mem_bio) => match cvt(ffi::ASN1_GENERALIZEDTIME_print(
-                    mem_bio.as_ptr(),
-                    self.as_ptr(),
-                )) {
-                    Err(_) => f.write_str(""),
-                    Ok(_) => f.write_str(str::from_utf8_unchecked(mem_bio.get_buf())),
-                },
+                Ok(_) => f.write_str(str::from_utf8_unchecked(mem_bio.get_buf())),
             }
         }
     }
@@ -211,12 +213,14 @@ impl<'a> PartialOrd<Asn1Time> for &'a Asn1TimeRef {
 impl fmt::Display for Asn1TimeRef {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         unsafe {
-            match MemBio::new() {
+            let mem_bio = match MemBio::new() {
+                Err(_) => return f.write_str("error"),
+                Ok(m) => m,
+            };
+            let print_result = cvt(ffi::ASN1_TIME_print(mem_bio.as_ptr(), self.as_ptr()));
+            match print_result {
                 Err(_) => f.write_str("error"),
-                Ok(mem_bio) => match cvt(ffi::ASN1_TIME_print(mem_bio.as_ptr(), self.as_ptr())) {
-                    Err(_) => f.write_str("error"),
-                    Ok(_) => f.write_str(str::from_utf8_unchecked(mem_bio.get_buf())),
-                },
+                Ok(_) => f.write_str(str::from_utf8_unchecked(mem_bio.get_buf())),
             }
         }
     }
--- a/openssl/src/asn1.rs
+++ b/openssl/src/asn1.rs
@@ -68,7 +68,7 @@ impl fmt::Display for Asn1GeneralizedTimeRef {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         unsafe {
             let mem_bio = match MemBio::new() {
-                Err(_) => return f.write_str(""),
+                Err(_) => return f.write_str("error"),
                 Ok(m) => m,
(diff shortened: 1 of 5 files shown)


openssl/src/asn1.rs
```diff
@@ -67,12 +67,16 @@ foreign_type_and_impl_send_sync! {
 impl fmt::Display for Asn1GeneralizedTimeRef {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         unsafe {
-            let mem_bio = MemBio::new()?;
-            cvt(ffi::ASN1_GENERALIZEDTIME_print(
-                mem_bio.as_ptr(),
-                self.as_ptr(),
-            ))?;
-            write!(f, "{}", str::from_utf8_unchecked(mem_bio.get_buf()))
+            match MemBio::new() {
```
sfackler, line 70: Can you split this up into multiple statements rather than one nested expression? You could also just propagate errors upstream, though I guess it makes sense to avoid failing the whole Display call.

openssl/src/asn1.rs
```diff
@@ -207,13 +211,23 @@ impl<'a> PartialOrd<Asn1Time> for &'a Asn1TimeRef {
 impl fmt::Display for Asn1TimeRef {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         unsafe {
-            let mem_bio = MemBio::new()?;
-            cvt(ffi::ASN1_TIME_print(mem_bio.as_ptr(), self.as_ptr()))?;
-            write!(f, "{}", str::from_utf8_unchecked(mem_bio.get_buf()))
+            match MemBio::new() {
```
sfackler, line 214: Similarly here

openssl/src/pkey.rs
```diff
@@ -286,6 +287,27 @@ where
...
+            Id::DSA => "DSA",
+            Id::DH => "DH",
+            Id::EC => "EC",
+            #[cfg(ossl111)]
+            Id::ED25519 => "Ed25519",
+            #[cfg(ossl111)]
+            Id::ED448 => "Ed448",
+            _ => "unknown",
+        };
+        fmt.debug_struct("public_key")
```
sfackler, line 304:
```suggestion
        fmt.debug_struct("PKey")
```

openssl/src/x509/mod.rs
```diff
@@ -671,6 +671,35 @@ impl Clone for X509 {
...
+                Err(_) => "".to_string(),
+            },
+            Err(_) => "".to_string(),
+        };
+        let mut debug_struct = formatter.debug_struct("X509");
+        debug_struct.field("serial_number", &serial);
+        debug_struct.field("signature_algorithm", &self.signature_algorithm().object());
+        debug_struct.field("issuer", &self.issuer_name());
+        debug_struct.field("subject", &self.subject_name());
+        if let Some(subject_alt_names) = &self.subject_alt_names() {
```
sfackler, line 688: Might also make sense to include the common name.

openssl/src/x509/mod.rs
```diff
@@ -1298,6 +1339,25 @@ impl GeneralNameRef {
...
+            return formatter.write_str(dnsname);
+        }
+        if let Some(uri) = self.uri() {
+            return formatter.write_str(uri);
+        }
+        if let Some(ipaddress) = self.ipaddress() {
+            let result = String::from_utf8_lossy(ipaddress);
+            return formatter.write_str(&result);
+        }
+        Ok(())
```
sfackler, line 1357: This should probably print something in the default case.

openssl/src/x509/mod.rs
jsha: The common name is printed as part of the Subject.

openssl/src/asn1.rs
```diff
@@ -67,12 +67,18 @@ foreign_type_and_impl_send_sync! {
 impl fmt::Display for Asn1GeneralizedTimeRef {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         unsafe {
-            let mem_bio = MemBio::new()?;
-            cvt(ffi::ASN1_GENERALIZEDTIME_print(
+            let mem_bio = match MemBio::new() {
+                Err(_) => return f.write_str(""),
```
sfackler, line 71: Probably best to write "error" here like the other impls.

openssl/src/asn1.rs
```diff
@@ -67,12 +67,18 @@ foreign_type_and_impl_send_sync! {
...
+                Ok(m) => m,
+            };
+            let print_result = cvt(ffi::ASN1_GENERALIZEDTIME_print(
                 mem_bio.as_ptr(),
                 self.as_ptr(),
-            ))?;
-            write!(f, "{}", str::from_utf8_unchecked(mem_bio.get_buf()))
+            ));
+            match print_result {
+                Err(_) => f.write_str(""),
```
sfackler, line 79: And here.

sfackler:
LGTM with the one last change and the tests fixed.

jsha:
Made those changes. I'm not sure what's up with the tests. For instance, the Windows build has this error:

```
..\src\vcpkg\metrics.cpp(44): error C2127: 'write_byte': illegal initialization of 'constexpr' entity with a non-constant expression [C:\projects\vcpkg\toolsrc\vcpkglib\vcpkglib.vcxproj]
  ..\src\vcpkg\metrics.cpp(44): note: failure was caused by evaluating a lambda expression
```

Any tips on how to figure out what's wrong?

sfackler:
Oh sorry, yeah you can ignore that Windows test. I was talking about the failing circle tests: https://github.com/sfackler/rust-openssl/pull/1282/checks?check_run_id=729802154

PR #1290, opened 2020-06-05
Fix private key serializtion in HasPublic impl for DsaRef

Leo1003:
The `private_key_to_pem!` macro for `Dsa` is wrongly placed in `HasPublic` impl block. I moved it to the correct `HasPrivate` block. People can no longer call `private_key_to_pem()` on `Dsa<Public>`.

P.s. By `git blame`, it seems that this was my fault...  :sweat_smile:

### DIFF
--- a/openssl/src/dsa.rs
+++ b/openssl/src/dsa.rs
@@ -80,26 +80,6 @@ impl<T> DsaRef<T>
 where
     T: HasPublic,
 {
-    private_key_to_pem! {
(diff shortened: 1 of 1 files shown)


sfackler:
Thanks! If someone called the methods with a key missing private components before, do you know what happened? Did it return an error or segfault?

sfackler:
Ok, it looks like it just serializes a little bit of garbage.

PR #1313, opened 2020-07-15
Fix SslStream drop bug when create by from_raw_parts function

vkill:
Hi there

Please review it, thank you.

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -3435,6 +3435,7 @@ impl<S> MidHandshakeSslStream<S> {
 pub struct SslStream<S> {
     ssl: ManuallyDrop<Ssl>,
     method: ManuallyDrop<BioMethod>,
+    skip_drop_ssl: bool,
     _p: PhantomData<S>,
 }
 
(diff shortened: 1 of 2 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -3483,7 +3487,9 @@ impl<S: Read + Write> SslStream<S> {
     /// The caller must ensure the pointer is valid.
     pub unsafe fn from_raw_parts(ssl: *mut ffi::SSL, stream: S) -> Self {
```
sfackler, line 3488: This method needs to take ownership of the SSL.

vkill: The following is an example, maybe the scene is wrong, the ssl ptr should from ffi. is that what you mean?

```rust
use std::io;
use std::net::TcpStream;

use foreign_types::ForeignTypeRef;
use openssl::ssl::{SslConnector, SslMethod, SslStream, SslVerifyMode};

fn main() -> io::Result<()> {
    let tcp_stream = TcpStream::connect("echo.websocket.org:443")?;

    let mut ssl_connector_build = SslConnector::builder(SslMethod::tls()).unwrap();
    ssl_connector_build.set_verify(SslVerifyMode::NONE);
    let ssl_connector = ssl_connector_build.build();
    let ssl_stream = ssl_connector
        .connect("echo.websocket.org", tcp_stream)
        .unwrap();

    let ssl_ptr = ssl_stream.ssl().as_ptr();
    let tcp_stream = ssl_stream.get_ref();

    let _ = unsafe { SslStream::from_raw_parts(ssl_ptr, tcp_stream) };

    Ok(())
}

```

sfackler: You cannot share an SSL between two streams.

vkill: Oh, I see, thank you.

PR #1329, opened 2020-08-11
Expose subject_name_hash

marcelbuesing:


openssl/src/x509/mod.rs
```diff
@@ -398,6 +398,11 @@ impl X509Ref {
         }
     }
 
+    /// Returns the hash of the certificates subject
+    pub fn subject_name_hash(&self) -> u64 {
```
sfackler, line 402: X509_subject_name_hash uses a SHA1 hash, so this should return a `u32` and cast below.

marcelbuesing: Thanks. I just updated this.

PR #1330, opened 2020-08-12
Make openssl-errors tests work on Windows

mati865:
With that change running `cargo test` in the root passes for both i686 and x86_64 windows-gnu targets.


sfackler:
Thanks! I should get CI running openssl-errors tests on the Windows builds.

PR #1332, opened 2020-08-25
Added support for HKDF

Gematik-Entwicklung:
Added support for HKDF calculation using a similar approach as the `Deriver` does.

Remark: Semse that the `Deriver` is missing a `Drop` implementation, isn't it?


openssl/src/hkdf.rs
```diff
@@ -0,0 +1,109 @@
...
+            .map(Hkdf)
+            .and_then(|ctx| cvt(ffi::EVP_PKEY_derive_init(ctx.0)).map(|_| ctx))?;
+
+            cvt(ffi::EVP_PKEY_CTX_set_hkdf_md(ret.0, hash.as_ptr()))?;
+
+            Ok(ret)
+        }
+    }
+
+    pub fn set_mode(self, mode: Mode) -> Result<Self, ErrorStack> {
```
adamreichold, line 36: I am not sure if by-value builds are a good choice here as existing builds like `X509Builder` use the by-mut-ref pattern instead. But more importantly, there is not way to recover the value if there is recoverable error (e.g. an unsupported and optional operation) while using a by-value fallible builder which does not return itself on the error arm as well (like [`try_send`](https://doc.rust-lang.org/stable/std/sync/mpsc/struct.SyncSender.html#method.try_send) from the standard library).

adamreichold:
There also seems to be a lot of overlap with #1234 here.

openssl/src/hkdf.rs
adamreichold: Looking at #1234, it seems questionable whether a builder pattern is applicable at all as these parameters are not optional or have sensible defaults. This could be encoded via a type state, but that feels like unhelpful ceremony compared to a free function.

openssl/src/hkdf.rs
```diff
@@ -0,0 +1,109 @@
...
+
+impl Hkdf {
+    pub fn new(hash: MessageDigest) -> Result<Self, ErrorStack> {
+        unsafe {
+            let ret = cvt_p(ffi::EVP_PKEY_CTX_new_id(
+                ffi::EVP_PKEY_HKDF,
+                ptr::null_mut(),
+            ))
+            .map(Hkdf)
+            .and_then(|ctx| cvt(ffi::EVP_PKEY_derive_init(ctx.0)).map(|_| ctx))?;
```
adamreichold, line 28: Why use combinators here but a regular early return in line 30?

openssl/src/hkdf.rs
```diff
@@ -0,0 +1,109 @@
...
+
+        unsafe {
+            cvt(ffi::EVP_PKEY_CTX_add1_hkdf_info(self.0, ptr, len as c_int))?;
+        }
+
+        Ok(self)
+    }
+
+    pub fn derive(&mut self, mut key_len: usize) -> Result<Vec<u8>, ErrorStack> {
+        let mut buf = Vec::new();
```
adamreichold, line 90: This could be written more succinctly as `let mut buf = vec![0; key_len]`.

PR #1335, opened 2020-08-31
Add from_der and to_der to PKCS7

m-seker:
https://www.openssl.org/docs/man1.1.0/man3/d2i_PKCS7.html
https://www.openssl.org/docs/man1.1.0/man3/i2d_PKCS7.html


sfackler:
Thanks!

PR #1336, opened 2020-09-01
Add wrappers for Time-Stamp Protocol (RFC 3161) implementation of OpenSSL

krisztian-kovacs:
The aim of these patch set is to cover enough functionality to implement a client interfacing with a time-stamping authority to request and verify timestamps.

This pull request is more of a request to review -- I'm a first time contributor to rust-openssl. It has only been tested with OpenSSL 1.1.0.


sfackler:
CI is unhappy with the use of TryFrom.

krisztian-kovacs:
Indeed, I've seen the CI issue yesterday evening -- will come up with a fix today. Just to confirm: is 1.31 the MSRV rust-openssl is targeting?

krisztian-kovacs:
It took me some time to get the FFI bindings right for all versions of OpenSSL and LibreSSL, but hopefully it's OK now.

krisztian-kovacs:
Does the code look reasonable? Or is there something I could make to improve it?

openssl/src/ts.rs
```diff
@@ -0,0 +1,457 @@
...
+    pub struct TsMsgImprint;
+
+    /// Reference to `TsMsgImprint`.
+    pub struct TsMsgImprintRef;
+}
+
+/// A builder used to construct a `TsMsgImprint`
+pub struct TsMsgImprintBuilder(TsMsgImprint);
+
+impl TsMsgImprintBuilder {
```
sfackler, line 33: TS_MSG_IMPRINT isn't reference counted, so there's no need for a separate builder type.

openssl/src/ts.rs
```diff
@@ -0,0 +1,457 @@
...
+            ffi::init();
+            let imprint: *mut ffi::TS_MSG_IMPRINT = cvt_p(ffi::TS_MSG_IMPRINT_new())?;
+            Ok(TsMsgImprintBuilder(TsMsgImprint::from_ptr(imprint)))
+        }
+    }
+
+    /// Sets the algorithm identifier of the message digest algorithm.
+    ///
+    /// This corresponds to `TS_MSG_IMPRINT_set_algo`.
+    pub fn set_algo(&mut self, digest: &MessageDigest) -> Result<(), ErrorStack> {
```
sfackler, line 46: It seems like this should take an X509Algorithm rather than a MessageDigest.

openssl/src/ts.rs
```diff
@@ -0,0 +1,457 @@
...
+        ///
+        /// [`i2d_TS_REQ`]: https://www.openssl.org/docs/man1.1.0/man3/i2d_TS_REQ.html
+        to_der,
+        ffi::i2d_TS_REQ
+    }
+}
+
+pub struct TsReqBuilder(TsReq);
+
+impl TsReqBuilder {
```
sfackler, line 125: Similarly, no need for a builder.

openssl/src/ts.rs
```diff
@@ -0,0 +1,457 @@
...
+    pub struct TsVerifyContextRef;
+}
+
+impl TsVerifyContext {
+    /// Construct a verify context from a timestamping request.
+    ///
+    /// Corresponds to `TS_REQ_to_TS_VERIFY_CTX`.
+    pub fn from_req(request: &TsReqRef) -> Result<TsVerifyContext, ErrorStack> {
+        unsafe {
+            ffi::init();
```
sfackler, line 267: This isn't necessary since a TsReqRef already exists.

openssl/src/ts.rs
```diff
@@ -0,0 +1,457 @@
...
+            let ctx: *mut ffi::TS_VERIFY_CTX = cvt_p(ffi::TS_VERIFY_CTX_new())?;
+            let ctx = cvt_p(ffi::TS_REQ_to_TS_VERIFY_CTX(request.as_ptr(), ctx))?;
```
sfackler, line 269: The TS_VERIFY_CTX is leaked if this fails.

openssl/src/ts.rs
```diff
@@ -0,0 +1,457 @@
...
+            let bio = ::bio::MemBioSlice::new(&der)?;
+            let response = cvt_p(ffi::TS_RESP_create_response(self.as_ptr(), bio.as_ptr()))?;
+            Ok(TsResp::from_ptr(response))
+        }
+    }
+}
+
+pub struct TsRespContextBuilder(TsRespContext);
+
+impl TsRespContextBuilder {
```
sfackler, line 302: Probably same here WRT not needing a builder.

openssl/src/ts.rs
krisztian-kovacs: Hmm, in that case I think it should be possible to easily construct an X509Algorithm from a MessageDigest. Now there's no public interface for constructing an X509Algorithm at all.
Also, according to RFC 3161 only message digest algorithms are valid here: "The hash algorithm indicated in the hashAlgorithm field SHOULD be a known hash algorithm (one-way and collision resistant)." -- that's why I thought that using a MessageDigest on this interface would be more precise and useful.

openssl/src/ts.rs
krisztian-kovacs: Fixed, thanks!

openssl/src/ts.rs
krisztian-kovacs: Fixed, thanks!

openssl/src/ts.rs
krisztian-kovacs: Removed, thanks!

openssl/src/ts.rs
krisztian-kovacs: Fixed, thanks. It seems there's no point in constructing a verify context and passing it to TS_REQ_to_TS_VERIFY_CTX() anyway and we're better off with passing in NULL.

openssl/src/ts.rs
krisztian-kovacs: Fixed, thanks for the review!

openssl/src/ts.rs
krisztian-kovacs: Oh, and OpenSSL is inconsistent here: TS_MSG_IMPRINT_set_algo() uses X509_ALGOR, but its counterpart on the server side, TS_RESP_CTX_add_md() takes an EVP_MD. The server-side only accepts imprints with digest algorithms explicitly added via TS_RESP_CTX_add_md().

krisztian-kovacs:
I've rebased these changes to the latest master.
@sfackler do you see any outstanding issues that prevent merging this PR?

Kijewski:
Hi, can you y'all tell me what the state of the PR is? I'd like to use this feature in a project of mine. (No rush, of course! :) )

dfintha:
Hi, I'm also interested, when this PR might get merged, it would be pretty useful. :)

JM4ier:
@sfackler does this PR still have work that needs to be done before it can be merged? (apart from the merge conflict I mean)

If so, I'll gladly implement it on a separate PR, as timestamping is a feature I need.

PR #1338, opened 2020-09-03
Add constructor for creating Asn1Object from a textual OID representation

krisztian-kovacs:
This is effectively a wrapper for OBJ_txt2obj.


sfackler:
Thanks!

PR #1339, opened 2020-09-07
Implement signing a CSR with a CA certificate.

### LINKED ISSUE #345: Creating an X.509 certificate from a CSR
Right now `X509Generator` only creates self-signed certificates. I'd like to be able to use a self-signed certificate to create a new certificate from a CSR. If I were to work on some new APIs to do this, does it make any sense to try to add this additional functionality to `X509Generator`? I'm imagining ultimately being able to do something like this:

``` rust
fn generate_server_cert() -> Result<X509> {
    let (ca_cert, ca_key) = X509Generator::new().generate().unwrap();
    let mut server_key = PKey::new();
    server_key.gen(2048);
    let csr = X509Generator::new().add_name("CN".to_owned(), "example.com".to_owned()).request(&server_key).unwrap();
    X509Generator::new().sign_cert(&ca_key, &csr)
}
```

`X509Generator` seems very awkward for this type of API—it's not clear exactly what its responsibility is. Any suggestions on how to approach this?

griwes:
The basic idea is taken from how openssl's x509 app does it: create a
new certificate from the data in the CSR and sign it with the CA's key.

Fixes #345.


sfackler:
I generally prefer to keep the API fairly close to direct bindings to OpenSSL functions. I think  this would make sense as documentation or an example though.

griwes:
Well, the downside of that approach is that it requires re-typing it for every project; and since there were already people asking for an API like this in issues on this repo, I thought that getting it in would make sense. But if you'd rather not merge this, I get it, and it's fine - it's just that everybody will need to end up with their own copy of this until someone else picks it up, because I personally have no intention of maintaining a package for just this one function.

sfackler:
You could always propose exposing this method in OpenSSL itself, which we could then pick up. I am hesitant to land complex aggregate operations like this because it often turns out that the thing that worked for whoever added it is subtly different from what someone else needs.

griwes:
Yeah, no, I'm not touching OpenSSL itself with a ten-foot pole ;)

Issue #1308, opened 2020-06-29
Add getter for x509 certificate version


robin-kunzler:
The x509 certificate version can be set e.g. using [set_version](https://docs.rs/openssl/0.10.30/openssl/x509/struct.X509Builder.html#method.set_version). 

However, as far as I understand, there is currently no way of reading the x509 certificate version. Openssl provides the function [get_version](https://www.openssl.org/docs/man1.1.0/man3/X509_get_version.html). Would it make sense to also add this to the Rust API?

sfackler:
PRs welcomed!

igorty:
AFAIU no one's going to add `X509Ref::version()` yet. If so, will take care of next weekend.

max-heller:
@sfackler this can be closed now, no?

PR #1341, opened 2020-09-12
Add X509Ref::version

igorty:
Implements #1308.

### DIFF
--- a/openssl-sys/src/x509.rs
+++ b/openssl-sys/src/x509.rs
@@ -225,6 +225,7 @@ extern "C" {
     pub fn X509_get_pubkey(x: *mut X509) -> *mut EVP_PKEY;
 
     pub fn X509_set_version(x: *mut X509, version: c_long) -> c_int;
+    #[cfg(ossl110)]
     pub fn X509_get_version(x: *const X509) -> c_long;
     pub fn X509_set_serialNumber(x: *mut X509, sn: *mut ASN1_INTEGER) -> c_int;
     pub fn X509_get_serialNumber(x: *mut X509) -> *mut ASN1_INTEGER;
--- a/openssl-sys/src/x509.rs
+++ b/openssl-sys/src/x509.rs
@@ -225,6 +225,13 @@ extern "C" {
     pub fn X509_get_pubkey(x: *mut X509) -> *mut EVP_PKEY;
 
     pub fn X509_set_version(x: *mut X509, version: c_long) -> c_int;
+    /// Returns certificate version previously set through `X509_set_version` or the default which
+    /// is `0`.
+    ///
+    /// **Important.** The corresponding OpenSSL method is not&nbsp;expected to fail under its
+    /// correct implementation. Despite that, the method invocation passes through scopes which
+    /// _may_ result in error. Having that it is encouraged to validate returned value.
+    pub fn X509_get_version(x: *const X509) -> c_long;
     pub fn X509_set_serialNumber(x: *mut X509, sn: *mut ASN1_INTEGER) -> c_int;
     pub fn X509_get_serialNumber(x: *mut X509) -> *mut ASN1_INTEGER;
     pub fn X509_set_issuer_name(x: *mut X509, name: *mut X509_NAME) -> c_int;
--- a/openssl-sys/src/x509.rs
+++ b/openssl-sys/src/x509.rs
@@ -225,12 +225,6 @@ extern "C" {
     pub fn X509_get_pubkey(x: *mut X509) -> *mut EVP_PKEY;
 
     pub fn X509_set_version(x: *mut X509, version: c_long) -> c_int;
-    /// Returns certificate version previously set through `X509_set_version` or the default which
-    /// is `0`.
-    ///
-    /// **Important.** The corresponding OpenSSL method is not&nbsp;expected to fail under its
-    /// correct implementation. Despite that, the method invocation passes through scopes which
-    /// _may_ result in error. Having that it is encouraged to validate returned value.
     pub fn X509_get_version(x: *const X509) -> c_long;
     pub fn X509_set_serialNumber(x: *mut X509, sn: *mut ASN1_INTEGER) -> c_int;
     pub fn X509_get_serialNumber(x: *mut X509) -> *mut ASN1_INTEGER;
--- a/openssl/src/x509/mod.rs
+++ b/openssl/src/x509/mod.rs
@@ -73,29 +73,6 @@ mod tests {
             "Default certificate version is incorrect",
         );
     }
-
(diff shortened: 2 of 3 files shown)


openssl-sys/src/x509.rs
```diff
@@ -225,6 +225,13 @@ extern "C" {
     pub fn X509_get_pubkey(x: *mut X509) -> *mut EVP_PKEY;
 
     pub fn X509_set_version(x: *mut X509, version: c_long) -> c_int;
+    /// Returns certificate version previously set through `X509_set_version` or the default which
+    /// is `0`.
+    ///
+    /// **Important.** The corresponding OpenSSL method is not&nbsp;expected to fail under its
+    /// correct implementation. Despite that, the method invocation passes through scopes which
+    /// _may_ result in error. Having that it is encouraged to validate returned value.
+    pub fn X509_get_version(x: *const X509) -> c_long;
```
sfackler, line 234: There's no need to document function definitions in openssl-sys.

openssl/src/x509/mod.rs
```diff
@@ -548,6 +589,20 @@ impl X509Ref {
...
+    ///
+    /// Note that `0` return value stands for version 1, `1`&nbsp;&ndash; for version 2 and so on.
+    ///
+    /// This corresponds to [`X509_get_version`].
+    ///
+    /// [`X509_get_version`]: https://www.openssl.org/docs/man1.1.1/man3/X509_get_version.html
+    pub fn version(&self) -> Result<i64, ErrorStack> {
+        // Covered with `x509_ref_version()`, `x509_ref_version_no_version_set()`,
+        // `x509_ref_version_incorrect_version_set()` tests
+        unsafe { cvt_long_n(ffi::X509_get_version(self.as_ptr())) }
```
sfackler, line 603: The man page for X509_get_version does not claim it ever returns an error.

openssl-sys/src/x509.rs
igorty: Removed redundant doc.

openssl/src/x509/mod.rs
igorty: Removed error check.

igorty:
It seems that tests on all platforms (except macos, which I haven't got available for run) [pass](https://app.circleci.com/pipelines/github/igorty/rust-openssl). There are still couple of concerns I'll raise as comments to the pull request changes.

openssl-sys/src/x509.rs
```diff
@@ -225,6 +225,8 @@ extern "C" {
     pub fn X509_get_pubkey(x: *mut X509) -> *mut EVP_PKEY;
 
     pub fn X509_set_version(x: *mut X509, version: c_long) -> c_int;
+    #[cfg(ossl110)]
```
igorty, line 228: It is unclear to me how comes that `X509_set_version` and `X509Builder::set_version` methods do not need `#[cfg(ossl110)]` constraint declared. OpenSSL docs for [v1.0.2](https://www.openssl.org/docs/man1.0.2/man3/) do not list a possibility to set X509 certificate version.

openssl/src/x509/mod.rs
```diff
@@ -40,7 +40,63 @@ pub mod extension;
...
+    /// before a call to the tested method.
+    ///
+    /// _Note._ As for now this test is disabled since OpenSSL erroneously allows to set any number
+    /// as a certificate version. See corresponding [`issue`].
+    ///
+    /// [`issue`]: https://github.com/openssl/openssl/issues/12138
+    #[cfg(ossl110)]
+    #[test]
+    #[ignore]
+    fn x509_ref_version_incorrect_version_set() {
```
igorty, line 87: Turns out there is a bug in OpenSSL `X509_set_version(X509 *x, long version)` function. It allows to set any number as described in the corresponding [issue](https://github.com/openssl/openssl/issues/12138). Probably it's worth to consider a temporary check in `X509Builder::set_version` and `X509ReqBuilder::set_version` while the mentioned bug is present?

openssl/src/x509/mod.rs
```diff
@@ -548,6 +604,21 @@ impl X509Ref {
...
+    /// Note that `0` return value stands for version 1, `1`&nbsp;&ndash; for version 2 and so on.
+    ///
+    /// This corresponds to [`X509_get_version`].
+    ///
+    /// [`X509_get_version`]: https://www.openssl.org/docs/man1.1.1/man3/X509_get_version.html
+    #[cfg(ossl110)]
+    pub fn version(&self) -> i32 {
+        // Covered with `x509_ref_version()`, `x509_ref_version_no_version_set()`,
+        // `x509_ref_version_incorrect_version_set()` tests
+        unsafe { ffi::X509_get_version(self.as_ptr()) as i32 }
```
igorty, line 619: In order `version()` method to work on both x32 and x64 OSs a cast to i32 is performed. Is this appropriate solution or should there be some kind of check performed in order to determine which `c_long` type definition was taken: from `libc-0.2.76/src/unix/linux_like/linux/gnu/b64/x86_64/not_x32.rs` or `libc-0.2.76/src/unix/linux_like/linux/gnu/b64/x86_64/x32.rs`?

openssl-sys/src/x509.rs
sfackler: OpenSSL's man pages do not cover every single function the library exposes.

openssl/src/x509/mod.rs
sfackler: I don't think it makes sense to write a test for a case that the underlying library does not handle.

openssl/src/x509/mod.rs
sfackler: This is fine.

openssl/src/x509/mod.rs
igorty: Removed this test.

openssl/src/x509/mod.rs
```diff
@@ -548,6 +581,20 @@ impl X509Ref {
         }
     }
 
+    /// Returns certificate version. If this certificate has no explicit version set, it defaults to
+    /// version&nbsp;1.
```
sfackler, line 585:
```suggestion
    /// version 1.
```

openssl/src/x509/mod.rs
```diff
@@ -548,6 +581,20 @@ impl X509Ref {
         }
     }
 
+    /// Returns certificate version. If this certificate has no explicit version set, it defaults to
+    /// version 1.
+    ///
+    /// Note that `0` return value stands for version 1, `1`&nbsp;&ndash; for version 2 and so on.
```
sfackler, line 587:
```suggestion
    /// Note that `0` return value stands for version 1, `1` for version 2 and so on.
```

PR #1343, opened 2020-09-21
Remove non-cargo directive output of environment variables

dfreese:
I'm not sure of the history of this output, so I'm happy to be corrected,
however, this output doesn't have any effect, and it ends up printing
quite a bit of output when I am building.

I not using cargo, but rules_rust with bazel, so the suppression of this
may be different.


sfackler:
This is done intentionally to aid in debugging when the build script fails to find the openssl installation.

PR #1095, opened 2019-04-16
Add support for reading more X.509 extensions

landaire:
This PR as it currently stands adds support for reading the key usage, extended key usage, subject key identifier, and authority key identifier extensions. These correspond to the following new APIs:

- `X509::key_usage() -> Option<KeyUsageFlags>`
- `X509::ext_key_usage() -> Option<Stack<Asn1Object>>`
- `X509::authority_keyid() -> Option<AuthorityKeyId>`
- `X509::subject_keyid() -> Option<Asn1String>`
- `X509::extensions() -> Extensions`
- `X509::extension_from_nid(nid) -> Option<X509ExtensionRef>`

I'll go ahead and call out the following potential issues with this PR:

- The objects returned are of the non-ref type. e.g. `subject_keyid()` returns an `Option<Asn1String>` instead of an `Option<&Asn1StringRef>`. From what I can tell it's up to the caller of `X509_get_ext_d2i` to free the object, so I *think* usage of these types is correct but I'm not terribly certain
- I've created a new type, `KeyUsageFlags` in the X509 mod that will probably need to be relocated to the `x509::extensions` module, but wanted to discuss this implementation first. Currently the `x509::extensions::KeyUsage` struct is used for *creating* extensions and the APIs don't lend themselves very nicely for *reading* the extensions without adding `get_*` methods for every flag. Let me know what the preferred approach here is.
- `Extensions` struct naming probably isn't the best

Issue #661, opened 2017-07-14
Don’t use strings when generating X509 extensions


DemiMarie:
rust-openssl generates X509 extensions as strings, then calls into OpenSSL to convert those strings to extensions.  This opens up the possibility of injection attacks and is slow.  In particular, DNS names and email addresses are passed to OpenSSL without checking for validity.

OpenSSL supports marking extensions as critical directly, and I suspect that it also allows methods for constructing extensions that don’t rely on strings.

sfackler:
When implementing the extension stuff I looked around but could not find ways of programmatically creating extensions. If they do exist I'd love to use them!

DemiMarie:
<https://www.openssl.org/docs/man1.1.0/crypto/X509_EXTENSION_get_object.html> might be of use.  But it requires that one provides the extension contents as raw ASN.1 data.

DemiMarie:
<https://www.openssl.org/docs/manmaster/man3/X509V3_EXT_i2d.html> is what we were looking for.

DemiMarie:
To elaborate:

* Each type of X.509 extension supported by OpenSSL has its own extension-specific C struct.
* We can also serialize our own extensions if we wish.

Issue #373, opened 2016-04-06
Read x509 Extensions


dbrgn:
Is there already support for reading x509 extensions from an existing certificate? In my case I need to be able to read the X509v3 Authority Key Identifier.

sfackler:
I've just recently added support for access to subject alt names. Each extension needs specific handling unfortunately, but a PR to add accessors for Authority Key Identifiers would be great!

ignatenkobrain:
This would be nice to have definitely since I need it in my project ;)

ignatenkobrain:
@sfackler Any news here? it's almost 2 years =(

sfackler:
Pull requests welcome! I'm probably not going to have a chance to do it myself in the near term.

ignatenkobrain:
My opensslFu is in very bad shape, last time I was using openssl to get some x509/asn stuff I got headache. So I'm not the one who can send PR unfortunately =(

landaire:
I came across this while looking for support for getting extended key usage. I've never interfaced with the OpenSSL APIs before, but I added the functions required for getting arbitrary extension data here: https://github.com/landaire/rust-openssl/commit/612c1e9e8075a161fb61f7e3aea7749df3e19b77

I'm not exactly sure what the correct way to implement this specific request would be. OpenSSL does not expose APIs that wrap as getters for these fields. If there were I think the solution would be:

1. Create `AuthorityKeyId` and `AuthorityKeyIdRef` structs and write wrapper methods for the `keyid`, `issuer`, and `serial` fields as defined [here](https://github.com/openssl/openssl/blob/9fdcc21fdc9d148f78d9cd5be34030f38cc45812/include/openssl/x509v3.h#L206-L211).functions that provide you with these fields.
2. `X509_get_ext_d2i(x, ffi::NID_authority_key_identifier, null_mut, null_mut)`
3. Return the result as an `AuthorityKeyIdRef`

Since this isn't the case, I'm not sure what other OpenSSL APIs are available for extracting this data. `X509_EXTENSION_get_data` looks like it may be the right approach to decode an `ASN1_STRING`, `GENERAL_NAMES`, and `ASN1_INTEGER`, but I'm unsure.

ignatenkobrain:
https://github.com/rpm-software-management/librhsm/blob/master/rhsm/rhsm-product-certificate.c#L86

This is the code I'm using now in C.

landaire:
[...]

PR #1344, opened 2020-09-26
Various x509 bindings

stbuehler:
Related to #373, #661 and pull requests #1095 and #1123.

I'm not quite happy with #1095 (I think it needs a more general approach to handle extensions in other types as well), also the PR seems to be stuck somehow :)

I think a good start would be to add the low-level bindings - hopefully this should be easy to review.


stbuehler:
Update: reused some CRL and REVOKED details from #1123 (just the -sys parts though), and hopefully fixed the cfgs :)

PR #1346, opened 2020-09-28
Fix ASN1_OCTET_STRING in bindings

stbuehler:
In #1344 I used `ASN1_STRING` in places where openssl actually uses `ASN1_OCTET_STRING`; the checks don't detect that as in C this is a simple typedef.

As #1344 shouldn't be part of a release yet this still can be fixed afaict.


sfackler:
Thanks!

PR #1357, opened 2020-10-10
Silence clippy::match_like_matches_macro (would require rust 1.42)

stbuehler:
Clippy from rust 1.47.0 [promotes](https://rust-lang.github.io/rust-clippy/master/index.html#match_like_matches_macro) the [`matches!`](https://doc.rust-lang.org/core/macro.matches.html) macro, but using it would require rust 1.42.0.


sfackler:
Thanks!

PR #1359, opened 2020-10-11
Add and use const_ptr_api macro in openssl-sys

stbuehler:
Alternative approach to #1348; actually tries to parse fn declarations and magically inserts `*const` or `*mut`.

### DIFF
--- a/openssl-sys/src/asn1.rs
+++ b/openssl-sys/src/asn1.rs
@@ -61,14 +61,8 @@ extern "C" {
     pub fn ASN1_TIME_set_string_X509(s: *mut ASN1_TIME, str: *const c_char) -> c_int;
 }
 
-cfg_if! {
-    if #[cfg(any(ossl110, libressl280))] {
-        extern "C" {
-            pub fn ASN1_STRING_to_UTF8(out: *mut *mut c_uchar, s: *const ASN1_STRING) -> c_int;
-        }
-    } else {
-        extern "C" {
-            pub fn ASN1_STRING_to_UTF8(out: *mut *mut c_uchar, s: *mut ASN1_STRING) -> c_int;
-        }
+const_ptr_api! {
+    extern "C" {
+        pub fn ASN1_STRING_to_UTF8(out: *mut *mut c_uchar, s: #[const_ptr_if(any(ossl110, libressl280))] ASN1_STRING) -> c_int;
     }
 }
--- a/openssl-sys/src/bio.rs
+++ b/openssl-sys/src/bio.rs
@@ -60,17 +60,10 @@ pub unsafe fn BIO_get_mem_data(b: *mut BIO, pp: *mut *mut c_char) -> c_long {
     BIO_ctrl(b, BIO_CTRL_INFO, 0, pp as *mut c_void)
 }
 
-cfg_if! {
-    if #[cfg(any(ossl110, libressl280))] {
-        extern "C" {
-            pub fn BIO_s_file() -> *const BIO_METHOD;
-            pub fn BIO_new(type_: *const BIO_METHOD) -> *mut BIO;
-        }
-    } else {
-        extern "C" {
-            pub fn BIO_s_file() -> *mut BIO_METHOD;
-            pub fn BIO_new(type_: *mut BIO_METHOD) -> *mut BIO;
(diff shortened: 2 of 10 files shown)


stbuehler:
It seems I actually got it working. Any opinions regarding the syntax?

I think putting the cfg flags on the specific `*const`/`*mut` part is the right way to go; but it might work to use a little bit different syntax (e.g. `*maybe_const($cfgflags)`).

If you really want the cfg flags as attribute on the fn item, it probably would have to be the first attribute (otherwise it will take more recursions just to find it).

Also I think ctest will ignore any macro expansions failures (I had `$fn_vis:vis` in there, and it didn't check the item anymore, as the `vis` pattern was experimental back then; not sure about hitting the recursion limit); which is dangerous, because it simply won't generate tests for those items - but they are still present.

openssl-sys/src/asn1.rs
```diff
@@ -61,14 +61,8 @@ extern "C" {
...
-        extern "C" {
-            pub fn ASN1_STRING_to_UTF8(out: *mut *mut c_uchar, s: *const ASN1_STRING) -> c_int;
-        }
-    } else {
-        extern "C" {
-            pub fn ASN1_STRING_to_UTF8(out: *mut *mut c_uchar, s: *mut ASN1_STRING) -> c_int;
-        }
+const_ptr_api! {
+    extern "C" {
+        pub fn ASN1_STRING_to_UTF8(out: *mut *mut c_uchar, s: #[const_ptr_if(any(ossl110, libressl280))] ASN1_STRING) -> c_int;
```
sfackler, line 66: Can the macro be changed to only apply to the individual extern definition rather than wrapping the whole extern "C" block?

sfackler:
Nice! Yeah, the cfg syntax in your macro seems fine.

Just to make sure I understand - ctest does still include tests for the macro-defined functions, right? It just doesn't fail to compile if the macro is misused but the normal openssl-sys build would fail at that point anyway?

openssl-sys/src/asn1.rs
stbuehler: Syntactically speaking I think the macro can do that, but rustc doesn't like it; [playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=677f536409246611f7feb259db9d56fd) fails with `error[E0428]: the name `ASN1_STRING_to_UTF8` is defined multiple times`; but when looking at the generated code through `Tools` -> `Expand macros` it seems to be doing the right thing.

I'm pretty sure that is the reason why the `cfg_if!` blocks have internal `extern "C" {...}` blocks too; probably the same limitation. And even if `rustc` got fixed and you raise the MSRV, ctest will still fail, as it uses an old fork if a rustc internal crates to do parsing and macro expansion.

Passing `extern "C" { ... }` can be removed of course, but it can't be part of another `extern "C" { ... }` block and needs to emit `extern "C" { ... }` for the fn items.

stbuehler:
> Nice! Yeah, the cfg syntax in your macro seems fine.
> 
> Just to make sure I understand - ctest does still include tests for the macro-defined functions, right? It just doesn't fail to compile if the macro is misused but the normal openssl-sys build would fail at that point anyway?

After I removed the `$fn_vis:vis` handling and just hardcoded it as `pub` it did, yes. I think if the macro expansion fails it pretends the macro expanded to an empty output.

As ctest uses an ([unmaintained](https://github.com/serde-deprecated/syntex)) [old fork of a rustc internal crate](https://docs.rs/crate/syntex_syntax2/0.0.2) the current rustc won't necessarily fail (also it seems to use a [hardcoded recursion limit of 128](https://github.com/alexcrichton/ctest/blob/4fe24c45c3b840ecd389ea28a4ee0363c33ec45e/src/lib.rs#L910)).

openssl-sys/src/asn1.rs
sfackler: Ah bummer. This seems good then!

stbuehler:
macos seems to be down in circleci...

I also opened https://github.com/gnzlbg/ctest/pull/101 - if this gets merged and released, I think `systest` could be made more reliable to check all items (i.e. fail if macro expansion fails).

PR #1333, opened 2020-08-27
Support LibreSSL 3.3.1

vishwin:
(Hopefully, after CircleCI run) closes #1309. Supersedes #1310.

PR #1362, opened 2020-10-18
compiles on openbsd-current (6.7-6.8 interregnum)

elmarsto:
This small adjustment allows rust-openssl to build without fault on the latest OpenBSD snapshot, circa today.


sfackler:
The newest released version is 3.2.1, which is not yet a stable release so the entire 3.2.x series cannot yet be allowed in its entirety. #1333 adds support for 3.2.1, but numerous tests appear to be failing.

Issue #1370, opened 2020-11-12
Set ASN1 object to PRINTABLESTRING / (string_mask = nombstr)


JoshuaNitschke:
Hi,

I have a need to make certificate where the CommonName, Org, OrgUnit are PRINTABLESTRING instead of UTFSTRING.  I am willing to tackle this if the maintainer would give some pointers as to where the function should be implemented in the rust-openssl library.  If there is already a way to do this, I apologize for making an issue and would appreciate some guidance.

From looking at the openssl source code, I believe we just need to add a binding to:
[ASN1_STRING_set_default_mask_asc(const char *p)](https://github.com/openssl/openssl/blob/225c9660a5a3435d9bcfc9166b9f79f132996249/crypto/asn1/a_strnid.c#L48)

Here is the example of the output of a certificate made through the rust-openssl library, as you see the dnqualifier is set to `PRINTABLESTRING` but the other objects are not.  I need all these `UTF8STRING` objects to be `PRINTABLESTRING`.

**openssl asn1parse -inform PEM -in test.pem** 
```
    0:d=0  hl=4 l=1121 cons: SEQUENCE          
    4:d=1  hl=4 l= 841 cons: SEQUENCE          
    8:d=2  hl=2 l=   3 cons: cont [ 0 ]        
   10:d=3  hl=2 l=   1 prim: INTEGER           :02
   13:d=2  hl=2 l=   8 prim: INTEGER           :51C8A26799DA5F8A
   23:d=2  hl=2 l=  13 cons: SEQUENCE          
   25:d=3  hl=2 l=   9 prim: OBJECT            :sha256WithRSAEncryption
   36:d=3  hl=2 l=   0 prim: NULL              
   38:d=2  hl=2 l= 126 cons: SEQUENCE          
   40:d=3  hl=2 l=  21 cons: SET               
   42:d=4  hl=2 l=  19 cons: SEQUENCE          
   44:d=5  hl=2 l=   3 prim: OBJECT            :organizationName
   49:d=5  hl=2 l=  12 prim: UTF8STRING        :orgName.com
   63:d=3  hl=2 l=  21 cons: SET               
   65:d=4  hl=2 l=  19 cons: SEQUENCE          
   67:d=5  hl=2 l=   3 prim: OBJECT            :organizationalUnitName
   72:d=5  hl=2 l=  12 prim: UTF8STRING        :orgUnitName
   86:d=3  hl=2 l=  39 cons: SET               
[...]

PR #1371, opened 2020-11-12
Add additional function so that x509 name with specific type can be added

JoshuaNitschke:
Fixes issue #1370


JoshuaNitschke:
Looks like an [upgrade to cc crate](https://github.com/alexcrichton/cc-rs/commit/c668e9e4770fac31515fb999630c867e80efe2c0) is causing failures.

openssl/src/x509/mod.rs
```diff
@@ -827,6 +827,33 @@ impl X509NameBuilder {
...
+    /// Add a field entry by str with a specific type. (ex: 19 for V_ASN1_PRINTABLESTRING)
+    ///
+    /// This corresponds to [`X509_NAME_add_entry_by_txt`].
+    ///
+    /// [`X509_NAME_add_entry_by_txt`]: https://www.openssl.org/docs/man1.1.0/crypto/X509_NAME_add_entry_by_txt.html
+    pub fn append_entry_by_text_with_type(
+        &mut self,
+        field: &str,
+        value: &str,
+        ty: i32,
```
sfackler, line 839: This argument should be some kind of real type rather than a raw number.

JoshuaNitschke: Just to make sure I understand: you would like an enum type defined?

sfackler: Not exactly an enum type, but something like https://docs.rs/openssl/0.10.30/src/openssl/ssl/mod.rs.html#595

JoshuaNitschke: Thanks for the example, I'll get a revision in either tonight or tomorrow.

JoshuaNitschke:
@sfackler how's this?

JoshuaNitschke:
@sfackler is there anything you'd like changed?  My org has a production app pointed to my branch, which is less than ideal, so I'm willing to make whatever changes needed to get this merged.   I appreciate your time.

openssl/src/x509/mod.rs
```diff
@@ -878,6 +905,32 @@ impl X509NameBuilder {
         }
     }
 
+    /// Add a field entry by NID with a specific type.  See [Asn1Tag](openssl::asn1::Asn1Tag)
```
sfackler, line 908: You should be able to have append_entry_by_nid above just call this method.

sfackler:
Sorry for the delay!

JoshuaNitschke:
Thanks @sfackler!  Happy Holidays!

PR #1372, opened 2020-11-16
Add AEAD SRTP protection profiles

HyeonuPark:
Add missing AEAD SRTP profile constants. I tried to mimic the openssl header file structure for the -sys file. I'll _fix_ it if you don't like it.


sfackler:
It looks like those are not available on libressl, so you'll have to cfg them out when building against that.

HyeonuPark:
Added `#[cfg(ossl110)]`s. Hope it fixes all the tests.

PR #1373, opened 2020-11-16
Support for enabling CRL-based certificate revocation checking

max-heller:
Adds bindings for APIs needed to enable revocation checking with CRLs looked up by the `X509_LOOKUP_hash_dir` method, namely:
* `X509_LOOKUP` and `X509_LOOKUP_METHOD`
* `X509_VERIFY_PARAM` (setting verification flags)

There's a lot of API surface that still isn't covered, especially around lookup methods--this is just the minimal set of additions needed to get things working.

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -1342,6 +1342,18 @@ impl SslContextBuilder {
         unsafe { X509StoreBuilderRef::from_ptr_mut(ffi::SSL_CTX_get_cert_store(self.as_ptr())) }
     }
 
+    /// Returns a mutable reference to the X509 verification configuration.
+    ///
+    /// Requires OpenSSL 1.0.2 or newer.
+    ///
+    /// This corresponds to [`SSL_CTX_get0_param`].
+    ///
+    /// [`SSL_CTX_get0_param`]: https://www.openssl.org/docs/man1.0.2/ssl/SSL_CTX_get0_param.html
+    #[cfg(any(ossl102, libressl261))]
+    pub fn verify_param_mut(&mut self) -> &mut X509VerifyParamRef {
+        unsafe { X509VerifyParamRef::from_ptr_mut(ffi::SSL_CTX_get0_param(self.as_ptr())) }
+    }
+
     /// Sets the callback dealing with OCSP stapling.
     ///
     /// On the client side, this callback is responsible for validating the OCSP status response
@@ -1994,18 +2006,6 @@ impl SslContextRef {
     pub fn verify_mode(&self) -> SslVerifyMode {
         let mode = unsafe { ffi::SSL_CTX_get_verify_mode(self.as_ptr()) };
         SslVerifyMode::from_bits(mode).expect("SSL_CTX_get_verify_mode returned invalid mode")
-    }
-
-    /// Returns a mutable reference to the X509 verification configuration.
-    ///
-    /// Requires OpenSSL 1.0.2 or newer.
-    ///
-    /// This corresponds to [`SSL_CTX_get0_param`].
-    ///
-    /// [`SSL_CTX_get0_param`]: https://www.openssl.org/docs/man1.0.2/ssl/SSL_CTX_get0_param.html
-    #[cfg(any(ossl102, libressl261))]
-    pub fn param_mut(&mut self) -> &mut X509VerifyParamRef {
-        unsafe { X509VerifyParamRef::from_ptr_mut(ffi::SSL_CTX_get0_param(self.as_ptr())) }
     }
 }
 
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -1342,6 +1342,18 @@ impl SslContextBuilder {
         unsafe { X509StoreBuilderRef::from_ptr_mut(ffi::SSL_CTX_get_cert_store(self.as_ptr())) }
     }
 
+    /// Returns a reference to the X509 verification configuration.
+    ///
+    /// Requires OpenSSL 1.0.2 or newer.
+    ///
+    /// This corresponds to [`SSL_CTX_get0_param`].
+    ///
+    /// [`SSL_CTX_get0_param`]: https://www.openssl.org/docs/man1.0.2/ssl/SSL_CTX_get0_param.html
+    #[cfg(any(ossl102, libressl261))]
+    pub fn verify_param(&self) -> &X509VerifyParamRef {
+        unsafe { X509VerifyParamRef::from_ptr(ffi::SSL_CTX_get0_param(self.as_ptr())) }
+    }
+
     /// Returns a mutable reference to the X509 verification configuration.
     ///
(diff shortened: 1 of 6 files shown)


openssl/src/x509/store.rs
```diff
@@ -94,6 +94,80 @@ impl X509StoreBuilderRef {
...
+    ///
+    /// [`X509_LOOKUP_hash_dir`]: https://www.openssl.org/docs/man1.1.0/crypto/X509_LOOKUP_hash_dir.html
+    pub fn hash_dir() -> &'static X509LookupMethodRef {
+        unsafe { X509LookupMethodRef::from_ptr(ffi::X509_LOOKUP_hash_dir()) }
+    }
+}
+
+impl X509LookupRef {
+    /// Specifies a directory from which certificates and CRLs will be loaded
+    /// on-demand. Must be used with `X509Lookup::hash_dir`.
```
sfackler, line 138: This is a potential soundness concern if any other lookup method uses CTL codes that collide with X509_L_ADD_DIR

openssl/src/x509/verify.rs
```diff
@@ -23,6 +23,41 @@ bitflags! {
     }
 }
 
+bitflags! {
+    /// Flags used to verify an `X509` certificate chain.
+    pub struct X509VerifyFlags: c_ulong {
+        const X509_V_FLAG_CB_ISSUER_CHECK = ffi::X509_V_FLAG_CB_ISSUER_CHECK;
```
sfackler, line 29: The `X509_V_FLAG_` prefix isn't necessary.

openssl/src/ssl/mod.rs
```diff
@@ -1342,6 +1342,18 @@ impl SslContextBuilder {
...
 
+    /// Returns a mutable reference to the X509 verification configuration.
+    ///
+    /// Requires OpenSSL 1.0.2 or newer.
+    ///
+    /// This corresponds to [`SSL_CTX_get0_param`].
+    ///
+    /// [`SSL_CTX_get0_param`]: https://www.openssl.org/docs/man1.0.2/ssl/SSL_CTX_get0_param.html
+    #[cfg(any(ossl102, libressl261))]
+    pub fn verify_param_mut(&mut self) -> &mut X509VerifyParamRef {
```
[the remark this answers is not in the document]
max-heller: Yep

openssl/src/x509/verify.rs
max-heller: Makes sense, went through and removed them

max-heller:
Looks like the `windows-vcpkg` action is broken:
```
Run echo "::set-env name=VCPKG_ROOT::$Env:VCPKG_INSTALLATION_ROOT"
Error: Unable to process command '::set-env name=VCPKG_ROOT::C:\vcpkg' successfully.
Error: The `set-env` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```

openssl/src/x509/store.rs
```diff
@@ -94,6 +94,80 @@ impl X509StoreBuilderRef {
...
+    /// Adds a lookup method to the store.
+    ///
+    /// This corresponds to [`X509_STORE_add_lookup`].
+    ///
+    /// [`X509_STORE_add_lookup`]: https://www.openssl.org/docs/man1.1.1/man3/X509_STORE_add_lookup.html
+    pub fn add_lookup(
+        &mut self,
+        method: &X509LookupMethodRef,
+    ) -> Result<&mut X509LookupRef, ErrorStack> {
+        let lookup = unsafe { ffi::X509_STORE_add_lookup(self.as_ptr(), method.as_ptr()) };
```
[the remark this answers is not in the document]
max-heller: Agreed

openssl/src/x509/store.rs
max-heller: If I'm understanding the concern correctly, would it make sense to make `X509Lookup` generic over a type indicating which CTL codes it supports? For instance, `hash_dir` would return something like `X509LookupRef<Directory>`, and `add_dir` would only be implemented on `X509LookupRef<Directory>`

sfackler:
vcpkg is fixed on master now.

max-heller:
> vcpkg is fixed on master now.

Awesome. Could you re-run it here? Don't think GitHub lets me.

sfackler:
You'll need to rebase or merge master back into this branch to get the updated CI config I think.

openssl/src/x509/store.rs
sfackler: Yeah, that would solve it (though I'd call the type `HashDir` rather than `Directory` to match the constructor).

max-heller: Alright, implemented this idea. Decided to name it `AddDir` after `X509_L_ADD_DIR` since I think it makes more sense to have the generic parameter correspond to the control commands supported rather than the exact method it came from. This might be problematic if any lookup methods support multiple control commands--in that case, maybe we'd need traits indicating support for each--but I think it should be sufficient for now.

max-heller: > This is a potential soundness concern if any other lookup method uses CTL codes that collide with X509_L_ADD_DIR

Rereading this, I might've taken a step in the wrong direction here. Is what you're saying that two lookup methods might interpret the same command to mean different things? In that case, I'll switch the marker type back to `HashDir`, since that's what we're actually looking for.

sfackler: Yeah, we'll want the parameter to indicate what kind of lookup algorithm it is (HashDir) in this case, since that's what determines how the X509_LOOKUP_ctrl behaves. There could be other ctrl APIs that the hash_dir method supports and they'd all end up under this type parameter for example.

max-heller: Gotcha, done

openssl/src/x509/store.rs
```diff
@@ -94,6 +94,85 @@ impl X509StoreBuilderRef {
...
+}
+
+/// Marker type corresponding to the [`X509_LOOKUP_hash_dir`] lookup method.
+///
+/// [`X509_LOOKUP_hash_dir`]: https://www.openssl.org/docs/man1.1.0/crypto/X509_LOOKUP_hash_dir.html
+pub struct HashDir;
+
+impl X509Lookup<HashDir> {
+    /// Lookup method that loads certificates and CRLs on demand and caches
+    /// them in memory once they are loaded. As of OpenSSL 1.0.0, it also
```
sfackler, line 129: We don't support 0.9.8, so you can remove the `As of OpenSSL 1.0.0` bit.

sfackler:
LGTM with that one last doc tweak!

max-heller:
> LGTM with that one last doc tweak!

Awesome!

sfackler:
Thanks!

PR #1374, opened 2020-11-17
Add private key export to PKCS#8 in DER format

lydiabritt1987:
I need to export private keys in PKCS#8 DER format, currently the bindings only support exporting DER as a raw RSA key, or PEM in PKCS#8. At present I'm working around this by chopping off the header and trailer, then parsing the base64 back into DER, which feels a bit silly.

This pull request isn't finished; it has a memory leak still, which I'm not sure how to fix. Anyone able to help?

### DIFF
--- a/openssl/src/pkey.rs
+++ b/openssl/src/pkey.rs
@@ -289,6 +289,21 @@ where
         private_key_to_der,
         ffi::i2d_PrivateKey
     }
+
+    pub fn private_key_to_der_pkcs8(&self) -> Result<Vec<u8>, ::error::ErrorStack> {
+        unsafe {
+            let p8inf = cvt_p(ffi::EVP_PKEY2PKCS8(self.as_ptr()))?;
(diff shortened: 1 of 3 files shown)


openssl/src/pkey.rs
```diff
@@ -289,6 +289,21 @@ where
         private_key_to_der,
         ffi::i2d_PrivateKey
     }
+
+    pub fn private_key_to_der_pkcs8(&self) -> Result<Vec<u8>, ::error::ErrorStack> {
+        unsafe {
+            let p8inf = cvt_p(ffi::EVP_PKEY2PKCS8(self.as_ptr()))?;
```
sfackler, line 295: This library typically tries to map individual OpenSSL functions to individual Rust methods, so this would be split into a method that returns a `Pkcs8PrivKeyInfo`, and a method on that to turn that into DER.

PR #1375, opened 2020-11-17
Added missing email_protection method to ExtendedKeyUsage struct.

Superhepper:
The code appeared to be prepared to have such method on the ExtendedKeyUsage struct though it was not present. So I assumed it had just not been added yet. One could argue that the flag could be set through the ```other``` method and this is not necessary but then the code that refers to the ```emailProtection``` should be removed.

Signed-off-by: Jesper Brynolf <jesper.brynolf@gmail.com>


sfackler:
Thanks!

PR #1378, opened 2020-11-21
Openssl 1.0.1 domain name matching tweaks

iximeow:
This is two distinct changes:
* [`76d996c`](https://github.com/sfackler/rust-openssl/pull/1378/commits/76d996c380334217efba23ceddc050c045ababbb) makes hostname comparison case-insensitive. Any poor soul using openssl 1.0.1 may like to see this. It also adds some checks to show what hostnames do/don't match by `matches_dns`.
* [`159922b`](https://github.com/sfackler/rust-openssl/pull/1378/commits/159922b514979aa437060e906981da9bea72268b) then removes partial wildcard support from the openssl 1.0.1 custom verify routine. This disagrees with the original libcurl implementation but is consistent with openssl 1.0.2+ behavior.

I'd recommend reading the commits one after the other, particularly because if there's any hesitation about partial wildcards the first commit should be fine on its own.

edit2: oh, clippy did not like this. I didn't touch openssl-sys and I don't see recent changes to `openssl-sys/build/main.rs` so I'm not sure why clippy chose me to be mad at? :(

### DIFF
--- a/openssl/src/ssl/connector.rs
+++ b/openssl/src/ssl/connector.rs
@@ -498,7 +498,7 @@ cfg_if! {
                     hostname = &hostname[..hostname.len() - 1];
                 }
 
-                matches_wildcard(pattern, hostname).unwrap_or_else(|| pattern == hostname)
+                matches_wildcard(pattern, hostname).unwrap_or_else(|| pattern.eq_ignore_ascii_case(hostname))
             }
 
             fn matches_wildcard(pattern: &str, hostname: &str) -> Option<bool> {
@@ -542,7 +542,7 @@ cfg_if! {
                 };
 
                 // check that the non-wildcard parts are identical
-                if pattern[wildcard_end..] != hostname[hostname_label_end..] {
+                if !pattern[wildcard_end..].eq_ignore_ascii_case(&hostname[hostname_label_end..]) {
                     return Some(false);
                 }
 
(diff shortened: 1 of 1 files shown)


sfackler:
I'll fix the clippy bits - we use the current stable compiler so there tend to be some extra warnings when new stable releases happen.

openssl/src/ssl/connector.rs
```diff
@@ -552,12 +552,19 @@ cfg_if! {
                 let hostname_label = &hostname[..hostname_label_end];
 
                 // check the prefix of the first label
-                if !hostname_label.starts_with(wildcard_prefix) {
+                if !hostname_label[..wildcard_location].eq_ignore_ascii_case(wildcard_prefix) {
```
sfackler: Do we know at this point that hostname_label is at least as long as wildcard_location?

openssl/src/ssl/connector.rs
```diff
@@ -502,11 +502,6 @@ cfg_if! {
             }
 
             fn matches_wildcard(pattern: &str, hostname: &str) -> Option<bool> {
-                // internationalized domains can't involved in wildcards
```
sfackler, line 505: Why is this check removed?

sfackler: Oh right, because the xn-- prefix implies that a wildcard has to be partial.

sfackler:
The clippy error should be fixed if you rebase.

openssl/src/ssl/connector.rs
iximeow: indeed, we do not. Good catch. There's [a check for that case](https://github.com/sfackler/rust-openssl/pull/1378/commits/76d996c380334217efba23ceddc050c045ababbb#diff-1b6ced8dfb91d53c0d3c93878fca256ae1431551161033aa3a264992e018829bR555-R559) and [more tests](https://github.com/sfackler/rust-openssl/pull/1378/commits/76d996c380334217efba23ceddc050c045ababbb#diff-1b6ced8dfb91d53c0d3c93878fca256ae1431551161033aa3a264992e018829bR601-R605) for partial wildcards now.

iximeow:
Done, and CI agrees with these changes now \o/

sfackler:
Thanks!

PR #1384, opened 2020-12-04
Add encrypt module and RSA OAEP support

### LINKED ISSUE #1383: Add support for RSA OAEP in openssl crate
[Josekit](https://github.com/hidekatsu-izuno/josekit-rs) heavily relies on `openssl` crate, but [directly implements RSA OAEP encryption/decryption](https://github.com/hidekatsu-izuno/josekit-rs/blob/d878348669c30d5bcd4d860019595e3dafe6830b/src/jwe/alg/rsaes.rs#L553-L698) using `openssl-sys` and, obviously, `unsafe` code.

It should be considered to implement the same feature directly in `openssl` crate. If the code does not have any particular issue, it could be implemented without particular efforts and, at the same time, it will avoid others to re-implement the same feature with unsafe code. On the other hand, if there are some concerns about the soundness or security, people involved in `rust-openssl` have surely a strong know-how to correctly implement the feature and expose a safe and sound abstraction for the ecosystem.

If you want, I could try to send a PR for this, but I am a bit intimidated by crypto issues -- I never worked on this field!

dodomorandi:
Introduce a new module `encrypt`, with the two structs `Encrypter` and `Decrypter`, pretty similar to `Signer` and `Verifier`. With the introduction of  `EVP_PKEY_CTX_set_rsa_oaep_md` in `openssl-sys`, this new abstraction allows to perform RSA OAEP encryption and decryption without directly relying of `unsafe` calls.

As already discussed with @sfackler, there is a pretty good amount of code duplication, there is a good opportunity to use traits to abstract away _sign_ and _encrypt_ algorithms (i.e. OAEP can only be used for encryption, PSS only for signing). This PR could be a good starting point to analyze the current situation and, maybe, to understand how in the future the abstractions could be improved.

I would like some feedback, mostly because I could have missed some potential problem. Let me know what you think and just tell me what's missing.

Closes #1383 

**EDIT**: I forgot to mention: I _copied_ the API from `Signer`/`Verifier`, but I think that having to preallocate a buffer without knowing the exact size needed is pretty inconvenient. IMHO two versions of `encrypt` and `decrypt` should be provided, one that returns a `Vec<u8>`, the other that takes a buffer. I would like to hear your opinion about naming, because if I follow something like `encrypt(in: &[u8]) -> Vec<u8>` and `encrypt_buf(in: &[u8], out: &mut [u8])` (which would be my approach), I will immediately break the API convention of `Signer`/`Verifier`. Moreover, I also think that taking a `&mut Vec<u8>` could be convenient, to allow automatic buffer resizing, but on the other hand some questions would rise: the `vec` should be automatically cleared if non-empty? Or it should follow the same principles of [`Read::read_to_end`](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_to_end) and similar?

P.S.: there is an issue about invalid C compiler parameters, which breaks the CI because `-Werror` is enabled, you probably already know that.

## TODO
- [x] Module documentation

### DIFF
--- a/openssl/src/encrypt.rs
+++ b/openssl/src/encrypt.rs
@@ -1,0 +1,342 @@
+use std::{marker::PhantomData, ptr};
+
+use error::ErrorStack;
+use foreign_types::ForeignTypeRef;
+use hash::MessageDigest;
+use pkey::{HasPrivate, HasPublic, PKeyRef};
+use rsa::Padding;
+use {cvt, cvt_p};
+
+pub struct Encrypter<'a> {
+    pctx: *mut ffi::EVP_PKEY_CTX,
+    _p: PhantomData<&'a ()>,
+}
+
+unsafe impl<'a> Sync for Encrypter<'a> {}
+unsafe impl<'a> Send for Encrypter<'a> {}
+
+impl<'a> Drop for Encrypter<'a> {
+    fn drop(&mut self) {
+        unsafe {
+            ffi::EVP_PKEY_CTX_free(self.pctx);
+        }
+    }
+}
+
+impl<'a> Encrypter<'a> {
+    /// Creates a new `Encrypter`.
+    ///
+    /// OpenSSL documentation at [`EVP_PKEY_encrypt_init`].
+    ///
+    /// [`EVP_PKEY_encrypt_init`]: https://www.openssl.org/docs/manmaster/man3/EVP_PKEY_encrypt_init.html
+    pub fn new<T>(pkey: &'a PKeyRef<T>) -> Result<Encrypter<'a>, ErrorStack>
+    where
+        T: HasPublic,
+    {
+        Self::new_intern(pkey)
+    }
+
+    fn new_intern<T>(pkey: &'a PKeyRef<T>) -> Result<Encrypter<'a>, ErrorStack>
+    where
+        T: HasPublic,
+    {
+        unsafe {
+            ffi::init();
+
+            let pctx = cvt_p(ffi::EVP_PKEY_CTX_new(pkey.as_ptr(), ptr::null_mut()))?;
+            let r = ffi::EVP_PKEY_encrypt_init(pctx);
+            if r != 1 {
+                ffi::EVP_PKEY_CTX_free(pctx);
+                return Err(ErrorStack::get());
+            }
+
+            Ok(Encrypter {
+                pctx,
+                _p: PhantomData,
+            })
+        }
+    }
+
+    /// Returns the RSA padding mode in use.
+    ///
+    /// This is only useful for RSA keys.
+    ///
+    /// This corresponds to `EVP_PKEY_CTX_get_rsa_padding`.
+    pub fn rsa_padding(&self) -> Result<Padding, ErrorStack> {
+        unsafe {
+            let mut pad = 0;
+            cvt(ffi::EVP_PKEY_CTX_get_rsa_padding(self.pctx, &mut pad))
+                .map(|_| Padding::from_raw(pad))
+        }
+    }
+
+    /// Sets the RSA padding mode.
+    ///
+    /// This is only useful for RSA keys.
+    ///
+    /// This corresponds to [`EVP_PKEY_CTX_set_rsa_padding`].
+    ///
+    /// [`EVP_PKEY_CTX_set_rsa_padding`]: https://www.openssl.org/docs/man1.1.0/crypto/EVP_PKEY_CTX_set_rsa_padding.html
+    pub fn set_rsa_padding(&mut self, padding: Padding) -> Result<(), ErrorStack> {
+        unsafe {
+            cvt(ffi::EVP_PKEY_CTX_set_rsa_padding(
+                self.pctx,
+                padding.as_raw(),
+            ))
+            .map(|_| ())
+        }
+    }
+
+    /// Sets the RSA MGF1 algorithm.
+    ///
+    /// This is only useful for RSA keys.
+    ///
+    /// This corresponds to [`EVP_PKEY_CTX_set_rsa_mgf1_md`].
+    ///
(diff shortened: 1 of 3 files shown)


sfackler:
It looks like `EVP_PKEY_CTRL_RSA_OAEP_MD` isn't defined in OpenSSL 1.0.1 and old versions of LibreSSL.

openssl/src/encrypt.rs
```diff
@@ -0,0 +1,342 @@
...
+    /// Creates a new `Encrypter`.
+    ///
+    /// OpenSSL documentation at [`EVP_PKEY_encrypt_init`].
+    ///
+    /// [`EVP_PKEY_encrypt_init`]: https://www.openssl.org/docs/manmaster/man3/EVP_PKEY_encrypt_init.html
+    pub fn new<T>(pkey: &'a PKeyRef<T>) -> Result<Encrypter<'a>, ErrorStack>
+    where
+        T: HasPublic,
+    {
+        Self::new_intern(pkey)
```
sfackler, line 36: Why is new_intern split out here?

openssl/src/encrypt.rs
```diff
@@ -0,0 +1,342 @@
...
+
+impl<'a> Encrypter<'a> {
+    /// Creates a new `Encrypter`.
+    ///
+    /// OpenSSL documentation at [`EVP_PKEY_encrypt_init`].
+    ///
+    /// [`EVP_PKEY_encrypt_init`]: https://www.openssl.org/docs/manmaster/man3/EVP_PKEY_encrypt_init.html
+    pub fn new<T>(pkey: &'a PKeyRef<T>) -> Result<Encrypter<'a>, ErrorStack>
+    where
+        T: HasPublic,
```
sfackler, line 34: Do we need to worry about encrypting with the private key and decrypting with the public key?

openssl/src/encrypt.rs
```diff
@@ -0,0 +1,342 @@
...
+            .map(|_| ())
+        }
+    }
+
+    /// Performs public key encryption.
+    ///
+    /// This corresponds to [`EVP_PKEY_encrypt`].
+    ///
+    /// [`EVP_PKEY_encrypt`]: https://www.openssl.org/docs/manmaster/man3/EVP_PKEY_encrypt.html
+    pub fn encrypt(&self, from: &[u8], to: &mut [u8]) -> Result<usize, ErrorStack> {
```
sfackler, line 129: What happens if `to` is too small?

openssl/src/encrypt.rs
dodomorandi: I used the same structure of `Signer`, which has two `new` fn. Now that I think about it, it should never be the case for `Encrypter` and `Decrypter`, I am going to merge `new_intern` with `new`.

openssl/src/encrypt.rs
dodomorandi: I did not check about this when I wrote this first stub. In any case, it should only make sense to encrypt with public data -- if I want to let anyone know that it's the owner of the private key that sent a message, I should _sign_ the message, not encrypt it. Right?

openssl/src/encrypt.rs
dodomorandi: That case will be checked by the underlying function, which will return an error. Should I unit test this case?

dodomorandi:
@sfackler Thank you so much for your review.

> It looks like `EVP_PKEY_CTRL_RSA_OAEP_MD` isn't defined in OpenSSL 1.0.1 and old versions of LibreSSL.

I totally forgot about checking if I needed to use specific `cfg`s, fixing that.

openssl/src/encrypt.rs
dodomorandi: Fixed in https://github.com/sfackler/rust-openssl/pull/1384/commits/411ad0f06653be32110e0d94e41418452df0e3dd

openssl/src/encrypt.rs
sfackler: Oh yeah I missed that outlen was initialized with to.len(). That should be fine - I was thinking of some other lower level functions where the caller needs to ensure the buffer size is sufficient.

openssl/src/encrypt.rs
sfackler: That's probably the case, yeah.

openssl/src/encrypt.rs
dodomorandi: However, playing devil's advocate, as a user I would ask myself how big the buffer should be. This is the main reason I would like to implement a better API (as I wrote in the first comment), but I am not sure what's the best (or at least, the most acceptable one) solution.

sfackler: EVP_PKEY_encrypt will tell that to you if you pass a null output pointer - you could add an `fn encrypt_len(&self, from: &[u8]) -> Result<usize, ErrorStack>` function that does that.

dodomorandi: Yes, my worries were more about the best way of exposing the feature than how to do that :relaxed:. At the end `encrypt_len/decrypt_len` is probably good enough: it is zero overhead (`encrypt(input: &[u8]) -> Vec<u8>` is not) and it is not ambiguous (`encrypt(input: &[u8], output: &mut Vec<u8>)` is ambiguous about output auto-cleanup). With `encrypt` + `encrypt_len` the input of the first call is _detached_ from the second, but I think that 99.99% of the use cases are trivial enough to make them straightforward to use :blush:. I'll go with this solution!

dodomorandi: Implemented in https://github.com/sfackler/rust-openssl/pull/1384/commits/d0676bf38d773989d60b342a9efc960c30a4b1fd

dodomorandi:
@sfackler I found that using `encrypt` and `decrypt` can be counterintuitive, because the size of the output buffer is greater than the output data. I added a bit of documentation to clarify how get an expected behavior.

sfackler:
Thanks!

PR #1385, opened 2020-12-04
Support OPENSSL_NO_RMD160, OPENSSL_NO_CMS, OPENSSL_NO_BF

KapJI:
Support linking against more OpenSSL configurations.


sfackler:
You will need to update https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/build/expando.c so these settings are actually discovered.

PR #1386, opened 2020-12-06
Stop manually specifying linkage type

sfackler:
This is not a very common thing to do, and in particular can cause
trouble when building against MUSL in an Alpine environment where the
logic can force the compiler to dynamically link to the system install
which is not compatible with a static CRT.

There's some risk that this will cause breakage, so we'll have to see
how things go.

### DIFF
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -8,7 +8,6 @@ extern crate pkg_config;
 #[cfg(target_env = "msvc")]
 extern crate vcpkg;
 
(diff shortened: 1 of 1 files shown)


sfackler:
It appears that rustc explicitly passes `-Wl,-Bdynamic` for some reason, which breaks the musl-vendored build as it ends up hitting the system install rather than the vendored copy.

sfackler:
I don't think this approach will work unfortunately.

PR #1390, opened 2020-12-16
Add additional x509 store verification bindings

naleba97:
Added some low level bindings to the X509 Store builder struct for expanded verification capabilities.

Also fixed `add_cert()` to take in an `&X509Ref` instead of a `X509`.


openssl/src/x509/store.rs
```diff
@@ -81,11 +88,53 @@ impl X509StoreBuilder {
 
 impl X509StoreBuilderRef {
     /// Adds a certificate to the certificate store.
-    // FIXME should take an &X509Ref
-    pub fn add_cert(&mut self, cert: X509) -> Result<(), ErrorStack> {
+    pub fn add_cert(&mut self, cert: &X509Ref) -> Result<(), ErrorStack> {
```
sfackler, line 91: This is a breaking change. That's why it was a FIXME.

openssl/src/x509/store.rs
```diff
@@ -81,11 +88,53 @@ impl X509StoreBuilder {
...
 
+    /// Sets the maximum verification depth, or the maximum number of intermediate CA certificates that can appear in a chain.
+    pub fn set_depth(&mut self, depth: c_int) -> Result<(), ErrorStack> {
+        unsafe { cvt(ffi::X509_STORE_set_depth(self.as_ptr(), depth)).map(|_| ()) }
+    }
+
+    cfg_if! {
+        if #[cfg(any(ossl102, libressl261))] {
+             /// Sets the verification flags used to verify a certificate against a chain.
+            pub fn set_flags(&mut self, flags: X509VerifyFlags) -> Result<(), ErrorStack> {
```
sfackler, line 103: Why would the argument type for this not be X509VerifyFlags on old versions?

openssl/src/x509/store.rs
```diff
@@ -81,11 +88,53 @@ impl X509StoreBuilder {
...
+        } else {
+            /// Sets the verification flags used to verify a certificate against a chain.
+            pub fn set_flags(&mut self, flags: c_ulong) -> Result<(), ErrorStack> {
+                unsafe { cvt(ffi::X509_STORE_set_flags(self.as_ptr(), flags)).map(|_| ()) }
+            }
+        }
+    }
+
+    /// Sets the purpose used to verify the certificate chain.
+    pub fn set_purpose(&mut self, purpose: c_int) -> Result<(), ErrorStack> {
```
sfackler, line 115: These should not take a raw integer.

naleba97: Will add the trust/purpose values as defined in https://github.com/openssl/openssl/blob/OpenSSL_1_1_1a/include/openssl/x509.h#L116 and https://github.com/openssl/openssl/blob/OpenSSL_1_1_1a/include/openssl/x509v3.h#L412

openssl/src/x509/store.rs
naleba97: X509VerifyFlags is in the `verify` module which is only compiled in the top level `x509` module for `any(ossl102, libressl261)`. I could relax/remove the version condition to use X509VerifyFlags for older versions.

openssl/src/x509/store.rs
naleba97: I see, would this just be fixed in the next major version?

sfackler: Yes

openssl/src/x509/store.rs
sfackler: If the type is used in APIs that exist in older versions, then it can be exposed on those older versions.

naleba97: Removed the top-level constraint on the `verify` module and added more fine-grained constraints on structs and bit flags in the `verify` module.

openssl/src/x509/store.rs
```diff
@@ -86,6 +90,40 @@ impl X509StoreBuilderRef {
...
+    }
+
+    /// Load trusted certificate(s) into the `X509Store` from a file or a directory.
+    ///
+    /// The certificates in the file or directory should be in a hashed format.
+    pub fn load_locations(&mut self, file: &str, dir: &str) -> Result<(), ErrorStack> {
+        unsafe {
+            cvt(ffi::X509_STORE_load_locations(
+                self.as_ptr(),
+                file.as_ptr() as *const c_char,
```
sfackler, line 120: Rust strings are not null terminated. This should also take Path arguments like https://github.com/sfackler/rust-openssl/blob/134a9de0f071cbbe4889a0ad7b040c234a76b7ae/openssl/src/x509/mod.rs#L873.

openssl/src/x509/verify.rs
```diff
@@ -58,6 +59,61 @@ bitflags! {
...
+    }
+}
+
+pub struct TrustId(c_int);
+
+impl TrustId {
+    #[cfg(not(any(ossl110)))]
+    pub const DEFAULT: TrustId = TrustId(ffi::X509_TRUST_DEFAULT);
+    #[cfg(any(ossl110))]
+    pub const DEFAULT: TrustId = TrustId(ffi::X509_TRUST_DEFAULT); /* Only valid in purpose settings */
```
sfackler, line 93: Aren't these two definitions identical?

openssl/src/x509/verify.rs
```diff
@@ -58,6 +59,61 @@ bitflags! {
     }
 }
 
+pub struct PurposeId(c_int);
```
sfackler, line 62: These seem to fit better in the store module since that's where they're actually used. I'd call them `X509Purpose` and `X509Trust` probably I think.

naleba97: Sure, I originally put them in the verify module since trust and purpose values are also used in `X509_VERIFY_PARAM`, in addition to `X509_STORE`.

openssl/src/x509/verify.rs
naleba97: Whoops, good catch, will remove.

eutampieri:
What’s missing in this PR? Can I do something to make it pass review?

naleba97:
Revisiting the PR: currently making fixes to build for 1.0.1

naleba97:
Hey @sfackler, please let me know if there's anything else blocking this PR. I've just rebased with the latest changes. Thanks!

eutampieri:
Any update on this?

minyung:
@sfackler 
I need `set_purpose` 😢😢. How is it going?

PR #1392, opened 2020-12-16
Support EVP_get_digestbyname binding

crab2313:
Fixes: #1391

### DIFF
--- a/openssl/src/hash.rs
+++ b/openssl/src/hash.rs
@@ -1,4 +1,5 @@
 use ffi;
+use std::ffi::CString;
 use std::fmt;
 use std::io;
 use std::io::prelude::*;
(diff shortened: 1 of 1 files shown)


openssl/src/hash.rs
```diff
@@ -46,6 +47,18 @@ impl MessageDigest {
         }
     }
 
+    pub fn from_name(name: &str) -> Option<MessageDigest> {
```
sfackler, line 50: Can you add a quick bit of documentation like what from_nid has?

sfackler:
LGTM other than adding the docs!

openssl/src/hash.rs
crab2313: Sure.

crab2313:
EVP_get_digestbyname should be called after OPENSSL_add_all_algorithms() with openssl 1.0.0.  The code is changed to fix the bug.

openssl/src/hash.rs
```diff
@@ -46,6 +47,25 @@ impl MessageDigest {
         }
     }
 
+    /// Returns the `MessageDigest` corresponding to an algorithm name.
+    ///
+    /// This corresponds to [`EVP_get_digestbyname`].
+    ///
+    /// [`EVP_get_digestbyname`]: https://www.openssl.org/docs/man1.1.0/crypto/EVP_DigestInit.html
+    pub fn from_name(name: &str) -> Option<MessageDigest> {
+        #[cfg(not(ossl110))]
```
sfackler, line 56:
```suggestion
```

Init is already a no-op on versions that don't need it, so it can be called unconditionally.

PR #1408, opened 2021-01-23
Find OpenSSL in pkgsrc on macOS

kiron1:
Look in `/opt/pkg` for OpenSSL when available. `/opt/pkg` is the root directory used by [pkgsrc](https://www.pkgsrc.org) which is [also available for macOS](https://pkgsrc.joyent.com/install-on-osx/).

### DIFF
--- a/openssl-sys/build/find_normal.rs
+++ b/openssl-sys/build/find_normal.rs
@@ -56,6 +56,16 @@ fn resolve_with_wellknown_homebrew_location(dir: &str) -> Option<PathBuf> {
     None
 }
 
+fn resolve_with_wellknown_pkgsrc_location() -> Option<PathBuf> {
+    let pkgsrc = Path::new("/opt/pkg");
+    let pkgsrc_include_openssl = pkgsrc.join("include/openssl");
+    if pkgsrc_include_openssl.exists() {
+        return Some(pkgsrc.to_path_buf());
+    }
+
+    None
+}
+
 fn find_openssl_dir(target: &str) -> OsString {
     let host = env::var("HOST").unwrap();
 
@@ -67,6 +77,8 @@ fn find_openssl_dir(target: &str) -> OsString {
             }
         } else if target.contains("apple-darwin") {
             if let Some(dir) = resolve_with_wellknown_homebrew_location("/usr/local/opt/openssl") {
+                return dir.into();
+            } else if let Some(dir) = resolve_with_wellknown_pkgsrc_location() {
                 return dir.into();
             }
         }


openssl-sys/build/find_normal.rs
```diff
@@ -68,6 +77,8 @@ fn find_openssl_dir(target: &str) -> OsString {
         } else if target.contains("apple-darwin") {
             if let Some(dir) = resolve_with_wellknown_homebrew_location("/usr/local/opt/openssl") {
                 return dir.into();
+            } else if let Some(dir) = resolve_with_wellknown_pkgsrc_location("/opt/pkg") {
```
sfackler, line 80: There's no need to pass the path in here since there's only one. You can just put it directly in the function.

openssl-sys/build/find_normal.rs
```diff
@@ -56,6 +56,15 @@ fn resolve_with_wellknown_homebrew_location(dir: &str) -> Option<PathBuf> {
     None
 }
 
+fn resolve_with_wellknown_pkgsrc_location() -> Option<PathBuf> {
```
sfackler, line 59: Is OpenSSL a mandatory install in pkgsrc? It seems like this could false-positive if you have an /opt/pkg but don't have OpenSSL installed through it.

kiron1: OpenSSL is included in the bootstrap package of pkgsrc:

```sh
curl -sSfL https://pkgsrc.joyent.com/packages/Darwin/bootstrap/bootstrap-macos14-trunk-x86_64-20200716.tar.gz | tar tzf - | grep ssl
```

The headers, libraries, `openssl` binary and `openssl.pc` is included. The `pkgin` tool also links to OpenSSL as we can see with `otool -L /opt/pkg/bin/pkgin`:

```
/opt/pkg/bin/pkgin:
[...]
	/opt/pkg/lib/libssl.1.1.dylib (compatibility version 1.1.0, current version 1.1.0)
	/opt/pkg/lib/libcrypto.1.1.dylib (compatibility version 1.1.0, current version 1.1.0)
[...]
```

Should we better check for presence of the following files:

- `/opt/pkg/include/openssl/ssl.h`
- `/opt/pkg/lib/libssl.a`
- `/opt/pkg/lib/libssl.dylib`
- `/opt/pkg/lib/libcrypto.a`
- `/opt/pkg/lib/libcrypto.dylib`

or would it be enough to test for `/opt/pkg/lib/pkgconfig/openssl.pc` and assume the rest is there as well (a `pkg-config` tool is not included with pkgsrc by default)?

sfackler: pkg-config isn't used if it's discovered in a well-known place. Might be best to explicitly look for `/opt/pkg/include/openssl` just in case.

PR #1413, opened 2021-02-09
Add support for MacPorts + fix scanning other directories on arm64

davidbailey00:
This PR makes the following changes:

- Scan `/opt/local` in addition to `/opt/pkg`, a well-known location where MacPorts installs packages
- Refactor `resolve_with_wellknown_pkgsrc_location` into generic `resolve_with_wellknown_location`
- Fix not scanning other directories on `aarch64-apple-darwin`: previously, only the Homebrew directory was scanned
- Remove unnecessary Homebrew path checks: [`openssl` is an alias of `openssl@1.1`](https://github.com/Homebrew/homebrew-core/blob/master/Aliases/openssl) so the package will always be installed to `<prefix>/opt/openssl@1.1` rather than the unversioned path
- Add docs for installing `openssl` in different ways on macOS

### DIFF
--- a/openssl-sys/build/find_normal.rs
+++ b/openssl-sys/build/find_normal.rs
@@ -25,27 +25,14 @@ fn resolve_with_wellknown_homebrew_location(dir: &str) -> Option<PathBuf> {
     // Check up default aarch 64 Homebrew installation location first
     // for quick resolution if possible.
     //  `pkg-config` on brew doesn't necessarily contain settings for openssl apparently.
-    let mut version_dir = dir.to_owned();
-    version_dir.push_str("@1.1");
-    let homebrew = Path::new(&version_dir);
+    let homebrew = Path::new(dir).join("opt/openssl@1.1");
     if homebrew.exists() {
         return Some(homebrew.to_path_buf());
     }
-    let homebrew = Path::new(dir);
-    if homebrew.exists() {
-        return Some(homebrew.to_path_buf());
-    }
+
     // Calling `brew --prefix <package>` command usually slow and
     // takes seconds, and will be used only as a last resort.
     let output = execute_command_and_get_output("brew", &["--prefix", "openssl@1.1"]);
-    if let Some(ref output) = output {
-        let homebrew = Path::new(&output);
-        if homebrew.exists() {
-            return Some(homebrew.to_path_buf());
-        }
-    }
-
-    let output = execute_command_and_get_output("brew", &["--prefix", "openssl"]);
     if let Some(ref output) = output {
         let homebrew = Path::new(&output);
         if homebrew.exists() {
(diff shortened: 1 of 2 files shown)


sfackler:
Is `/opt/local` the canonical place for macports? Is there a fallback to something like `brew --prefix` we should try?

davidbailey00:
> Is `/opt/local` the canonical place for macports?

Yes, this is the standard installation directory; see https://guide.macports.org/#installing.macports

> Is there a fallback to something like `brew --prefix` we should try?

I've had a look through the MacPorts guide (see link above) and manpage, and from what I can tell there is no way to get the install prefix from outside a Portfile, so I think defaulting to `/opt/local` is best for most users, given that most will be using the standard location.

davidbailey00:
I've also updated the PR to remove unnecessary checks for Homebrew paths and updated the description accordingly.

openssl-sys/build/find_normal.rs
```diff
@@ -25,16 +25,11 @@ fn resolve_with_wellknown_homebrew_location(dir: &str) -> Option<PathBuf> {
     // Check up default aarch 64 Homebrew installation location first
     // for quick resolution if possible.
     //  `pkg-config` on brew doesn't necessarily contain settings for openssl apparently.
-    let mut version_dir = dir.to_owned();
-    version_dir.push_str("@1.1");
-    let homebrew = Path::new(&version_dir);
+    let homebrew = Path::new(dir).join("opt/openssl@1.1");
```
sfackler, line 28: After these changes it will no longer find an OpenSSL 1.0.2 install (under `/usr/local/opt/openssl`).

davidbailey00: This is intentional as only OpenSSL 1.1 is provided by Homebrew [as of November 2019](https://brew.sh/2019/11/27/homebrew-2.2.0/), however can split into a separate PR if this is worth discussing to focus on the initial changes in this PR :)

sfackler: Oh nice, I was not aware of that!

sfackler: Do you know if that change actively removed existing 1.0.2 installs or just prevented new ones? If it's the former, there is a chance someone who was "grandfathered" in could have their builds broken by this change.

davidbailey00: Hmm, although I do not primarily use Homebrew myself, I believe that running `brew upgrade` would cause it to upgrade to the latest version, and it seems unlikely that someone would have skipped updating their packages for that long a period of time.

sfackler: openssl@1.1 was actually a separate package from openssl, so I'm not sure the automatic upgrade would have happened. In any case though, anyone in that situation really needs to bump to a version that's actually supported, so we can probably do this anyway!

PR #1414, opened 2021-02-15
Add ec point validation functions

nomick:
Adds two more functions to validate EC points received from untrusted locations.

### DIFF
--- a/openssl/src/ec.rs
+++ b/openssl/src/ec.rs
@@ -529,14 +529,22 @@ impl EcPointRef {
     }
 
     /// Checks if point is infinity
-    pub fn is_infinity(&self, group: &EcGroupRef) -> Result<bool, ErrorStack> {
+    ///
+    /// OpenSSL documentation at [`EC_POINT_is_at_infinity`]
+    ///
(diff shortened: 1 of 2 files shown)


openssl/src/ec.rs
```diff
@@ -527,6 +527,30 @@ impl EcPointRef {
             .map(|_| ())
         }
     }
+
+    /// Checks if point is infinity
```
sfackler, line 531: The docs should mention what these correspond to like on the other methods in the module.

openssl/src/ec.rs
```diff
@@ -527,6 +527,30 @@ impl EcPointRef {
...
+    pub fn is_infinity(&self, group: &EcGroupRef) -> Result<bool, ErrorStack> {
```
sfackler, line 532: This is not documented to ever return an error, just 0 or 1.

openssl/src/ec.rs
nomick: added

openssl/src/ec.rs
nomick: Just wanted to make sure, but I see that it does not correspond to other functions..

PR #1415, opened 2021-02-15
Expose weak-crypto feature from openssl-src

mloebel:
Added to openssl-src in https://github.com/alexcrichton/openssl-src-rs/pull/68

### DIFF
--- a/openssl-sys/Cargo.toml
+++ b/openssl-sys/Cargo.toml
@@ -13,6 +13,7 @@ build = "build/main.rs"
 
 [features]
 vendored = ['openssl-src']
+weak-crypto = ["openssl-src/weak-crypto"]
 
 [dependencies]
 libc = "0.2"
--- a/openssl/Cargo.toml
+++ b/openssl/Cargo.toml
@@ -18,6 +18,7 @@ v110 = []
 v111 = []
 
 vendored = ['ffi/vendored']
+weak-crypto = ["ffi/weak-crypto"]
 
 [dependencies]
 bitflags = "1.0"


sfackler:
I'm not sure this is necessary - someone that desires this can add their own dependency on openssl-src and enable the feature.

mloebel:
Hm, it makes sense not to add the `openssl` feature because there is no high level API that makes use of the weak crypto algorithms, but it would be good for `openssl-sys` to have it. Or at least, if I'm not mistaken then some of the bindings I added in https://github.com/sfackler/rust-openssl/pull/1416 are for legacy algorithms.

sfackler:
Having it in openssl-sys is just kind of confusing though, IMO. openssl and openssl-sys both enable APIs based on what the underlying OpenSSL build has, but someone may expect that they need to enable this feature. Instead, it switches them over to linking to a vendored OpenSSL which they may not want at all!

mloebel:
Hm, I might have misunderstood how the feature and cfg system is set up here. But based on what you said, is this right?

- `openssl_sys` detects `#defines` of the used version of OpenSSL and enables internal `cfg` directives as needed.
- Which means it can automatically detect if weak algorithms are enabled, and does not need to be told up-front.
- That way we can either use an system OpenSSL, or a self-build one via the vendored flag, and it recognizes what is possible automatically.
- In case that we want to vendor OpenSSL, we can enable `openssl_src` features by depending on it, without going through `openssl_sys`, because cargo de-duplicates dependencies and merges features.

If this is right, then this PR is indeed unneeded, and I'm sorry for the trouble. :)

sfackler:
Yep, that's correct. It definitely makes sense to add features to openssl-src to control various build-time flags in OpenSSL, but they just don't need to be propagated upstream in Cargo.

mloebel:
Alright, I'll close this then, and remove the same commits from https://github.com/sfackler/rust-openssl/pull/1416

PR #1416, opened 2021-02-15
WIP: Add additional bindings to openssl-sys for asymmetric algorithms

mloebel:
This mostly copies over a bunch of `#define` constants from openssl 1.1.1i headers, as well as a few missing function prototypes and macros.

It is not clear to me so far how to exactly handle OpenSSL version differences exactly, so this PR should be considered as a WIP so far. Guidance would be appreciated. 😄

This includes the commits from https://github.com/sfackler/rust-openssl/pull/1415, which I'd expect to be mergable before this.


sfackler:
If you look through the rest of openssl-sys, you'll see that bits of AIP that differ across versions are flagged of appropriately via cfgs. There is also a special macro to handle the relatively common case of a pointer argument to a function becoming const.

openssl-sys/src/kdf.rs
```diff
@@ -0,0 +1,220 @@
...
+const EVP_PKEY_CTRL_SCRYPT_N: c_int = EVP_PKEY_ALG_CTRL + 10;
+const EVP_PKEY_CTRL_SCRYPT_R: c_int = EVP_PKEY_ALG_CTRL + 11;
+const EVP_PKEY_CTRL_SCRYPT_P: c_int = EVP_PKEY_ALG_CTRL + 12;
+const EVP_PKEY_CTRL_SCRYPT_MAXMEM_BYTES: c_int = EVP_PKEY_ALG_CTRL + 13;
+
+const EVP_PKEY_HKDEF_MODE_EXTRACT_AND_EXPAND: c_int = 0;
+const EVP_PKEY_HKDEF_MODE_EXTRACT_ONLY: c_int = 1;
+const EVP_PKEY_HKDEF_MODE_EXPAND_ONLY: c_int = 2;
+
+pub unsafe extern "C" fn EVP_PKEY_CTX_set_tls1_prf_md(
```
sfackler, line 23: Functions defined from macros elsewhere in the crate use the standard Rust ABI rather than the C ABI. Probably doesn't matter too much, but it seems best to match that here.

sfackler:
Something appears to have made systest unhappy but I'm not sure what off the top of my head. Might require some binary searching through the changes to track down.

PR #1422, opened 2021-03-02
Authority Information Access x509 extension

cmars:
Add support for parsing and accessing the [Authority Information Access
extension](https://tools.ietf.org/html/rfc5280#section-4.2.2.1).


sfackler:
Thanks!

PR #1423, opened 2021-03-03
Updated CMS (PKCS#7) to match OpenSSL spec

shanecurran:
The [OpenSSL spec for CMS (PKCS#7)](https://www.openssl.org/docs/man1.0.2/man3/CMS_decrypt.html) allows for NULL to be provided as the `cert` value to try all possible recipients. This is useful in scenarios where a recipient certificate is not available, like in the [AWS Nitro Enclaves KMS flow](https://docs.aws.amazon.com/enclaves/latest/user/verify-root.html).

It does, however, potentially leave the implementation vulnerable to Bleichenbacher's attack on PKCS#1 v1.5 RSA padding as outlined in the OpenSSL specs. Included documentation to outline this.

### DIFF
--- a/openssl/src/cms.rs
+++ b/openssl/src/cms.rs
@@ -289,12 +289,12 @@ mod test {
             let encrypted_der = encrypt.to_der().expect("failed to create der from cms");
             let decrypt =
                 CmsContentInfo::from_der(&encrypted_der).expect("failed read cms from der");
-            
+
             let decrypt_with_cert_check = decrypt
                 .decrypt(&priv_cert.pkey, &priv_cert.cert)
                 .expect("failed to decrypt cms");
-            let decrypt_with_cert_check =
-                String::from_utf8(decrypt_with_cert_check).expect("failed to create string from cms content");
+            let decrypt_with_cert_check = String::from_utf8(decrypt_with_cert_check)
+                .expect("failed to create string from cms content");
 
             let decrypt_without_cert_check = decrypt
                 .decrypt_without_cert_check(&priv_cert.pkey)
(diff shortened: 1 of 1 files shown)


openssl/src/cms.rs
```diff
@@ -67,19 +67,23 @@ foreign_type_and_impl_send_sync! {
...
     /// decrypt the data in `self`.
     ///
-    /// OpenSSL documentation at [`CMS_decrypt`]
+    /// *Warning*: Not providing a certificate may leave you vulnerable to Bleichenbacher's attack on PKCS#1 v1.5 RSA padding.
+    /// See the [`OpenSSL docs`] for more information.
     ///
-    /// [`CMS_decrypt`]: https://www.openssl.org/docs/man1.1.0/crypto/CMS_decrypt.html
-    pub fn decrypt<T>(&self, pkey: &PKeyRef<T>, cert: &X509) -> Result<Vec<u8>, ErrorStack>
+    /// [`OpenSSL docs`]: https://www.openssl.org/docs/man1.1.0/crypto/CMS_decrypt.html
+    pub fn decrypt<T>(&self, pkey: &PKeyRef<T>, cert: Option<&X509>) -> Result<Vec<u8>, ErrorStack>
```
sfackler, line 77: Directly changing the signature is a breaking change. This should probably be added as a separate `decrypt2` method or something like that.

shanecurran: Thanks for the review, @sfackler!

I've added a new, separate `decrypt_without_cert_check()` method to avoid a breaking change. Let me know if there's anything else you think needs to be updated 👍

sfackler:
Thanks!

PR #1425, opened 2021-03-06
expose rsa OAEP label methods

baloo:
I've been trying to play with a TPM lately and spec uses OAEP labels.

This PR exposes the methods.

### DIFF
--- a/openssl/src/encrypt.rs
+++ b/openssl/src/encrypt.rs
@@ -39,6 +39,7 @@
 //! decrypted.truncate(decrypted_len);
 //! assert_eq!(&*decrypted, data);
 //! ```
+use libc::{c_int, c_void};
 use std::{marker::PhantomData, ptr};
 
 use crate::error::ErrorStack;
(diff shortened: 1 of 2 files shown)


openssl/src/encrypt.rs
```diff
@@ -157,6 +158,32 @@ impl<'a> Encrypter<'a> {
...
+    ///
+    /// This corresponds to [`EVP_PKEY_CTX_set0_rsa_oaep_label`].
+    ///
+    /// [`EVP_PKEY_CTX_set0_rsa_oaep_label`]: https://www.openssl.org/docs/manmaster/man3/EVP_PKEY_CTX_set0_rsa_oaep_label.html
+    #[cfg(any(ossl102, libressl310))]
+    pub fn set_rsa_oaep_label(&mut self, label: CString) -> Result<(), ErrorStack> {
+        unsafe {
+            let label = label.into_bytes_with_nul().into_boxed_slice();
+            // Openssl takes ownership of the label and will free it if needed
+            let label = Box::leak(label);
```
sfackler, line 173: Allocations performed by Rust are not compatible with those performed by OpenSSL. This needs to call CRYPTO_malloc manually, like here: https://github.com/sfackler/rust-openssl/blob/9872a5dd7f476170372b9e119b8a6b681396b434/openssl/src/ssl/mod.rs#L3126-L3130.

openssl/src/encrypt.rs
```diff
@@ -157,6 +158,32 @@ impl<'a> Encrypter<'a> {
...
 
+    /// Sets the RSA OAEP label.
+    ///
+    /// This is only useful for RSA keys.
+    ///
+    /// This corresponds to [`EVP_PKEY_CTX_set0_rsa_oaep_label`].
+    ///
+    /// [`EVP_PKEY_CTX_set0_rsa_oaep_label`]: https://www.openssl.org/docs/manmaster/man3/EVP_PKEY_CTX_set0_rsa_oaep_label.html
+    #[cfg(any(ossl102, libressl310))]
+    pub fn set_rsa_oaep_label(&mut self, label: CString) -> Result<(), ErrorStack> {
```
sfackler, line 169: This should just take a `&str` rather than a CString.

baloo: How about a `&[u8]`? Or a `&CStr`?

openssl/src/encrypt.rs
baloo: makes sense.

openssl/src/encrypt.rs
baloo: My spec (https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p59_Part1_Architecture_pub.pdf section 4) ask for quite specific sequence of bytes, and ideally I'd like to stick as close as possible to it.

sfackler: A string is composed of a specific sequence of bytes, but if OAEP labels are not constrained to be textual, `&[u8]` seems fine.

baloo: Yeah, I know. I'm not too sure what's expected and what's not for an oaep label to be fair. I'll try to dig up some documentation.

baloo: https://tools.ietf.org/html/rfc8017#section-2
defines the OAEP label as:
> L              optional RSAES-OAEP label, an octet string

https://tools.ietf.org/html/rfc8017#section-4
then defines the octet string as:
> For the purposes of this document, and consistent with ASN.1 syntax,
   an octet string is an ordered sequence of octets (eight-bit bytes).
   The sequence is indexed from first (conventionally, leftmost) to last
   (rightmost).  For purposes of conversion to and from integers, the
   first octet is considered the most significant in the following
   conversion primitives.

So my opinion is that it should be an `&[u8]`.

sfackler: Cool - that sounds right to me as well.

openssl/src/encrypt.rs
```diff
@@ -157,6 +158,43 @@ impl<'a> Encrypter<'a> {
...
+
+            cvt(ffi::EVP_PKEY_CTX_set0_rsa_oaep_label(
+                self.pctx,
+                p as *mut c_void,
+                label.len() as c_int,
+            ))
+            .map(|_| ())
+            .map_err(|e| {
+                #[cfg(not(ossl110))]
+                ::ffi::CRYPTO_free(p as *mut c_void);
```
sfackler, line 186: This is unfortunate, but looking at the OpenSSL source is required :(

PR #1264, opened 2020-04-29
Initial support for OpenSSL 3.0.0-alpha17

sfackler:
Closes #1263

PR #1426, opened 2021-03-08
Add support for openssl3 KDF methods

baloo:
This PR tries to add support for the KDF api openssl3 brings.

The current implementation works, but only supports the KDF I need. I'm sharing early see if you'd have any opinions about the proposed API before I add support for other KDFs.

From my understanding, KDF only works with `OSSL_PARAMS` which are not the easiest thing to use from rust. This is an attempt, but I'm not necessarily convinced.

Please don't mind the extra OAEP label commit, It's merged in master, it's only here because it's easier for me to develop and test the whole api. I'll drop the commit eventually.


openssl-sys/src/evp.rs
```diff
@@ -30,6 +30,7 @@ pub unsafe fn EVP_get_digestbynid(type_: c_int) -> *const EVP_MD {
 }
 
 extern "C" {
+    pub fn EVP_MD_name(md: *const EVP_MD) -> *const c_char;
```
baloo, line 33: This looks like an ossl300 specific according to the test-suite.

wongsyrone:
Why not send PR based on https://github.com/sfackler/rust-openssl/pull/1264

baloo:
> Why not send PR based on #1264

it's based on openssl-300 and targets it, so I'm not sure what you're referring to.

jstarks:
@sfackler, you closed this without comment--are you open to adding support for this KDF functionality?

baloo:
@jstarks it automatically closed because the target branch has been removed.

I did not have time to re-target and make the adjustments necessary. feel free to take over.

jstarks:
Ah, I see. OK, I may follow up here later.

PR #1427, opened 2021-03-08
Add sm3 cryptographic hash support

jiegec:
SM3 wikipedia: https://en.wikipedia.org/wiki/SM3_(hash_function)
SM3 rfc draft: https://datatracker.ietf.org/doc/html/draft-sca-cfrg-sm3

SM3 support is added since OpenSSL 1.1.1: https://www.openssl.org/news/openssl-1.1.1-notes.html

And it's guarded by `OPENSSL_NO_SM3` config in openssl/evp.h:

```cpp
# ifndef OPENSSL_NO_SM3
const EVP_MD *EVP_sm3(void);
# endif
```

NID:

```cpp
#define SN_sm3          "SM3"
#define LN_sm3          "sm3"
#define NID_sm3         1143
#define OBJ_sm3         OBJ_sm_scheme,401L

#define SN_sm3WithRSAEncryption         "RSA-SM3"
#define LN_sm3WithRSAEncryption         "sm3WithRSAEncryption"
#define NID_sm3WithRSAEncryption                1144
#define OBJ_sm3WithRSAEncryption                OBJ_sm_scheme,504L
```

libressl nid: https://github.com/libressl-portable/openbsd/commit/03add4789c7ebb91d58d856b1b80624ae9177bbf


Harry-Chen:
SM3 is supportted from libreSSL 2.9.1, so I also add `libressl291` to building condition.

Harry-Chen:
I also add NIDs related to SM2 in `obj_macs.rs`. SM2 is a signature algorithm utilizing SM3 (see <https://www.openssl.org/docs/manmaster/man7/SM2.html>), which does not involve additional functions but some new NIDs.

sfackler:
You'll need to add OPENSSL_NO_SM3 to https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/build/expando.c if you want the osslconf bit of the cfg to work.

jiegec:
> You'll need to add OPENSSL_NO_SM3 to https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/build/expando.c if you want the osslconf bit of the cfg to work.

Thanks for that!

PR #1431, opened 2021-03-10
Use DH_generate_parameters_ex()

nospam3089:
https://www.openssl.org/docs/man1.1.0/man3/DH_generate_parameters.html
states that DH_generate_parameters() is deprecated. It has been so since
OpenSSL 0.9.8 was released in 2005.

### DIFF
--- a/openssl/src/dh.rs
+++ b/openssl/src/dh.rs
@@ -77,7 +77,6 @@ impl Dh<Params> {
     ///
     /// This corresponds to [`DH_generate_parameters_ex`].
     ///
-    /// [`DH_new`]: https://www.openssl.org/docs/man1.1.0/crypto/DH_new.html
     /// [`DH_generate_parameters_ex`]: https://www.openssl.org/docs/man1.1.0/crypto/DH_generate_parameters.html
     pub fn generate_params(prime_len: u32, generator: u32) -> Result<Dh<Params>, ErrorStack> {
         unsafe {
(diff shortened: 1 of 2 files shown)


openssl/src/dh.rs
```diff
@@ -77,15 +77,18 @@ impl Dh<Params> {
     ///
     /// This corresponds to [`DH_generate_parameters`].
```
sfackler, line 78:
```suggestion
    /// This corresponds to [`DH_generate_parameters_ex`].
```

nospam3089: Good catch! That's an excellent improvement. I've re-pushed the branch with this change included. Thank you!

openssl/src/dh.rs
```diff
@@ -75,17 +75,20 @@ impl Dh<Params> {
 
     /// Generates DH params based on the given `prime_len` and a fixed `generator` value.
     ///
-    /// This corresponds to [`DH_generate_parameters`].
+    /// This corresponds to [`DH_generate_parameters_ex`].
     ///
-    /// [`DH_generate_parameters`]: https://www.openssl.org/docs/man1.1.0/crypto/DH_generate_parameters.html
+    /// [`DH_new`]: https://www.openssl.org/docs/man1.1.0/crypto/DH_new.html
```
sfackler, line 80:
```suggestion
```

PR #1439, opened 2021-03-18
Add wrapper for X509_STORE_set_flags()

kkovaacs:
This makes it possible to properly set verification flags (like CRL check) for the store.


sfackler:
CI's failing when building against 1.0.1 looks like.

kkovaacs:
Yeah, I'm fixing it right now...

kkovaacs:
OK, I had to make X509StoreBuilderRef::set_flags() openssl 1.0.2+ only, since X509VerifyFlags is also dependent on those versions.

The MacOS check is still failing but I don't think it's related to this change.

kkovaacs:
I've just found out that this covers a tiny part of #1390, which is a more complete solution.

PR #1440, opened 2021-03-21
pkey: add conversions to specific key types

petreeftime:
Adds TryFrom conversions for PKey into EcKey, Rsa, Dsa and Dh and
reverse.

Signed-off-by: Petre Eftime <petre.eftime@gmail.com>


sfackler:
Thanks!

PR #1442, opened 2021-03-25
fix swapped IGNORE_CRITICAL/X509_STRICT flag definitions

kallsyms:
Noticed these were swapped while looking at uses of `X509_V_FLAG_X509_STRICT`.


sfackler:
Good find, thanks!

PR #1450, opened 2021-04-08
ec: Add EcKey affine_coordinates

puiterwijk:
In OpenSSL, new functions were introduced[1] to get/set the affine
coordinates of a point, where OpenSSL itself determines what version of
the _GFp or _GF2m underlying function to call.
This makes that function available.

[1]: https://github.com/openssl/openssl/commit/8e3cced75fb5fee5da59ebef9605d403a999391b

Signed-off-by: Patrick Uiterwijk <patrick@puiterwijk.org>

### DIFF
--- a/openssl/src/ec.rs
+++ b/openssl/src/ec.rs
@@ -480,6 +480,32 @@ impl EcPointRef {
     /// Place affine coordinates of a curve over a prime field in the provided
     /// `x` and `y` `BigNum`s
     ///
+    /// OpenSSL documentation at [`EC_POINT_get_affine_coordinates`]
+    ///
+    /// [`EC_POINT_get_affine_coordinates`]: https://www.openssl.org/docs/man1.1.1/man3/EC_POINT_get_affine_coordinates.html
+    #[cfg(ossl111)]
+    pub fn affine_coordinates(
+        &self,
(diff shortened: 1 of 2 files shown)


openssl/src/ec.rs
```diff
@@ -477,6 +477,32 @@ impl EcPointRef {
...
+    /// `x` and `y` `BigNum`s
+    ///
+    /// OpenSSL documentation at [`EC_POINT_get_affine_coordinates`]
+    ///
+    /// [`EC_POINT_get_affine_coordinates`]: https://www.openssl.org/docs/man1.1.1/man3/EC_POINT_get_affine_coordinates.html
+    #[cfg(ossl111)]
+    pub fn affine_coordinates(
+        &self,
+        group: &EcGroupRef,
+        x: &mut BigNumRef,
```
sfackler, line 490: Passing in these pointers will cause memory leaks, since EC_POINT_get_affine_transformation does not free the old value. The arguments are intended to be used as out-pointers, where this method would return `Result<(&BigNumRef, &BigNumRef), ErrorStack>`.

sfackler: Oh wait no I'm wrong, sorry.

PR #1456, opened 2021-04-14
Add support for retrieving SignerInfo structures from PKCS7 signatures

chenxiaolong:
This commit adds support for querying most fields from the SignerInfo structures contained within a PKCS7 signature:

* Subject name of signer
* Serial number of signer
* Digest algorithm
* Digest encryption algorithm
* Raw signature

### DIFF
--- a/openssl/src/pkcs7.rs
+++ b/openssl/src/pkcs7.rs
@@ -4,14 +4,101 @@ use libc::c_int;
 use std::mem;
 use std::ptr;
 
+use crate::asn1::{Asn1IntegerRef, Asn1StringRef};
 use crate::bio::{MemBio, MemBioSlice};
 use crate::error::ErrorStack;
 use crate::pkey::{HasPrivate, PKeyRef};
-use crate::stack::{Stack, StackRef};
+use crate::stack::{Stack, StackRef, Stackable};
 use crate::symm::Cipher;
+use crate::util::ForeignTypeRefExt;
 use crate::x509::store::X509StoreRef;
-use crate::x509::{X509Ref, X509};
+use crate::x509::{X509AlgorithmRef, X509NameRef, X509Ref, X509};
 use crate::{cvt, cvt_p};
+
+foreign_type_and_impl_send_sync! {
+    type CType = ffi::PKCS7_SIGNER_INFO;
+    fn drop = ffi::PKCS7_SIGNER_INFO_free;
+
+    /// A PKCS#7 SignerInfo structure.
+    pub struct Pkcs7SignerInfo;
+
+    /// Reference to `Pkcs7SignerInfo`
+    pub struct Pkcs7SignerInfoRef;
+}
+
+impl Pkcs7SignerInfoRef {
+    /// Returns the issuer's subject name.
+    ///
+    /// This corresponds to `PKCS7_SIGNER_INFO`'s `issuer_and_serial.issuer` field.`
+    pub fn subject_name(&self) -> &X509NameRef {
+        unsafe {
+            let ias = (*self.as_ptr()).issuer_and_serial;
+            assert!(!ias.is_null());
+
+            let issuer = (*ias).issuer;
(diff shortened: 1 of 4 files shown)


chenxiaolong:
I'm not sure if there's a better way to express:
```rust
let ptr = (*self.as_ptr()).enc_digest as *mut ffi::ASN1_STRING;
```
in `Pkcs7SignerInfo.signature()`. `ASN1_OCTET_STRING` and `ASN1_STRING` are both typedefs to the same thing in the openssl headers. Maybe `pub type ASN1_OCTET_STRING = ASN1_STRING` would make more sense, though I don't know if that's considered a breaking change.

Also, `Pkcs7.signer_infos()` has a weird name, but it's how RFC5652 refers to the list of SignerInfo structures. I'm happy to rename this function if desired.

chenxiaolong:
I'm going to mark this as a draft while I fix the CI issues with older versions of openssl.

chenxiaolong:
I believe all the issues with older versions of openssl and libressl have been resolved. This PR should be ready for review.

By the way, I'm planning on making a similar change to get the SignerInfos from the CMS API. Should that go in a new PR or this one?

openssl/src/pkcs7.rs
```diff
@@ -311,11 +398,28 @@ impl Pkcs7Ref {
             Ok(stack)
         }
     }
+
+    /// Retrieve the SignerInfo entries from the PKCS#7 structure.
+    ///
+    /// This corresponds to [`PKCS7_get_signer_info`].
+    ///
+    /// [`PKCS7_get_signer_info`]: https://man.archlinux.org/man/community/libressl/libressl-PKCS7_get_signer_info.3.en
+    pub fn signer_infos(&self) -> Result<&StackRef<Pkcs7SignerInfo>, ErrorStack> {
```
sfackler, line 407: From looking at the implementation, this should return an `Option<StackRef<Pkcs7SignerInfo>>` rather than a `Result`: https://github.com/openssl/openssl/blob/c7d848e220ecb432faa05d2198ae689298d71728/crypto/pkcs7/pk7_lib.c#L528.

chenxiaolong: Fixed, thanks!

openssl/src/pkcs7.rs
```diff
@@ -4,15 +4,102 @@ use libc::c_int;
...
+                .expect("digest encryption algorithm must not be null")
+        }
+    }
+
+    /// Returns the raw signature.
+    ///
+    /// This corresponds to `PKCS7_SIGNER_INFO`'s `enc_digest` field.
+    pub fn signature(&self) -> &Asn1StringRef {
+        unsafe {
+            // ASN1_OCTET_STRING is a typedef of ASN1_STRING
```
sfackler, line 92: They seem to be distinct types as far as I can tell: https://github.com/openssl/openssl/blob/master/include/openssl/asn1.h.in

chenxiaolong: Hmm, this is what I was looking at: https://github.com/openssl/openssl/blob/6d2e0076e6a8146f3bdaa60151b2fa6e4df0d1bf/include/openssl/types.h#L48

Both `ASN1_OCTET_STRING` and `ASN1_STRING` are typedefs of `struct asn1_string_st`.

sfackler: Sure, but they have distinct APIs in the asn1.h header, and we currently define separate types mapping to many of those already: https://docs.rs/openssl/0.10.34/openssl/asn1/index.html

bkstein: We should add `Asn1OctetString`. Suggestion:  As this PR is not automatically mergeable any more due to changes in the directory structure (-> handwritten/...) and as I will need signer info from PKCS7, too, I can integrate the changes in my [PR](https://github.com/sfackler/rust-openssl/pull/1598). @chenxiaolong @sfackler: OK?

chenxiaolong: Yes, please. Thank you!

PR #1452, opened 2021-04-12
keep it to OpenBSD -current

gonzalo-:
This makes sense? :D

PR #1459, opened 2021-04-20
Support LibreSSL 3.3.2

vishwin:
DTLS support is now exposed. Opcode values are different from OpenSSL.
While here, re-enable SRTP tests that should have been done during the 2.9.2 cycle.

Supersedes #1452


sfackler:
Thanks!

PR #1469, opened 2021-05-04
Make openssl-sys no_std

IGI-111:
I have been writing some embedded Rust that makes heavy use of openssl and I'd like to use the good wrappers instead of writing or generating my own but can't because openssl-sys requires std.

Turns out there's very little reason to.

openssl-sys's only use of std is `std::sync::Once` to run `OPENSSL_init_ssl` a single time.
Replacing the standard implementation by conquer_once's spinlock, the
wrappers can now be used in no_std context with no compatibility loss.

I didn't think it pertinent to modify the feature-set of the crate for something so slight, but if the small extra dependency is a pain, putting no_std behind a flag is always a possibility.

### DIFF
--- a/openssl-sys/Cargo.toml
+++ b/openssl-sys/Cargo.toml
@@ -16,6 +16,7 @@ vendored = ['openssl-src']
 
 [dependencies]
 libc = "0.2"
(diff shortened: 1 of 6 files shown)


theoparis:
Why is this closed?

PR #1473, opened 2021-05-17
Add SSL_peek support

NerLOR:


openssl/src/ssl/mod.rs
```diff
@@ -3704,6 +3709,49 @@ impl<S: Read + Write> SslStream<S> {
...
+            Ok(ret as usize)
+        } else {
+            Err(self.make_error(ret))
+        }
+    }
+
+    /// Receives data on the socket from the remote address to which it is connected, without removing that data from the queue.
+    ///
+    /// Successive calls return the same data.
+    pub fn peek(&mut self, buf: &mut [u8]) -> io::Result<usize> {
```
sfackler, line 3737: I don't think this variant is necessary since we don't need to conform to an existing trait.

NerLOR: Yeah, you're right. First I thought, it'd be rather convenient to have a method which behaves like the corresponding method of `TcpStream` (as `.read()` and `.write()` do). But if no trait requires it, it is unnecessary to "ignore" possible error codes...

I'll remove it and update the PR sometime later today :)

NerLOR:
Forgot to `git add` changes before amending... Sorry

openssl/src/ssl/mod.rs
```diff
@@ -3704,6 +3709,26 @@ impl<S: Read + Write> SslStream<S> {
         }
     }
 
+    /// It is particularly useful with a nonblocking socket, where the error value will identify if
+    /// OpenSSL is waiting on read or write readiness.
```
sfackler, line 3713:
```suggestion
    /// Reads data from the stream, without removing it from the queue.
```

PR #1474, opened 2021-05-17
Add PKeyMethod

brandonweeks:
Seeking an initial review, as I'm not sure how to best idiomatically wrap `EVP_PKEY_METHOD` in the existing style of this crate.

Typical usage is either going to be calling `EVP_PKEY_meth_new` and setting all applicable function for a given algorithm or calling `EVP_PKEY_meth_new` + `EVP_PKEY_meth_find` + `EVP_PKEY_meth_copy` to make a mutable copy of the default struct and setting a subset of the functions. Then `EVP_PKEY_meth_add0` is called to replace the existing struct with ours.

Some observations:
- `EVP_PKEY_meth_find` returns `*const EVP_PKEY_METHOD` and does not have an `up_ref` function, so I don't believe it is possible to implement `ToOwned`.
- `EVP_PKEY_meth_add0` replaces the existing struct for a given algorithm globally. 
- I couldn't figure out a way to adapt the callbacks to Rust closures in the style of openssl/src/ssl/callbacks.rs.

https://www.openssl.org/docs/man1.1.1/man3/EVP_PKEY_meth_new.html

### DIFF
--- a/openssl-sys/src/evp.rs
+++ b/openssl-sys/src/evp.rs
@@ -25,6 +25,8 @@ pub const EVP_CTRL_GCM_SET_IVLEN: c_int = 0x9;
 pub const EVP_CTRL_GCM_GET_TAG: c_int = 0x10;
 pub const EVP_CTRL_GCM_SET_TAG: c_int = 0x11;
 
+pub const EVP_PKEY_FLAG_AUTOARGLEN: c_int = 2;
+
 pub unsafe fn EVP_get_digestbynid(type_: c_int) -> *const EVP_MD {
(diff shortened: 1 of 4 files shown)


wongsyrone:
It seems the repo author intends to add support to OpenSSL 3 and defer feature PRs after OpenSSL 3.

FYI, EVP_PKEY_meth_new() has been deprecated from OpenSSL 3.0. https://github.com/openssl/openssl/blob/master/doc/man7/migration_guide.pod#engines-and-method-apis

brandonweeks:
Sounds like I'm writing an OpenSSL 3 provider instead. Thanks @wongsyrone.

PR #1477, opened 2021-05-27
Add d2i_X509_bio to openssl-sys

coolreader18:
I wasn't sure where to find `der` keys to use to test, so I just concatenated `cms_pubkey.der` with itself to test it.

### DIFF
--- a/openssl-sys/src/ossl_typ.rs
+++ b/openssl-sys/src/ossl_typ.rs
@@ -1010,7 +1010,41 @@ cfg_if! {
     }
 }
 
-pub enum COMP_METHOD {}
+pub enum COMP_CTX {}
+
+cfg_if! {
+    if #[cfg(ossl110)] {
+        pub enum COMP_METHOD {}
+    } else {
+        #[repr(C)]
+        pub struct COMP_METHOD {
+            pub type_: c_int,
+            pub name: *const c_char,
+            init: Option<unsafe extern "C" fn(*mut COMP_CTX) -> c_int>,
+            finish: Option<unsafe extern "C" fn(*mut COMP_CTX)>,
+            compress: Option<
+                unsafe extern "C" fn(
+                    *mut COMP_CTX,
+                    *mut c_uchar,
+                    c_uint,
+                    *mut c_uchar,
+                    c_uint,
+                ) -> c_int,
+            >,
+            expand: Option<
+                unsafe extern "C" fn(
+                    *mut COMP_CTX,
+                    *mut c_uchar,
+                    c_uint,
+                    *mut c_uchar,
+                    c_uint,
+                ) -> c_int,
+            >,
+            ctrl: Option<unsafe extern "C" fn() -> c_long>,
+            callback_ctrl: Option<unsafe extern "C" fn() -> c_long>,
+        }
+    }
+}
 
 cfg_if! {
     if #[cfg(any(ossl110, libressl280))] {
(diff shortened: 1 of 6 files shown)


sfackler:
Could you give some background on the motivation for this change? I have never heard of storing concatenated DER-encoded certificates before.

coolreader18:
In Python, on windows, the `ssl` module loads certificates from the windows certificate store by iterating through them and just concatenating the DER certificate data together, then passing that to the wrapper for `X509_STORE_add_cert`, which reads either PEM or DER as concatenated through a BIO. Admittedly I don't know of other examples of concatenated DER, but that's the use case I have.

sfackler:
How would this crate be interacting with Python's internal implementation details?

coolreader18:
Oh this is for use in RustPython, the Python code concatenates the certificates and then passes it to Rust. Also in general that's part of the API that the sal module provides, that you can pass DER-encoded data and if there are multiple certificates it will extract them all.

coolreader18:
I suppose that's pretty theoretical in terms of usage, but I guess it's something common enough that it was deemed useful to put in Python's ssl library. Also it mirrors the existing pair of functions; from_pem+stack_from_pem, and now from_der+stack_from_der

coolreader18:
Actually, while I'm doing this, could I just dump the other manual bindings we have? Not all of them would necessarily make any sense wrapping in the `openssl` crate, but given that `openssl-sys` is just for the entire raw API it should be fine.

sfackler:
Sure!

Looks like some need to be flagged off from 1.0.1 though.

alex:
Is the goal here to support implementing something with the same API as Python's `ssl` module, using rust-openssl? Or just to parse the data that Python's capturing for Windows CAs?

coolreader18:
@alex I realized after I typed it that I could just edit the `ssl.py` module to call `add_verify_locations` individually per CA, so short-term I'm just trying to get windows CAs loaded into the SSLContext, but yeah, we're re-implementing the `_ssl` module's API in Rust.

alex:
Coolio! If you're implementing the full `SSLContext` in rust then yeah, you'll need to support the many-DERs `cadata` arg (and I defer to @sfackler on what the best API for that is). If your goal was just to parse the Winows CA data, there's a function `enum_certificates` (IIRC) that'd get you the windows CAs one at a time.

coolreader18:
Ok, looks like CI's finally passing, I swear I wasn't mining crypto :upside_down_face:

openssl-sys/src/ossl_typ.rs
```diff
@@ -1010,7 +1010,41 @@ cfg_if! {
...
-pub enum COMP_METHOD {}
+pub enum COMP_CTX {}
+
+cfg_if! {
+    if #[cfg(ossl110)] {
+        pub enum COMP_METHOD {}
+    } else {
+        #[repr(C)]
+        pub struct COMP_METHOD {
+            pub(crate) type_: c_int,
```
sfackler, line 1021: Why pub(crate) here? IIRC nothing in openssl-sys would ever be looking at this field in particular.

sfackler:
LGTM other than the one question!

openssl-sys/src/ossl_typ.rs
coolreader18: Oh the shim for `COMP_get_type` uses that, but maybe shims should be in consumers of `openssl-sys` rather than `openssl-sys` itself?

sfackler: Yeah, unless there's a `COMP_get_type` macro defined in the OpenSSL sources, I don't think we'd want to make a shim in openssl-sys.

It's fine to just make all fields of openssl-sys structs public unless there's some specific field or type that you don't want to bother properly defining.

sfackler:
Thanks!

PR #1480, opened 2021-05-28
Add extern C functions for ENGINE to use hardware private keys

TheEnbyperor:
Needed this functions to use PKCS#11 to access private keys using libp11, so am now suggesting they're merged into upstream.

More work would be needed to expose a safe interface for these, although given that loading an engine results in telling openssl to randomly open .so files I'm sure a safe interface is even possible / desirable.

Hope these are useful contributions.


sfackler:
It looks like you'll need to add another header here to get systest compiling: https://github.com/sfackler/rust-openssl/blob/master/systest/build.rs#L47

openssl-sys/src/engine.rs
```diff
@@ -0,0 +1,62 @@
...
+    meth: *const UI_METHOD,
+    strings: *mut c_void,
+    user_data: *mut c_void,
+    ex_data: CRYPTO_EX_DATA,
+    flags: c_int,
+}
+
+#[repr(C)]
+pub struct UI_STRING {
+    string_type: UI_string_types,
```
sfackler, line 48: We should be able to just use a c_int here since it's not publicly exposed, which also avoids potential UB issues if new variants are added to UI_string_types on OpenSSL's end.

TheEnbyperor: It is documented on the openssl website (https://www.openssl.org/docs/man1.1.1/man3/UI_string_types.html) so I assumed this is something that should be public.

sfackler: But none of the fields in the structs you've defined are public.

TheEnbyperor: Ah yeah, my bad, will fix!

artemist:
On my system UI_METHOD is an incomplete type. I have a quick fix in my repo at https://github.com/artemist/rust-openssl/commit/43af0af52491d49d4ce7f9e05a508a8b6c161d45 but I don't believe this is ready for upstream (I've only tested one OpenSSL version)

taikulawo:
Same here, any progress? We need openssl engine do some intel QAT offload

larrydewey:
Would there be any objections if I were to pick this up and finish the work here?

IniterWorker:
It seems like the previous struct comment is missing:

```c

/* .... */

/*
 * Duplicate the ui_data that often comes alongside a ui_method. This
 * allows some backends to save away UI information for later use.
 */
void *(*ui_duplicate_data)(UI *ui, void *ui_data);
void (*ui_destroy_data)(UI *ui, void *ui_data);

/* .... */

/*
 * UI_METHOD specific application data.
 */
CRYPTO_EX_DATA ex_data;
```

See [OpenSSL](https://github.com/openssl/openssl/blob/ca1d2db291530a827555b40974ed81efb91c2d19/crypto/ui/ui_local.h#L20).

I checked the rest, and I think with the current state of this crate, we have everything to test it. The remaining question is: do we need to provide a safe interface that will take care of allocating and deallocating the resources?

IniterWorker:
It looks like you'll need to add another header to get `systest` compiling: [build.rs#L51](https://github.com/sfackler/rust-openssl/blob/master/systest/build.rs#L51).

Specifically, it is missing `openssl/engine.h`.

### Details

I am working a bit on this and hit a wall addressing the support of the `UI_METHOD` type.

The following error is encountered:

```
invalid application of ‘sizeof’ to incomplete type ‘UI_METHOD’ {aka ‘struct ui_method_st’}
```

### Relevant Code

Here is the current definition of `UI_METHOD` in Rust:

```rust
#[repr(C)]
pub struct UI_METHOD {
    name: *const c_char,
    ui_open_session: extern "C" fn(ui: *mut UI) -> c_int,
    ui_write_string: extern "C" fn(ui: *mut UI, uis: *mut UI_STRING) -> c_int,
    ui_flush: extern "C" fn(ui: *mut UI) -> c_int,
    ui_read_string: extern "C" fn(ui: *mut UI, uis: *mut UI_STRING) -> c_int,
    ui_close_session: extern "C" fn(ui: *mut UI) -> c_int,
    ui_duplicate_data: extern "C" fn(ui: *mut UI, ui_data: *mut c_void) -> *mut c_void,
    ui_destroy_data: extern "C" fn(ui: *mut UI, ui_data: *mut c_void),
    ui_construct_prompt: extern "C" fn(
        ui: *mut UI,
        object_desc: *const c_char,
        object_name: *const c_char,
    ) -> *mut c_char,
    ex_data: CRYPTO_EX_DATA,
}
```

PR #1493, opened 2021-06-24
Add support for PKey raw byte representation

tomleavy:
Added x25519, ed25519, x448, and ed448 PKey support for getting a raw byte representation of their public and private keys, along with being able to construct these keys from the raw byte representation.

These keys are different than standard EC keys in Openssl because they do not work with the standard EC key functions. The methods added are the only way you can get access to the actual bytes of the key without additional encoding.

### DIFF
--- a/openssl/src/pkey.rs
+++ b/openssl/src/pkey.rs
@@ -250,6 +250,34 @@ where
     {
         unsafe { ffi::EVP_PKEY_cmp(self.as_ptr(), other.as_ptr()) == 1 }
     }
+
+    /// Raw byte representation of a public key
+    ///
+    /// This function only works for algorithms that support raw public keys.
+    /// Currently this is: X25519, ED25519, X448 or ED448
+    ///
+    /// This corresponds to [`EVP_PKEY_get_raw_public_key`].
+    ///
+    /// [`EVP_PKEY_get_raw_public_key`]: https://www.openssl.org/docs/man1.1.1/man3/EVP_PKEY_get_raw_public_key.html
+    #[cfg(ossl111)]
+    pub fn raw_public_key(&self) -> Result<Vec<u8>, ErrorStack> {
+        unsafe {
+            let mut len = 0;
+            cvt(ffi::EVP_PKEY_get_raw_public_key(
+                self.as_ptr(),
+                ptr::null_mut(),
+                &mut len,
+            ))?;
+            let mut buf = vec![0u8; len];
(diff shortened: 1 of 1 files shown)


openssl/src/pkey.rs
```diff
@@ -250,6 +250,34 @@ where
...
+    /// Raw byte representation of a public key
+    ///
+    /// This corresponds to [`EVP_PKEY_get_raw_public_key`].
+    ///
+    /// [`EVP_PKEY_get_raw_public_key`]: https://www.openssl.org/docs/man1.1.1/man3/EVP_PKEY_get_raw_public_key.html
+    ///
+    /// This function only works for algorithms that support raw public keys.
+    /// Currently this is: X25519, ED25519, X448 or ED448
+    #[cfg(ossl111)]
+    pub fn public_key_to_raw_bytes(&self) -> Result<Vec<u8>, ErrorStack> {
```
sfackler, line 263: I would just call these `raw_public_key` and `raw_private_key` to match the function names.

tomleavy: Sure, no problem. Are those tests failing due to something in my code changes? Or is it a problem with the CI?

sfackler: You need to add the raw FFI bindings to openssl-sys.

sfackler: If the methods are only available on 1.1.1, you need to add cfgs to the tests as well.

tomleavy: Just pushed an update that should address everything

openssl/src/pkey.rs
```diff
@@ -627,6 +683,63 @@ impl PKey<Private> {
             .map(|p| PKey::from_ptr(p))
         }
     }
+
+    #[cfg(ossl111)]
+    fn private_key_from_raw_bytes(
```
sfackler, line 688: It seems like we can just make this public (taking in the `Id` type: https://docs.rs/openssl/0.10.35/openssl/pkey/struct.Id.html) and not need to have a bunch of different constructors.

tomleavy:
Ok @sfackler this one should hopefully pass the CI and has only one function based on Id

tomleavy:
@sfackler can you approve the workflow so I can confirm I got the flags around the test cases correct?

tomleavy:
@sfackler sorry about that, tested this time locally with older openssl. Should be good now to pass CI

sfackler:
Thanks!

PR #1498, opened 2021-07-02
Added HKDF derivation functionality

tomleavy:
Added `HkdfDeriver` to match the API from OpenSSL `EVP_PKEY_derive`, as well as one-shot functions, which is how HKDF is typically used. Supplied tests based on the SHA-256 test from RFC 5869


openssl-sys/src/evp.rs
```diff
@@ -1,3 +1,4 @@
+use core::ptr;
```
sfackler, line 1: You can just use std::ptr instead.

sfackler:
I have been leaning towards moving away from workflow-specific wrappers like `Deriver` and `Verifier` to a "raw" `PKeyContext` type which just directly exposes all of the relevant methods. It might make sense to start that here rather than adding more types in the old approach.

No need to cover all of the methods it will eventually need to replace `Deriver` etc - just the ones required for HKDF derivation.

tomleavy:
I think that makes sense for sure, when looking at it initially I figured adding methods to the existing `Deriver` would make more sense given it's the same thing under the hood, but then I didn't want to make the API confusing given the specific nature of the existing `Deriver`

How do you want me to handle it? Make a new `PKeyContext` and support both what exists in `Deriver` and what exists here in `HkdfDeriver`. I don't want to break existing code obviously unless you decide it can be a breaking change between versions.

tomleavy:
@sfackler this should be in the direction that you are looking for, I imported the existing functionality from the `Deriver` struct as well so this module can now do both HKDF and ECDH functions within the PKeyContext. The windows-vcpkg CI you might want to just re-run, looks like it had a download failure

tomleavy:
@sfackler any way you can look at this?

openssl/src/pkey_ctx.rs
```diff
@@ -0,0 +1,384 @@
+//! PKeyContext operations
+use crate::error::ErrorStack;
+#[cfg(ossl110)]
+use crate::hash::MessageDigest;
+use crate::pkey::{HasPrivate, HasPublic, PKeyRef};
+use crate::{cvt, cvt_p};
+use core::ptr;
```
sfackler, line 7: `std::ptr`

openssl-sys/src/lib.rs
```diff
@@ -12,6 +12,7 @@
 #![doc(html_root_url = "https://docs.rs/openssl-sys/0.9")]
 #![recursion_limit = "128"] // configure fixed limit across all rust versions
 
+extern crate core;
```
sfackler, line 15: Shouldn't need to do this.

openssl/src/pkey_ctx.rs
```diff
@@ -0,0 +1,384 @@
...
+    ///
+    /// [`EVP_PKEY_derive_init`]: https://www.openssl.org/docs/man1.0.2/crypto/EVP_PKEY_derive_init.html
+    pub fn new_pkey<T>(key: &'a PKeyRef<T>) -> Result<PKeyContext<'a>, ErrorStack>
+    where
+        T: HasPrivate,
+    {
+        unsafe {
+            cvt_p(ffi::EVP_PKEY_CTX_new(key.as_ptr(), ptr::null_mut()))
+                .map(|p| PKeyContext(p, PhantomData))
+                .and_then(|ctx| cvt(ffi::EVP_PKEY_derive_init(ctx.0)).map(|_| ctx))
```
sfackler, line 44: I think we should split construction and derive_init into separate functions - you can reuse a pkey ctx for different operations.

openssl/src/pkey_ctx.rs
```diff
@@ -0,0 +1,384 @@
...
+    /// This corresponds to [`EVP_PKEY_derive`].
+    ///
+    /// # Warning
+    ///
+    /// When using this `PKeyContext` for HKDF, this function is only allowed when using HKDF with
+    /// [`Extract`](HkdfMode::Extract) mode.
+    ///
+    /// [`PKeyContext::derive`]: #method.derive
+    /// [`EVP_PKEY_derive`]: https://www.openssl.org/docs/man1.0.2/crypto/EVP_PKEY_derive_init.html
+    pub fn len(&mut self) -> Result<usize, ErrorStack> {
```
sfackler, line 196: This should probably be `derive_len`.

openssl/src/pkey_ctx.rs
```diff
@@ -0,0 +1,384 @@
...
+use crate::{cvt, cvt_p};
+use core::ptr;
+use foreign_types::ForeignTypeRef;
+use std::marker::PhantomData;
+#[cfg(ossl110)]
+use std::os::raw::{c_int, c_uchar};
+
+/// HKDF modes of operation. See [`hkdf_mode`](PKeyContext::hkdf_mode)
+#[cfg(ossl111)]
+pub enum HkdfMode {
```
sfackler, line 15: This should be a newtype around an integer with associated constants rather than an enum. See bn::MsbOption as an example.

sfackler:
Sorry for forgetting about this!

openssl/src/pkey_ctx.rs
```diff
@@ -0,0 +1,384 @@
...
+
+    /// Creates a new `PKeyContext` in HKDF derivation mode using the digest provided.
+    ///
+    /// This corresponds to [`EVP_PKEY_CTX_new_id`] using `EVP_PKEY_HKDF` followed by
+    /// [`EVP_PKEY_derive_init`].
+    ///
+    /// [`EVP_PKEY_CTX_new_id`]: https://www.openssl.org/docs/man1.1.1/man3/EVP_PKEY_CTX_new_id.html
+    /// [`EVP_PKEY_derive_init`]: https://www.openssl.org/docs/man1.1.1/man3/EVP_PKEY_derive_init.html
+    #[cfg(ossl110)]
+    pub fn new_hkdf() -> Result<Self, ErrorStack> {
```
sfackler, line 68: Rather than having this specific to hkdf, I'd instead have a constructor taking `pkey::Id`.

openssl/src/pkey_ctx.rs
tomleavy: I split into just a `new` method and `derive_init` method

openssl/src/pkey_ctx.rs
tomleavy: turned into `new_id` which requires an Id and the user will manually call derive_init after

openssl/src/pkey_ctx.rs
tomleavy: updated

tomleavy:
Hey @sfackler looks like there is a conflict. Made the changes you asked for, you can likely just merge in the HKDF specific stuff from my branch

PR #1499, opened 2021-07-06
CI: speed up windows-vcpkg build

wongsyrone:
Based on my daily experience, windows-vcpkg installs `vcpkg` each time when the github action runs, switch to `lukka/run-vcpkg` to benefit from the `vcpkg` cache.

To test it truly makes sense, I added several commits to test CI behavior, if you want to merge this PR, I will rebase them into a single commit.

Testing result:

Current:  10m 36s
https://github.com/wongsyrone/rust-openssl/runs/2994266452?check_suite_focus=true
First time using this PR: 11m 37s
https://github.com/wongsyrone/rust-openssl/runs/2994422238?check_suite_focus=true
Second time using this PR: 2m 39s
https://github.com/wongsyrone/rust-openssl/runs/2994476156?check_suite_focus=true

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,8 +86,15 @@ jobs:
       - uses: sfackler/actions/rustup@master
       - run: echo "::set-output name=version::$(rustc --version)"
         id: rust-version
-      - run: echo "VCPKG_ROOT=$env:VCPKG_INSTALLATION_ROOT" | Out-File -FilePath $env:GITHUB_ENV -Append
-      - run: vcpkg install openssl:x64-windows
+      - name: Prepare vcpkg
+        uses: lukka/run-vcpkg@v7
+        id: runvcpkg
+        with:
+          vcpkgArguments: openssl:x64-windows
+          vcpkgDirectory: ${{ runner.workspace }}/vcpkg/
+          vcpkgTriplet: x64-windows
+          # 2021.05.12 release
+          vcpkgGitCommitId: 5568f110b509a9fd90711978a7cb76bae75bb092
       - uses: actions/cache@v1
         with:
           path: ~/.cargo/registry/index


.github/workflows/ci.yml
```diff
@@ -86,8 +87,15 @@ jobs:
...
       - run: echo "::set-output name=version::$(rustc --version)"
         id: rust-version
-      - run: echo "VCPKG_ROOT=$env:VCPKG_INSTALLATION_ROOT" | Out-File -FilePath $env:GITHUB_ENV -Append
-      - run: vcpkg install openssl:x64-windows
+      - name: Prepare vcpkg
+        uses: lukka/run-vcpkg@v7
+        id: runvcpkg
+        with:
+          vcpkgArguments: openssl:x64-windows
+          vcpkgDirectory: ${{ runner.workspace }}/vcpkg/
```
sfackler, line 95: `runner.workspace` isn't documented as a key in the Actions docs.

I think this option only applies when using vcpkg as a submodule of the repository though, which we aren't doing.

.github/workflows/ci.yml
```diff
@@ -86,8 +87,15 @@ jobs:
...
+          vcpkgTriplet: x64-windows
+          # 2021.05.12 release
+          vcpkgGitCommitId: 5568f110b509a9fd90711978a7cb76bae75bb092
```
sfackler, line 98: I think I'd rather not pin this so we keep up with the OpenSSL that users would end up with.

wongsyrone:
Another benefit when I skimmed the building result:

> I think I'd rather not pin this so we keep up with the OpenSSL that users would end up with.

I just want to write something on this. In theory, https://github.com/sfackler/rust-openssl/commit/291f35e176b98d1c6031863a02bc22d79864249b has nothing to do with windows-vcpkg, but it fails magically (https://github.com/sfackler/rust-openssl/runs/2992387706). The root cause is 
```
Error: Failed to download from mirror set:
https://strawberryperl.com/download/5.30.0.1/strawberry-perl-5.30.0.1-32bit.zip: WinHttpSendRequest() failed: 12002
```

If you check more, the vcpkg version is outdated and never got updated during CI workflow.

```
Error: Building package openssl:x64-windows failed with: BUILD_FAILED
Please ensure you're using the latest portfiles with `.\vcpkg update`, then
submit an issue at https://github.com/Microsoft/vcpkg/issues including:
  Package: openssl:x64-windows
  Vcpkg version: 2021-05-05-b110959e13f562c2a17d7bd9e352e2dd8d6dc603
```

The commit I used is just a `stable` version that passes all `vcpkg` test. You can indeed use vcpkg's latest commit (https://github.com/microsoft/vcpkg/commit/b361c2eefa3966cb7cec45275aff32e90430aaa6), but it doesn't even build (https://github.com/microsoft/vcpkg/runs/2980699203) and fails with 12 errors

.github/workflows/ci.yml
wongsyrone: Posted my thoughts below.

.github/workflows/ci.yml
wongsyrone: The `best practices` part documents

```
To sum up, you need to pin the specific version of vcpkg you want to use to keep a consistent development experience between local and remote build environments. This is accomplished by using vcpkg as submodule of your Git repository; this way the version of vcpkg used is implied by the commit id specified by the submodule for vcpkg.
```

If you remove it, it will download the  vcpkg git repository each time, which makes no sense

https://github.com/marketplace/actions/run-vcpkg#best-practices

wongsyrone:
It seems the vcpkg commit id is a must-have.

```
Error: 'vcpkgGitCommitId' input parameter must be provided when the specified vcpkg directory (D:\a\rust-openssl\vcpkg) is not a submodule.
    at VcpkgRunner.<anonymous> (D:\a\_actions\lukka\run-vcpkg\v7\dist\index.js:5774:31)
    at Generator.next (<anonymous>)
    at fulfilled (D:\a\_actions\lukka\run-vcpkg\v7\dist\index.js:5511:58)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

.github/workflows/ci.yml
sfackler: But the directory you are pointing to here is not a submodule containing a vcpkg repository.

sfackler:
Does that commit reference just the version of the vcpkg tool or also the versions of openssl it uses?

wongsyrone:
> Does that commit reference just the version of the vcpkg tool or also the versions of openssl it uses?

It includes both of them. OpenSSL building rules locates at https://github.com/microsoft/vcpkg/tree/master/ports/openssl

wongsyrone:
You can check commits related to openssl at https://github.com/microsoft/vcpkg/commits/master/ports/openssl, I want to use https://github.com/microsoft/vcpkg/commit/06f8fd637d928837e9484047f1e41c1b2159835c before this PR, unfortunately, I don't know whether it passes vcpkg tests or not.

wongsyrone:
Commits to test CI has been removed and squashed into one commit. Ready to merge if you think it's okay.

.github/workflows/ci.yml
wongsyrone: Yes, I think adding the whole vcpkg repo as git submodule bloats more and not git clone friendly, that's why the github actions requires to pin at a specific commit.

wongsyrone:
I checked all real world examples from https://github.com/marketplace/actions/run-vcpkg#real-world-project-samples

https://github.com/otland/forgottenserver/blob/master/.github/workflows/build-vcpkg.yml
https://github.com/libevent/libevent/blob/master/.github/workflows/windows.yml
https://github.com/zealdocs/zeal/blob/master/.github/workflows/build-check.yml
https://github.com/marian-nmt/marian-dev/blob/master/.github/workflows/windows.yml
https://github.com/GrinPlusPlus/GrinPlusPlus/blob/master/.github/workflows/ci.yml
https://github.com/scummvm/scummvm/blob/master/.github/workflows/ci.yml

Links above pins git commit

https://github.com/OpenTTD/OpenTTD/pull/8718 opt-out this github action 

Others add vcpkg as git submodules to their repo.

sfackler:
The point of the vcpkg test is to make sure the library builds properly via vcpkg as people would actually use it, not against an arbitrarily old version of vcpkg. If that action can't support this workflow, then it seems like we can't use it directly.

Issue #1447, opened 2021-04-06
No implementation of KISA Seed encryption algorithm


nickkma:
Is ffi for  [SEED](https://www.openssl.org/docs/man1.1.1/man3/EVP_seed_cbc.html)  implemented in this library? If not is there a framework/template for doing so?

sfackler:
It does not appear to be, but PRs welcomed.

PR #1513, opened 2021-08-14
Seed cipher

uniqueNullptr2:
added the SEED cipers as requested in #1447


sfackler:
It looks like these will need to be conditionally compiled based when OPENSSL_NO_SEED is not set.

uniqueNullptr2:
hm I forgot to pull first. I hope I didn't mess up git.
was #[cfg(not(osslconf = "OPENSSL_NO_SEED"))] correct?
how would I test such a compiler flag locally?

Also there was te case of EVP_seed_cfb() which caused a linker error on my machine. the cfb128 works fine but both were listed in the docs. I have however not found what the difference is supposed to be.

sfackler:
You'll also need to update this file to get the build system to look for OPENSSL_NO_SEED: https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/build/expando.c

uniqueNullptr2:
So I configured the tests to not run on ossl300 like the other legacy algorithms defined in this [list](https://www.openssl.org/docs/manmaster/man7/OSSL_PROVIDER-legacy.html).
On that occasion I noticed, that these other legacy algorithms are configured in a way where only the tests don't build on versions that dont these algorithms and I was wondering ig I should remove the cfg flags from the method definitions.

Also I would really appreciate it if you could point me to some ressources on how I can build rust-openssl with different flags (Like the OPENSSL_NO_SEED) and how to easily switch version.

sfackler:
I think it looks right for now - when we add support to load the legacy provider we can re-enable the tests on 3.0.0.

sfackler:
Thanks!

PR #1515, opened 2021-08-16
add OBJ_get0_data()

ericchiang:
This allows encoding of an OBJECT IDENTIFIER, which uses complex base
128 encoding rules.

https://luca.ntop.org/Teaching/Appunti/asn1.html

Same as_slice() API as ASN1_STRING_get0_data().

Test data generated with the following Go program:
https://play.golang.org/p/vA1PBcx1R4v


sfackler:
It looks like these aren't defined in 1.0.1.

sfackler:
It's also fine to just only expose this method when not compiling against 1.0.1.

ericchiang:
Didn't realize that was an option :) I accidentally forced pushed on this branch so I can't reopened. Opened https://github.com/sfackler/rust-openssl/pull/1516 instead

PR #1516, opened 2021-08-17
add OBJ_get0_data()

ericchiang:
Follow up from https://github.com/sfackler/rust-openssl/pull/1515

I accidentally force pushed to my branch so I can't reopen the original one :(

This allows encoding of an OBJECT IDENTIFIER, which uses complex base
128 encoding rules.

https://luca.ntop.org/Teaching/Appunti/asn1.html

Same as_slice() API as ASN1_STRING_get0_data().

Test data generated with the following Go program:
https://play.golang.org/p/vA1PBcx1R4v


sfackler:
Thanks!

Issue #909, opened 2018-05-07
Use accessor functions on LibreSSL 2.7.2+


sfackler:
We can finally move beyond the direct struct access nightmare!

sfackler:
I did most of this in #946, but a couple of functions are still defined as macros :(

vext01:
What's the status of this please? Is there any hope that openssl-sys can be used with new versions of LibreSSL?

sfackler:
Last time I checked, there are still 3 or 4 structures that don't have a proper stable ABI.

ruuda:
The macro mentioned in https://github.com/sfackler/rust-openssl/issues/1278#issuecomment-633727309 [is no longer a macro in the current version](https://github.com/libressl-portable/openbsd/blob/36c8919b69b46cf5f20524ede4602a2e7a381968/src/lib/libcrypto/bn/bn.h#L394). Since 3.5.0 (released February 2022), [many functions are now available that were previously macros](https://github.com/libressl-portable/portable/blob/bf23b6b37255b232f7768555f4b6dd7153de2dce/ChangeLog#L365-L366).

Are there any structs left now without a stable ABI? Would it finally be possible to resolve #1278?

PR #1517, opened 2021-08-20
Set the maximum libressl version to 3.x.x

orbea:
This set the maximum libressl version to `3.x.x`.

This would be very helpful as it is constantly tedious to have to patch rust to use the latest libressl version which to my knowledge has always worked for the `3.x` series. In the event a future version does cause problems I suggest reexamining this issue then if necessary.

### DIFF
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -226,17 +226,7 @@ See rust-openssl README for more information:
             (2, 8, _) => ('2', '8', 'x'),
             (2, 9, 0) => ('2', '9', '0'),
             (2, 9, _) => ('2', '9', 'x'),
-            (3, 0, 0) => ('3', '0', '0'),
-            (3, 0, 1) => ('3', '0', '1'),
-            (3, 0, _) => ('3', '0', 'x'),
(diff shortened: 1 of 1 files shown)


sfackler:
See #909.

sfackler:
LibreSSL does not have a stable ABI and has in the past made changes which have caused UB in this library. If they get their shit together and create a stable ABI we can loosen the version restrictions at that time.

orbea:
@sfackler Can you show some examples? I think you are mistaken, libressl says this:

> LibreSSL's installed library version numbers are incremented to account for ABI and API changes.

So in the event the ABI or API is broken for any `3.x.x` release that would be considered a libressl bug which I am unaware of ever happening before. If and when `4.0.0` is released then there may be incompatible changes.

sfackler:
The layout of SSL_SESSION changed between 2.5.0 and 2.5.1: https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/src/ssl.rs#L30.

Additionally, the first several releases of any series are "development" releases, which apparently means the may have unstable ABIs.

PR #1479, opened 2021-05-28
Keep LibreSSL 3.4.0 up-to-date for OpenBSD -current

gonzalo-:
I hope this is enough:

```
test result: ok. 46 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 8.95s

   Doc-tests openssl-sys

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

Cheers.-

PR #1524, opened 2021-09-21
 Keep LibreSSL 3.4.0 up-to-date for OpenBSD -current, continuation

hargoniX:
This PR is a continuation of #1479 since the original author doesn't have time to fix it at the moment. I addressed the review comments and ran tests on an OpenBSD snapshot from a few days ago, everything is working fine still.


sfackler:
Thanks!

PR #1528, opened 2021-09-30
Support LibreSSL 3.5 (and fix up other LibreSSL bits)

vishwin:
Continuing from #1479 and #1524, sync more bits with LibreSSL 3.4.0. While here, sync bits from previous LibreSSL versions.

I added a CI for the current stable branch, as 3.4.x is not stable yet, but mainly to compare tests between branches. Two tests ignored from LibreSSL 3.2.1 now pass, but the other four still fail in the same fashion. I'm inclined to ignore them again until we can figure out at a deeper level what behaviour changed between 3.2.0 and 3.2.1.

### DIFF
--- a/openssl/src/ssl/connector.rs
+++ b/openssl/src/ssl/connector.rs
@@ -4,6 +4,8 @@ use std::ops::{Deref, DerefMut};
 
 use crate::dh::Dh;
 use crate::error::ErrorStack;
+#[cfg(libressl340)]
+use crate::ssl::SslVersion;
 use crate::ssl::{
     HandshakeError, Ssl, SslContext, SslContextBuilder, SslContextRef, SslMethod, SslMode,
     SslOptions, SslRef, SslStream, SslVerifyMode,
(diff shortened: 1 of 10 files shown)


openssl/src/ssl/connector.rs
```diff
@@ -247,13 +249,16 @@ impl SslAcceptor {
...
     ///
     /// [docs]: https://wiki.mozilla.org/Security/Server_Side_TLS
-    #[cfg(ossl111)]
+    #[cfg(any(ossl111, libressl340))]
     pub fn mozilla_modern_v5(method: SslMethod) -> Result<SslAcceptorBuilder, ErrorStack> {
         let mut ctx = ctx(method)?;
+        #[cfg(ossl111)]
         ctx.set_options(SslOptions::NO_SSL_MASK & !SslOptions::NO_TLSV1_3);
+        #[cfg(libressl340)]
+        ctx.set_min_proto_version(Some(SslVersion::TLS1_3))?;
```
sfackler, line 261: Can't openssl use this as well?

sfackler: Why can't LibreSSL use the set_options approach?

vishwin: `NO_SSL_MASK` doesn't exist in LibreSSL.

sfackler: Is LibreSSL no longer just trying to be an "improved" implementation of the OpenSSL API? If it is going to go off and do its own thing, it is not clear to me that this crate should continue supporting it.

vishwin: According to the front page, the goals have not changed.

orbea: @sfackler I would hope supporting it so that OpenBSD and other libressl users don't have unnecessary blockers for using rust would be reason enough.

sfackler:
#1628

vishwin:
We still need to re-run tests that were skipped starting in the 3.2 cycle. That was the main blocker here, not any actual updating.

vishwin:
Also #1628 is missing configs for 3.4.

PR #1531, opened 2021-10-03
openssl-sys: Clean up old libressl version checks

orbea:
Trivial clean up patch, I think this makes it easier to read without all of the redundant versions.


sfackler:
Some of these subversions are explicitly checked for. For example: https://github.com/sfackler/rust-openssl/blob/4722f97437c4151ad5d479ce138174cc934079a9/openssl-sys/src/ssl.rs#L270

orbea:
Does it really matter? If I understand correctly the version checks will match those versions regardless and these more mature libressl releases are very unlikely to get API changes.

An alternative would be to add the those specific versions back, I can do this if you feel its better?

orbea:
Additionally it might be a good idea to remove some of these subversion checks from rust-openssl? For example does it really need to check for `2.7.0`, `2.7.1` ad `2.7.3` rather than just `2.7.3`? Are people expected to be using the older patch versions? I haven't looked closely yet, but it might be possible to reduce some of code paths for easier maintainability?

sfackler:
What maintenance burden do these definitions add?

PR #1532, opened 2021-10-04
Add BigNumRef::to_vec_padded

rshearman:
Wire protocols sometimes define a fixed size for fields that may be
derived from a BigNum, so it is useful to be able to obtain a Vec<u8>
that has been padded accordingly.

So add a BN_bn2binpad -sys function and a to_vec_padded wrapper around
this for BigNumRef.


sfackler:
Thanks!

PR #1534, opened 2021-10-05
Add X509Name to/from DER methods

rshearman:
Since X509Name is more complex than a single value (it's a a sequence
of entries) it's useful to be able to serialise/deserialise to/from
flat data, and DER is a natural form for this.

So add a {i2d,d2i}_X509_NAME -sys functions, and to_der/from_der
wrappers in X509NameRef and X509Name respectively.

### DIFF
--- a/openssl-sys/src/x509.rs
+++ b/openssl-sys/src/x509.rs
@@ -478,6 +478,7 @@ const_ptr_api! {
             loc: c_int,
             set: c_int,
         ) -> c_int;
+        pub fn i2d_X509_NAME(n: #[const_ptr_if(ossl300)] X509_NAME, buf: *mut *mut u8) -> c_int;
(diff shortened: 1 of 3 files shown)


rshearman:
Failing CI job seems to be due to a network issue and unrelated to my changes:

`Warning: read ECONNRESET
events.js:187
      throw er; // Unhandled 'error' event
      ^

Error: EBADF: bad file descriptor, read
Emitted 'error' event on ReadStream instance at:
    at internal/fs/streams.js:167:12
    at FSReqCallback.wrapper [as oncomplete] (fs.js:470:5) {
  errno: -9,
  code: 'EBADF',
  syscall: 'read'
}`

However, I don't have permissions to trigger a re-run of the job.

sfackler:
Thanks!

PR #1537, opened 2021-10-09
Add support for secure and const time flags on BigNum

tomleavy:
Introduces support for `BN_FLG_CONSTTIME ` and `BN_FLG_SECURE ` with related functions


sfackler:
Thanks!

PR #1540, opened 2021-10-18
Support OpenSSL configured with `no-chacha`

bossmc:
Pretty much as the title says, if the underlying `libcrypto.so` is configured with `no-chacha` then the `EVP_chacha20` and `EVP_chacha20_poly1305` symbols are not available, so `openssl-sys` should not declare them (and thus `openssl-rs` should not expose them either).

I also added a line to ensure that the build script gets rerun when `expando.c` is modified (since that caught me out when testing my changes).


sfackler:
Thanks!

PR #1548, opened 2021-11-01
Print out a version directive for OpenSSL 3.0.0

alexcrichton:
This was printed out for previous OpenSSL versions for downstream
consumers but it looks like this was accidentally omitted for the 3.0.0
branch.

### DIFF
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -266,6 +266,7 @@ See rust-openssl README for more information:
         if openssl_version >= 0x4_00_00_00_0 {
             version_error()
         } else if openssl_version >= 0x3_00_00_00_0 {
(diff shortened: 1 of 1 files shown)


sfackler:
I think this was actually intentional - the recommended way to check what version you're linking against is to load the exact version from `DEP_OPENSSL_VERSION_NUMBER`. 

Example: https://github.com/sfackler/rust-openssl/blob/master/openssl-errors/build.rs

alexcrichton:
Ah ok makes sense! I'll update [these lines](https://github.com/alexcrichton/curl-rust/blob/df64ee4fcb1e3e2927d04e6bb6d8a1c0feedc844/build.rs#L7) to use that instead.

Issue #1555, opened 2021-11-10
Link to libatomic for android static 32 bit 3.0.0 builds


ValHeimer:
The pull request https://github.com/sfackler/rust-openssl/pull/1547/files resolved an issue about linking libatomic on 32 bit architectures but it only works for linux platforms.

The same issue is happening when building on android target.

PR #1556, opened 2021-11-10
Also link to libatomic for android static 32 bit 3.0.0 builds

ValHeimer:
I only added the android target beside linux

Resolves https://github.com/sfackler/rust-openssl/issues/1555

### DIFF
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -111,7 +111,8 @@ fn main() {
     // https://github.com/openssl/openssl/pull/15086
     if version == Version::Openssl3xx
         && kind == "static"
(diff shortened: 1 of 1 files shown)


sfackler:
Just to confirm, what's the compilation error you run into without this change?

ValHeimer:
> Just to confirm, what's the compilation error you run into without this change?

You can find the compilation error described in the following pull request: https://github.com/rusqlite/rusqlite/pull/1037

Initially I thought it was about rusqlite but actually the problem seems to come from this crate since it is where the linking process takes place.

sfackler:
Cool, thanks! I will cut a new release tonight.

sfackler:
Done!

PR #1557, opened 2021-11-11
Pkcs12Builder: Add option to specify MAC digest algorithm

tpambor:


openssl/src/pkcs12.rs
```diff
@@ -186,10 +200,22 @@ impl Pkcs12Builder {
...
                 nid_cert,
                 self.iter,
-                self.mac_iter,
+                -1,
                 keytype,
             ))
-            .map(Pkcs12)
+            .map(Pkcs12)?;
+
+            cvt(ffi::PKCS12_set_mac(
```
sfackler, line 208: It seems like this can just be a method on `Pkcs12Ref` rather than being part of the builder.

tpambor: I think it should go to `Pkcs12Builder` as the MAC digest algorithm is closely related to the other parameters `mac_iter` and `password` when building the PKCS12 structure. `PKCS12_create()` internally calls `PKCS12_set_mac()` with `mac_iter` and `password` if `mac_iter` is not `-1`. (https://github.com/openssl/openssl/blob/c7d848e220ecb432faa05d2198ae689298d71728/crypto/pkcs12/p12_crt.c#L121)

If `-1` is passed for `mac_iter` and afterwards `PKCS12_set_mac()` is called, as in this PR, the behavior equals the previous behavior but furthermore gives access to specify the MAC digest algorithm, whose default changed recently in openssl 3.0 from `SHA-1` to `SHA-256`. This is the same approach the PKCS12 app uses. (https://github.com/openssl/openssl/blob/00cf3a2d30fc7642bf9f816a7c545115985a8c0c/apps/pkcs12.c#L683)

sfackler: Given that the builder only allows a u32 to be provided for mac_iter this seems fine for now, but I think I'm going to remove the builder in favor of a more direct interface in the next release.

PR #1578, opened 2021-12-12
Downgrade our vendored OpenSSL version back to 1.1.1

sfackler:
3.0.0 unfortunately seems to have shipped with several major performance regressions. It seems best to temporarily revert the version used by the `vendored` feature back to 1.1.1 for now. Hopefully an upcoming 3.x.x release will fix the performance issues, at which point we'll switch back.

### DIFF
--- a/openssl-sys/Cargo.toml
+++ b/openssl-sys/Cargo.toml
@@ -21,7 +21,7 @@ libc = "0.2"
 
 [build-dependencies]
 cc = "1.0"
-openssl-src = { version = "300", optional = true }
+openssl-src = { version = "111", optional = true }
 pkg-config = "0.3.9"
 autocfg = "1.0"
 


bdbai:
May I have some details about the performance regression?

Our work requires build support for Windows UWP targets which is introduced in OpenSSL 3, therefore I have to pin version 0.9.71 as a workaround. Would it be better if a feature can be used to allow users to opt-in for OpenSSL 3?

sfackler:
https://github.com/openssl/openssl/issues/16871
https://github.com/openssl/openssl/issues/15199
https://github.com/openssl/openssl/issues/17064

The vendored feature is intended to just provide a reasonable version for people that want to get a thing building - I don't think we'll want to formally support multiple variants of it for different versions. If you have specific needs there, I'd recommend not using the vendored feature and instead point openssl-sys to an existing install that you manage.

bdbai:
@sfackler Thanks for the information! Reusing a prebuilt OpenSSL from vcpkg sounds like a good idea for my case.

wez:
> The vendored feature is intended to just provide a reasonable version for people that want to get a thing building 

If the vendored feature is aimed to make it convenient to build, rather than the recommended way to pull in the best available openssl, could we revisit and reverse this PR?

In wezterm I use the vendored feature for macOS and Windows.  While wezterm itself doesn't have a preference on the openssl version, one of the deps is https://github.com/wez/libssh-rs/tree/main/libssh-rs-sys which needs a build of openssl that includes the `engine` related functions that are only built on windows starting in `openssl-src-300` and up: https://github.com/alexcrichton/openssl-src-rs/pull/109

https://github.com/wez/libssh-rs/commit/3dc392c2b6c1a4104d1478e744c6ae8f12a4e9de pinned `libssh-rs` to `openssl-sys`  `=0.9.71` to workaround #1578 which was "fine" until I ran cargo update in wezterm today and ran into https://github.com/sfackler/rust-openssl/issues/1630.  To resolve that, I need to pin wezterm to `openssl` `=0.10.38`.

I'd love to remove this pinning!  However, it is a huge PITA to do this by providing my own openssl build (as suggested in the comments above), as I'd need to contrive for this to happen for the various combinations/permutations of:

* CI, non-CI
* Windows, macOS
* libssh-rs, wezterm

It seems to me that the vendored feature is ideal for resolving this sort of situation, and that if performance were a motivating factor, that would be a good reason to not use the vendored feature and instead use either the system openssl or some other externally provided build.

What do you think about this? :)

chipsenkbeil:
Want to echo that this was an unexpected breakage to [distant](https://github.com/chipsenkbeil/distant) and indirectly [distant.nvim](https://github.com/chipsenkbeil/distant.nvim), which consumes libssh-rs and ssh2-rs and uses the vendored feature.

PR #1582, opened 2021-12-13
check for libraries in lib64, if it exists

### LINKED ISSUE #1581: `OPENSSL_DIR` does not work with OpenSSL 3.0
Setting `OPENSSL_DIR` implies `OPENSSL_LIB_DIR=$OPENSSL_DIR/lib` which isn't true for OpenSSL 3.0, OpenSSL switched to using `lib64` by default in 3.0.0-beta2.

https://www.openssl.org/news/changelog.html#openssl-30

> Add support for --libdir option and LIBDIR variable in makefiles. This makes it possible to install openssl libraries in locations which have names other than "lib", for example "/usr/lib64" which some systems need.

alex:
fixes #1581

### DIFF
--- a/openssl-sys/build/find_normal.rs
+++ b/openssl-sys/build/find_normal.rs
@@ -5,16 +5,27 @@ use std::process::{self, Command};
 
 use super::env;
 
-pub fn get_openssl(target: &str) -> (PathBuf, PathBuf) {
+pub fn get_openssl(target: &str) -> (Vec<PathBuf>, PathBuf) {
     let lib_dir = env("OPENSSL_LIB_DIR").map(PathBuf::from);
     let include_dir = env("OPENSSL_INCLUDE_DIR").map(PathBuf::from);
 
     match (lib_dir, include_dir) {
-        (Some(lib_dir), Some(include_dir)) => (lib_dir, include_dir),
+        (Some(lib_dir), Some(include_dir)) => (vec![lib_dir], include_dir),
         (lib_dir, include_dir) => {
             let openssl_dir = env("OPENSSL_DIR").unwrap_or_else(|| find_openssl_dir(target));
             let openssl_dir = Path::new(&openssl_dir);
-            let lib_dir = lib_dir.unwrap_or_else(|| openssl_dir.join("lib"));
+            let lib_dir = lib_dir.map(|d| vec![d]).unwrap_or_else(|| {
+                let mut lib_dirs = vec![];
+                // OpenSSL 3.0 now puts it's libraries in lib64/ by default,
+                // check for both it and lib/.
+                if openssl_dir.join("lib64").exists() {
+                    lib_dirs.push(openssl_dir.join("lib64"));
+                }
+                if openssl_dir.join("lib").exists() {
+                    lib_dirs.push(openssl_dir.join("lib"));
+                }
+                lib_dirs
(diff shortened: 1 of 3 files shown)


openssl-sys/build/find_normal.rs
```diff
@@ -14,7 +14,15 @@ pub fn get_openssl(target: &str) -> (PathBuf, PathBuf) {
         (lib_dir, include_dir) => {
             let openssl_dir = env("OPENSSL_DIR").unwrap_or_else(|| find_openssl_dir(target));
             let openssl_dir = Path::new(&openssl_dir);
-            let lib_dir = lib_dir.unwrap_or_else(|| openssl_dir.join("lib"));
+            let lib_dir = lib_dir.unwrap_or_else(|| {
+                // OpenSSL 3.0 now puts it's libraries in lib64/ by default.
+                // Check if that exists and if not, use lib/.
+                if openssl_dir.join("lib64").exists() {
```
messense, line 20: Is it possible that user installs OpenSSL to `/usr` and OpenSSL libs reside in `/usr/lib` but `/usr/lib64` directory also exists?

```bash
# ls /usr/lib64
ld-linux-x86-64.so.2  ld-lsb-x86-64.so.2  ld-lsb-x86-64.so.3
```

alex: Hmm, yes it is. In the `/usr` case I'd kind of expect it to be covered by `pkg-config`. What do you think @sfackler ?

sfackler: pkg-config would cover a /usr install in normal contexts, but not if `OPENSSL_DIR=/usr` was set explicitly. We could just add both `$OPENSSL_DIR/lib` and `$OPENSSL_DIR/lib64` as search paths?

alex: Sure, I'll do that.

alex: Done.

openssl-sys/build/find_normal.rs
```diff
@@ -5,16 +5,27 @@ use std::process::{self, Command};
...
+        (Some(lib_dir), Some(include_dir)) => (vec![lib_dir], include_dir),
         (lib_dir, include_dir) => {
             let openssl_dir = env("OPENSSL_DIR").unwrap_or_else(|| find_openssl_dir(target));
             let openssl_dir = Path::new(&openssl_dir);
-            let lib_dir = lib_dir.unwrap_or_else(|| openssl_dir.join("lib"));
+            let lib_dir = lib_dir.map(|d| vec![d]).unwrap_or_else(|| {
+                let mut lib_dirs = vec![];
+                // OpenSSL 3.0 now puts it's libraries in lib64/ by default,
+                // check for both it and lib/.
+                if openssl_dir.join("lib64").exists() {
```
sfackler, line 21: It should be fine to just unconditionally return both paths I think.

alex: That's how I originally wrote the code, but then `determine_kind` calls `read_dir` on it and I'd have to add error handling there so just checking for existence on the front seemed easier and more consistent.

sfackler: Ugh, determine_kind is such an unfortunate function...

PR #1584, opened 2021-12-15
openssl: set ossl110h cfg

Fabian-Gruenbichler:
this is used as conditional guard for disabling renegotiation via
SslOptions.

Fixes: 9189b6732646e2af9d0847ef4230aec1d1e61730

Signed-off-by: Fabian Grünbichler <f.gruenbichler@proxmox.com>


sfackler:
Thanks!

PR #1591, opened 2022-01-12
Some miscellaneous documentation fixes

nashley:
Mostly fixing typos, uncommon spellings, etc. ~I also marked some insecure items as `deprecated`.~

Feel free to pick and choose which edits you think are useful ~(e.g., you may not want to be in the business of deciding which features should be `deprecated` or not)~.

- [ ] Do any of these changes need to be upstreamed to [`openssl`](https://github.com/openssl/openssl) itself (i.e., are they generated from `openssl`'s comments)?


sfackler:
Thanks! The documentation here is separate from the main OpenSSL repo, so these typos probably wouldn't be present there.

Issue #1593, opened 2022-01-16
AES-CCM with Tag Length of 16 bytes fails


kedars:
Specifically, the `{encrypt, decrypt}_aead` function returns TLS stack error. 

The following is a test code from test case 225 of NIST Test Vectors:
```
    #[test]
    fn test_aes128_ccm_tag16() {
        // Test case 225 from
        // https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Algorithm-Validation-Program/documents/mac/ccmtestvectors.zip
        let key = "26511fb51fcfa75cb4b44da75a6e5a0e";
        let nonce = "5a8aa485c316e9403aff859fbb";
        let aad = "a16a2e741f1cd9717285b6d882c1fc53655e9773761ad697a7ee6410184c7982";

        let pt = "8739b4bea1a099fe547499cbc6d1b13d849b8084c9b6acc5";
        let ct = "50038b5fdd364ee747b70d00bd36840ece4ea19998123375";
        let tag = "c0a458bfcafa3b2609afe0f825cbf503";

        let mut actual_tag = [0; 16];
        let out = encrypt_aead(
            Cipher::aes_128_ccm(),
            &Vec::from_hex(key).unwrap(),
            Some(&Vec::from_hex(nonce).unwrap()),
            &Vec::from_hex(aad).unwrap(),
            &Vec::from_hex(pt).unwrap(),
            &mut actual_tag,
        )
        .unwrap();

        assert_eq!(ct, hex::encode(out));
        assert_eq!(tag, hex::encode(actual_tag));

        let out = decrypt_aead(
            Cipher::aes_128_ccm(),
            &Vec::from_hex(key).unwrap(),
            Some(&Vec::from_hex(nonce).unwrap()),
            &Vec::from_hex(aad).unwrap(),
            &Vec::from_hex(ct).unwrap(),
            &Vec::from_hex(tag).unwrap(),
        )
        .unwrap();
        assert_eq!(pt, hex::encode(out));
    }
```

PR #1594, opened 2022-01-16
Fix Issue with AES CCM with tag size of 16 bytes

kedars:
Potential Fix for Issue #1593 

It appears that AES CCM requires set_tag_length() to be performed _before_ setting the key/IV lengths.
I am not sure if this PR would be the best way to address it, but I wanted to create a PoC that makes the test succeed.


sfackler:
I would recommend using the lower level CipherCtx type directly instead of Crypter.

kedars:
> I would recommend using the lower level CipherCtx type directly instead of Crypter.

Updated. 
Hrmph, seems libressl requires a separate sequence for this...

PR #1596, opened 2022-02-01
Enable setting RSA OAEP label in Decrypter

ipaljak-tbtl:
Hi, 

I've tried to use RSA OAEP encryption with a custom OAEP label, but couldn't figure out how to easily decrypt such ciphertext using the existing API. To the best of my knowledge, it looks like the method for setting a custom OAEP label should be present in the Decrypter object as well.

This patch should fix the issue. 

Apologies if I'm misusing the API in any way.

### DIFF
--- a/openssl/src/encrypt.rs
+++ b/openssl/src/encrypt.rs
@@ -363,6 +363,32 @@ impl<'a> Decrypter<'a> {
         }
     }
 
+    /// Sets the RSA OAEP label.
+    ///
+    /// This is only useful for RSA keys.
+    ///
+    /// This corresponds to [`EVP_PKEY_CTX_set0_rsa_oaep_label`].
+    ///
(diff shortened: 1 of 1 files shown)


sfackler:
Thanks! Crypter/Decrypter are on the way to being deprecated in favor of PkeyCtx - could you also expose this over there while you're at it? https://github.com/sfackler/rust-openssl/blob/master/openssl/src/pkey_ctx.rs

ipaljak-tbtl:
> Thanks! Crypter/Decrypter are on the way to being deprecated in favor of PkeyCtx - could you also expose this over there while you're at it? https://github.com/sfackler/rust-openssl/blob/master/openssl/src/pkey_ctx.rs

The implementation looks fine there (I've tried locally). I can cover this in tests if you want.

sfackler:
Ah - I forgot we already added it there!

PR #1598, opened 2022-02-04
Kletterstein/pkcs7

bkstein:
This branch extends the wrapper mainly for PKCS7 applications. I use it for:
- creation of a CSR with a challenge password attribute
- creation ofa signed PKCS7 containing an enveloped PKCS7 with an encrypted CSR
- extraction of certificates from a PKCS7
- extraction of content from a signed PKCS7

As this is my first contribution to rust-openssl, I'd appreciate a reviewer's opinion for this:
- I understand, the PKCS7 should be opaque, but I had to expose the inner structure of openssl's `PKCS7` to the wrapper. There is no other way to get access to the certificates in a PKCS7. openssl's api misses something like `PKCS7_get0_certificates()`. All I found was the implementation of `openssl pkcs -in p7.pem -print-certs`, which also does `certs = p7->d.sign->cert;`
- I did my best to ensure the `const_ptr_api!` correctness for openssl, but not for libressl.
- I did not quite get the spirit of `Type` vs `TypeRef`. E.g. I would have preferred to implement `Pkcs7Ref.certificates(&self)`, but as the ref type is opaque, I had to implement it for `Pkcs7`. Looks wrong to me...
- `Pkcs7Ref.get_content()` works for `Nid::PKCS7_SIGNED`, but I'm not sure for other types. Should we handle this smarter?


openssl/src/pkcs7.rs
```diff
@@ -146,6 +183,82 @@ impl Pkcs7 {
...
+    pub fn set_signed_attributes(
+        signer_info: &Pkcs7SignerInfoRef,
+        attributes: Stack<X509Attribute>,
+    ) -> Result<(), ErrorStack> {
+        unsafe {
+            cvt(ffi::PKCS7_set_signed_attributes(
+                signer_info.as_ptr(),
+                attributes.as_ptr(),
+            ))?;
+            mem::forget(attributes);
```
sfackler, line 201: `PKCS7_set_signed_attributes` does not take ownership of the stack - this function should take `&StackRef<X509Attribute>`.

openssl/src/pkcs7.rs
```diff
@@ -146,6 +183,82 @@ impl Pkcs7 {
...
+    ///
+    //  Note: `value` is immutable, but the OpenSSL function takes a (non-const) void pointer. Thus,
+    //  we have to cast to mutable in the unsafe block.
+    //  OpenSSL takes ownership of `value`.
+    //
+    #[corresponds(PKCS7_add_signed_attribute)]
+    pub fn add_signed_attribute(
+        signer_info: &Pkcs7SignerInfoRef,
+        nid: Nid,
+        atrtype: Asn1Type,
```
sfackler, line 219: What happens if `atrtype` disagrees with the contents of `Asn1Object`?

openssl/src/pkcs7.rs
```diff
@@ -146,6 +183,82 @@ impl Pkcs7 {
...
+        }
+    }
+
+    /// Get the certificates stored to a PKCS#7 structure
+    ///
+    // Unfortunately, there is no corresponding function in openssl. Thus, we have to enter
+    // openssl's internal PKCS7 struct. This is also, what openssl does, when e.g.
+    // `openssl pkcs7 -in pkcs7.pem -print-certs` is called.
+    //
+    pub fn certificates(&self) -> Result<Stack<X509>, ErrorStack> {
```
sfackler, line 240: I would prefer to not build nontrivial APIs that don't have an equivalent in OpenSSL itself. It might be preferable to expose the various components of the signer info and allow this logic to be written externally.

openssl/src/pkcs7.rs
```diff
@@ -146,6 +183,82 @@ impl Pkcs7 {
             .map(Pkcs7)
         }
     }
+
+    /// Set signed attributes in a PKCS#7 structure.
+    ///
+    /// `attributes` is a stack of the attributes to be added.
+    ///
+    #[corresponds(PKCS7_set_signed_attributes)]
+    pub fn set_signed_attributes(
```
sfackler, line 192: Shouldn't this be a method on `Pkcs7SignerInfoRef`?

openssl/src/pkcs7.rs
```diff
@@ -146,6 +183,82 @@ impl Pkcs7 {
...
+    ///
+    /// `nid` is the Nid of the attribute, `atrtype` is the attribute's type (an ASN.1 tag
+    /// value) and `value` is the ASN1 object to be added.
+    ///
+    //  Note: `value` is immutable, but the OpenSSL function takes a (non-const) void pointer. Thus,
+    //  we have to cast to mutable in the unsafe block.
+    //  OpenSSL takes ownership of `value`.
+    //
+    #[corresponds(PKCS7_add_signed_attribute)]
+    pub fn add_signed_attribute(
```
sfackler, line 216: Same here.

openssl/src/pkcs7.rs
```diff
@@ -279,17 +392,191 @@ impl Pkcs7Ref {
...
+    ///     signer_key_clone,
+    ///     MessageDigest::sha256()
+    /// );
+    ///  ```
+    ///
+    #[corresponds(PKCS7_add_signature)]
+    pub fn add_signature<PT>(
+        &self,
+        cert: &X509Ref,
+        pkey: PKey<PT>,
```
sfackler, line 475: This does not take ownership of the pkey.

openssl/src/pkcs7.rs
```diff
@@ -279,17 +392,191 @@ impl Pkcs7Ref {
...
+    /// `content` is the payload of type `content_type`.
+    ///
+    pub fn add_content(&self, content_type: Nid, content: &[u8]) -> Result<MemBio, ErrorStack> {
+        unsafe {
+            // Initialize content
+            cvt(ffi::PKCS7_content_new(self.as_ptr(), content_type.as_raw())).map(|_| ())?;
+            // Write content
+            let bcont_bio = ptr::null_mut();
+            let bio = cvt_p(ffi::PKCS7_dataInit(self.as_ptr(), bcont_bio))
+                .map(|bio| MemBio::from_ptr(bio))?;
```
sfackler, line 514: MemBio does not seem like the right type to use here.

openssl/src/pkcs7.rs
bkstein: But [OpenSSL increments reference count](https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/pkcs7/pk7_lib.c#L324) on `pkey`. How do we handle that? From OpenSSLs point of view, we have 2 owners after `ffi::PKCS7_add_signature` was called: the `PKey`, that was passed and the `Pkcs7`.

openssl/src/pkcs7.rs
bkstein: This case is a little delicate. As I mentioned in my first comment, OpenSSL api does not provide access to the certificates `STACK` of a `PKCS7` of type `PKCS7_SIGNED` or `PKCS7_SIGNEDANDENVELOPED`. But I don't think, that's missing on purpose. OpenSSL itself accesses this directly and SSCEP does that, too. Because they need it and there's no other way. I chose to return a (new) `Stack<X509>` (instead of a `&StackRef<X509>`), because I wanted to avoid a dangling reference. But probably, there is a better way?

openssl/src/pkcs7.rs
bkstein: Well, inside openssl this comes down to
```C
void ASN1_TYPE_set(ASN1_TYPE *a, int type, void *value)
{
    if (a->type != V_ASN1_BOOLEAN
            && a->type != V_ASN1_NULL
            && a->value.ptr != NULL) {
        ASN1_TYPE **tmp_a = &a;
        asn1_primitive_free((ASN1_VALUE **)tmp_a, NULL, 0);
    }
    a->type = type;
    if (type == V_ASN1_BOOLEAN)
        a->value.boolean = value ? 0xff : 0;
    else
        a->value.ptr = value;
}
```
(https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/asn1/a_type.c#L26)
In the last line, all values are accepted without control. But I could check for `V_ASN1_BOOLEAN` and `V_ASN1_NULL`.
Is that, what you intended? Or should we check against all `V_ASN1_<Type>`?

openssl/src/x509/mod.rs
```diff
@@ -693,6 +693,54 @@ impl Stackable for X509 {
...
+}
+
+impl Stackable for X509Attribute {
+    type StackType = ffi::stack_st_X509_ATTRIBUTE;
+}
+
+impl X509Attribute {
+    /// Creates an X509 attribute, which can be added to X509 certificates, signing requests and
+    /// PKCS7 messages.
+    pub fn from_string(nid: Nid, value: &str) -> Result<X509Attribute, ErrorStack> {
```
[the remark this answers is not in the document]
bkstein: I can do that. We need to implement `Asn1String::new()` for that, but sooner or later, we'll need that anyway, correct? However, usage will be more costly, as he or she will need to do something like
```rust
fn make_asn1string(value: &str) -> Result<Asn1String> {
    let value_len = value.len() as c_int;
    let value = CString::new(value).unwrap();
    let value = value.as_ptr() as *mut _;
    let asn1_string = Asn1String::new();
    asn1string.set(value, value_len).unwrap();
    
    asn1string
}

let transaction_id_asn1 = make_asn1string(&transaction_id);
let transaction_id_attr = X509Attribute::from_string(*TRANS_ID, &transaction_id_asn1).unwrap();
```

in opposite to just call

```rust
let transaction_id_attr = X509Attribute::from_string(*TRANS_ID, &transaction_id).unwrap();
```
What do you think?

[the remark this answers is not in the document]
bkstein: Another thought: I think we should hide the `ffi` stuff from a user of rust-openssl. But in order to create an `Asn1String`, we need to provide an `Asn1String.set()` function which either takes a `ffi::CString` or a `&str`. In the latter case, we have the same situation as now (just moved into another function).

openssl/src/pkcs7.rs
sfackler: This method should take a `&PKeyRef<PT>`.

openssl/src/pkcs7.rs
sfackler: I am concerned about what the worst case outcome is of someone passing the wrong type into this function. Is it just weird attributes? Is it UB?

openssl/src/x509/mod.rs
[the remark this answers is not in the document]
sfackler: We don't have the same situation - there is one Rust method that corresponds to `ASN1_STRING_set` and one that corresponds to `X509_ATTRIBUTE_create`.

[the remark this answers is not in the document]
bkstein: If understand you correctly, we will provide this:

```rust
let transaction_id_asn1 = Asn1String::new();
transaction_id_asn1.set(String::from("string value"))?;  // string will be consumed by srd::ffi::CString::new()
let transaction_id_attr = X509Attribute::from_string(*TRANS_ID, &transaction_id_asn1).unwrap();
```
That would be quite close to OpenSSL.

openssl/src/pkcs7.rs
bkstein: `PKCS7_dataInit()` returns a `*BIO`. So I think, `MemBio` is exactly what we need here. But you're the expert. Do you have another type in mind? Semantically, I would prefer a reference like `MemBioSlice`, but it's not possible to get a `MemBioSlice` from a pointer.

openssl/src/pkcs7.rs
```diff
@@ -279,17 +392,191 @@ impl Pkcs7Ref {
...
+            let content_length = content.len() as c_int;
+            let len = ffi::BIO_write(
```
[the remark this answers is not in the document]
bkstein: 😳 I'm very sorry, but what's a "short write"? Writing when `content.len() == 0`? I googled and asked around, but I didn't find that term.

openssl/src/pkcs7.rs
```diff
@@ -1,19 +1,46 @@
...
+    /// A `PKCS_SIGNER_INFO` signer info strucuture
+    pub struct Pkcs7SignerInfo;
+
+    /// Reference to `PKCS7_SIGNER_INFO`.
+    pub struct Pkcs7SignerInfoRef;
+}
+
+impl Pkcs7SignerInfo {
+    pub fn as_ptr(&self) -> *mut ffi::PKCS7_SIGNER_INFO {
+        &self.0 as *const _ as *mut _
```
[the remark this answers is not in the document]
bkstein: Fixed.

openssl/src/pkcs7.rs
bkstein: Fixed.

openssl/src/pkcs7.rs
bkstein: Yes, that's better. Fixed.

openssl/src/pkcs7.rs
```diff
@@ -279,17 +392,191 @@ impl Pkcs7Ref {
...
+    /// - Nid::PKCS7_SIGNED
+    /// - Nid::PKCS7_DATA
+    /// - Nid::PKCS7_SIGNEDANDENVELOPED
+    /// - Nid::PKCS7_ENVELOPED
+    /// - Nid::PKCS7_ENCRYPTED
+    /// - Nid::PKCS7_DIGEST
+    ///
+    #[corresponds(PKCS7_set_type)]
+    ///
+    pub fn set_type(&self, nid: Nid) -> Result<(), ErrorStack> {
```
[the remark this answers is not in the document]
bkstein: Fixed.

openssl/src/pkcs7.rs
```diff
@@ -279,17 +392,191 @@ impl Pkcs7Ref {
...
+    ///     MessageDigest::sha256()
+    /// );
+    ///  ```
+    ///
+    #[corresponds(PKCS7_add_signature)]
+    pub fn add_signature<PT>(
+        &self,
+        cert: &X509Ref,
+        pkey: PKey<PT>,
+        algorithm: MessageDigest,
```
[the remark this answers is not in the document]
bkstein: Yes, I fixed that. (Do you know OpenSSL by heart? This is not obvious from the api. You have to read the code...)

openssl/src/pkcs7.rs
bkstein: Got it. Fixed.

bkstein:
@sfackler How do we handle issues resolved by me? You are the maintainer, so I think you should make the last review. I currently comment them as "Fixed.", if I think it's fixed.

bkstein:
From my side, there is just one open question the return value (`SignerInfo`) of `Pkcs7Ref::add_signature()` (see above).

bkstein:
Reverting to draft. I think I can probably solve my last question by myelf.

openssl/src/pkcs7.rs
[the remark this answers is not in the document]
sfackler: A short write is when the call writes some but not all of the buffer you provided.

bkstein:
@sfackler There is something strange with the CI. windows-vcpkg and i686-unknown-linux-gnu-openssl-3.0.1-true are failing for reasons, I can't relate to my changes. I also observed spurious failures in the past, that vanished after simply re-triggering the CI with an empty commit , e.g.  PR#1612
- [Failed](https://github.com/sfackler/rust-openssl/actions/runs/2155210711)
- [Succeeded](https://github.com/sfackler/rust-openssl/actions/runs/2155570223)

bkstein:
This PR contains changes from PR#1612.

bkstein:
Converting to draft again, as I need more functions in my project and I don't want to many unmerged branches, that depend on each other.

bkstein:
Hi @sfackler , I try to avoid bothering you, but the failing check is misterious, as it is not reliably reproducable. If you have a minute, I would appreciate your eye on [the test implementations](https://github.com/bkstein/rust-openssl/blob/e2aadd836ac48c258acef7dba02682a1fa09ecc3/openssl/src/asn1.rs#L1030). In preparation of the actual check, I want to create Asn1Types. I use a `CString` to feed `ASN1_generate_v3()`.
Second, as this PR is not yet merged. Is this due to lack of time on your side or is there anything I can do? Best Regards, Bernd.

Skepfyr:
Hi, this PR is very large so reviewing and merging it will take a pretty large time commitment from one of us. I may be able to look at it at some point but I can't make any promises. Also, it's marked as a draft and has some git conflict markers in it, which makes me very hesitant to review. Is this ready (barring the merge conflicts)?

bkstein:
The only reason for the Draft marker is a race condition in OpenSSL, which is encountered by a test (`asn1_type_type_compatibility`). See #1730. I still want to analyze the problem. But note, that the code changes in rust-openssl are not affected. It's only some preparation code for the test, that crashes in multithreaded environments. This could also be a flaw in OpenSSL (actually, I assume it is).
In short: if I can see a realistic chance, that my PR might be reviewed and possibly be merged in the near future :), I offer to
analyze that race condition to make sure, we have no skeletons in the closet.
I will also resolve the merge conflicts (I do this anyway from time to time).
Would it help, if I separated the PR into multiple smaller PRs?

Skepfyr:
Splitting it up into smaller PRs would be helpful. Obviously investigating that flaky test would be useful, but don't feel obligated.

bkstein:
@skepfyr I will try to split the PR, but I'm not sure, if I will be able to detangle this without dropping something. My plan is to make 4 PRs:
1. a small one with X509Purpose stuff, see #1782.
2. one with the openssl-sys changes
3. one for X509 certificates and other basics
4. one for PKCS7 related changes
Please note, that the changes will depend on each other. 4 on 3 on 2 on 1.

Skepfyr:
That sounds reasonable. It's probably sensible to leave this PR up so that we can check it's all there once everything's merged.

bkstein:
I completed the split. I think it's best to review the PR's in order:
- #1788
- #1789 
- #1790 
- #1791
Each PR contains the predecessor(s), so merging #1791 would be sufficient, but more work at a single blow.

PR #1605, opened 2022-02-21
Add X509_PUBKEY related api

Leo1003:
The `X509_PUBKEY` is the structure represents the ASN.1 SubjectPublicKeyInfo which is used in certificates and certificate requests. 

Since I need to parse/serialize such data structure in DER format, I implemented the types and the related methods.


Leo1003:
It seems that some functions are implemented as macros in some versions, I'll try to figure out how to fix...

openssl/src/x509/mod.rs
```diff
@@ -451,6 +451,15 @@ impl X509Ref {
         }
     }
 
+    /// Returns the X509 public key of the certificate request.
+    #[corresponds(X509_get_X509_PUBKEY)]
+    pub fn x509_public_key(&self) -> Result<X509PublicKey, ErrorStack> {
```
sfackler, line 456: From the documentation, it looks like this should return a `&X509PublicKeyRef`.

sfackler, line 456: This needs to return a `&X509PublicKeyRef` according to the OpenSSL documentation.

openssl/src/x509/mod.rs
```diff
@@ -1251,6 +1260,15 @@ impl X509ReqRef {
         }
     }
 
+    /// Returns the X509 public key of the certificate request.
+    #[corresponds(X509_REQ_get_X509_PUBKEY)]
+    pub fn x509_public_key(&self) -> Result<X509PublicKey, ErrorStack> {
```
sfackler, line 1265: Same thing here.

openssl-sys/src/handwritten/x509.rs
```diff
@@ -291,6 +304,17 @@ extern "C" {
     pub fn X509_set_serialNumber(x: *mut X509, sn: *mut ASN1_INTEGER) -> c_int;
     pub fn X509_get_serialNumber(x: *mut X509) -> *mut ASN1_INTEGER;
 }
+cfg_if! {
+    if #[cfg(ossl110)] {
+        extern "C" {
+            pub fn X509_get_X509_PUBKEY(x: *const X509) -> *mut X509_PUBKEY;
+        }
+    } else {
+        pub unsafe fn X509_get_X509_PUBKEY(x: *const X509) -> *mut X509_PUBKEY {
```
sfackler, line 313: This macro redefinition will need to live in the top-level x509.rs file so it exists even when building with bindgen.

Leo1003:
Why do [here](https://github.com/sfackler/rust-openssl/blob/6be9b761de89a918ed9e55082d0b2dc8a17b3dbd/openssl-sys/src/handwritten/types.rs#L256) have `libressl280`? The latest libreSSL source codes still have fields definitions in its public header...
```rust
if #[cfg(any(ossl110, libressl280))] {
        pub enum X509 {}
    } else if #[cfg(libressl)] {
// ...
```

sfackler:
I think because 2.8.0 added the accessor functions required to support everything we currently used.

Leo1003:
> I think because 2.8.0 added the accessor functions required to support everything we currently used.

Should we remove this?

sfackler:
If you need to access fields that aren't exposed via functions, you could change that.

sfackler:
It looks like the LibreSSL 3.4.2 build is failing.

Leo1003:
I'm still trying to figure out why it failed. It looks like I need to prepare a test environment with LibreSSL installed.

PR #1608, opened 2022-03-01
WIP: Add is_ca method to X509 struct

Richterrettich:
This adds the ability to tell the difference between CA certificates and leaf certificates by exposing the `is_ca` method on `X509` structs. 
I've marked this as an WIP PR since I am unsure about the implementation as is and I would appreciate your feedback.

The current implementation introduces a new `BasicConstraints` type that has nothing to do with the existing `x509::extension::BasicConstraints` type. This seems wrong, but I am unsure how to combine those types in an idiomatic way.

I will add more documentation after an initial review.

### DIFF
--- a/openssl/src/x509/mod.rs
+++ b/openssl/src/x509/mod.rs
@@ -47,6 +47,20 @@ pub mod store;
 mod tests;
 
 foreign_type_and_impl_send_sync! {
+    type CType = ffi::BASIC_CONSTRAINTS;
+    fn drop = ffi::BASIC_CONSTRAINTS_free;
+
+    pub struct BasicConstraints;
+    pub struct BasicConstraintsRef;
+}
+
+impl BasicConstraintsRef {
+    pub fn is_ca(&self) -> bool {
+        unsafe { (*self.as_ptr()).ca > 0 }
+    }
+}
+
+foreign_type_and_impl_send_sync! {
     type CType = ffi::X509_STORE_CTX;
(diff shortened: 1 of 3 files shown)


Richterrettich:
I've pushed a new version which exposes the basic constraints directly.

openssl/src/x509/extension.rs
```diff
@@ -81,6 +81,18 @@ impl BasicConstraints {
...
+        self.pathlen
+    }
+
+    pub fn get_critical(&self) -> bool {
+        self.critical
+    }
+
+    pub fn get_ca(&self) -> bool {
+        self.ca
+    }
```
sfackler, line 95: I'm not necessarily opposed to adding these, but what's the use case for them?

openssl/src/x509/mod.rs
```diff
@@ -46,6 +47,17 @@ pub mod store;
 #[cfg(test)]
 mod tests;
 
+// this mod is only used to make the BasicConstraints glue type private.
```
sfackler, line 50: Why would we want to make the BasicConstraints type private?

openssl/src/x509/mod.rs
```diff
@@ -565,6 +577,53 @@ impl X509Ref {
         }
     }
 
+    pub fn basic_constraints(&self) -> Result<Option<extension::BasicConstraints>, ErrorStack> {
```
sfackler, line 580: Why is this reconstructing the type after retrieving it?

PR #1610, opened 2022-03-09
WIP: Support X509 CRLs

Alvenix:
I am not sure if adding support for CRLs is needed in this repo. However, I wanted to handle x509 CRLs in rust and found that openssl the best option for me. (Though I needed to add some binding)

This is work in progress because I don't know how to proceed further. Should I add some documentation? Testing? How? In addition, to be honest I am mimicking the existing code. I do not understand the underlying macros and I do not trust myself to write correct unsafe code without review.

### DIFF
--- a/openssl/src/crl.rs
+++ b/openssl/src/crl.rs
@@ -1,0 +1,177 @@
+use std::{marker::PhantomData, mem, os::raw::c_int, ptr};
+
+use libc::c_void;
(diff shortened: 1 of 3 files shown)


Alvenix:
I did not know that there is an existing (and old pull request) for X509 CRLs. I am closing the pull request.

PR #1612, opened 2022-03-16
Asn1Type completed

bkstein:
With this PR `Asn1Type` will represent the complete OpenSSL `ASN1_TYPE` including the `union value` part.

**Important:** this introduces a small incompatibility, as the former `Asn1Type` was implemented as `struct Asn1Type(c_int) {}` and now it is

```rust
pub struct ASN1_TYPE {
    pub type_: c_int,
    pub value: ASN1_TYPE_value,
}
```

See also the changes in ...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,5 @@ target/
 Cargo.lock
 .idea/
 *.iml
+*.tmp
 .vscode/
--- a/openssl/Cargo.toml
+++ b/openssl/Cargo.toml
@@ -25,6 +25,7 @@ bitflags = "1.0"
 cfg-if = "1.0"
 foreign-types = "0.3.1"
 libc = "0.2"
+num_enum = "0.5"
 once_cell = "1.5.2"
 
 openssl-macros = { path = "../openssl-macros" }
--- a/openssl/src/asn1.rs
+++ b/openssl/src/asn1.rs
@@ -895,6 +895,19 @@ mod tests {
         }
     }
 
+    // Check (deprecated) `pub const Asn1Type::...` et al.
+    #[test]
+    fn asn1_type_type_compatibility() {
+        let null = null_mut();
+        unsafe {
+            // Create an ASN.1 type object
+            let s = CString::new("UTF8String:Hällö Test").unwrap();
+            let at: Asn1Type = cvt_p(ffi::ASN1_generate_v3(s.as_ptr(), null))
+                .map(|p| Asn1Type::from_ptr(p)).unwrap();
+            assert_eq!(at.typ().unwrap(), Asn1Type::UTF8STRING);
+        }
+    }
+
     #[test]
     fn asn1_string_from_asn1_type() {
         let null = null_mut();
--- a/openssl/src/asn1.rs
+++ b/openssl/src/asn1.rs
@@ -29,11 +29,15 @@ use foreign_types::{ForeignType, ForeignTypeRef};
 use libc::{c_char, c_int, c_long, time_t};
 #[cfg(ossl102)]
 use std::cmp::Ordering;
+use std::convert::TryFrom;
+use std::error::Error;
 use std::ffi::CString;
 use std::fmt;
 use std::ptr;
 use std::slice;
 use std::str;
+
+use num_enum::TryFromPrimitive;
 
 use crate::bio::MemBio;
 use crate::bn::{BigNum, BigNumRef};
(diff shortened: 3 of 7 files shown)


bkstein:
Fixing the CI pipeline.

bkstein:
@sfackler The CI pipeline is red, as a test does not work for OpenSSL 3 anymore. I did not touch any code related to this test, so I see 2 ways out of this:
- Skip the test for OpenSSL 3
- Fix it
Unfortunately, I currently can't investigate that.

.gitignore
```diff
@@ -2,4 +2,5 @@ target/
 Cargo.lock
 .idea/
 *.iml
+*.tmp
```
sfackler, line 5: ?

openssl-errors/tests/test.rs
```diff
@@ -16,6 +16,7 @@ openssl_errors::openssl_errors! {
 }
 
 #[test]
+#[allow(clippy::single_char_pattern)]
```
sfackler, line 19: Probably better to just fix the thing it's complaining about.

openssl/src/asn1.rs
```diff
@@ -84,77 +109,129 @@ impl fmt::Display for Asn1GeneralizedTimeRef {
     }
 }
 
-/// The type of an ASN.1 value.
-#[derive(Debug, Copy, Clone, PartialEq, Eq)]
-pub struct Asn1Type(c_int);
+/// See ASN.1 specification for the meaning of Asn1TagValues (e.g. https://www.itu.int/en/ITU-T/asn1)
+#[derive(Clone, Copy, Debug, Eq, PartialEq, TryFromPrimitive)]
+#[repr(isize)]
+pub enum Asn1TagValue {
```
sfackler, line 115: We don't use enums for these kinds of types: https://github.com/sfackler/rust-openssl/blob/master/openssl/src/rsa.rs#L45-L64

openssl/src/asn1.rs
```diff
@@ -84,77 +109,129 @@ impl fmt::Display for Asn1GeneralizedTimeRef {
...
+        }
     }
+}
 
-    /// Returns the raw OpenSSL value represented by this type.
-    pub fn as_raw(&self) -> c_int {
-        self.0
-    }
+/// Must be implemented by all ASN.1 object structs
+pub trait FromAsn1Type<T: ForeignTypeRefExt> {
```
sfackler, line 232: ForeignTypeRefExt shouldn't be in the public API.

openssl/src/asn1.rs
```diff
@@ -480,6 +557,40 @@ impl fmt::Debug for Asn1StringRef {
...
+                Asn1TagValue::OctetString => Ok(from_asn1type_ptr(ty)),
+                Asn1TagValue::PrintableString => Ok(from_asn1type_ptr(ty)),
+                Asn1TagValue::T61String => Ok(from_asn1type_ptr(ty)),
+                Asn1TagValue::UniversalString => Ok(from_asn1type_ptr(ty)),
+                Asn1TagValue::UtcTime => Ok(from_asn1type_ptr(ty)),
+                Asn1TagValue::Utf8String => Ok(from_asn1type_ptr(ty)),
+                Asn1TagValue::VideotexString => Ok(from_asn1type_ptr(ty)),
+                Asn1TagValue::VisibleString => Ok(from_asn1type_ptr(ty)),
+                _ => Err(Asn1Error {
+                    message: String::from("Not a string type. Conversion not supported."),
```
sfackler, line 587: I'd rather this just return an option than make a new error type.

.gitignore
bkstein: I wanted to avoid checking in temporary files. I removed it.

openssl-errors/tests/test.rs
bkstein: When I remove the clippy-`allow`'s, I get errors in the CI, but not locally (running the same command `cargo clippy --all --all-targets`). I'll try to fix the errors blind, then.

openssl/src/asn1.rs
bkstein: Fixed.

openssl/src/asn1.rs
bkstein: Fixed.

openssl/src/asn1.rs
bkstein: Fixed by turning `Result` to `Option`. But we don't report error reasons this way...

bkstein:
@sfackler CI fails for the same reason as in PR #1598. Something seems to have changed in the CI environment.

openssl/Cargo.toml
```diff
@@ -25,6 +25,7 @@ bitflags = "1.0"
 cfg-if = "1.0"
 foreign-types = "0.3.1"
 libc = "0.2"
+num_enum = "0.5"
```
awidegreen, line 28: not used anywhere, is it?

bkstein: Right, not anymore. I had enums before, but Steven preferred constants.

PR #1616, opened 2022-03-29
Add X509*_cmp/_dup methods to openssl-sys

Skepfyr:
The cmp methods are a bit of a pain to actually use, in OpenSSL 1.1.1 they're documented to return a positive number when larger, 0 if equal and negative when smaller, and -2 if there's an error. However, I'm pretty sure -2 also means smaller in some situations so I've got no idea what you do if it actually ever returns -2.

I have not verified the `#[cfg(any(ossl110, libressl270))]` or the `#[const_ptr_if(ossl300)]`, because I haven't been able to work out how, however it does match what's already there for `X509_REVOKED_dup` and I see no reason for the other dups to diverge from that.

The compare methods appear to have existed since OpenSSL 1.0.0 and LibreSSL 2.0.0.


sfackler:
Thanks!

PR #1629, opened 2022-05-02
Fix up LibreSSL bits

vishwin:
Rebased on #1628

Expose stuff introduced since the 3.4 cycle and fix some before then. Re-add 3.4 to CI to verify, also since it is not yet considered EOL.

I'm inclined to re-ignore the three tests that continue to fail. May be LibreSSL bugs rather than anything on our end.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,12 +178,22 @@ jobs:
               bindgen: true
               library:
                 name: libressl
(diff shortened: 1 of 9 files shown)


sfackler:
If the tests don't pass, then we shouldn't enable them, yeah.

sfackler:
Thanks!

Issue #1641, opened 2022-05-24
X509_NAME based comparison


Skepfyr:
Being able to compare `X509_NAME` (and some related types) would be really useful, however `X509_NAME_cmp` has very unhelpful semantics: it returns a negative number for less than, 0 for equal, and a positive number for greater than or -2 for an error. -2 can also mean less than so I believe it's completely impossible to know if an error occurred. This is fixed in OpenSSL 3 (success is only -1,0,1), but I'd quite like it on earlier versions too.

I think that the error is only thrown if a memory allocation fails (although that's not documented) so we could just treat it as infallible, especially as unsafe code isn't allowed to depend on `Ord` and friends being sensible, but it could be very confusing.

One option is not implementing `Ord` and friends, and just providing a compare function with these caveats.

sfackler:
Yeah, I think it probably makes the most sense to just provide fallible comparison operations rather than implementing the traits unfortunately.

Skepfyr:
The openssl crate doesn't appear to know what version of OpenSSL it's running on? I ask because it would be nice to have a fallible method for OpenSSL 3 that returns an error in the -2 case, but for any other OpenSSL version it will just have to hope that an error hasn't occurred. If there's no way of telling then I'll just have to implement an infallible method with a caveat that it might occasionally return the wrong answer...

sfackler:
Yep, that can definitely be done - for example https://github.com/sfackler/rust-openssl/blob/14d93244f6326a9747d8899941140bf60bb2cb38/openssl/src/cipher.rs#L120.

I'm not sure how I feel about having a conditionally compiled `PartialOrd`/`Ord` implementation, though.

PR #1642, opened 2022-05-24
Implement Ord & Eq for X509 and X509Ref

Skepfyr:
X509_cmp is very helpfully infallible, I wanted to implement this for X509_NAME too but had to open #1641 instead. Hopefully I'll also implement these for Asn1Integers too at some point.


sfackler:
Thanks! It may be worth also adding `impl PartialEq<X509> for X509Ref` and vice versa? Same with `PartialOrd`.

Skepfyr:
I've added those cross implementations just delegating to the X509Ref implementation, being nice and explicit so nothing ends up being recursive.

sfackler:
Looks good!

PR #1644, opened 2022-05-26
Add add_alias method to X509Extension

tpambor:


sfackler:
It looks like the OpenSSL implementation of this function modifies global state without any synchronization, so the wrapper here would need to be marked unsafe.

Issue #1647, opened 2022-05-31
X509_V_ERR_INVALID_CA seems to be missing


BiagioFesta:
Maybe I am missing something, but after upgrading my dependencies to openssl 3 the constant `X509_V_ERR_INVALID_CA` seems to be missing in the sys-crate.

Indeed, the constant is not defined in the sys-crate when openssl >3 ([see here](https://github.com/sfackler/rust-openssl/blob/e76289f6addb9e5e11f640c590ae13a0b87dc557/openssl-sys/src/x509_vfy.rs#L34)).

However, the error constant is still in the [openssl library v3](https://www.openssl.org/docs/man3.0/man3/X509_STORE_CTX_get_current_cert.html). (and [in the latest code version](https://github.com/openssl/openssl/blob/4d346a188c27bdf78aa76590c641e1217732ca4b/include/openssl/x509_vfy.h.in#L301)).

Is this missing made on purpose?

sfackler:
It must have been missing in one of the earlier alpha releases of 3.0.0 or something. Should be easy to add back!

BiagioFesta:
Sorry for the trivial question.

Since my project depends on this (this error code is pretty common in X509 validation), I am wondering whether you already have a planning/date for the a new release in the short term?

sfackler:
Done!

PR #1648, opened 2022-05-31
openss-sys: define X509_V_ERR_INVALID_CA errno for openssl v3

BiagioFesta:
From https://github.com/sfackler/rust-openssl/issues/1647


sfackler:
Thanks!

Issue #1519, opened 2021-08-31
BoringSSL Support


benbrittain:
### BoringSSL

I'd like to investigate adding `BoringSSL` support to the `rust-openssl` crate. `BoringSSL` is a fork of `OpenSSL` that's designed to be harder to use incorrectly, but it also lacks the same backwards compatibility guarantees and has an unstable API/ABI. Some companies use `BoringSSL` exclusively (e.g. Google or Cloudflare) and since `rust-openssl` is the de facto way to interface with TLS/SSL in the rust ecosystem it would be nice to have support through this crate.

There exists some prior art in the rust ecosystem, notably [boring](https://github.com/cloudflare/boring) and [mundane](https://github.com/google/mundane).

### Constraints
* Minimal cfg additions to `rust-openssl`
  * `rust-openssl` already contains many `cfg`s for various openssl versions. Boringssl support should be at most minimally more invasive.
  * Should not make upgrading `rust-openssl` harder.
* No way for ABI drift to occur with boringssl
  * This is a scary problem because a subtle ABI violation could introduce a security hole.
  * The lack of stability in boringssl means that copy/paste headers isn't a viable solution.
* Less relevant to `rust-openssl`, but we should also avoid putting an undue burden on the `BoringSSL` maintainers. 

### Implementation
I'd like to solve this in a more integrated way by adding "*-sys" binding generation support into the build system of [boringssl](https://boringssl.googlesource.com/boringssl/) directly. This will guarantee the ABI doesn't drift from top-of-tree. The primary maintainers of `BoringSSL` have expressed willingness to use this approach. As much as possible should be generated by `bindgen` directly, as opposed to how `openssl-sys` handrolls a chunk of the bindings.

We'll likely want a boringssl-sys/boringssl-source setup similar to openssl-sys however so that we can provide a quick and easy static linking solution for people who desire it.

I'm in the middle of a quick prototype of this at the moment and it seems pretty tractable. Concerns and/or thoughts?

sfackler:
That seems reasonble to me at a high level. I'd like to keep bindgen out of the default dependency graph of openssl-sys (pulling it in as an optional feature is fine).

sfackler:
Now that OpenSSL 3.0.0 is out, I'm going to start working on a new major version release of the openssl crate (openssl-sys will be staying at 0.9). If there are API breaks that would help with the BoringSSL integration, let me know!

tomleavy:
[...]

PR #1649, opened 2022-06-02
Adds support for linking against boringssl

maurer:
The boringssl-sys crate is specific to the revision of boringssl you are
using, so you will need to generate it from a boringssl build tree and
point the openssl crate at it via cargo source replacement, or place the
build tree next to rust-openssl while building.

Co-authored-by: Matthew Maurer <mmaurer@google.com>
Co-authored-by: Andrew Scull <ascull@google.com>
Fixes: #1519


maurer:
We're currently using We're currently using these patches in [Android](https://cs.android.com/android/platform/superproject/+/master:external/rust/crates/openssl/) with a bit of extra glue to connect with our build system.

We're open to restructuring this patch (even significantly), but this is a working starting point.

openssl/src/ssl/mod.rs
```diff
@@ -57,6 +57,27 @@
...
+// use std::io::prelude::*;
+// use std::marker::PhantomData;
+// use std::mem::{self, ManuallyDrop};
+// use std::ops::{Deref, DerefMut};
+// use std::panic::resume_unwind;
+// use std::path::Path;
+// use std::ptr;
+// use std::slice;
+// use std::str;
+// use std::sync::{Arc, Mutex};
```
sfackler, line 80: ?

sfackler:
The path-based bssl-sys isn't going to work since we can't publish with that setup.

openssl-sys now optionally uses bindgen to generate its bindings - could we avoid the separate sys crate by using that for boringssl? It seems fine to require bindgen for boringssl support.

openssl/src/hkdf.rs
```diff
@@ -0,0 +1,89 @@
+use crate::cvt;
```
sfackler, line 1: This is a drive-by feature addition? Not opposed, just want to make sure I understand.

openssl/src/ssl/mod.rs
maurer: This was removed in the version we've got in Android, but didn't make it into the merge to produce this patch, I'll remove it.

openssl/Cargo.toml
```diff
@@ -28,7 +30,8 @@ libc = "0.2"
 once_cell = "1.5.2"
 
 openssl-macros = { version = "0.1.0", path = "../openssl-macros" }
-ffi = { package = "openssl-sys", version = "0.9.73", path = "../openssl-sys" }
+ffi = { package = "openssl-sys", version = "0.9.73", path = "../openssl-sys", optional = true }
+bssl-ffi = { package = "bssl-sys", version = "0.1.0", path = "../../boringssl/build/rust", optional=true }
```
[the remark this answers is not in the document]
maurer: That works for me, I can make that adjustment.

openssl/src/hkdf.rs
maurer: OpenSSL exposes HKDF via the `EVP_PKEY` API. BoringSSL doesn't, so this was to reach parity there. It's easy to split this back into a separate patch (it was in our repo), I had it squashed because it was part of bringing BoringSSL up to parity.

maurer:
>The path-based bssl-sys isn't going to work since we can't publish with that setup.
>
> openssl-sys now optionally uses bindgen to generate its bindings - could we avoid the separate sys crate by using that for boringssl? It seems fine to require bindgen for boringssl support.

One of the parts of the negotiation with the boringssl folks to add the Rust bindings there was to have the generation [inside boringssl](https://github.com/google/boringssl/blob/ea46caf2656ec1207bd895fd4da9113275366916/rust/CMakeLists.txt) rather than as a standalone or external package so that the bindings would always be updated in sync. Effectively, upstream wants to deliver the bindings with their package, not have them be installed alongside.

That said, there's nothing stopping us from having a `bssl-sys` crate with a specific boringssl and its corresponding generated sources. I'd want to chat with them before doing this (boringssl does not have "releases", so there's no obvious cadence to update such a crate on). What do you think?

openssl/src/hkdf.rs
davidben: (Drive-by comment, I haven't looked at any of this PR yet.)

FWIW we'll probably add OpenSSL's HKDF API in BoringSSL soon as a compatibility shim for legacy code. That might make this simpler.

It's not a good API... jamming HKDF into the same API as ECDH is absurd and forced other API mistakes (too many copies and some completely ad hoc notion of "modes" that is unsupported by the HKDF specification), but apparently that's what other projects expect so okay whatever. :-)

davidben:
> That said, there's nothing stopping us from having a bssl-sys crate with a specific boringssl and its corresponding generated sources. I'd want to chat with them before doing this (boringssl does not have "releases", so there's no obvious cadence to update such a crate on).

We can discuss this more during the week, but we tried this with mundane. I think it was a mistake and we should undo that. It's possible there are other ways to do it that will work less badly, but my sense so far is cargo is unsuitable for shipping a non-Rust library like this.

sfackler:
I don't know how boringssl is distributed, but one potential possibility could be to actually install the generated Rust sources in /etc or wherever and have openssl-sys pull those in directly. That way we could be confident that they're matching with the version we're linking against but not be forced to bundle the BoringSSL C sources in bssl-sys.

openssl/src/hkdf.rs
sfackler: I'm happy to expose any part of OpenSSL's public API as long as the bindings are sound :)

davidben: OpenSSL often designs bad APIs. You really should expose, say, a good HKDF API and just happen to implement it with OpenSSL.

Ideally those two are one and the same, but not here in the slightest. To that end...

These are reasonable APIs
https://docs.rs/openssl-hkdf/0.2.0/openssl_hkdf/hkdf/fn.hkdf_expand.html
https://docs.rs/openssl-hkdf/0.2.0/openssl_hkdf/hkdf/fn.hkdf_extract.html

These are not and should not be exported.
https://docs.rs/openssl-hkdf/0.2.0/openssl_hkdf/hkdf/struct.HkdfDeriver.html
This is simply not how HKDF works at all. Please consult RFC 5869. Indeed even OpenSSL are moving away from stuffing KDFs into EVP_PKEY_CTX (see EVP_KDF), though their new one still makes similar mistakes.

sfackler: I am absolutely aware that OpenSSL's APIs contain significant bits of jank. However, generally speaking I see the role of this library to as directly as possible expose those APIs, jank and all.

davidben: I mean, that makes sense for something complex like TLS where you need to expose it as-is to get full fidelity. For something like this, OpenSSL's APIs are essentially a really messy builder pattern. You don't get any better fidelity by exposing it, just an incorrect interpretation of OpenSSL's design patterns and unsafe APIs. (OpenSSL's API here mixes together two completely different kinds of secrets in this notion of "mode". You actually break domain separation---meaning protocols built this way aren't secure---by doing this.)

maurer:
Notable pieces that should probably be considered beyond simple review:

1. I changed the `-1` mac_iter value to use `self.mac_iter`, as `boringssl` rejects it. I believe this to be correct and tests pass, but I wanted to call attention to it.
2. I am currently adding a boringssl-master test in addition to a specific revision of boringssl. The intention is that you only make the specific version a blocking test, and leave master as a warning/informative
3. I've removed the boringssl HKDF patch, we can land that in a follow-up CL
4. The intended way to use boringssl is displayed in `ci.yml` - basically you patch a in a replacement for the placeholder I've put on crates.io. If this approach is acceptable, I can add instructions to `README.md`.

maurer:
@sfackler Gentle ping

.github/workflows/ci.yml
```diff
@@ -233,14 +241,24 @@ jobs:
             key: openssl-${{ matrix.target }}-${{ matrix.library.name }}-${{ matrix.library.version }}-2
           if: matrix.library.version != 'vendored'
           id: openssl-cache
+        - run: |
+            echo "RUST_TEST_THREADS=1" >> $GITHUB_ENV
```
sfackler, line 245: Leftover from debugging?

.github/workflows/ci.yml
```diff
@@ -233,14 +241,24 @@ jobs:
...
+            echo BINDGEN_EXTRA_CLANG_ARGS="--sysroot /usr/arm-linux-gnueabihf" >> $GITHUB_ENV
```
sfackler, line 246: Is this necessary if bindgen + boringssl isn't being tested?

openssl/src/bn.rs
```diff
@@ -839,7 +841,7 @@ impl BigNumRef {
     /// # use openssl::bn::BigNum;
     /// let s = -BigNum::from_u32(0x99ff).unwrap();
     ///
-    /// assert_eq!(&**s.to_hex_str().unwrap(), "-99FF");
+    /// assert_eq!(s.to_hex_str().unwrap().to_owned().to_uppercase(), "-99FF");
```
sfackler, line 844: the to_owned shouldn't be necessary here

openssl/src/ssl/test/mod.rs
```diff
@@ -1314,7 +1320,9 @@ fn psk_ciphers() {
     let mut server = Server::builder();
     server.ctx().set_cipher_list(CIPHER).unwrap();
     server.ctx().set_psk_server_callback(|_, identity, psk| {
+        eprintln!("hello???!");
```
sfackler, line 1323: Some stray debug printlns here

openssl/src/x509/mod.rs
```diff
@@ -1452,8 +1457,13 @@ impl GeneralNameRef {
                 return None;
             }
 
-            let ptr = ASN1_STRING_get0_data((*self.as_ptr()).d as *mut _);
-            let len = ffi::ASN1_STRING_length((*self.as_ptr()).d as *mut _);
+            #[cfg(boringssl)]
+            let d: *const ffi::ASN1_STRING = std::mem::transmute((*self.as_ptr()).d);
```
sfackler, line 1461: Why is this transmuting into a raw pointer type?

sfackler:
Using patch to hook up to the installed bindings seems a bit janky. Couldn't we have openssl-sys pull them in automatically with the build script finding the right directory and an `include!(...)`?

openssl/src/x509/mod.rs
maurer: It was transmuting because there's a union there, not a pointer. In your handwritten bindings you've pretended it is a void*, then cast it to an ASN1_STRING, and blocklist it from the bindgen.

The transmute still wasn't the correct call, as there's a union member that gives you raw pointer access. I've switched to use that instead in the boringssl side of the ifdef, and you could consider using it if you want to stop blocklisting GENERAL_NAME in your bindgen openssl path.

maurer:
> Using patch to hook up to the installed bindings seems a bit janky. Couldn't we have openssl-sys pull them in automatically with the build script finding the right directory and an `include!(...)`?

I didn't do it that way for a few reasons:
1. The rust crate doesn't get installed anywhere, it just lives in the build output directory. This means that users will never successfully autodetect through `build.rs`, they will always need to tell it where to find it somehow, likely an environment variable.
2. The crate has two files right now, and might have more in the future (to deal with some functions that are header-only in boringssl but are not in openssl). This means that `include!` would not be sufficient by itself, we'd also need a local mimic of the directory / module structure, and this could easily get out of sync.
3. If we built with environment variable based detection, then when we (Android) consume this crate, we'd have to carry a patch patching that back out, because users can't set environment variable in their builds, and we'd have to do a bunch of hackery. I know this isn't a constraint on your development, but I included it in the interests of completeness.

All that said, if you'd strongly prefer I do this with `include!` and `env!` I can do it.

sfackler:
1. Could it be installed somewhere?
2. I think `include!` can handle submodules properly, and if it can't `#[path = "..."] mod foo;` definitely can.
3. If it can be installed somewhere that's specified in e.g. a pkg-config variable we wouldn't need the env var.

I guess more broadly I'm wondering about the intended consumers of the BoringSSL support. If it's just Google internal stuff then I guess I don't care how janky the integration is :). If it's something that people in the broader ecosystem could pick up then I'm going to be more picky about the setup. I know BoringSSL has historically been a "if you aren't in Google, don't use this" kind of situation but it seems like that is empirically not really limiting external users too much.

davidben:
BoringSSL is not something we support installing system-wide in that way. (Not unlike typical Rust libraries!) We don't promise a stable ABI or API and instead expect that consumers match header and link version, build against known revisions, and keep those revisions regularly updated. The so called "live at head" model.

This means the overall project (the root of the dependency tree) using it will have _some_ pre-existing story for building BoringSSL and _some_ pre-existing opinion as to which instance they're building. This is especially important when you don't want to link two copies of a dependency together. (Either because the language, like C/C++, does not support it, or because the dependency's types leak through another library, such as rust-openssl, or because you don't want to worry about security bugs in both copies.)

Alas, tools for cross-language dependency management, especially if one is C/C++, are so fragmented that there isn't much for a library build to do but to provide _some_ way for the user to point at a dependency instance and let them pick. Auto-detection will just risk picking up the wrong version, possibly resulting in multiple linking and thus memory errors in C/C++.

> I know BoringSSL has historically been a "if you aren't in Google, don't use this" kind of situation but it seems like that is empirically not really limiting external users too much.

Please see the README for what it actually says. There are some non-Google users of BoringSSL, but they're one that can work with the deployment model. That they can work with it is not a reason for rust-openssl to disregard it. :-)

davidben:
I'm not familiar enough with Cargo to know how best to translate that to Cargo (I'll leave that to you all), but that's the deployment model to aim for.

maurer:
I'm not sure I follow why having `OPENSSL_RUST_DIR` as an environment variable is considered cleaner than a `.cargo/config.toml` setting, but I don't feel strongly about this. I'll adjust the patches to use an environment variable instead.

sfackler:
Oh, if there's a build script involved in bssl-sys, then we may not be able to take that approach then, unfortunately.

We should be able to go with the `.cargo/config` approach. I'll take another look over the PR tonight, but we can probably land it.

openssl-sys/Cargo.toml
```diff
@@ -15,9 +15,11 @@ build = "build/main.rs"
 
 [features]
 vendored = ['openssl-src']
+boringssl = ['bssl-sys']
```
sfackler, line 18: Can we rename this feature to `unstable_boringssl` or something like that for now? I just want to have some time for external consumers to pick this up before committing to the bssl-sys patching approach.

sfackler:
I think this looks good other than renaming the Cargo feature temporarily! I will merge and cut a release once that's done.

maurer:
I renamed the feature, and added two extra cleanup commits to keep the build happy:
1. Fix up compiler lints (necessary)
2. Fix up clippy lints (bonus, just did it while I was at it)

sfackler:
Thanks for sticking with it! I will cut a release tonight.

openssl-sys/src/lib.rs
```diff
@@ -13,162 +13,174 @@
...
+        use std::ptr;
+        use std::sync::Once;
+
+        // explicitly initialize to work around https://github.com/openssl/openssl/issues/3505
+        static INIT: Once = Once::new();
+
+        #[cfg(not(ossl111b))]
+        let init_options = OPENSSL_INIT_LOAD_SSL_STRINGS;
+        #[cfg(ossl111b)]
+        let init_options = OPENSSL_INIT_LOAD_SSL_STRINGS | OPENSSL_INIT_NO_ATEXIT;
```
eloff, line 107: This fixes some crashes I was experiencing in diesel using libpq, which uses openssl. The atexit handler was racing the r2d2 connection pool worker threads when main() exits, and that caused a seg fault.

See: https://github.com/sfackler/r2d2/issues/137

thomasmost: 👍🏼  Reproduced using MySql also

PR #1661, opened 2022-07-09
Add SSL_get_psk_identity

newAM:
This adds the functions described here: https://www.openssl.org/docs/man1.0.2/man3/SSL_get_psk_identity.html

This is useful for extracting the PSK identity from the callback in `SslContextBuilder::set_psk_client_callback`.


sfackler:
Thanks!

PR #1665, opened 2022-07-21
Add feature `vendored-engine` for engine support in openssl-src

neuronull:
NOTE this change depends on the following PR being merged into `openssl-src` : https://github.com/alexcrichton/openssl-src-rs/pull/150 , and thus CI will fail until that is made available. Posting this PR in anticipation of that for visibility.

One of our project dependency crates added a hard dependency on the openssl Engine module.

This created a problem when cross compiling with musl due to the disabling of the Engine module by openssl-src for the musl target.

By adding the option to enable Engine support in openssl-src , and adding this feature flag to openssl-sys to make use of that option, together this would unblock us .

### DIFF
--- a/openssl-sys/Cargo.toml
+++ b/openssl-sys/Cargo.toml
@@ -15,6 +15,7 @@ build = "build/main.rs"
 
 [features]
 vendored = ['openssl-src']
+vendored-engine = ['openssl-src']
 
 [dependencies]
 libc = "0.2"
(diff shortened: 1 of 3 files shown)


sfackler:
It seems like it would be better for this to be modeled as a Cargo feature in openssl-src like all of the other configuration.

jszwedko:
@sfackler Agreed this should be a feature flag on `openssl-src`. Once that is added, we will need a new in `openssl-sys` to enable the flag on `openssl-src` though, right? Would you be open to that?

I.e. something like:

```diff
[features]
vendored = ['openssl-src']
+ vendored-engine = ['openssl-src/engine']
```

neuronull:
I submitted a new PR to `openssl-src` for the feature flag (https://github.com/alexcrichton/openssl-src-rs/pull/151)

New PR because I needed to change to the 111 release.

neuronull:
This was covered in https://github.com/alexcrichton/openssl-src-rs/pull/151.

Closing

PR #1666, opened 2022-07-22
Added missing NIDs for SHA3 hashes

bkstein:
The `Nid`s for SHA3 hash algorithm were missing. I added them, as SHA3 is already supported in `MessageDigest`. Also added `Nid` checks in the tests to make sure, that the NIDs set by OpenSSL match the constants defined by rust-openssl.

### DIFF
--- a/openssl/src/x509/mod.rs
+++ b/openssl/src/x509/mod.rs
@@ -228,7 +228,7 @@ impl X509Builder {
     /// the X.509 standard should pass `2` to this method.
     #[corresponds(X509_set_version)]
     pub fn set_version(&mut self, version: i32) -> Result<(), ErrorStack> {
-        unsafe { cvt(ffi::X509_set_version(self.0.as_ptr(), version.into())).map(|_| ()) }
+        unsafe { cvt(ffi::X509_set_version(self.0.as_ptr(), version as c_long)).map(|_| ()) }
     }
 
     /// Sets the serial number of the certificate.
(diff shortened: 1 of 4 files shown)


openssl/src/x509/mod.rs
```diff
@@ -228,7 +228,7 @@ impl X509Builder {
     /// the X.509 standard should pass `2` to this method.
     #[corresponds(X509_set_version)]
     pub fn set_version(&mut self, version: i32) -> Result<(), ErrorStack> {
-        unsafe { cvt(ffi::X509_set_version(self.0.as_ptr(), version.into())).map(|_| ()) }
+        unsafe { cvt(ffi::X509_set_version(self.0.as_ptr(), version as c_long)).map(|_| ()) }
```
sfackler, line 231: Why?

bkstein: Clippy emits a warning, if `.into()` and the target type is the same as the source type:
```
warning: useless conversion to the same type: `i32`
   --> openssl\src\x509\mod.rs:231:61
    |
231 |         unsafe { cvt(ffi::X509_set_version(self.0.as_ptr(), version.into())).map(|_| ()) }
    |                                                             ^^^^^^^^^^^^^^ help: consider removing `.into()`: `version`
    |
    = note: `#[warn(clippy::useless_conversion)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
```

PR #1669, opened 2022-08-05
Fix generic bound to allow verification with public keys

wiktor-k:
Previously the generic bound on `MdCtxRef::digest_verify_init` required initialization with a key possessing the private material. This bound is too strict as it disallowed the more common verification of signatures with only public key bits.
    
This patch adjusts the bound to require only public bits (`HasPublic`).
    
Additionally two tests have been added: the first one making sure that the previous behavior of using private keys still works. The second test extracts public key components and uses them for signature verification.


sfackler:
Oops, thanks!

PR #1670, opened 2022-08-09
Add SslOptions::PRIORITIZE_CHACHA

james58899:
This option can improve performance on clients without AES hardware acceleration.

The `SSL_OP_PRIORITIZE_CHACHA` added in OpenSSL 1.1.1, see [SSL_CTX_set_options](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_options.html#HISTORY).


sfackler:
Thanks!

PR #1672, opened 2022-08-10
Added `.to-text()` for `X509Req`.

bkstein:
Added `.to_text()` for `X509Req`, which provides human readable representation of CSR's as in
```
openssl req -in csr.pem -text -noout
```


sfackler:
Thanks!

PR #1673, opened 2022-08-11
Add Clone implementation for SslContextBuilder

Gor027:
I needed to clone `SslContextBuilder` before calling `build()` as it will consume the builder object. Need to keep the prepared state of the builder to easily make some changes and build another `SslContext` object.


sfackler:
This would allow unsound concurrent modifications to the underlying `SSL_CTX`.

Gor027:
Yes, I should have paid proper attention to what the `Clone` implementation actually does for `SslContext`. I will close this PR, then.

PR #1676, opened 2022-08-23
Add `try_cmp` function to `X509NameRef`s

### LINKED ISSUE #1641: X509_NAME based comparison

Skepfyr:
Fixes #1641.

This adds a `X509NameRef::try_cmp(&self, other: &X509NameRef) -> Result<Ordering, ErrorStack>` function that is sensible on OpenSSL 3.0.0 and up but can occasionally return a spurious less than in older versions. I believe that spurious failure can only occur if a memory allocation fails so is vanishingly unlikely (or at least is unlikely to occur on a system that isn't about to fall over anyway).

### DIFF
--- a/openssl/src/x509/mod.rs
+++ b/openssl/src/x509/mod.rs
@@ -10,7 +10,7 @@
 use cfg_if::cfg_if;
 use foreign_types::{ForeignType, ForeignTypeRef};
 use libc::{c_int, c_long};
-use std::cmp;
+use std::cmp::{self, Ordering};
 use std::error::Error;
 use std::ffi::{CStr, CString};
 use std::fmt;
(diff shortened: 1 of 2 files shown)


openssl/src/x509/mod.rs
```diff
@@ -1028,6 +1028,21 @@ impl X509NameRef {
...
+    /// Compare two names, like [`Ord`] but it may fail.
+    ///
+    /// With OpenSSL versions from 3.0.0 this may return an error if the underlying `X509_NAME_cmp`
+    /// call fails.
+    /// For OpenSSL versions before 3.0.0 it will never return an error, but due to a bug it may
+    /// spuriously return `Ordering::Less` if the `X509_NAME_cmp` call fails.
+    #[corresponds(X509_NAME_cmp)]
+    pub fn try_cmp(&self, other: &X509NameRef) -> Result<Ordering, ErrorStack> {
+        let cmp = unsafe { ffi::X509_NAME_cmp(self.as_ptr(), other.as_ptr()) };
+        if cfg!(ossl300) && cmp == -2 {
```
sfackler, line 1040: It looks like X509_NAME_cmp is documented to return -2 on failure as well in 1.1.1 at least: https://www.openssl.org/docs/man1.1.1/man3/X509_NAME_cmp.html

Skepfyr: It does return -2 on failure, however [this line](https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/x509/x509_cmp.c#L183) means that it also returns -2 if the first argument is two bytes shorter than the second. That issue is what I was trying to get at in the doc comment, if that's not clear do you have any suggestions to improve it?

sfackler: Wow. I guess that works then...

sfackler:
Thanks!

PR #1678, opened 2022-08-28
Added support for `EVP_MD_CTX_reset`

mjb3279:
Signed-off-by: Michael Boquard <michael.j.boquard@gmail.com>

### DIFF
--- a/openssl/src/md_ctx.rs
+++ b/openssl/src/md_ctx.rs
@@ -380,6 +380,16 @@ impl MdCtxRef {
     pub fn size(&self) -> usize {
         unsafe { ffi::EVP_MD_CTX_size(self.as_ptr()) as usize }
     }
+
+    /// Resets the underlying EVP_MD_CTX instance
+    #[corresponds(EVP_MD_CTX_init)]
+    #[inline]
+    pub fn reset(&self) -> Result<(), ErrorStack> {
+        unsafe {
+            let _ = cvt(ffi::EVP_MD_CTX_reset(self.as_ptr()))?;
+            Ok(())
+        }
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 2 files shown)


openssl/src/md_ctx.rs
```diff
@@ -380,6 +380,17 @@ impl MdCtxRef {
     pub fn size(&self) -> usize {
         unsafe { ffi::EVP_MD_CTX_size(self.as_ptr()) as usize }
     }
+
+    /// Resets the underlying EVP_MD_CTX instance
+    #[corresponds(EVP_MD_CTX_init)]
+    #[cfg(ossl111)]
+    #[inline]
+    pub fn reset(&self) -> Result<(), ErrorStack> {
```
sfackler, line 388: `&mut self`.

openssl/src/md_ctx.rs
```diff
@@ -380,6 +380,17 @@ impl MdCtxRef {
     pub fn size(&self) -> usize {
         unsafe { ffi::EVP_MD_CTX_size(self.as_ptr()) as usize }
     }
+
+    /// Resets the underlying EVP_MD_CTX instance
+    #[corresponds(EVP_MD_CTX_init)]
```
sfackler, line 385: `EVP_MD_CTX_reset`.

sfackler:
Thanks!

davidben:
What's the use case here? I've never seen this function used except in sloppy ports of OpenSSL 1.0.x code to OpenSSL 1.1.x, where it was unnecessary.

If you're about to destroy the EVP_MD_CTX, this is unnecessary because EVP_MD_CTX_free will do all the cleanup.
If you're about to reuse the EVP_MD_CTX, this is unnecessary because you can call EVP_DigestInit_ex without resetting. (Arguably reuse is itself unnecessary because you can just create a new EVP_MD_CTX... a lot of this is because OpenSSL's own API didn't evolve correctly from 1.0.x to 1.1.x.)

davidben:
I'll should add that, in the second case, not calling reset is more efficient anyway because EVP_MD_CTX can internally reuse its allocation if you aren't changing hash types.

PR #1679, opened 2022-08-28
Added Clone implementation for `MdCtx`

mjb3279:
Signed-off-by: Michael Boquard <michael.j.boquard@gmail.com>

### DIFF
--- a/openssl-sys/src/handwritten/evp.rs
+++ b/openssl-sys/src/handwritten/evp.rs
@@ -260,6 +260,7 @@ const_ptr_api! {
 extern "C" {
     pub fn EVP_CIPHER_CTX_new() -> *mut EVP_CIPHER_CTX;
     pub fn EVP_CIPHER_CTX_free(ctx: *mut EVP_CIPHER_CTX);
+    pub fn EVP_MD_CTX_copy(dst: *mut EVP_MD_CTX, src: *const EVP_MD_CTX) -> c_int;
     pub fn EVP_MD_CTX_copy_ex(dst: *mut EVP_MD_CTX, src: *const EVP_MD_CTX) -> c_int;
     pub fn EVP_CIPHER_CTX_set_key_length(ctx: *mut EVP_CIPHER_CTX, keylen: c_int) -> c_int;
     pub fn EVP_CIPHER_CTX_set_padding(ctx: *mut EVP_CIPHER_CTX, padding: c_int) -> c_int;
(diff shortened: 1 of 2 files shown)


openssl-sys/src/handwritten/evp.rs
```diff
@@ -260,6 +260,7 @@ const_ptr_api! {
 extern "C" {
     pub fn EVP_CIPHER_CTX_new() -> *mut EVP_CIPHER_CTX;
     pub fn EVP_CIPHER_CTX_free(ctx: *mut EVP_CIPHER_CTX);
+    pub fn EVP_MD_CTX_copy(dst: *mut EVP_MD_CTX, src: *const EVP_MD_CTX) -> c_int;
```
sfackler, line 263: According to the docs, it seems like EVP_MD_CTX_copy_ex should be used instead of EVP_MD_CTX_copy.

sfackler:
Given the falibility, it might be preferrable to just have a `pub fn copy(&mut self, other: &MdCtxRef) -> Result<(), ErrorStack>`

openssl-sys/src/handwritten/evp.rs
mjb3279: Fair enough, missed that comment, will update

mjb3279:
> Given the falibility, it might be preferrable to just have a `pub fn copy(&mut self, other: &MdCtxRef) -> Result<(), ErrorStack>`

I can understand that sentiment.  I'll remove the `impl Clone` and just provide `copy` function.

openssl/src/md_ctx.rs
```diff
@@ -118,6 +118,17 @@ impl MdCtx {
 }
 
 impl MdCtxRef {
+    /// Performs a copy of this instance of `MdCtxRef` to another `MdCtxRef`
+    #[corresponds(EVP_MD_CTX_copy_ex)]
+    #[inline]
+    pub fn copy(&self, other: &mut MdCtxRef) -> Result<(), ErrorStack> {
```
sfackler, line 124: Nit, I'd probably reverse the arguments to both match the OpenSSL function signature and line up with the standard library's `clone_from`.

LGTM otherwise.

haroldbruintjes:
Since we're in need of this functionality as well, any chance in the near future of merging this, or can I otherwise help make that happen?

alex:
Looks like there's an outstanding piece of review feedback. With that resolved we could merge.

Issue #1693, opened 2022-09-22
Confusing list of supported Openssl versions in documentation


puzzlewolf:
The docs of the openssl crate state
> This crate provides a safe interface to the popular OpenSSL cryptography library. OpenSSL versions 1.0.1 through 1.1.1 and LibreSSL versions 2.5 through 3.4.1 are supported.

This reads as if OpenSSL > 1.1.1 is not supported, but OpenSSL 3 does work and the code agrees :)

https://github.com/sfackler/rust-openssl/blob/f8c321fbbc530f0fb607b890264b16a5b631dffe/openssl-sys/build/main.rs#L153

It would be great if the docs could be adjusted, or am I missing something?

sfackler:
Yeah, that comment just needs to be updated.

puzzlewolf:
I can make a PR if you like. Would this be correct?
> OpenSSL versions 1.0.1 through 1.1.1 or >= 3.0.0 and LibreSSL versions 2.5 through 3.4.1 are supported.

sfackler:
It can just say `OpenSSL versions 1.0.1 through 3.x.x`

sfackler:
Or really, that comment should just be removed since it is inevitably going to get out of date again.

puzzlewolf:
Hm, as a user I like being able to check if the library I have on my system will work with a `-sys` crate. But if you don't expect to remember to update that line, then removing it is probably for the best.

sfackler:
> Hm, as a user I like being able to check if the library I have on my system will work with a -sys crate.

Source diving for a random comment seems like a bad way of determining that though.

puzzlewolf:
Yeah, that's why I looked at the docs generated from 
https://github.com/sfackler/rust-openssl/blob/f8c321fbbc530f0fb607b890264b16a5b631dffe/openssl/src/lib.rs#L3-L4
which don't talk about OpenSSL 3. That's what I want to fix, sorry if I was unclear :)

sfackler:
Ah yeah, that should definitely be updated!

PR #1694, opened 2022-09-22
Fix supported OpenSSL versions

puzzlewolf:
`openssl-sys` does support OpenSSL version 3. Update the documentation to reflect that.

Closes https://github.com/sfackler/rust-openssl/issues/1693


sfackler:
Thanks!

PR #1695, opened 2022-09-26
Recognize LibreSSL 3.6

niklasha:
This is a minimal patch to make openssl-sys work on OpenBSD-current (soon to be 7.2).  LibreSSL has already been versioned there as 3.6.0, although the portable version is not yet released (portable LibreSSL releases are synchronized with OpenBSD releases).  When the portable release is done, I guess there will be more changes needed for finding the correct library on e.g. Linux.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -201,6 +201,11 @@ jobs:
               library:
                 name: libressl
                 version: 3.5.2
+            - target: x86_64-unknown-linux-gnu
+              bindgen: false
+              library:
+                name: libressl
+                version: 3.6.0
           exclude:
             - library:
                 name: boringssl
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -200,7 +200,7 @@ jobs:
               bindgen: false
               library:
                 name: libressl
-                version: 3.5.2
+                version: 3.5.3
             - target: x86_64-unknown-linux-gnu
               bindgen: false
               library:
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -280,6 +280,7 @@ See rust-openssl documentation for more information:
             (3, 4, 0) => ('3', '4', '0'),
             (3, 4, _) => ('3', '4', 'x'),
             (3, 5, _) => ('3', '5', 'x'),
+            (3, 6, _) => ('3', '6', 'x'),
             _ => version_error(),
         };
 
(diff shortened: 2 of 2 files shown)


alex:
I think you should add a CI job to verify this really works: https://github.com/sfackler/rust-openssl/blob/master/.github/workflows/ci.yml#L199

niklasha:
> I think you should add a CI job to verify this really works: https://github.com/sfackler/rust-openssl/blob/master/.github/workflows/ci.yml#L199

Well, good call.  The regression test actually seems to uncover a bug in the LibreSSL 3.6.0 code, either in the new, stricter, ASN.1 parsing implementation, or in the old ASN.1 generation code (which I believe is still coming from OpenSSL proper) not generating ASN.1 per the RFC 5280 requirements, which is what the new ASN.1 parsing code enforces.  I will feed the test case upstream.

sfackler:
You can ignore the boringssl failures - I think that's unrelated to this change.

3.6.0 is a developmental release, so we can't yet allow all 3.6.x versions.

niklasha:
> You can ignore the boringssl failures - I think that's unrelated to this change.
> 
> 3.6.0 is a developmental release, so we can't yet allow all 3.6.x versions.

I have spoken to the LibreSSL team.  They will make an errata for OpenBSD 7.2, and include a fix in LibreSSL portable 3.6.1 .  I.e. the specific rust-openssl tests won't ever pass 3.6.0 on Linux, or unpatched OpenBSD 7.2 (which of course isn't a tier 1 platform so doesn't affect GitHub CI).  I don't know what you want to do with this information?  Wait for merge until 3.6.1 is out, and not support 3.6.0 ?  It is a pretty unusual API use, to use the streaming API for PKCS#7, LibreSSL has been tested pretty widely for 2 months without anyone noticing the bug.  I find a bit "over-the-top" to use an edge API for just testing basic works of rust-openssl.  The normal PKCS#7 workings would probably be better tested with something like [this](https://github.com/sfackler/rust-openssl/commit/aec3bb30c40230c4d55da66f1a09c845be0ca872).

niklasha:
FYI: OpenBSD-current has got a [fix](https://marc.info/?l=openbsd-cvs&m=166602943014106&w=2), which I have run the rust-openssl test-suite against with positive results.  It has yet to show up as an errata in the OPENBSD_7_2 branch or in LibreSSL portable 3.6.1.

sfackler:
It's probably easiest to ignore the tests on libressl for now.

vishwin:
Earlier releases of a version series are supported, but tests may not necessarily pass for them as CI only tracks latest versions of each series.

niklasha:
Superseded by #1705 which contains all of this PR, closing.

PR #1705, opened 2022-10-18
LibreSSL 3.6.0

vishwin:
Limit to 3.6.0 only, as the 3.6 series is still considered developmental. Pet CI, add 3.6.0 and 3.6.1 definitions in `openssl` and use them to ignore two failing PKCS#7 tests due to a bug. Incorporates #1695.

There have been some API additions and removals for 3.6.0 but nothing concerning what we implement/expose.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -201,6 +201,11 @@ jobs:
               library:
                 name: libressl
                 version: 3.5.2
+            - target: x86_64-unknown-linux-gnu
+              bindgen: false
+              library:
+                name: libressl
+                version: 3.6.0
           exclude:
             - library:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -185,7 +185,12 @@ jobs:
               bindgen: true
               library:
                 name: libressl
-                version: 3.5.2
+                version: 3.5.3
```
sfackler, line 188: We can remove the 3.5.3 tests here.

vishwin:
3.5 is still the latest stable series for now, so better to leave it until 3.6 is marked stable.

sfackler:
Thanks!

PR #1706, opened 2022-10-18
Provide method to skip initialization

FreeMasen:
This PR creates the ability to disable openssl initialization in the openssl_sys crate. In a C code base where Rust is being consumed as a library, the consumer may want to perform the initialization explicitly and since this crate's initialization is lazy, it would end up clobbering any global registration already performed.

### DIFF
--- a/openssl-sys/src/lib.rs
+++ b/openssl-sys/src/lib.rs
@@ -85,6 +85,10 @@ mod openssl {
     mod x509_vfy;
     mod x509v3;
 
+    use std::sync::Once;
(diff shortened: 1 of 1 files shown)


sfackler:
I think it'd just look like moving the once static outside of the existing `init` function and then adding something like

```rust
pub unsafe fn assume_init() {
    INIT.call_once(|| {});
}
```

sfackler:
Nope

FreeMasen:
Thanks again for the guidance, I have updated the `lib.rs` to expose `unsafe pub fn assume_init`

PR #1708, opened 2022-10-20
support LibreSSL 3.6.x

timkuijsten:
Successfully compiles on OpenBSD 7.2 amd64.

### DIFF
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -280,6 +280,7 @@ See rust-openssl documentation for more information:
             (3, 4, 0) => ('3', '4', '0'),
             (3, 4, _) => ('3', '4', 'x'),
             (3, 5, _) => ('3', '5', 'x'),
+            (3, 6, _) => ('3', '6', 'x'),
             _ => version_error(),
         };
 


timkuijsten:
I only now see #1695 and #1705. Please note that OpenBSD has been released today and so LibreSSL 3.6.0 is stable now (apart from the upcoming errata).

vishwin:
The 3.6.0 release notes specifically state that it is a development release for the 3.6 branch.

timkuijsten:
Yeah it's a bit confusing. LibreSSL 3.6.0 was released ahead of OpenBSD 7.2 on the 6th of October as a new development release with the following note:
> We have released LibreSSL 3.6.0, which will be arriving in the
LibreSSL directory of your local OpenBSD mirror soon. This is a
development release for the 3.6.x branch, and we appreciate additional testing
and feedback before the final release coming soon with OpenBSD 7.2.
https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.6.0-relnotes.txt

Yesterday OpenBSD 7.2 (stable) was released which includes LibreSSL 3.6.0: https://www.openbsd.org/72.html

vishwin:
Does not mean the API for the series is stable yet. Have been many past instances where stability didn't happen until .3 or later, so we have to account for each point release until then.

sfackler:
Fixed by #1705

PR #1709, opened 2022-10-24
Add openssl::dsa::DsaSig

wiktor-k:
Adds safe interface for parsing and constructing DSA signatures and their components: `r` and `s`.

This mirrors the interface offered by `EcdsaSig` with one omission: there is no `sign` method on `DsaSig` (as in `EcdsaSig`) since [`DSA_do_sign`](https://www.openssl.org/docs/manmaster/man3/DSA_do_sign.html) seems to be deprecated.

### DIFF
--- a/openssl/src/dsa.rs
+++ b/openssl/src/dsa.rs
@@ -344,6 +344,159 @@ cfg_if! {
     }
 }
 
+foreign_type_and_impl_send_sync! {
+    type CType = ffi::DSA_SIG;
+    fn drop = ffi::DSA_SIG_free;
+
+    /// Object representing DSA signature.
+    ///
+    /// DSA signatures consist of two components: `r` and `s`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use std::convert::TryInto;
+    ///
+    /// use openssl::bn::BigNum;
+    /// use openssl::dsa::{Dsa, DsaSig};
+    /// use openssl::hash::MessageDigest;
+    /// use openssl::pkey::PKey;
+    /// use openssl::sign::{Signer, Verifier};
+    ///
+    /// const TEST_DATA: &[u8] = &[0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
+    /// let dsa_ref = Dsa::generate(1024).unwrap();
+    ///
+    /// let pub_key: PKey<_> = dsa_ref.clone().try_into().unwrap();
+    /// let priv_key: PKey<_> = dsa_ref.try_into().unwrap();
+    ///
+    /// let mut signer = if let Ok(signer) = Signer::new(MessageDigest::sha256(), &priv_key) {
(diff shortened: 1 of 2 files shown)


wiktor-k:
Sorry for the flood of force-pushes but the CI pipeline here is really comprehensive and I can't (or don't know how) to test that locally :sweat_smile: 

Thank you for your time :bow:

openssl/src/dsa.rs
```diff
@@ -344,6 +344,156 @@ cfg_if! {
...
+            DSA_SIG_set0(sig, r.as_ptr(), s.as_ptr());
+            mem::forget((r, s));
+            Ok(DsaSig::from_ptr(sig))
+        }
+    }
+
+    from_der! {
+        /// Decodes a DER-encoded DSA signature.
+        #[corresponds(d2i_DSA_SIG)]
+        signature_from_der,
```
sfackler, line 420: The `signature` part of the name here seems a bit redundant given that it's being called from `DsaSig`. Any reason not to just call it `from_der`?

sfackler:
Looks good to me other than the one function name!

openssl/src/dsa.rs
wiktor-k: Hmm... nope... I thought I copied the same name from `EcdsaSig` but apparently not :)

Should be fixed now!

Thanks for the review.

wiktor-k:
Function name adjusted.

openssl/src/dsa.rs
```diff
@@ -344,6 +344,156 @@ cfg_if! {
...
+        #[corresponds(d2i_DSA_SIG)]
+        from_der,
+        DsaSig,
+        ffi::d2i_DSA_SIG
+    }
+}
+
+impl fmt::Debug for DsaSig {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "DSA Sig")
```
Skepfyr, line 428: It's a shame this isn't better but it's better than not having a debug implementation, could this at least be:
```suggestion
        write!(f, "DsaSig")
```
for consistency with other debug implementations (generally it looks like Rust code).

wiktor-k: For the record I'm ok with adding more data to debug output like the R and S components. I'm just not sure what's the policy in this crate of (potentially) calling native functions from debug formatter and I just mimicked the other types.

Skepfyr: Given the r and s functions are infallible and don't allocate (BigNum's Debug impl does but seems fine) I suspect we should use debug_struct here and show both of those fields. I can't think of any reason not to.

wiktor-k: Good points! I've rewritten the Debug impl using `debug_struct` and added a test case.

Skepfyr:
Thanks!

PR #1710, opened 2022-10-24
Add support for X509_VERIFY_PARAM_set_time and X509_VERIFY_PARAM_set_…

alexanderjordanbaker:
…depth


openssl/src/x509/store.rs
```diff
@@ -122,6 +122,13 @@ impl X509StoreBuilderRef {
     pub fn set_flags(&mut self, flags: X509VerifyFlags) -> Result<(), ErrorStack> {
         unsafe { cvt(ffi::X509_STORE_set_flags(self.as_ptr(), flags.bits())).map(|_| ()) }
     }
+
+    /// Sets certificate chain validation related parameters.
+    #[corresponds[X509_STORE_set1_param]]
+    #[cfg(any(ossl102, libressl261))]
+    pub fn set_param(&mut self, param: X509VerifyParam) -> Result<(), ErrorStack> {
+        unsafe { cvt(ffi::X509_STORE_set1_param(self.as_ptr(), param.as_ptr())).map(|_| ()) }
```
sfackler, line 130: This method doesn't take ownership of the param, so it should take `&X509VerifyParamRef`.

openssl/src/x509/verify.rs
```diff
@@ -139,4 +150,16 @@ impl X509VerifyParamRef {
             .map(|_| ())
         }
     }
+
+    /// Set the verification time
+    #[corresponds(X509_VERIFY_PARAM_set_time)]
+    pub fn set_time(&mut self, time: time_t) {
```
sfackler, line 156: If nothing else, this should at least document the units of `time`.

alexanderjordanbaker: The format of time_t is not defined in the C standard, per https://en.cppreference.com/w/c/chrono/time_t, however it is usually seconds since epoch. I'm not sure we can reliably state that that will always be true in Rust, though

openssl/src/x509/store.rs
alexanderjordanbaker: Updated, thank you

openssl/src/x509/verify.rs
alexanderjordanbaker: Updated to add additional description to the comment

openssl/src/x509/tests.rs
```diff
@@ -543,3 +545,92 @@ fn test_name_cmp() {
...
+    };
+    verify_params.set_depth(expected_depth);
+    store_bldr.set_param(&verify_params).unwrap();
+    let store = store_bldr.build();
+
+    let mut context = X509StoreContext::new().unwrap();
+    assert!(context
+        .init(&store, &cert, &chain, |c| c.verify_cert())
+        .unwrap());
+    assert!(context
```
Skepfyr, line 633: I might be missing something, but why are you calling assert twice with the same arguments? I'd also be much happier with these tests if they also checked a failing cert or depth (you can do it in the same function).

Skepfyr (approved):

openssl/src/x509/tests.rs
alexanderjordanbaker: test_verify_cert in this file calls it twice, I assumed that was best practice. On the failing side, I do checking failing certs with respect to set_time in test_verify_param_set_time_fails_verification. For set_depth, I don't believe there are any test certs in the repo with a 3-deep chain. What is best practice for creating such a chain? Should I just add it to the repo?

Skepfyr: `test_verify_cert` calling it twice is somewhere between a fluke of history and a test about cleanup, in this case you only need one. Yes, just create a new certificate and throw it in the directory with the rest of them (`openssl/test`), I think this change would be fine without that test but if you're willing to do that it would be nice to have the coverage.

alexanderjordanbaker: Removed the double call, and added new certificates into the repository to create a 3-long chain, which is then used in the set_depth tests to check failure

alexanderjordanbaker:
@sfackler Anything else I need to do for this PR?

sfackler:
Thanks! Sorry for the delay.

PR #1711, opened 2022-10-25
Fix reference to `Self::cipher_final` in docs

wiktor-k:
This patch changes the position of backtick and square brackets so that the reference is rendered as a link in generated documentation.


sfackler:
Thanks!

PR #1713, opened 2022-10-27
Add `BigNum::copy_from_slice`

wiktor-k:
Unfortunately `BigNum::from_slice` always creates a new BigNum and cannot be used to initialize an already existing BigNum thus it is not possible to have a secure BigNum initialized from a slice.
    
This patch adds the function for overwriting existing BigNum with given slice bytes. Thus the BigNum can be created using `BigNum::new_secure` and then overwritten using `BigNum::copy_from_slice`.

### DIFF
--- a/openssl/src/bn.rs
+++ b/openssl/src/bn.rs
@@ -1070,7 +1070,7 @@ impl BigNum {
     pub fn from_slice(n: &[u8]) -> Result<BigNum, ErrorStack> {
         unsafe {
             ffi::init();
-            assert!(n.len() <= c_int::max_value() as usize);
+            assert!(n.len() <= LenType::max_value() as usize);
 
             cvt_p(ffi::BN_bin2bn(
                 n.as_ptr(),
--- a/openssl/src/bn.rs
+++ b/openssl/src/bn.rs
@@ -1080,6 +1080,30 @@ impl BigNum {
             .map(|p| BigNum::from_ptr(p))
         }
     }
+
+    /// Copies data from a slice overwriting what was in the BigNum.
+    ///
+    /// This function can be used to copy data from a slice to a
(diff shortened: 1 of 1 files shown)


sfackler:
I think I'd prefer a method taking `&mut self` and overwriting the bignum like you mentioned.

wiktor-k:
Ok, I'll rewrite the PR. Thanks!

openssl/src/bn.rs
```diff
@@ -1080,6 +1080,31 @@ impl BigNum {
...
+    /// # use openssl::bn::BigNum;
+    /// let mut bignum = BigNum::new().unwrap();
+    /// bignum.copy_from_slice(&[0x12, 0x00, 0x34]).unwrap();
+    ///
+    /// assert_eq!(bignum, BigNum::from_u32(0x120034).unwrap());
+    /// ```
+    #[corresponds(BN_bin2bn)]
+    pub fn copy_from_slice(&mut self, n: &[u8]) -> Result<(), ErrorStack> {
+        unsafe {
+            ffi::init();
```
sfackler, line 1101: init here is unnecessary since we've already constructed the BigNum.

openssl/src/bn.rs
```diff
@@ -1080,6 +1080,31 @@ impl BigNum {
...
+            assert!(n.len() <= c_int::max_value() as usize);
```
sfackler, line 1102: Should be checking against LenType, not c_int.

wiktor-k: Yep. I actually copied the code from `BigNum::from_slice` so I've added another commit to fix it there too.

Thanks for your time! :pray:

sfackler:
Thanks!

PR #1714, opened 2022-10-27
Add NID constants for Brainpool curves

wiktor-k:
Brainpool curves are used by European government organizations that are suspicious of the NIST curves. Brainpool curves are very similar to NIST curves with one critical difference: the parameters to Brainpool are [nothing-up-my-sleeve numbers][0].

The actual values of the NID constants have been taken from the [OpenSSL source code][1].

[0]: https://github.com/veorq/numsgen#seeds

[1]: https://github.com/openssl/openssl/blob/4e6647506331fc3b3ef5b23e5dbe188279ddd575/include/openssl/obj_mac.h#L4759

### DIFF
--- a/openssl-sys/src/obj_mac.rs
+++ b/openssl-sys/src/obj_mac.rs
@@ -94,6 +94,12 @@ pub const NID_sect409k1: c_int = 731;
 pub const NID_sect409r1: c_int = 732;
 pub const NID_sect571k1: c_int = 733;
 pub const NID_sect571r1: c_int = 734;
+#[cfg(ossl110)]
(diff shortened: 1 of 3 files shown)


wiktor-k:
Thanks for the hint the error messages can be a little bit confusing at first...

Is there a way to check which version of OpenSSL introduced which curves? I've tried to get the `cfg(ossl...)` version correct but it seems it's not that easy (or I'm holding the search engine wrong :sweat_smile: ).

sfackler:
I usually just look through the releases manually to see when they ended up in the headers the first time.

Issue #1715, opened 2022-10-28
Error formatting issues


rukai:
I investigated the formatting of `openssl::error::Error` and observed 3 bugs:
1. fmt::Display of `openssl::ssl::error::Error` and the error it returns via `.source()` are identical.
This results in weird looking error chains where the same error is repeated twice.
Internally what happens is `openssl::ssl::error::Error` forwards its formatting logic onto its inner source value.
We either need to return None for `.source()`, or provide a much more vague error for the parent error to avoid the error being confusingly duplicated.
2. fmt::Display implementation does not provide enough context for a human to make sense of it.
3. the data field is never populated due to passing a NULL pointer for `data` and `flags` to `ERR_get_error_all` as documented here: https://www.openssl.org/docs/man3.0/man3/ERR_get_error_line_data.html

I will be opening PRs for some of these shortly.

sfackler:
3. It is not passing a null pointer to ERR_get_error_all, it is passing a pointer to a null pointer that ERR_get_error_all will fill in. The data field is absolutely populated and is in fact tested here for example: https://github.com/sfackler/rust-openssl/blob/master/openssl-errors/tests/test.rs#L59-L74

rukai:
1. I cant think of a suitable solution.
If we set .source() to None then the user loses access to the inner values.
If we make the parent error return a generic message like "SSL error" then this causes a silent breaking change where everyone who only used the parent error suddenly gets really bad error messages.

2. was rejected https://github.com/sfackler/rust-openssl/pull/1716

3. You are absolutely right, I'm not used to reading C API's, my bad.
I am suspicious there might be something broken here as another python application using openssl was able to report extra information in a specific openssl failure case that isnt included in the error my application received from openssl-rs in that same case. (and I am yet to see a single `data() == Some` in any of the errors I get)
But that is just a suspicion on my part and I dont have anything else to go on, so lets consider this invalid.

## My solution
For anyone who comes across this issue, this was my solution.
It uses anyhow but you should be able to easily adjust it to your own error types.
It keeps the same information provided in the Display trait but formats it in a human readable way.
```rust
// Always use these openssl_* conversion methods instead of directly directly converting to anyhow

pub fn openssl_ssl_error_to_anyhow(error: openssl::ssl::Error) -> anyhow::Error {
    if let Some(stack) = error.ssl_error() {
        openssl_stack_error_to_anyhow(stack.clone())
    } else {
        anyhow!("{error}")
    }
}

[...]

PR #1716, opened 2022-10-28
Improve fmt::Display impl for Error

rukai:
Resolves point 2 in https://github.com/sfackler/rust-openssl/issues/1715

A bit of bikeshedding is welcome but this change is already a lot more readable than the previous implementation.

## Example 1
An error in the old format:
`error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:1919:`

That same error with the new format:
`error 0x1416F086 'certificate verify failed' occurred in function 'tls_process_server_certificate' in file 'ssl/statem/statem_clnt.c:1919' in library 'SSL routines'`

## Example 2
If we had a working data implementation (point 3 in https://github.com/sfackler/rust-openssl/issues/1715 ) the previous example would look like this.

An error in the old format:
`error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:1919:DATA_HERE`

That same error with the new format:
`error 0x1416F086 'certificate verify failed: DATA_HERE' occurred in function 'tls_process_server_certificate' in file 'ssl/statem/statem_clnt.c:1919' in library 'SSL routines'`

### DIFF
--- a/openssl/src/error.rs
+++ b/openssl/src/error.rs
@@ -315,8 +315,8 @@ impl fmt::Display for Error {
         }
         write!(fmt, "in file '{}:{}' ", self.file(), self.line())?;
         match self.library() {
-            Some(l) => write!(fmt, "in library '{}' ", l),
-            None => write!(fmt, "in library 'Unknown' "),
+            Some(l) => write!(fmt, "in library '{}'", l),
+            None => write!(fmt, "in library 'Unknown'"),
         }
     }
 }
--- a/openssl/src/error.rs
+++ b/openssl/src/error.rs
@@ -298,26 +298,26 @@ impl fmt::Debug for Error {
 
 impl fmt::Display for Error {
     fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(fmt, "error:{:08X}", self.code())?;
-        match self.library() {
-            Some(l) => write!(fmt, ":{}", l)?,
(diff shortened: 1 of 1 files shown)


sfackler:
The current implementation is designed to mirror the output of OpenSSL's own error formatting: https://www.openssl.org/docs/man1.1.1/man3/ERR_print_errors_fp.html

rukai:
I see, is openssl's error formatting a generally understood output format? Or is there some other advantage to matching openssl's format that I'm missing?

For my own use case I bubble up openssl errors to the user of the application.
I cant hope to handle every possible openssl error and provide useful handling or advice directly to the user.
So I want the errors to be as readable as possible for them.
I think this would be a really common need so it makes sense for us to change the error message in rust where we value different trade offs to those that are valued in C.

sfackler:
The objective of this library is to expose OpenSSL's APIs as directly as possible. Mirroring its error output is part of that. If you would like a different format then I'd recommend converting the `openssl::ErrorStack` into your own error type.

rukai:
The Display trait is documented as for user facing output: https://doc.rust-lang.org/std/fmt/trait.Display.html

While the Debug trait is documented as for programmer facing output: https://doc.rust-lang.org/std/fmt/trait.Debug.html

I think its pretty silly that openssl-rs's implementation of the Debug trait is better suited for user facing output than the Display trait due to actually conveying what the fields mean.

But if its openssl-rs's goal to expose that format to the end user, then so be it.

PR #1717, opened 2022-10-31
Expose Camellia, CAST5 and IDEA ciphers

wiktor-k:
All of them are used in the [OpenPGP specification for symmetric ciphers][0].

IDEA and CAST5 may be old but they are provided for users for decryption of archive data. Camellia is used in the new version of the OpenPGP specification as an alternative to AES.

Camellia has [more variants available](https://www.openssl.org/docs/man1.1.1/man3/EVP_camellia_128_cfb.html). I've exposed only the ones I need but maybe it's a good idea to expose more for completeness (just like AES).

(Just for context I'm working on OpenSSL cryptographic backend for an OpenPGP library: https://gitlab.com/sequoia-pgp/sequoia/-/merge_requests/1361)

[0]: https://openpgp-wg.gitlab.io/rfc4880bis/#name-symmetric-key-algorithms


wiktor-k:
:thinking: I'm wondering if this needs cipher detection code just like ChaCha. (Interestingly most of the failures are from either `vendored=true` or `vendored=false` jobs)...

openssl-sys/src/handwritten/evp.rs
```diff
@@ -364,6 +364,29 @@ extern "C" {
     #[cfg(all(any(ossl111, libressl291), not(osslconf = "OPENSSL_NO_SM4")))]
     pub fn EVP_sm4_ctr() -> *const EVP_CIPHER;
 
+    #[cfg(ossl111)]
```
Skepfyr, line 367: As far as I can tell, these functions have existed since OpenSSL 0.9.8, why are you restricting the to OpenSSL 1.1.1+? These also appear to be in all versions of libressl. I'm also slightly confused by the boringssl annotations (it appears to not include any of these?) but I'm much less familiar with boringssl.

wiktor-k: To be honest I was trying to get the CI green even with restrictive attributes and then slowly relaxing them (I don't see a way of checking easily which version of OpenSSL introduced which function... The man pages on openssl home page are a little bit scarce).

Thanks for your help!

Skepfyr: Yeah it's a pain, the best way I've found is looking at what commit introduced them to the header files.

Skepfyr:
> :thinking: I'm wondering if this needs cipher detection code just like ChaCha. (Interestingly most of the failures are from either `vendored=true` or `vendored=false` jobs)...

The naming of the jobs isn't vendored=true ofr vendored=false the true or false on the end indicates if bindgen is involved. All the jobs that are failing are the vendored jobs because openss-src isn't built with camellia or idea support by default (and I assume CAST5 but that's not even mentioned). I suspect we will need something like the ChaCha optional support but I'm not entirely certain about that.

openssl-sys/src/handwritten/evp.rs
wiktor-k: Good point, git blame'ing and then following to release tag should be quite easy. For these particular ciphers it seems they were introduced long ago enough that all tested versions either contain it or filter them out with `OPENSSL_NO_xxx` defines.

I've updated the `cfg`s to be as broad as possible.

Thanks for help!

wiktor-k:
Thanks for all your suggestions @Skepfyr, I really appreciate them :bow: !

I was able to mimic what was done for other optional ciphers and it seems to be working, phew :sweat_smile:

sfackler:
Thanks!

LeoRowan:
Hi, it seems that this change has broken future builds for our production pipeline because we are using a hardened version of OpenSSL. Our version does not allow CAST5 ciphers, and the addition of this commit causes a requirement on CAST5. I looked for a way to turn it off, but it looks like this commit doesn't have any cipher detection code for CAST5, only the other added ciphers. Am I missing something here?

wiktor-k:
Hmm... Indeed, now I found a `OPENSSL_NO_CAST` symbol that seems to be relevant. If I made a PR fixing this would you be able to verify the fix @LeoRowan ?

LeoRowan:
Yeah, I would be happy to, I'll move conversation to that PR

PR #1725, opened 2022-11-08
[openssl-sys] adding no-bundle option to openssl

lightyear15:
Definining ``OPENSSL_STATIC_NOBUNDLE=1`` environment variable, so
openssl libraries are linked via ``static-nobundle`` options.
See
https://github.com/rust-lang/rfcs/blob/master/text/1717-dllimport.md#unbundled-static-libs-optional

##### Some more context:
I am trying to build a static library 
```toml
[lib]
crate-type = ["staticlib"]
```

that includes openssl as a dependency.
This rust-based static library (let's call it ``librust.a`` for the sake of the explanation) is part of a bigger project that includes some C/C++ code as well, which itself uses openssl.
The problem I am having is at the linking phase where, linking the C++ code against both  ``libopenssl.a`` and ``librust.a``, gives me an error of ``duplicate symbol`` as all the openssl symbols are present in both static libraries.

This change would allow me to leave all the openssl symbols within ``librust.a`` as undefined, and let the linker solve them at the very final phase where the project-wide library is built and all symbols resolved.

### DIFF
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -364,11 +364,16 @@ fn parse_new_version(version: &str) -> u64 {
 /// statically or dynamically.
 fn determine_mode(libdirs: &[PathBuf], libs: &[&str]) -> &'static str {
     // First see if a mode was explicitly requested
-    let kind = env("OPENSSL_STATIC");
-    match kind.as_ref().and_then(|s| s.to_str()) {
+    let _static = env("OPENSSL_STATIC");
+    match _static.as_ref().and_then(|s| s.to_str()) {
         Some("0") => return "dylib",
         Some(_) => return "static",
         None => {}
+    }
+    let _static_nobundle = env("OPENSSL_STATIC_NOBUNDLE");
+    match _static_nobundle.as_ref().and_then(|s| s.to_str()) {
+        Some("0") | None => {}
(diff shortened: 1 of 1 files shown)


sfackler:
I think using the static-nobundle link kind may be a better option?

lightyear15:
> 

Didn't know about that option :thinking: 
I will give it a try
thanks

sfackler:
Are there any other sys crates that have an option like this?

lightyear15:
> Are there any other sys crates that have an option like this?

After a quick search, I don't think there are many out there that allow somehow the user to take the path of ``static-nobundle`` linking.

I was just thinking, :thinking:  maybe I can trick the rust linker to do what I want by instructing openssl-sys to link dynamically, so that when building my ``staticlib`` all the openssl symbols will be left undefined. 
But then when finally linking my project library using gcc/clang I can explicitly say I want to link ``libssl`` and ``libcrypto`` statically.
Of course I need to be sure to have those ``libssl`` and ``libcrypto`` as both static and dynamic libraries...

Would that work?? :thinking:

sfackler:
I don't think you'll be able to dynamically-and-then-statically link.

I think it'd probably be best to figure out how to enable this configuration on the pkg-config-rs repo and then mirror that behavior here.

PR #1733, opened 2022-11-14
Make `CipherCtx::cipher_update` more flexible

### LINKED ISSUE #1729: `CipherCtx::cipher_update` is too pessimistic
Hi,

I'm using the symmetric cipher encryption API and it seems the constraints in `CipherCtx::cipher_update` are too strict for block ciphers compared to what OpenSSL requires.

Currently the code checks if the output size has sufficient size and for streaming ciphers it's OK but for block ciphers it always assumes pessimistic case of partial block updates - that is feeding inputs that are smaller than the block size. Then, and only then the output needs to be big enough to contain the extra block.

I'm using block ciphers in a very strict case of **always feeding data one entire block at a time** and I'd like the output buffer to also be one block of length (this would make it consistent with other cryptographical backends that we use and also consistent with OpenSSL itself).

I see the following solutions:
  1. introduce an `unsafe` function that would not require the output buffer to be big enough for additional block. I've written a sample code and tested that against our data:

```rust
    pub unsafe fn unchecked_cipher_update(
        &mut self,
        input: &[u8],
        output: &mut [u8],
    ) -> Result<usize, ErrorStack> {
        let inlen = LenType::try_from(input.len()).unwrap();

        let block_size = self.block_size();
        if block_size > 1 {
            // block cipher
            assert_eq!(block_size, input.len());
            assert!(output.len() >= block_size);
        } else {
            // stream cipher
            assert_eq!(input.len(), output.len());
        }

        let mut outlen = 0;

        cvt(ffi::EVP_CipherUpdate(
            self.as_ptr(),
            output.as_mut_ptr(),
            &mut outlen,
            input.as_ptr(),
            inlen,
        ))?;

        Ok(outlen as usize)
    }
```

(Full code at https://github.com/wiktor-k/rust-openssl/commit/7d0fff8aec8813d51319ba8c15ca80972c5d8034). Note that this function is unsafe only when mixed with partial block updates with `CipherCtx::cipher_update`. The disadvantage of this approach is exposing unsafe functions.

  2. Create other struct that has almost identical API but only allows stream ciphers or block ciphers where `input.len() == block_size == output.len()`. This has the advantage of offering safe API but the disadvantage of duplicating code (although for my use case only parts are needed).

  3. In `CipherCtx` keep track of how many bytes were fed into update and thus always require the output buffer to be of correct size. The advantage is that no public API would change and the `cipher_update` would just become more flexible. The disadvantage: additional complexity of tracking bytes modulo block size.

  4. Do not change anything and just make clients use raw `ffi` API. I'd like to avoid it but if there's no better solution I'd have to do that.

Maybe there are some other better ideas, if so I'm all ears!

Thank you for your time and sorry that this issue is so long :bow: 

-------------------

For reference I got the exact algorithm description from Matt from OpenSSL ([source](https://mta.openssl.org/pipermail/openssl-users/2022-November/015623.html)):

<blockquote>

EVP_EncryptUpdate() can be called repeatedly, incrementally feeding in 
the data to be encrypted. The ECB mode (when used with AES-128) will 
encrypt input data 16 bytes at a time, and the output size will also be 
16 bytes per input block. If the data that you feed in to 
EVP_EncryptUpdate() is not a multiple of 16 bytes then the amount of 
data that is over a multiple of 16 bytes will be cached until a 
subsequent call where it does have 16 bytes.

Let's say you call EVP_EncryptUpdate() with 15 bytes of data. In that 
case all 15 bytes will be cached and 0 bytes will be output.

If you then call it again with 17 bytes of data, then added to the 15 
bytes already cached we have a total of 32 bytes. This is a multiple of 
16, so 2 blocks (32 bytes) will be output, so:

(inl + cipher_block_size - 1) = (17 + 16 - 1) = 32
</blockquote>

wiktor-k:
This change relaxes constraints on the output buffer size when it can be safely determined how many bytes will be put in the output buffer.

For supported cryptographic backends (OpenSSL >= 1.1) the cipher's `num` parameter will be consulted for the number of bytes in the block cache. For unsupported backends the behavior will not change (the code will assume full block in the cache).

For callers that do the check themselves and want to use other backends (e.g. BoringSSL or LibreSSL) and unsafe `unchecked_cipher_update` function is added.

Additionally a `CipherCtx::minimal_output_size` function is added for letting the callers know how big should the output buffer be for the next `cipher_update` call.

Fixes #1729.

See: https://mta.openssl.org/pipermail/openssl-users/2022-November/015623.html

### DIFF
--- a/openssl/src/cipher_ctx.rs
+++ b/openssl/src/cipher_ctx.rs
@@ -363,6 +363,22 @@ impl CipherCtxRef {
         unsafe { ffi::EVP_CIPHER_CTX_iv_length(self.as_ptr()) as usize }
     }
 
+    /// Returns the `num` parameter of the cipher.
+    ///
+    /// Built-in ciphers typically use this to track how much of the
+    /// current underlying block has been "used" already.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the context has not been initialized with a cipher.
+    #[corresponds(EVP_CIPHER_CTX_num)]
+    #[cfg(ossl110)]
+    pub fn num(&self) -> usize {
+        self.assert_cipher();
+
+        unsafe { ffi::EVP_CIPHER_CTX_num(self.as_ptr()) as usize }
+    }
+
     /// Sets the length of the IV expected by this context.
     ///
     /// Only some ciphers support configurable IV lengths.
--- a/openssl/src/cipher_ctx.rs
+++ b/openssl/src/cipher_ctx.rs
@@ -379,6 +379,49 @@ impl CipherCtxRef {
         unsafe { ffi::EVP_CIPHER_CTX_num(self.as_ptr()) as usize }
     }
 
+    /// Returns number of bytes cached in partial block update.
+    #[cfg(ossl110)]
+    fn used_block_size(&self) -> usize {
+        self.num()
+    }
+
+    /// Returns maximum number of bytes that could be cached.
+    #[cfg(not(ossl110))]
+    fn used_block_size(&self) -> usize {
+        self.block_size()
+    }
+
+    /// Calculate the minimal size of the output buffer given the
+    /// input buffer size.
+    ///
+    /// For streaming ciphers the minimal output size is the same as
+    /// the input size. For block ciphers the minimal output size
+    /// additionally depends on the partial blocks that might have
+    /// been written in previous calls to [`Self::cipher_update`].
+    ///
+    /// This function takes into account the number of partially
+    /// written blocks for block ciphers for supported targets
+    /// (OpenSSL >= 1.1). For unsupported targets the number of
+    /// partially written bytes is assumed to contain one full block
+    /// (pessimistic case).
+    ///
+    /// # Panics
+    ///
+    /// Panics if the context has not been initialized with a cipher.
+    pub fn minimal_output_size(&self, inlen: usize) -> usize {
+        let block_size = self.block_size();
+        if block_size > 1 {
+            // block cipher
+            let num = self.used_block_size();
+            let total_size = inlen + num;
+            let num_blocks = total_size / block_size;
+            num_blocks * block_size
+        } else {
+            // streaming cipher
+            inlen
+        }
+    }
+
     /// Sets the length of the IV expected by this context.
     ///
     /// Only some ciphers support configurable IV lengths.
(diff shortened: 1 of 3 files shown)


openssl/src/cipher_ctx.rs
```diff
@@ -363,6 +363,65 @@ impl CipherCtxRef {
...
+    /// This function takes into account the number of partially
+    /// written blocks for block ciphers for supported targets
+    /// (OpenSSL >= 1.1). For unsupported targets the number of
+    /// partially written bytes is assumed to contain one full block
+    /// (pessimistic case).
+    ///
+    /// # Panics
+    ///
+    /// Panics if the context has not been initialized with a cipher.
+    pub fn minimal_output_size(&self, inlen: usize) -> usize {
```
sfackler, line 411: I usually don't like to add extra methods on top of OpenSSL APIs, but it seems like it may be reasonable to make an exception here.

openssl/src/cipher_ctx.rs
```diff
@@ -501,33 +560,61 @@ impl CipherCtxRef {
...
+    /// cache data (e.g. OpenSSL < 1.1, BoringSSL, LibreSSL).
+    ///
+    /// SAFETY: The caller is expected to provide `output` buffer
+    /// large enough to contain correct number of bytes. For streaming
+    /// ciphers the output buffer size should be at least as big as
+    /// the input buffer. For block ciphers the size of the output
+    /// buffer depends on the state of partially updated blocks (see
+    /// [`Self::minimal_output_size`]).
+    #[corresponds(EVP_CipherUpdate)]
+    pub unsafe fn unchecked_cipher_update(
```
sfackler, line 602: Nit: I believe the Rust naming convention for this would be `cipher_update_unchecked` to match e.g. `[T]::get_unchecked`.

sfackler:
Looks good to me other than the method naming!

openssl/src/cipher_ctx.rs
wiktor-k: Understood!

openssl/src/cipher_ctx.rs
wiktor-k: Yep, good catch! Thanks.

wiktor-k:
Rebased on top of `master` and fixed the naming issue.

Thanks for your reviews! I really appreciate it :pray: 

:thinking: It seems I missed a release with this PR... if you don't mind I'll file a couple of small documentation + example PRs.

sfackler:
PRs are always welcome!

sfackler:
I'm happy to cut a new release as well if you'd like.

wiktor-k:
> I'm happy to cut a new release as well if you'd like.

Thank you! I'll be back with one more documentation PR and then I'll kindly ask you for a new release (it will also look better on docs.rs). Thanks in advance, I'll be back :)

Issue #1721, opened 2022-11-02
Attempt to use the 'ParsedPkcs12.chain' member for keystores without a key chain results in segfault


ancwrd1:
I have reproduced it by attempting to read the Java truststore file (PKCS12 with only certificates, no private key chains).

Code example:

```rust
let pkcs12 = Pkcs12::from_der(data)?;
let parsed = pkcs12.parse(password)?;
if let Some(chain) = parsed.chain {
    for cert in chain {
        println!("{:?}", cert.subject_name()); // <-- SEGFAULT
    }
}
```
The chain is returned as `Some` value and the X509 cert in there has a null internal pointer (checked it in the debugger).

sfackler:
Do you have an example PKCS12 archive I can use to test?

ancwrd1:
Sure, here is the (zipped) file which triggers this behavior. Password is 'changeit'.
[truststore.zip](https://github.com/sfackler/rust-openssl/files/9926795/truststore.zip)

sfackler:
Hmm, I can't reproduce that crash:

```rust
   #[test]
    fn chain() {
        let der = include_bytes!("/Users/sfackler/Downloads/truststore.p12");
        let pkcs12 = Pkcs12::from_der(der).unwrap();
        let parsed = pkcs12.parse("changeit").unwrap();
        for cert in &parsed.chain.unwrap() {
            println!("{:?}", cert.subject_name());
        }
    }
```

```
running 1 test
[domainComponent = "lan", domainComponent = "home", domainComponent = "inforce", commonName = "inforce-W12K-CA"]
test pkcs12::test::chain ... ok
```

ancwrd1:
Ah, I think it's not the `parsed.chain`, but `parsed.cert`.
Try `println!("{:?}", parsed.cert.subject_name());`

sfackler:
Ok, yeah I see. We'll have to deprecate `ParsedPkcs12` and make a new version with an optional cert and private key.

ancwrd1:
Thanks. FYI, the current (ugly) workaround is to check if private key has zero bits size, then the cert is assumed to not be valid.

PR #1735, opened 2022-11-22
PKCS#12 create without pkey or cert

### LINKED ISSUE #1734: PKCS#12 Create without pkey or cert
The current bindings do not have support for building a PKCS#12 archive without a `pkey` or `cert` entry.

The OpenSSL docs state that either entry can be set `NULL` in order to do this.
https://www.openssl.org/docs/man1.0.2/man3/PKCS12_create.html

jxtur:
In the build function, make the cert and key entries Options, which are set to ptr::null_mut when None and otherwise the inner ptr when calling the FFI.

Fixes #1734 

#1721 prevents full testing from crate functions. However, the produced DER-encoded PKCS#12 has been parse by the OpenSSL binary 1.1.1.

### DIFF
--- a/openssl/src/pkcs12.rs
+++ b/openssl/src/pkcs12.rs
@@ -164,8 +164,8 @@ impl Pkcs12Builder {
         self,
         password: &str,
         friendly_name: &str,
-        pkey: &PKeyRef<T>,
-        cert: &X509Ref,
+        pkey: Option<&PKeyRef<T>>,
+        cert: Option<&X509Ref>,
     ) -> Result<Pkcs12, ErrorStack>
     where
         T: HasPrivate,
@@ -173,8 +173,8 @@ impl Pkcs12Builder {
         unsafe {
             let pass = CString::new(password).unwrap();
             let friendly_name = CString::new(friendly_name).unwrap();
-            let pkey = pkey.as_ptr();
-            let cert = cert.as_ptr();
+            let pkey = pkey.map_or_else(ptr::null_mut, |k| k.as_ptr());
+            let cert = cert.map_or_else(ptr::null_mut, |c| c.as_ptr());
             let ca = self
                 .ca
                 .as_ref()
@@ -304,7 +304,7 @@ mod test {
 
         let pkcs12_builder = Pkcs12::builder();
         let pkcs12 = pkcs12_builder
-            .build("mypass", subject_name, &pkey, &cert)
+            .build("mypass", subject_name, Some(&pkey), Some(&cert))
             .unwrap();
         let der = pkcs12.to_der().unwrap();
 


openssl/src/pkcs12.rs
```diff
@@ -164,17 +164,17 @@ impl Pkcs12Builder {
         self,
         password: &str,
         friendly_name: &str,
-        pkey: &PKeyRef<T>,
-        cert: &X509Ref,
+        pkey: Option<&PKeyRef<T>>,
```
wiktor-k, line 167: The parameter type here could be `Into<Option<...>>` so that the existing code continues to work and `None` is a new allowed value to be passed as this parameter.

More elaborate discussion of this technique is in http://web.archive.org/web/20170930063008/http://www.suspectsemantics.com/blog/2016/11/29/the-into-trick/

As for real code using this see: https://gitlab.com/sequoia-pgp/sequoia/-/blob/main/openpgp/src/cert/builder.rs#L443

wiktor-k: Yeah I didn't mean it will be a silver-bullet, just more backwards-compatibility-proof.

> As far as I can tell, you'd still lose the ability to automatically ref a `PKey<_>` as `&PKeyRef<_>` without defining the trait `From<&pkey::PKey<_>>` for `Option<&PKeyRef<_>>`.

Another option (no pun intended) would be to accept an `Into<Option<AsRef` parameter and impl that for `PKey(Ref)`... but at this point the function signature looks extremely ugly.

> So calling code may have to change, do you still think the change should stand?

I'm just thinking about options and trade-offs. I don't have any strong opinion either way. (Sorry if that's not the answer you were looking for :pray:).

jxtur: > Another option (no pun intended) would be to accept an `Into<Option<AsRef` parameter and impl that for `PKey(Ref)`... but at this point the function signature looks extremely ugly.

I've done this where the function signature becomes (can't seem to stack the traits - if you know how I can, please tell me...):
```rust
pub fn build<T, U, V, W, X>(
    self,
    password: &str,
    friendly_name: &str,
    pkey: U,&pkey
    cert: V,
) -> Result<Pkcs12, ErrorStack>
where
    T: HasPrivate,
    U: Into<Option<W>>,
    V: Into<Option<X>>,
    W: AsRef<PKeyRef<T>>,
    X: AsRef<X509Ref>,
```
That can desugar the following types/examples for `pkey`, (generic ignored):
* `PKey`
* `&PKey`
* `PKeyRef`
* `&PKeyRef`
* `None`
But when passing `None`, you have to define the generic types that it cannot infer:
```rust
let pkcs12 = pkcs12_builder
    .build::<Private, _, _, PKeyRef<_>, _>("mypass", subject_name, None, &cert)
    .unwrap();
```

The equivalent applies to the `cert` parameter.

In my opinion, this seems like a poor trade-off the sake of backwards compatibility :woozy_face:  
To me it becomes a bit less obvious the values that _can_ be passed to it, what do you think?

sfackler: I think the better option would be to add `name`, `pkey` and `cert` setters to the builder type, and then a `build2` method that only takes the truly required parameters. I *think* `pass` is required (?) but a test would confirm.

jxtur: I agree with the setters, that matches with the optional CA stack.

On the assumption you'd like a backwards compatible build function now, I've added a `build2` function in a feature branch in my repo. What do you think? To me it seems like a lot duplication to for the purpose of keeping a backwards compatible `build` function. The setters take a `mut` PKCS12 builder, which wouldn't be passed to the `build` function without changing the signature so they couldn't be used to refactor the backward compatible build function.

sfackler: Changing `self` to `mut self` doesn't change the signature of the method - just what the implementation of the method can do with self. It should be fine to have build delegate to build2.

PR #1736, opened 2022-11-22
Improve `PKey` documentation related to ed25519/ed448

wiktor-k:
This patch adds doctests with common usage to the `generate_*` functions as well as cross links to other pieces of documentation.


sfackler:
Thanks!

PR #1740, opened 2022-11-29
Add support for X509_load_cert_file

alexanderjordanbaker:


openssl/src/x509/store.rs
```diff
@@ -176,6 +176,37 @@ impl X509LookupRef<HashDir> {
     }
 }
 
+/// Marker type corresponding to the [`X509_LOOKUP_file`] lookup method.
+///
+/// [`X509_LOOKUP_file`]: https://www.openssl.org/docs/man1.1.1/man3/X509_LOOKUP_file.html
+pub struct ASN1;
```
sfackler, line 182: Why is this called `ASN1` and not e.g. `File`?

alexanderjordanbaker: Thank you for pointing that out, that was a reference that I didn't replace with the appropriate name when creating the PR. I've updated the name to be `File` to match the `HashDir` struct above

openssl/src/x509/store.rs
```diff
@@ -176,6 +176,37 @@ impl X509LookupRef<HashDir> {
...
+    pub fn file() -> &'static X509LookupMethodRef<File> {
+        unsafe { X509LookupMethodRef::from_ptr(ffi::X509_LOOKUP_file()) }
+    }
+}
+
+#[cfg(not(boringssl))]
+impl X509LookupRef<File> {
+    #[corresponds(X509_load_cert_file)]
+    /// Specifies a file from which certificates will be loaded
+    pub fn load_cert_file(&mut self, file: &str, file_type: SslFiletype) -> Result<(), ErrorStack> {
```
sfackler, line 197: The type of `file` should be `impl AsRef<Path>` like here: https://docs.rs/openssl/latest/src/openssl/ssl/mod.rs.html#902-912

sfackler:
LGTM other than the method signature!

sfackler:
Thanks!

PR #1755, opened 2022-12-16
Add binding for X509_load_crl_file

steffen-eiden:
Signed-off-by: Steffen Eiden <seiden@linux.ibm.com>

### DIFF
--- a/openssl/src/x509/store.rs
+++ b/openssl/src/x509/store.rs
@@ -194,8 +194,9 @@ impl X509Lookup<File> {
 
 #[cfg(not(boringssl))]
 impl X509LookupRef<File> {
-    #[corresponds(X509_load_cert_file)]
     /// Specifies a file from which certificates will be loaded
+    #[corresponds(X509_load_cert_file)]
+    // FIXME should return 'Result<i32, ErrorStack' like load_crl_file
     pub fn load_cert_file<P: AsRef<Path>>(
         &mut self,
         file: P,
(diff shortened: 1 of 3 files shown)


steffen-eiden:
I can add at test if #1123 is merged, as before there is no crl in `openssl/test`

Or did I miss one?

openssl/src/x509/store.rs
```diff
@@ -211,6 +211,24 @@ impl X509LookupRef<File> {
...
         }
     }
+
+    #[corresponds(X509_load_crl_file)]
+    /// Specifies a file from which certificate revocation lists will be loaded
+    pub fn load_crl_file<P: AsRef<Path>>(
+        &mut self,
+        file: P,
+        file_type: SslFiletype,
+    ) -> Result<(), ErrorStack> {
```
sfackler, line 221: Probably makes sense to propagate the return value since it indicates the number of loaded CRL entries.

sfackler:
Yeah I don't think we have an existing CRL for tests.

openssl/src/x509/store.rs
steffen-eiden: I second that. However, the corresponding cert function does also not return that value, so I kept it returning `void`

I can change `load_cert_file` to returning ` Result<i32, ErrorStack` as well If that is OK for you.

sfackler: That'd be a breaking change so I wouldn't want to do it now. Can you just add a FIXME to change that signature to match load_crl_file?

steffen-eiden: OK I'll do that

PR #1760, opened 2022-12-19
Adding support for X509_get_ext_by_NID and OBJ_txt2nid

alexanderjordanbaker:


openssl/src/nid.rs
```diff
@@ -78,6 +78,22 @@ impl Nid {
         }
     }
 
+    /// Returns the `Nid` for the supplied `s` which can be the
+    /// numerical format, short name, or long name
+    #[corresponds(OBJ_txt2nid)]
+    pub fn txt_to_nid(s: &str) -> Result<Nid, ErrorStack> {
```
sfackler, line 84: Should this be a `Parse` implementation instead?

alexanderjordanbaker: Do you mean changing the function name to something like `from_str`, or what Parse implementation are you referencing?

alexanderjordanbaker: An important point about this function, it will not create that NID if it does not exist, so parse might be somewhat of a misnomer. It really is more like fetch nid if it exists

sfackler: Oh sorry, `FromStr`, not `Parse`: https://doc.rust-lang.org/stable/std/str/trait.FromStr.html

alexanderjordanbaker: I've updated to implement the `FromStr` trait for `Nid`

openssl/src/x509/mod.rs
```diff
@@ -405,6 +405,18 @@ impl X509Ref {
         unsafe { ffi::X509_issuer_name_hash(self.as_ptr()) as u32 }
     }
 
+    /// Returns the extension index for the nid, starting at lastpos, if it exists.
+    #[corresponds(X509_get_ext_by_NID)]
+    pub fn ext_by_nid(&self, nid: Nid, lastpos: c_int) -> Result<c_int, ErrorStack> {
```
sfackler, line 410: This seems like a very raw API. What's the use case?

alexanderjordanbaker: This is used as an alternative to x509_get_ext_d2i when you don’t need the actual extension value, just to see if it is present

sfackler: If there's desire to do something like this externally I think I'd like a cleaner API, and if this is only for a test I don't think it's worth adding to the public API.

alexanderjordanbaker: Are you looking for exposing x509_get_ext_d2i itself, so we just have one method to get an extension (which implicitly see's if it is present), or are you looking for this API to expose less of OpenSSL's internal structure in this API. I could modify this API to be higher level, and just return a boolean if the extension was detected, dropping lastpos

PR #1762, opened 2022-12-20
Fix broken relaxation

wiktor-k:
Hi Steven,

After a round of extensive integration testing and discussions on the [OpenSSL mailing list](https://mta.openssl.org/pipermail/openssl-users/2022-December/015727.html) it seems my "improvement" in #1733 was actually based on unsound promise. That is the `num` parameter does *not* reflect the internal cache size (it's always zero and used only internally).

I've basically reverted the old PR but kept the introduced function not to break the API (if you think it would be OK to remove `minimal_output_size` or `#[deprecate]` it I'll adjust). As for my client code I'll build a safe API around `*_unchecked` calls by keeping the additional cache state in my structs.

I'm super, super sorry for the annoyance I caused :bow: I should've spent more time integration testing my previous PR but now I've caused you more trouble than necessary. Once again: my deepest apologies :pray: !

### DIFF
--- a/openssl/src/cipher_ctx.rs
+++ b/openssl/src/cipher_ctx.rs
@@ -607,14 +607,34 @@ impl CipherCtxRef {
             assert!(output.len() >= block_size);
         }
 
+        unsafe { self.cipher_final_unchecked(output) }
+    }
+
+    /// Finalizes the encryption or decryption process.
+    ///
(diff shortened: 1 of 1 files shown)


sfackler:
minimal_output_size is only in one release, so I think we can just yank that release and remove the function.

PR #1770, opened 2022-12-27
Add some ssl api

iamwwc:


iamwwc:
lack some unsafe ffi, I will add them soon

iamwwc:
@sfackler I'm trying add `#cfg(any(osslxxx))` to every api. where I can find api min level requirement for openssl :)

sfackler:
I usually search through the header files in OpenSSL releases to find the first one it was added.

openssl/src/ssl/mod.rs
```diff
@@ -3114,6 +3112,142 @@ impl SslRef {
...
+                ssl_file_type.as_raw(),
+            ))?;
+        };
+        Ok(())
+    }
+    #[corresponds(SSL_use_PrivateKey)]
+    pub fn use_private_key_pem(&mut self, key: &[u8]) -> Result<(), ErrorStack> {
+        let pkey = PKey::private_key_from_pem(key)?;
+        unsafe {
+            // take ownership of pkey
```
Skepfyr, line 3143: `SSL_use_PrivateKey` does not take ownership of the pkey afaict, it appears to call up_ref, so we should drop pkey not forget it.

openssl/src/ssl/mod.rs
```diff
@@ -3114,6 +3112,142 @@ impl SslRef {
...
+            cvt(ffi::SSL_use_PrivateKey(self.as_ptr(), pkey.as_ptr()))?;
+        };
+        mem::forget(pkey);
+        Ok(())
+    }
+    #[corresponds(SSL_use_certificate)]
+    pub fn use_certificate_pem(&mut self, cert: &[u8]) -> Result<(), ErrorStack> {
+        let cert = X509::from_pem(cert)?;
+        unsafe {
+            // take ownership of cert
```
Skepfyr, line 3153: Similarly this doesn't appear to take ownership of the cert, it clones it.

openssl/src/ssl/mod.rs
```diff
@@ -3114,6 +3112,142 @@ impl SslRef {
...
+            cvt(ffi::SSL_use_certificate(self.as_ptr(), cert.as_ptr()))?;
+        };
+        mem::forget(cert);
+        Ok(())
+    }
+
+    #[cfg(ossl110)]
```
Skepfyr, line 3160: This can also gain `libressl332`.

openssl/src/ssl/mod.rs
```diff
@@ -3114,6 +3112,142 @@ impl SslRef {
...
+            cvt(ffi::SSL_use_PrivateKey_file(
+                self.as_ptr(),
+                key_file.as_ptr(),
+                ssl_file_type.as_raw(),
+            ))?;
+        };
+        Ok(())
+    }
+    #[corresponds(SSL_use_PrivateKey)]
+    pub fn use_private_key_pem(&mut self, key: &[u8]) -> Result<(), ErrorStack> {
```
Skepfyr, line 3140: Why does this function take a `&[u8]` rather than a `&PKeyRef` (and be called `set_private_key`)? That would be strictly more general.

openssl/src/ssl/mod.rs
```diff
@@ -3114,6 +3112,142 @@ impl SslRef {
...
+        let pkey = PKey::private_key_from_pem(key)?;
+        unsafe {
+            // take ownership of pkey
+            cvt(ffi::SSL_use_PrivateKey(self.as_ptr(), pkey.as_ptr()))?;
+        };
+        mem::forget(pkey);
+        Ok(())
+    }
+    #[corresponds(SSL_use_certificate)]
+    pub fn use_certificate_pem(&mut self, cert: &[u8]) -> Result<(), ErrorStack> {
```
Skepfyr, line 3150: Similarly, this function should take a `&X509Ref` and be called `set_certificate`

openssl/src/ssl/mod.rs
```diff
@@ -3114,6 +3112,142 @@ impl SslRef {
...
+                cipher_list.as_ptr() as *const _,
+            ))
+            .map(|_| ())
+        }
+    }
+    #[cfg(ossl111)]
+    #[cfg(not(libressl))]
+    #[cfg(not(boringssl))]
+    #[corresponds(SSL_set_cert_store)]
+    pub fn set_cert_store(&mut self, cert_store: X509Store) {
```
Skepfyr, line 3245: This function should be called `set_verify_cert_store` like [SslContextBuilder::set_verify_cert_store](https://docs.rs/openssl/latest/openssl/ssl/struct.SslContextBuilder.html#method.set_verify_cert_store) and should return a `Result<(), ErrorStack>`.

openssl/src/ssl/mod.rs
```diff
@@ -3114,6 +3112,142 @@ impl SslRef {
...
+                self.as_ptr(),
+                cipher_list.as_ptr() as *const _,
+            ))
+            .map(|_| ())
+        }
+    }
+    #[cfg(ossl111)]
+    #[cfg(not(libressl))]
+    #[cfg(not(boringssl))]
+    #[corresponds(SSL_set_cert_store)]
```
Skepfyr, line 3244: This should be:
```rust
    #[corresponds(SSL_set_cert_store)]
    #[cfg(ossl102)]
```
(There should be only one cfg for simplicity, `cfg(not(..))` doesn't do anything here, and reordered for consistency.)

openssl/src/ssl/test/mod.rs
```diff
@@ -1422,3 +1422,50 @@ fn add_chain_cert() {
     let mut ssl = Ssl::new(&ctx).unwrap();
     assert!(ssl.add_chain_cert(cert).is_ok());
 }
+#[test]
+#[cfg(any(ossl111))]
+#[cfg(not(libressl))]
+#[cfg(not(boringssl))]
```
Skepfyr, line 1428: This should just be:
```suggestion
#[cfg(ossl111)]
```
Also can you get away with `ossl102`? I can't see what 1.1.1 functions you're using.

openssl/src/ssl/mod.rs
```diff
@@ -3114,6 +3112,142 @@ impl SslRef {
         }
         Ok(())
     }
+    #[cfg(not(boringssl))]
+    pub fn set_method(&mut self, method: SslMethod) -> Result<(), ErrorStack> {
```
Skepfyr, line 3116: It would be good to document the functions you're adding, you can mostly get away with stealing the docs from [SslContextBuilder](https://docs.rs/openssl/latest/openssl/ssl/struct.SslContextBuilder.html) although they may need a couple of changes.

Skepfyr (changes requested):
Thanks for the PR!

openssl/src/ssl/mod.rs
iamwwc: But openssl still hold a reference to it https://github.com/openssl/openssl/blob/1708e3e85b4a86bae26860aa5d2913fc8eff6086/ssl/ssl_rsa.c#L169. drop `pkey` will cause use after free

openssl/src/ssl/mod.rs
iamwwc: I have change &[u8] to PKey<Private>, but I don't think we should pass as &PKeyRef

openssl/src/ssl/mod.rs
```diff
@@ -3114,6 +3112,142 @@ impl SslRef {
         }
         Ok(())
     }
+    #[cfg(not(boringssl))]
```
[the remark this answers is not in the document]
iamwwc: I have also add `#[cfg(not(boringssl))]` to ffi function

openssl-sys/src/handwritten/ssl.rs
```diff
@@ -515,6 +517,13 @@ extern "C" {
         ctx: *mut SSL_CTX,
         cert_chain_file: *const c_char,
     ) -> c_int;
+    pub fn SSL_use_PrivateKey_file(ssl: *mut SSL, file: *const c_char, type_: c_int) -> c_int;
+    pub fn SSL_use_PrivateKey(ssl: *mut SSL, pkey: *mut EVP_PKEY) -> c_int;
+    pub fn SSL_use_certificate(ssl: *mut SSL, x: *mut X509) -> c_int;
+    #[cfg(ossl110)]
+    pub fn SSL_use_certificate_chain_file(ssl: *mut SSL, file: *const c_char) -> c_int;
```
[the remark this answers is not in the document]
iamwwc: Added `libressl332` to both `SSL_use_certificate` and `SSL_use_certificate_chain_file`. But I can't find a result in https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.3.2-relnotes.txt for `SSL_use_PrivateKey`

iamwwc:
By the way, could you help me investgate why check fails on windows/cvpkg?

All tests have passed in linux platform but windows not

openssl/src/ssl/mod.rs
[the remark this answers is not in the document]
Skepfyr: Why does this function have that attribute? None of the other functions do, and I can't find anything boringssl specific about this method.

openssl/src/ssl/mod.rs
Skepfyr: `EVP_PKEY`s are reference counted, the line just above the one you linked is calling `up_ref` which increments that reference count. That means we still need to call `free` on our copy so that the reference count is decremented.

openssl/src/ssl/mod.rs
Skepfyr: Could you document the other functions too? (and put some newlines in?)

Skepfyr (changes requested):

openssl/src/ssl/mod.rs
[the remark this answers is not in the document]
iamwwc: boringssl don't have this method. without `#[cfg(not(boringssl))]`, it will cause CI failed

openssl/src/ssl/mod.rs
iamwwc: I got it! PKey<Private> 's drop call `EVP_PKEY_free` instead `libc::free`

iamwwc:
all tests passed with some workaround. it's ready for review again :) @Skepfyr

openssl/src/ssl/mod.rs
```diff
@@ -3114,6 +3112,178 @@ impl SslRef {
...
+    /// Sets the private key.
+    pub fn set_private_key(&mut self, pkey: PKey<Private>) -> Result<(), ErrorStack> {
+        unsafe {
+            cvt(ffi::SSL_use_PrivateKey(self.as_ptr(), pkey.as_ptr()))?;
+        };
+        Ok(())
+    }
+
+    #[corresponds(SSL_use_certificate)]
+    #[cfg(any(ossl110, libressl332))]
```
Skepfyr, line 3154: Given the above change I think you can delete this too as `SSL_use_certificate` exists in all versions of OpenSSL.

openssl/src/ssl/mod.rs
```diff
@@ -3114,6 +3112,178 @@ impl SslRef {
...
+    }
+
+    #[corresponds(SSL_use_certificate_chain_file)]
+    #[cfg(any(ossl110, libressl332))]
+    /// Loads a certificate chain from a file.
+    ///
+    /// The file should contain a sequence of PEM-formatted certificates, the first being the leaf
+    /// certificate, and the remainder forming the chain of certificates up to and including the
+    /// trusted root certificate.
+    pub fn set_certificate_chain_file<P: AsRef<Path>>(
```
Skepfyr, line 3170: Sorry for being picky, but these need to be the other way around, like:
```rust
    /// Loads a certificate chain from a file.
    ///
    /// The file should contain a sequence of PEM-formatted certificates, the first being the leaf
    /// certificate, and the remainder forming the chain of certificates up to and including the
    /// trusted root certificate.
    #[corresponds(SSL_use_certificate_chain_file)]
    #[cfg(any(ossl110, libressl332))]
```
This applies for all the methods you've added. This is partially just for consistency's sake but is also required to make the documentation look sensible because the `#[corresponds(..)]` attribute expands into a doc comment so needs to go below so that the documentation is in the right order.

openssl/src/ssl/mod.rs
```diff
@@ -3114,6 +3112,178 @@ impl SslRef {
...
+        unsafe {
+            cvt(ffi::SSL_use_PrivateKey(self.as_ptr(), pkey.as_ptr()))?;
+        };
+        Ok(())
+    }
+
+    #[corresponds(SSL_use_certificate)]
+    #[cfg(any(ossl110, libressl332))]
+    /// Sets the certificate
+    pub fn set_certificate_pem(&mut self, cert: X509) -> Result<(), ErrorStack> {
```
Skepfyr, line 3156: This can now take a `&X509Ref`

openssl/src/ssl/mod.rs
```diff
@@ -3114,6 +3112,178 @@ impl SslRef {
...
+                key_file.as_ptr(),
+                ssl_file_type.as_raw(),
+            ))?;
+        };
+        Ok(())
+    }
+
+    #[corresponds(SSL_use_PrivateKey)]
+    /// Sets the private key.
+    pub fn set_private_key(&mut self, pkey: PKey<Private>) -> Result<(), ErrorStack> {
```
Skepfyr, line 3146: This can now take a `&PKeyRef<Private>`.

openssl/src/ssl/test/mod.rs
```diff
@@ -1422,3 +1422,58 @@ fn add_chain_cert() {
...
+    ssl.set_private_key(pkey).unwrap();
+    {
+        let pkey = String::from_utf8(
+            ssl.private_key()
+                .unwrap()
+                .private_key_to_pem_pkcs8()
+                .unwrap(),
+        )
+        .unwrap();
+        assert!(pkey.lines().eq(key_str.lines()));
```
Skepfyr, line 1456: This is a great solution!

Skepfyr (changes requested):
Sorry for all the back-and-forth. OpenSSL is a pain with versioning and there are still some bits of this library I don't understand. I should be able to merge this once you've addressed these final comments.

Skepfyr (approved):

Skepfyr:
Thanks!

PR #1776, opened 2023-01-04
Make load_cert_file return an i32 like load_crl_file

alexanderjordanbaker:


alexanderjordanbaker:
Since this was returning `()` before, this should not affect the functioning of this method

sfackler:
> Since this was returning `()` before, this should not affect the functioning of this method

That's not true. You can write code that only compiles if this method returns `()`.

PR #1778, opened 2023-01-07
brew: prefer to install openssl@3

chenrui333:
Since both `openssl@3` and `openssl@1.1` are supported, it would be good to recommend to use `openssl@3`.

https://github.com/sfackler/rust-openssl/blob/8621407a3268071e30312dfe40098b574554ea1f/openssl-sys/build/find_normal.rs#L36

### DIFF
--- a/openssl/src/lib.rs
+++ b/openssl/src/lib.rs
@@ -29,7 +29,7 @@
 //!
 //! ```not_rust
 //! # macOS (Homebrew)
-//! $ brew install openssl@1.1
+//! $ brew install openssl@3
 //!
 //! # macOS (MacPorts)
 //! $ sudo port install openssl


sfackler:
In my experience OpenSSL 3 is strictly worse than 1.1 - what's the rationale for wanting to switch this?

chenrui333:
> In my experience OpenSSL 3 is strictly worse than 1.1 - what's the rationale for wanting to switch this?

Can you elaborate?

For us:
- We are trying to have most of the formulae building with openssl@3 as possible, https://github.com/Homebrew/homebrew-core/labels/openssl-3-migration
- Also for the other package managers in the docs, the openssl@3 is one get recommended.

sfackler:
If it's part of a broader Homebrew migration then that seems fine.

For reference, some 3.0.0 issues are linked here: https://github.com/sfackler/rust-openssl/pull/1578#issuecomment-992760346

chenrui333:
> If it's part of a broader Homebrew migration then that seems fine.
> 
> For reference, some 3.0.0 issues are linked here: [#1578 (comment)](https://github.com/sfackler/rust-openssl/pull/1578#issuecomment-992760346)

that is good to know, thanks for the note.

PR #1785, opened 2023-01-11
openssl: Add AES_decrypt and AES_encrypt bindings

anayw2001:
As unfortunate as exposing these bindings is, we do unfortunately need them.


openssl/src/aes.rs
```diff
@@ -258,6 +258,36 @@ pub fn unwrap_key(
     }
 }
 
+/// Encrypts an aes-128/256 block from `in` to `out`.
+/// * `key`: The key-encrypting-key to decrypt the wrapped key. Must be a decrypting key
+/// * `out`: The buffer to write the plaintext to
+/// * `in_`: The input plaintext
+#[corresponds(AES_encrypt)]
+pub fn aes_encrypt(key: &AesKey, out: &mut [u8], in_: &[u8]) {
```
sfackler, line 266: How big do these buffers need to be? What happens if they're not big enough?

openssl/src/aes.rs
```diff
@@ -258,6 +258,36 @@ pub fn unwrap_key(
     }
 }
 
+/// Encrypts an aes-128/256 block from `in` to `out`.
+/// * `key`: The key-encrypting-key to decrypt the wrapped key. Must be a decrypting key
```
sfackler, line 262: key-encrypting-key? wrapped key? decrypting key?

openssl/src/aes.rs
```diff
@@ -258,6 +258,36 @@ pub fn unwrap_key(
...
+/// * `out`: The buffer to write the plaintext to
+/// * `in_`: The input plaintext
```
sfackler, line 264: plaintext?

openssl/src/aes.rs
sfackler: These docs should more generally mirror the format used in the rest of the crate.

openssl/src/aes.rs
```diff
@@ -258,6 +258,36 @@ pub fn unwrap_key(
...
+        );
+    }
+}
+
+/// Decrypts an aes-128/256 block from `in` to `out`.
+/// * `key`: The key-encrypting-key to decrypt the wrapped key. Must be a decrypting key
+/// * `out`: The buffer to write the plaintext to
+/// * `in_`: The input plaintext
+#[corresponds(AES_decrypt)]
+pub fn aes_decrypt(key: &AesKey, out: &mut [u8], in_: &[u8]) {
```
sfackler, line 281: See above.

sfackler:
This needs tests.

What is your use case for needing these? Why is the `CipherCtx` type not sufficient?

openssl/src/aes.rs
anayw2001: sorry, copy pasted docs, fixed now.

openssl/src/aes.rs
anayw2001: fixed

anayw2001:
We're using AES as the underlying block cipher in a length doubler, which uses the block cipher as a secure PRP for exactly one block.

openssl/src/aes.rs
anayw2001: specified via types

openssl/src/aes.rs
```diff
@@ -258,6 +258,47 @@ pub fn unwrap_key(
     }
 }
 
+/// The standard AES block
+type AesBlock = [u8; ffi::AES_BLOCK_SIZE as usize];
```
sfackler, line 262: `pub`

sfackler:
Does using AES-ECB with the existing APIs not cover that?

PR #1786, opened 2023-01-12
Add check for OPENSSL_NO_CAST

wiktor-k:
This check detects at build time if CAST5 algorithm has been disabled in current OpenSSL library build. The constant name has been checked with OpenSSL: https://github.com/openssl/openssl/blob/5adca946c3b6f779eb593bff6dbefe0a85238d84/crypto/evp/c_allc.c#L116

See: https://github.com/sfackler/rust-openssl/pull/1717#issuecomment-1379474589

Note that since all environments I'm using do have CAST5 I don't have a way to test this but I hope @LeoRowan can help here :)


LeoRowan:
Thanks for this! Going to try this out today and I'll update with what we find

sfackler:
This looks good to me, but I'll hold off merging until @LeoRowan confirms it fixes things.

LeoRowan:
@sfackler Hey, we were a little delayed, but just confirmed that it's working for us! Thanks so much :D

sfackler:
Great!

PR #1787, opened 2023-01-12
X509 store crl

steffen-eiden:
includes commits from #1123 which needs to be merged in beforehand.
Therefore, marked as _Draft_

Commits from that PR included, as this PR extends the CRL handling regarding the  introduced there


steffen-eiden:
`x509: refractoring` and `Add CRL related bindings for X509_STORE` are the relevant commits for this PR

PR #1788, opened 2023-01-13
X509_PURPOSE

bkstein:
- `X509_PURPOSE` id's  were implemented as bitfields, but they are just constants.
- Renamed `X509PurposeFlags` to `X509PurposeId`, as this seems to be the naming used in OpenSSL.
- Moved the definitions from x509/verify.rs to x509/mod.rs, as `X509_PURPOSE` is not specific to verification (`X509_STORE` also uses this).
- This PR is also part 1/4 of the split PR #1598, which grew too big over time.

### DIFF
--- a/openssl-sys/src/x509v3.rs
+++ b/openssl-sys/src/x509v3.rs
@@ -58,25 +58,15 @@ pub const EXFLAG_FRESHEST: u32 = 0x1000;
 #[cfg(any(ossl102, libressl261))]
 pub const EXFLAG_SS: u32 = 0x2000;
 
-#[cfg(not(boringssl))]
 pub const X509v3_KU_DIGITAL_SIGNATURE: u32 = 0x0080;
-#[cfg(not(boringssl))]
 pub const X509v3_KU_NON_REPUDIATION: u32 = 0x0040;
-#[cfg(not(boringssl))]
 pub const X509v3_KU_KEY_ENCIPHERMENT: u32 = 0x0020;
-#[cfg(not(boringssl))]
 pub const X509v3_KU_DATA_ENCIPHERMENT: u32 = 0x0010;
-#[cfg(not(boringssl))]
 pub const X509v3_KU_KEY_AGREEMENT: u32 = 0x0008;
-#[cfg(not(boringssl))]
 pub const X509v3_KU_KEY_CERT_SIGN: u32 = 0x0004;
-#[cfg(not(boringssl))]
 pub const X509v3_KU_CRL_SIGN: u32 = 0x0002;
-#[cfg(not(boringssl))]
 pub const X509v3_KU_ENCIPHER_ONLY: u32 = 0x0001;
-#[cfg(not(boringssl))]
 pub const X509v3_KU_DECIPHER_ONLY: u32 = 0x8000;
-#[cfg(not(boringssl))]
 pub const X509v3_KU_UNDEF: u32 = 0xffff;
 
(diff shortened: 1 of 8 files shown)


openssl-sys/src/x509v3.rs
```diff
@@ -58,15 +58,25 @@ pub const EXFLAG_FRESHEST: u32 = 0x1000;
 #[cfg(any(ossl102, libressl261))]
 pub const EXFLAG_SS: u32 = 0x2000;
 
+#[cfg(not(boringssl))]
```
sfackler, line 61: We don't use the openssl-sys bindings with boringssl.

openssl/src/x509/mod.rs
```diff
@@ -1737,3 +1739,98 @@ cfg_if! {
         }
     }
 }
+
+pub struct X509PurposeId(i32);
```
sfackler, line 1743: `#[derive(Copy, Clone, PartialEq, Eq)]`

openssl/src/x509/mod.rs
```diff
@@ -1737,3 +1739,98 @@ cfg_if! {
...
+    }
+}
+
+impl From<i32> for X509PurposeId {
+    fn from(id: i32) -> Self {
+        X509PurposeId(id)
+    }
+}
+
+/// fake free method, since X509_PURPOSE is static
```
sfackler, line 1767: If that's the case then I think you don't want an X509Purpose type at all, just X509PurposeRef which you can hand-write.

openssl/src/x509/mod.rs
```diff
@@ -1737,3 +1739,98 @@ cfg_if! {
...
+    ///  - "nssslserver",
+    ///  - "smimesign",
+    ///  - "smimeencrypt",
+    ///  - "crlsign",
+    ///  - "any",
+    ///  - "ocsphelper",
+    ///  - "timestampsign"
+    /// The index can be used with `X509Purpose::from_idx()` to get the purpose.
+    #[allow(clippy::unnecessary_cast)]
+    pub fn get_by_sname(sname: &str) -> Result<i32, ErrorStack> {
```
sfackler, line 1794: Why is this defined on X509Purpose if it doesn't involve that type at all?

openssl/src/x509/store.rs
```diff
@@ -125,6 +126,19 @@ impl X509StoreBuilderRef {
...
     }
 
+    /// Sets the certificate purpose.
+    /// The purpose value can be obtained by `X509Purpose::get_by_sname()`
+    #[corresponds(X509_STORE_set_purpose)]
+    pub fn set_purpose(&mut self, purpose: X509PurposeId) -> Result<(), ErrorStack> {
+        unsafe {
+            cvt(ffi::X509_STORE_set_purpose(
+                self.as_ptr(),
+                purpose.value() as c_int,
```
sfackler, line 136: If you have X509PurposeId store a c_int you can avoid this conversion.

openssl-sys/src/x509v3.rs
bkstein: Removed the bindings.

openssl/src/x509/mod.rs
bkstein: Done.

openssl/src/x509/mod.rs
bkstein: Done.

openssl/src/x509/store.rs
bkstein: Done.

PR #1790, opened 2023-01-13
X509 and X509Req extensions

bkstein:
- This is part 3/4 of the split #1598.
- It contains implementations for creating and reading X509 certs and requests. This handles X509 extensions and attributes.
- #1788 and #1789 are prerequisites for this PR.

### DIFF
--- a/openssl/src/symm.rs
+++ b/openssl/src/symm.rs
@@ -1248,6 +1248,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(not(ossl300))]
     fn test_des_ede3_cbc() {
         let pt = "54686973206973206120746573742e";
         let ct = "6f2867cfefda048a4046ef7e556c7132";
--- a/openssl/src/x509/mod.rs
+++ b/openssl/src/x509/mod.rs
@@ -9,7 +9,7 @@
 
 use cfg_if::cfg_if;
 use foreign_types::{ForeignType, ForeignTypeRef};
-use libc::{c_int, c_long, c_uint};
+use libc::{c_char, c_int, c_long, c_uchar, c_uint, c_void};
 use std::cmp::{self, Ordering};
 use std::error::Error;
 use std::ffi::{CStr, CString};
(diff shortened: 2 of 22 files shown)


openssl/src/symm.rs
```diff
@@ -1248,6 +1248,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(not(ossl300))]
```
Skepfyr, line 1251: It seems surprising to me that this PR needs to add this, why is it not affecting everything else?

openssl/src/x509/mod.rs
```diff
@@ -467,6 +476,34 @@ impl X509Ref {
         }
     }
 
+    /// Returns this certificate's key usage extension value, if it exists.
+    #[corresponds(X509_get_ext_d2i)]
+    pub fn key_usage(&self) -> Option<&Asn1BitStringRef> {
+        unsafe {
+            let ptr = ffi::X509_get_ext_d2i(
```
Skepfyr, line 483: For consistency with the OpenSSL API we're now implementing these APIs with a single `extension` function, like this: https://docs.rs/openssl/latest/openssl/x509/struct.X509Revoked.html#method.extension

openssl/src/x509/mod.rs
```diff
@@ -791,6 +905,75 @@ impl PartialEq<X509Ref> for X509 {
...
+        unsafe {
+            let object_ptr = cvt_p(ffi::X509_ATTRIBUTE_get0_object(self.as_ptr()))?;
+            Ok(Asn1ObjectRef::from_ptr(object_ptr))
+        }
+    }
+
+    /// Get an item from the list of `ASN1_TYPE`s.
+    pub fn typ(&self, idx: isize) -> Result<&Asn1TypeRef, ErrorStack> {
+        unsafe {
+            let type_ptr = cvt_p(ffi::X509_ATTRIBUTE_get0_type(self.as_ptr(), idx as c_int))?;
```
Skepfyr, line 971: This must test the length of the stack first so it doesn't do an out-of-bounds read.

openssl/src/x509/mod.rs
```diff
@@ -1156,6 +1402,14 @@ impl X509NameRef {
         to_der,
         ffi::i2d_X509_NAME
     }
+
+    /// Compares the X509NameRef with an`other` X509NameRef
+    /// Returns 0 if equal.
+    #[corresponds(X509_NAME_cmp)]
+    #[allow(clippy::unnecessary_cast)]
+    pub fn cmp_with(&self, other: &X509NameRef) -> i32 {
```
Skepfyr, line 1410: This already (now?) exists as `try_cmp`.

openssl/src/x509/mod.rs
```diff
@@ -1341,6 +1595,38 @@ impl X509ReqBuilder {
...
+                name.as_ptr() as *const c_char,
+                ffi::MBSTRING_UTF8,
+                value.as_ptr() as *const c_uchar,
+                len as c_int,
+            ))
+            .map(|_| ())
+        }
+    }
+
+    pub fn add_attribute_by_nid(&mut self, nid: Nid, value: &str) -> Result<(), ErrorStack> {
```
Skepfyr, line 1615: This pair of functions worries me, are we going to need wrappers for each type of value? Should they actually be taking an Asn1Type or Asn1String (I'm not familiar with attributes).

Skepfyr (changes requested):

PR #1795, opened 2023-01-20
Add X509Name::to_owned()

AndrewScull:
The X509_NAME_dup() function can fail but that isn't compatible with the ToOwned trait. Follow the pattern used in BigNum to add a custom, fallible to_owned() function.

### DIFF
--- a/openssl/src/x509/mod.rs
+++ b/openssl/src/x509/mod.rs
@@ -1064,6 +1064,13 @@ impl X509NameRef {
         Ok(cmp.cmp(&0))
     }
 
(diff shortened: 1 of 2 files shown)


sfackler:
It looks like you also need to add X509_NAME_dup to the openssl-sys bindings.

AndrewScull:
Thanks for the quick reply! I found it was already in the openssl-sys bindings but only for certain versions of the library so I've copied those restrictions onto the new function and test.

sfackler:
Thanks!

PR #1799, opened 2023-01-20
Fix debug formatting of ipaddress for GeneralName

### LINKED ISSUE #1798: Debug implementation of GeneralName does not correctly format IP addresses
The implementation of `Debug` for `GeneralName` uses `String::from_utf8_lossy`, but the bytes represent a raw IP address and may not be valid UTF-8, and even if they are, the resulting string is not helpful.

This may also mislead readers of the documentation as the `Debug` implementation is shown as an "Example found in repository". Readers may incorrectly conclude that the bytes returned by `ipaddress` are a string representation of an IP address.

20230120_16h57m19s_grim

stephaneyfx:
`from_utf8_lossy` is not appropriate as the bytes are the raw IP address (e.g. 4 bytes for IPv4).

Resolves #1798


sfackler:
Good catch, thanks!

PR #1806, opened 2023-02-02
fixes #1768 -- integrate boringssl into the build process more naturally

### LINKED ISSUE #1768: Detect boringssl in expando.c?
Right now if you compile against BoringSSL (i.e., point `OPENSSL_DIR` at it), but don't specify `unstable_boringssl` it will silently build, but many things won't work (e.g. https://github.com/pyca/cryptography/pull/7933#issuecomment-1364703766).

At a minimum, would it make sense to detect this situation with `expando.c` and fail the build, rather than silently proceeding?

Even better would be if it could detect this _and work_ -- perhaps by BoringSSL putting the generated `bssl-sys` crate somewhere that could be detected.

cc: @davidben, @maurer, @benbrittain, @durin42

alex:
This PR uses the in-development bindgen support for static inline functions (https://github.com/rust-lang/rust-bindgen/pull/2335) + an in-development boringssl patch (https://boringssl-review.googlesource.com/c/boringssl/+/56505) to allow using boringssl with rust-openssl without needing a .cargo/config override

This is not mergable as-is (requires a bindgen release + boringssl to merge my patch), putting this up for visibility only.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -311,24 +311,14 @@ jobs:
               make install_sw
               ;;
             "boringssl")
-              sed -i rust/CMakeLists.txt -e '1s%^%include_directories(../include)\n%'
-              cpu=`echo ${{ matrix.target }} | cut -d - -f 1`
-              echo "set(CMAKE_SYSTEM_NAME Linux)" > toolchain.cmake
-              echo "set(CMAKE_SYSTEM_PROCESSOR $cpu)" >> toolchain.cmake
-              echo "set(triple ${{ matrix.target }})" >> toolchain.cmake
-              echo 'set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} '$OS_FLAGS '" CACHE STRING "c++ flags")' >> toolchain.cmake
-              echo 'set(CMAKE_C_FLAGS   "${CMAKE_C_FLAGS}   '$OS_FLAGS '" CACHE STRING "c flags")' >> toolchain.cmake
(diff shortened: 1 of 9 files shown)


alex:
One substantive issue I see:

BoringSSL uses `core::ffi` for its types, which requires the 2018 edition for `-sys`. The existing `-sys` crate has `use *;` which is not compatible with 2018. I suppose the solution is to just replace all the `use *;`? Any blocker to that?

sfackler:
Should be totally fine to update -sys to the 2018 edition and just change the import to `use ::*;` I think?

alex:
Ok! I'll endeavour to do a stand-alone PR for that.

sfackler:
Our MSRV is at 1.56.0 so we can probably go all the way to the 2021 edition really.

alex:
pyca/cryptography still has a lower MSRV (1.48 ATM), so I want to avoid creating a hard dep on 2021.

sfackler:
Very much in favor of cleaning up the bssl-sys setup, but obviously I'll defer to @davidben and @maurer here.

sfackler:
If you're depending on a 1.48 MSRV then we should probably change the existing CI check (or add a new one?) to enforce that.

pvdrz:
This feature is now available on bindgen `0.64.0` :tada:

alex:
🎉 🎉 🎉 -- as soon as it's available in brew (https://github.com/Homebrew/homebrew-core/pull/122575) I plan to rebase my boringssl CL and hopefully advance the conversation there.

PR #1808, opened 2023-02-03
Re-lower the MSRV to 1.48

alex:


alex:
Clippy failures are addressed by https://github.com/sfackler/rust-openssl/pull/1807

alex:
The no-op bssl-sys crate from crates.io has a 2021 edition, I've reached out to the maintainer to see about lowering that.

.github/workflows/ci.yml
```diff
@@ -57,11 +57,16 @@ jobs:
   min-version:
     name: min-version
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # Current MSRV and expected next MSRV
+        RUST: ["1.48.0", "1.56.0"]
```
sfackler, line 64: Is 1.56 actually an expected next MSRV? that was IIRC just the one that one of those dependency minor versions required. I don't think it was any more intentional than that.

alex: It's pyca/cryptography's next expected MSRV (based on our usage numbers). So we'd be happy with that reality :-)

sfackler: Fair enough!

PR #1809, opened 2023-02-03
Add a variant of add_extra_chain_cert for X509Ref

kgraney:
There's not reason this function needs to take ownership of the X509 certificate, and it's more consistent with set_certificate to accept instead a &X509Ref.


sfackler:
This is unsound. SSL_CTX_add_extra_chain_cert takes ownership over the certificate passed to it.

kgraney:
Whoops, you're right!  I totally missed that sentence in the manpage.

PR #1814, opened 2023-02-08
Proper homebrew prefix detection

### LINKED ISSUE #1813: openssl-sys | mac | Doesn't respect `brew --prefix`
# Expected

It should just work.

# Encountered

Tries to link to unrelated brew environment with a different architecture and fails to compile. Several utility crates are affected.

# Use case

There is an installation at `/opt/homebrew/` for a different architecture that absolutely cannot be moved. Cannot do magic environment variables workarounds because this is fragile and not scalable.

# Solution

Trying random "well-known" locations particular to tools is fragile, breaks repeatability, separations of concerns, and isolation of builds and dev environments.

Instead of guessing at locations, if the `brew` command exists, ask it for the source of truth `brew --prefix openssl@{{n}}`.

steakknife:
Resolves #1813

### DIFF
--- a/openssl-sys/build/find_normal.rs
+++ b/openssl-sys/build/find_normal.rs
@@ -31,18 +31,8 @@ pub fn get_openssl(target: &str) -> (Vec<PathBuf>, PathBuf) {
     }
 }
 
-fn resolve_with_wellknown_homebrew_location(dir: &str) -> Option<PathBuf> {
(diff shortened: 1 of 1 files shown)


steakknife:
And if `brew` isn't in the `PATH`, it also shouldn't go tinkering around the system and making guesses.

PR #1820, opened 2023-02-10
Expand documentation on PkeyCtxRef's HKDF APIs

### LINKED ISSUE #1818: HKDF API documentation lacks warnings about unsafe APIs
It looks like, for HKDF, this crate chose to just copy over OpenSSL's EVP_PKEY-based HKDF API without taking the underlying subject matter into account.

HKDF is defined in [RFC 5869](https://www.rfc-editor.org/rfc/rfc5869). It is two _separate_ functions, HKDF-Extract and HKDF-Expand. These functions and their inputs are _not_ interchangeable. That is, you cannot take an HKDF-Extract IKM and then also use it as an HKDF-Expand PRK.

Due to various design quirks (and IMO mistakes) that have no reason to apply to rust-openssl, OpenSSL's API has a notion of an "HKDF key" and then a separate "HKDF mode". This leads one to believe that there is such a thing as an HKDF key, and that one can use multiple modes with it. rust-openssl has carried over that mistake:

"Sets the input keying material for HKDF generation."
https://docs.rs/openssl/latest/openssl/pkey_ctx/struct.PkeyCtxRef.html#method.set_hkdf_key

This is wrong. When the mode is EXTRACT_AND_EXPAND or EXTRACT_ONLY, this "HKDF key" is an HKDF-Extract IKM. When it is EXPAND_ONLY, this "HKDF key" is an HKDF-Expand PRK. These are not interchangeable. When we implemented this incorrect API in BoringSSL, for compatibility with legacy software, we made sure to document these pitfalls. 
https://commondatastorage.googleapis.com/chromium-boringssl-docs/kdf.h.html#EVP_PKEY_CTX_hkdf_mode
https://commondatastorage.googleapis.com/chromium-boringssl-docs/kdf.h.html#EVP_PKEY_CTX_set1_hkdf_key

OpenSSL also documents this to some degree, though not as clearly.
https://www.openssl.org/docs/man3.0/man7/EVP_KDF-HKDF.html

rust-openssl does neither. These pitfalls should be documented in rust-openssl as well, and then this unsafe API replaced with an API that better reflects the underlying subject matter.

mauricelam:
Fixes #1818

### DIFF
--- a/openssl/src/pkey_ctx.rs
+++ b/openssl/src/pkey_ctx.rs
@@ -84,8 +84,23 @@ pub struct HkdfMode(c_int);
 
 #[cfg(ossl111)]
 impl HkdfMode {
+    /// This is the default mode. Calling [`derive`][PkeyCtxRef::derive] on a [`PkeyCtxRef`] set up
+    /// for HKDF will perform an extract followed by an expand operation in one go. The derived key
+    /// returned will be the result after the expand operation. The intermediate fixed-length
+    /// pseudorandom key K is not returned.
     pub const EXTRACT_THEN_EXPAND: Self = HkdfMode(ffi::EVP_PKEY_HKDEF_MODE_EXTRACT_AND_EXPAND);
+
+    /// In this mode calling [`derive`][PkeyCtxRef::derive] will just perform the extract operation.
+    /// The value returned will be the intermediate fixed-length pseudorandom key K.
+    ///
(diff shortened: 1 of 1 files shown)


Skepfyr:
Thanks, this looks great. Even though there have been a bunch of test failures, it's clear these doc changes didn't cause them so I'm going to merge it anyway and go and fix the failures separately.

PR #1823, opened 2023-02-15
Add SSL_CTX_set_num_tickets and friends

canarysnort01:
These are required to disable session tickets on TLS 1.3 connections.

Note that libressl added these functions in 3.5.2, but they are for compatibility only and have no effect, so I didn't include these definitions for libressl. Let me know if I should add support them too.

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -1687,6 +1687,16 @@ impl SslContextBuilder {
         }
     }
 
+    /// Sets the number of TLS 1.3 session tickets that will be sent to a client after a full
+    /// handshake.
+    ///
+    /// Requires OpenSSL 1.1.1 or newer.
+    #[corresponds(SSL_CTX_set_num_tickets)]
+    #[cfg(ossl111)]
+    pub fn set_num_tickets(&mut self, num_tickets: usize) -> Result<(), ErrorStack> {
+        unsafe { cvt(ffi::SSL_CTX_set_num_tickets(self.as_ptr(), num_tickets)).map(|_| ()) }
+    }
+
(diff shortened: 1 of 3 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -1880,6 +1890,16 @@ impl SslContextRef {
...
         SslVerifyMode::from_bits(mode).expect("SSL_CTX_get_verify_mode returned invalid mode")
     }
+
+    /// Gets the number of TLS 1.3 session tickets that will be sent to a client after a full
+    /// handshake.
+    ///
+    /// Requires OpenSSL 1.1.1 or newer.
+    #[corresponds(SSL_CTX_get_num_tickets)]
+    #[cfg(ossl111)]
+    pub fn get_num_tickets(&self) -> usize {
```
sfackler, line 1900: Getters in Rust don't usually use the `get_` prefix, so this would just be `pub fn num_tickets`.

openssl/src/ssl/mod.rs
```diff
@@ -3283,6 +3303,26 @@ impl SslRef {
...
+        unsafe { cvt(ffi::SSL_set_num_tickets(self.as_ptr(), num_tickets)).map(|_| ()) }
+    }
+
+    /// Gets the number of TLS 1.3 session tickets that will be sent to a client after a full
+    /// handshake.
+    ///
+    /// Requires OpenSSL 1.1.1 or newer.
+    #[corresponds(SSL_get_num_tickets)]
+    #[cfg(ossl111)]
+    pub fn get_num_tickets(&self) -> usize {
```
sfackler, line 3323: Same here

sfackler:
LGTM other than the naming nits!

PR #1829, opened 2023-02-28
Add DTLS 1.2 support in newer releases of SSL libs.

dbodden-pesa:
Both SSL 1.1/3.0 and LibreSSL 3.3.2 and up support the DTLS 1.2 methods.

### DIFF
--- a/openssl-sys/src/handwritten/ssl.rs
+++ b/openssl-sys/src/handwritten/ssl.rs
@@ -712,8 +712,6 @@ extern "C" {
     pub fn DTLSv1_2_method() -> *const SSL_METHOD;
 }
 
-
-
 extern "C" {
     pub fn SSL_get_error(ssl: *const SSL, ret: c_int) -> c_int;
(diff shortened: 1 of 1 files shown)


openssl-sys/src/handwritten/ssl.rs
```diff
@@ -679,6 +679,10 @@ cfg_if! {
             pub fn TLS_server_method() -> *const SSL_METHOD;
 
             pub fn TLS_client_method() -> *const SSL_METHOD;
+
+            // DTLS 1.2 support doesn't exist in LibresSSL 2.9.1
+            #[cfg(ossl110)]
+            pub fn DTLSv1_2_method() -> *const SSL_METHOD;
```
sfackler, line 685: Given that this symbol's cfg doesn't align with the rest of the cfg_if block, I'd just move it out to the non-cfg-if block below

dbodden-pesa: Done - plese let me know if that's acceptable.

PR #1831, opened 2023-03-04
Added support for building boringssl with bindgen

alex:
This allows building it without the bssl-sys crate. This is an alternative approach to fixing #1768 (in contrast to #1806).

This maintains support for using the bssl-sys crate.

### DIFF
--- a/openssl-sys/build/run_bindgen.rs
+++ b/openssl-sys/build/run_bindgen.rs
@@ -1,13 +1,17 @@
+#[cfg(feature = "bindgen")]
 use bindgen::callbacks::{MacroParsingBehavior, ParseCallbacks};
-use bindgen::RustTarget;
-use std::env;
+#[cfg(feature = "bindgen")]
+use bindgen::{MacroTypeVariation, RustTarget};
+use std::io::Write;
 use std::path::PathBuf;
+#[cfg(not(feature = "bindgen"))]
+use std::process;
+use std::{env, fs};
 
 const INCLUDES: &str = "
 #include <openssl/aes.h>
 #include <openssl/asn1.h>
 #include <openssl/bio.h>
-#include <openssl/comp.h>
 #include <openssl/conf.h>
 #include <openssl/crypto.h>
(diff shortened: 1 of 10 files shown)


alex:
FYI: @davidben @maurer

openssl-sys/build/run_bindgen.rs
```diff
@@ -94,9 +103,107 @@ pub fn run(include_dirs: &[PathBuf]) {
...
+            &out_dir
+                .join("boring_static_wrapper.h")
+                .display()
+                .to_string(),
+        )
+        .compile("boring_static_wrapper");
+}
+
+#[cfg(not(feature = "bindgen"))]
+pub fn run_boringssl(include_dirs: &[PathBuf]) {
```
sfackler, line 154: It seems a bit unfortunate to be duplicating all of the bindgen flags here and above. If we end up having to tweak them in the future we should probably refactor.

alex: Agreed, and yes, makes sense.

openssl-sys/build/run_bindgen.rs
```diff
@@ -94,9 +103,107 @@ pub fn run(include_dirs: &[PathBuf]) {
...
+        .arg("--rust-target=1.47")
+        .arg("--ctypes-prefix=::libc")
+        .arg("--no-derive-default")
+        .arg("--enable-function-attribute-detection")
+        .arg("--size_t-is-usize")
+        .arg("--default-macro-constant-type=signed")
+        .arg("--rustified-enum=point_conversion_form_t")
+        .arg("--allowlist-file=.*/openssl/[^/]+\\.h")
+        .arg("--experimental")
+        .arg("--wrap-static-fns")
```
davidben, line 175: Based on, it doesn't look like this flag works in bindgen yet, so from BoringSSL's perspective, this bindings strategy is unsupported for now.
https://boringssl-review.googlesource.com/c/boringssl/+/56505/8/rust/CMakeLists.txt#37

PR #1832, opened 2023-03-04
bump actions/cache version

alex:
this is needed to deal with a GHA deprecation


sfackler:
It looks like we also need to bump checkout to v3 and migrate some set-outputs: https://github.com/sfackler/rust-openssl/actions/runs/4331365034

sfackler:
Thanks!

PR #1835, opened 2023-03-08
Add new feature maybe-vendored to openssl & openssl-sys

NobodyXu:
so that crates depending on openssl or openssl-sys can build successfully even if their system does not have openssl installed.

The motivation for me to submit this PR is due to
https://github.com/cargo-bins/cargo-quickinstall/issues/171

### DIFF
--- a/openssl-sys/Cargo.toml
+++ b/openssl-sys/Cargo.toml
@@ -15,8 +15,6 @@ build = "build/main.rs"
 edition = "2018"
 
 [features]
-default = ["maybe-vendored"]
-
 maybe-vendored = ["openssl-src"]
 vendored = ['openssl-src']
 unstable_boringssl = ['bssl-sys']
--- a/openssl/Cargo.toml
+++ b/openssl/Cargo.toml
@@ -21,7 +21,7 @@ vendored = ['ffi/vendored']
 maybe-vendored = ["ffi/maybe-vendored"]
 bindgen = ['ffi/bindgen']
 unstable_boringssl = ["ffi/unstable_boringssl"]
-default = ["maybe-vendored"]
+default = []
 
 [dependencies]
 bitflags = "1.0"


NobodyXu:
The msrv failure is from openssl-src:
 - The stdlib feature `command_access` is stablised in 1.57
 - impl<S: [Borrow](https://doc.rust-lang.org/std/borrow/trait.Borrow.html)<[OsStr](https://doc.rust-lang.org/std/ffi/struct.OsStr.html)>> [Join](https://doc.rust-lang.org/std/slice/trait.Join.html)<&[OsStr](https://doc.rust-lang.org/std/ffi/struct.OsStr.html)> for [[S]](https://doc.rust-lang.org/std/primitive.slice.html) is implemented on 1.63

NobodyXu:
> [`impl<S: Borrow<OsStr>> Join<&OsStr> for [S]`](https://doc.rust-lang.org/std/ffi/struct.OsStr.html#impl-Join%3C%26OsStr%3E-for-%5BS%5D) is implemented on 1.63

I can open a PR in [`openssh-src-rs`](https://github.com/alexcrichton/openssl-src-rs) to remove use of this API so that this PR has a msrv of 1.57

sfackler:
People should be explicitly making a choice about where their OpenSSL is coming from. I don't really want things to "work" but secretly via a vendored OpenSSL because they had a typo in their environment setup or whatever.

NobodyXu:
> People should be explicitly making a choice about where their OpenSSL is coming from.

I wish I can make that choice, but unfortunately when building binary crates, they don't necessary expose that as a feature and in cargo-quickinstall where we build many third-party crates it is even harder to do, though not impossible.

> I don't really want things to "work" but secretly via a vendored OpenSSL

Other `*-sys` crates, such as `zstd-sys`, `lzma-sys`, `bzip2-sys`, `libz-sys`, `libzng-sys`, they all fallback to vendored version when they fail to find a pre-installed version on the system.

> because they had a typo in their environment setup or whatever.

For cross-compilation using `cross`/`cargo-zigbuild` or whatever technique, Installing libopenssl for another arch and configure `pkg-config` to recognize here is hard and having a vendored OpenSSL is much easier to work with.

sfackler:
That is a convenient approach for something like zlib that has had a stable interface for like 20 years.

This project supports several major versions of multiple variants of OpenSSL, all exposing different features and APIs. I am Extremely Not Interested in dealing with tons of user confusion from people who think they're dynamically linking to their LibreSSL installation but end up with a statically linked OpenSSL instead.

NobodyXu:
> That is a convenient approach for something like zlib that has had a stable interface for like 20 years.
> 
> This project supports several major versions of multiple variants of OpenSSL, all exposing different features and APIs. I am Extremely Not Interested in dealing with tons of user confusion from people who think they're dynamically linking to their LibreSSL installation but end up with a statically linked OpenSSL instead.

Thanks for the explanation, I've removed `maybe-vendored` from the default feature set.

sfackler:
What is the use case for enabling this feature vs the existing vendored feature?

NobodyXu:
> What is the use case for enabling this feature vs the existing vendored feature?

It provides the same behavior as other crates such as zstd-rs: It tries to use the system-wide library and fallback to vendored one if failed.

sfackler:
Does the cargo-quickinstall CI system not know whether or not it wants to link to a preexisting installation?

NobodyXu:
> Does the cargo-quickinstall CI system not know whether or not it wants to link to a preexisting installation?

`cargo-quickinstall` tries to enable `vendored-openssl` if it's supported by the crate and has openssl installed on Linux, but for cross compilation, but a lot of crates actually don't provide `vendored-openssl` and thus fail to build.

I believe this feature is useful for build other than `cargo-quickinstall`, as people could have a incompatible version of openssl installed or have no openssl installed on their system, using `maybe-vendored` would ensure that the build would success.

After all, not everybody cares whether their binary is linked with system-wide openssl or vendored version, many just want to build and run it.

sfackler:
If they don't care what OpenSSL they link to, then why wouldn't they just enable the existing vendored feature? That is exactly what that feature is for.

NobodyXu:
> If they don't care what OpenSSL they link to, then why wouldn't they just enable the existing vendored feature? That is exactly what that feature is for.

This feature provides the same behavior as other crates: Tries to use system-wide library and fallback to vendored, so if there's a system-wide installation it can skip building.

But that's just my own opinion, if you think the existing feature is good enough, feel free to close this PR.

Issue #1836, opened 2023-03-09
Potential missing of `OPENSSL_free` in error case of `set_ocsp_status`


shinmao:
I am trying to call `set_ocsp_status` from Rust and look into its [implementation](https://github.com/sfackler/rust-openssl/blob/1ab42213942eb58f9293d597169086ebbbf11d22/openssl/src/ssl/mod.rs#L2873).
```rust
pub fn set_ocsp_status(&mut self, response: &[u8]) -> Result<(), ErrorStack> {
    unsafe {
        assert!(response.len() <= c_int::max_value() as usize);
        let p = cvt_p(ffi::OPENSSL_malloc(response.len() as _))?;
        ptr::copy_nonoverlapping(response.as_ptr(), p as *mut u8, response.len());
        cvt(ffi::SSL_set_tlsext_status_ocsp_resp(
            self.as_ptr(),
            p as *mut c_uchar,
            response.len() as c_long,
        ) as c_int)
        .map(|_| ())
    }
}
```
The implementation used `OPENSSL_malloc()` to allocate a heap memory at first. However, when an error occurs in `SSL_set_tlsext_status_ocsp_resp()`, it will not call `OPENSSL_free(p)` to release the memory which might lead to a memory leak!

Should the following code be added?
```rust
.map_err(p: |e: ErrorStack|) {
      ffi::OPENSSL_free(p);
      e
})
```
This can also make it consistent with other places such as `encrypt.rs`, `pkey_ctx.rs`, `string.rs`.

If I have any misunderstanding, please let me know :)

sfackler:
I believe you are right (though the current implementation won't ever return an error). Happy to take a PR fixing it.

shinmao:
Just submitted the PR. Please check. THX!

PR #1837, opened 2023-03-10
add missed free() on error

shinmao:
add free statement if error occurs in `SSL_set_tlsext_status_ocsp_resp()`
reference to the issue: https://github.com/sfackler/rust-openssl/issues/1836

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -2882,7 +2882,7 @@ impl SslRef {
             ) as c_int)
             .map(|_| ())
             .map_err(|e| {
-                ffi::OPENSSL_free(p)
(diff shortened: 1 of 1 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -2881,6 +2881,10 @@ impl SslRef {
                 response.len() as c_long,
             ) as c_int)
             .map(|_| ())
+            .map_err(|e| {
+                ffi::OPENSSL_free(p)
```
sfackler, line 2885: Missing a `;`

shinmao: Thanks. I will resubmit a new PR.

PR #1838, opened 2023-03-10
add missed free() on error and semicolon

shinmao:
update previous PR


sfackler:
You can just modify the existing PR fyi

PR #1843, opened 2023-03-14
Recognise LibreSSL 3.7.x

niklasha:
OpenBSD 7.3-beta currently features LibreSSL 3.7.1.
I am not sure about the policy here, 3.7.1 is not "released" per se yet.
The odd OpenBSD releases are normally happening around May 1st (give or take a few weeks).
But anyway the uname has changed to 7.3-beta, which usually triggers testers to build stuff,
and currently Rust code using openssl-sys will fail to build, without this fix.


sfackler:
We'll pick up 3.7.1 once a release is cut.

vishwin:
While 3.7.1 has now been cut, we cannot wildcard the 3.7 branch right now because 3.7.1 is still a development release.

PR #1847, opened 2023-03-17
Add issuer_name and reason_code to X509RevokedRef

Skepfyr:


openssl/src/x509/mod.rs
```diff
@@ -1421,6 +1422,37 @@ impl X509ReqRef {
     }
 }
 
+/// The reason that a certificate was revoked.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct CrlReason(i64);
```
Skepfyr, line 1427: This feels messy but the ASN1_ENUMERATED `get_int64` methods forces my hand slightly.

openssl/src/asn1.rs
```diff
@@ -661,6 +661,32 @@ cfg_if! {
...
+
+    /// An ASN.1 enumerated.
+    pub struct Asn1Enumerated;
+    /// A reference to an [`Asn1Enumerated`].
+    pub struct Asn1EnumeratedRef;
+}
+
+impl Asn1EnumeratedRef {
+    /// Get the value, if it fits in the required bounds.
+    #[corresponds(ASN1_ENUMERATED_get)]
```
sfackler, line 676: ASN1_ENUMERATED_get_int64

openssl/src/x509/mod.rs
```diff
@@ -1472,6 +1511,46 @@ impl X509RevokedRef {
...
+                self.as_ptr() as *const _,
+                // NID_crl_reason is an X509_REVOKED extension that is an ASN1_ENUMERATED
+                ffi::NID_crl_reason,
+                // Only one instance of the extension is permissable
+                ptr::null_mut(),
+                // Don't care if the extension is critical
+                ptr::null_mut(),
+            ) as *mut _)
+        }?;
+        Some(reason_code.get_i64().map(CrlReason::from_raw))
```
sfackler, line 1552: I'm not sure how I feel about this doing the conversion to CrlReason internally.

openssl/src/x509/mod.rs
```diff
@@ -1472,6 +1511,46 @@ impl X509RevokedRef {
             Asn1IntegerRef::from_ptr(r as *mut _)
         }
     }
+
+    /// Get the issuer name of the revoked certificate
+    #[corresponds(X509_REVOKED_get_ext_d2i)]
+    pub fn issuer_name(&self) -> Option<Stack<GeneralName>> {
```
sfackler, line 1517: In keeping with the general attempt to have 1:1 bindings to OpenSSL, how about switching to a single generic method that retrieves extensions? Something like:

```rust
impl X509RevokedRef {
    pub fn extension<T>(&self) -> Option<T::Output>
        where T: X509Extension
    { ... }
}

pub unsafe trait X509Extension {
    const NID: Nid;

    type Output: ForeignType;
}

pub enum CertificateIssuer {}

unsafe impl X509Extension for CertificateIssuer {
    const NID: Nid = Nid::CERTIFICATE_ISSUER;

    type Output = Stack<GeneralName>;
}
```

openssl/src/x509/mod.rs
Skepfyr: I've changed this to use `c_int` as it's now the caller's problem.

openssl/src/x509/mod.rs
Skepfyr: I like it as it feels a little more type-safe with no downsides I can see (CRLReason has a from_raw and to_raw). It's moot now though as the trait based system can't do it.

openssl/src/x509/mod.rs
Skepfyr: I've swapped over to this, I think it's a nicer match for the OpenSSL API and we should probably do the same thing to all the others but it is definitely a bit more complicated. (I did also deal with the crit parameter correctly at the same time which didn't help).

openssl/src/x509/mod.rs
```diff
@@ -50,6 +50,15 @@ pub mod store;
 #[cfg(test)]
 mod tests;
 
+/// A type of X509 extension.
+///
+/// # Safety
+/// The value of NID and Output must match those in OpenSSL so that
```
sfackler, line 56: Docs got cut off.

sfackler:
Nice - looks good! Can you add a few quick tests just confirming that the extension impls added work as expected?

openssl/src/x509/mod.rs
Skepfyr: Whoops, thanks, fixed.

PR #1852, opened 2023-03-23
Improve reliability of some tests

smoelius:
This PR proposes the following changes to the tests:
- Add `CALLED_BACK` flags in several places to ensure callbacks are actually executed.
- Call `unwrap` on the result of the `set_ssl_context` call to ensure the call was successful.
- Rename `connector_no_hostname_can_disable_verify` to `connector_can_disable_verify`, and furthermore:
  - remove the call to `verify_hostname` (unnecessary since verification of the server's certificate is disabled);
  - change the connected-to hostname from `foobar` to `fizzbuzz` to emphasize that it is not checked.


sfackler:
Thanks!

PR #1858, opened 2023-03-28
Add ASN.1 Integer comparison and to_owned

Skepfyr:
Comparing integers is actually infallible excitingly so we can implement the proper traits for it.

### DIFF
--- a/openssl/src/asn1.rs
+++ b/openssl/src/asn1.rs
@@ -790,6 +790,28 @@ mod tests {
     }
 
     #[test]
+    fn integer_to_owned() {
+        let a = Asn1Integer::from_bn(&BigNum::from_dec_str("42").unwrap()).unwrap();
+        let b = a.to_owned().unwrap();
+        assert_eq!(
+            a.to_bn().unwrap().to_dec_str().unwrap().to_string(),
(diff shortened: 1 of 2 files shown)


openssl/src/asn1.rs
```diff
@@ -512,6 +511,23 @@ impl Asn1Integer {
...
 }
 
+impl Ord for Asn1Integer {
+    fn cmp(&self, other: &Self) -> Ordering {
+        Asn1IntegerRef::cmp(self, other)
+    }
+}
+impl PartialOrd for Asn1Integer {
+    fn partial_cmp(&self, other: &Asn1Integer) -> Option<Ordering> {
+        Some(self.cmp(other))
```
steffen-eiden, line 521: Couldn't you just use `Asn1IntegerRef::partial_cmp` here as well?

Skepfyr: I could, but I couldn't see it making any difference and I'd already written this code.

steffen-eiden: So the stuff looks more similar.
yeah, its more a nit.

steffen-eiden (approved):

sfackler:
Thanks!

PR #1870, opened 2023-04-05
Raise the minimum CC version

alex:


openssl-sys/Cargo.toml
```diff
@@ -24,7 +24,7 @@ bssl-sys = { version = "0.1.0", optional = true }
 
 [build-dependencies]
 bindgen = { version = "0.64.0", optional = true, features = ["experimental"] }
-cc = "1.0"
+cc = "1.0.52"
```
sfackler, line 27: Looks like this needs to be 1.0.61: https://github.com/rust-lang/cc-rs/commit/cc8a24982b56675d4f7410416c36b518e33618ce

Issue #1484, opened 2021-06-11
Is there a way to encrypt/decrypt in place?


FH0:
https://docs.rs/openssl/0.10.34/openssl/symm/index.html

For example, I need encrypt 100 bytes data. Allocate 116 bytes, the last 16 bytes are used for `tag`.
```rust
let mut buf = vec![0u8; 116];
encrypt_aead_in_place(cipher, &key, iv, &aad, &mut buf);
```

manojgop:
It'll be nice to have this in-place encrypt/decrypt feature.

redpandamonium:
This would be a great addition. Especially if have a cipher with an 8 bit block length. There really is little reason to allocate temporary storage for that.

PR #1871, opened 2023-04-06
Add in-place cipher update method

haroldbruintjes:
This addresses #1484, for which we have a need as well. Especially useful if the input is not of a bound size. The OpenSSL documentation is not very explicit about this, but complete buffer overlap should be OK, just not partial overlap.
Note that the FFI function is called directly and does not go through cipher_update_unchecked, as the input buffer need to be provided twice to that function as a pointer.

### DIFF
--- a/openssl/src/cipher_ctx.rs
+++ b/openssl/src/cipher_ctx.rs
@@ -591,6 +591,50 @@ impl CipherCtxRef {
         Ok(len)
     }
 
+    /// Like [`Self::cipher_update`] except that it writes output into the
+    /// `data` buffer. The `inlen` parameter specifies the number of bytes in
+    /// `data` that are considered the input. For streaming ciphers, the size of
+    /// `data` must be at least the input size. Otherwise, it must be at least
+    /// an additional block size larger.
+    ///
+    /// Note: Use [`Self::cipher_update`] with no output argument to write AAD.
+    ///
+    /// # Panics
+    ///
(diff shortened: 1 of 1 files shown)


openssl/src/cipher_ctx.rs
```diff
@@ -591,6 +591,34 @@ impl CipherCtxRef {
...
+    /// Like [`Self::cipher_update`] except that it writes output into the `data`
+    /// buffer. This function can only be used for stream ciphers.
+    /// Use [`Self::cipher_update`] with no output argument to write AAD.
+    ///
+    /// # Panics
+    ///
+    /// This function panics if the underlying cipher is not a stream cipher.
+    pub fn cipher_update_inplace(&mut self, data: &mut [u8]) -> Result<usize, ErrorStack> {
+        assert!(
+            self.block_size() == 1,
```
sfackler, line 603: Are you sure in place updates aren't supported for block ciphers? I'd expect you may just need to have an oversized buffer - something like `pub fn cipher_update_inplace(&mut self, buf: &mut [u8], in_bytes: usize) -> Result<usize, ErrorStack>` where `buf.len() > in_bytes + self.block_size()` or whatever.

haroldbruintjes: I hadn't considered that (extra parameter), makes much more sense. Left the name as `inlen` to match the openssl API.

sfackler:
Thanks!

Issue #1878, opened 2023-04-12
Remove X509Extension::add_alias


davidben:
It is impossible to use this function without introducing a security vulnerability in your program. It should simply be removed. See https://crbug.com/boringssl/590 for details.

As noted in that bug, not only is this always function incorrect to call, it's also pointless. I suspect rust-openssl added it because `X509Extension::new` was based on the string-based system, which was itself a security vulnerability in the library. (RUSTSEC-2023-0023 and https://github.com/sfackler/rust-openssl/pull/1854) Had rust-openssl provided the correct extension construction API (i.e. the underlying DER bytes, not the ad-hoc, OpenSSL-specific, and ambiguous text format), `add_alias` would not have been needed to add custom extensions.

(Strictly speaking, it's already not needed because of the `ASN1:whatver` mode of the mini-language, but you all should stop using the mini-language because it's a security bug. It's particularly unreasonable when purporting to provide APIs in a safe language. 😄  https://github.com/sfackler/rust-openssl/pull/1854 fixed the worst of it, but it just added a warning on `X509Extension::new` without providing a replacement.)

Once you've removed that, I think you'll also have removed the only use case for `OBJ_create` and can remove that too. `OBJ_create` has similar problems with global state. To that end, `Nid::create` is missing a warning about how unsafe it is... I'll file a bug there too. See [BoringSSL's warning](https://commondatastorage.googleapis.com/chromium-boringssl-docs/obj.h.html#Adding-objects-at-runtime).

I hope to remove these unsafe functions from BoringSSL altogether, so if I've cleared the other blockers by that point, you all may need to ifdef this API out of your BoringSSL port altogether.

CC @alex and @reaperhulk, since we've been talking about various security problems caused by rust-openssl binding the wrong APIs.

sfackler:
Yep, that seems right to me.

sfackler:
Once we add the new extensions API there are a fair number of ad-hoc methods like `X509Ref::subject_alt_names` that we can deprecate as well.

davidben:
As long as the extensible extensions API doesn't require process-global knowledge like the OpenSSL ones do, I don't have a _correctness_ objection. (I.e. no NIDs, and two different libraries can have different preferences on how to process the same extension.) But, based on my experience with X.509 itself, I think an extensible API is not the way to go.

[...]

PR #1880, opened 2023-04-13
Introduce X509Extension::new_from_der and deprecate the bad APIs

alex:
Refs #1878

### DIFF
--- a/openssl/src/x509/mod.rs
+++ b/openssl/src/x509/mod.rs
@@ -24,8 +24,8 @@ use std::slice;
 use std::str;
 
 use crate::asn1::{
-    Asn1BitStringRef, Asn1Enumerated, Asn1IntegerRef, Asn1Object, Asn1ObjectRef, Asn1StringRef,
-    Asn1TimeRef, Asn1Type,
+    Asn1BitStringRef, Asn1Enumerated, Asn1IntegerRef, Asn1Object, Asn1ObjectRef,
+    Asn1OctetStringRef, Asn1StringRef, Asn1TimeRef, Asn1Type,
 };
 use crate::bio::MemBioSlice;
 use crate::conf::ConfRef;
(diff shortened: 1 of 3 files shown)


alex:
Clippy job failing because there are deprecation warnings... because there are deprecated APIs, and we still need to test them. Not sure how to resolve?

openssl/src/x509/mod.rs
```diff
@@ -842,6 +842,10 @@ impl X509Extension {
     /// mini-language that can read arbitrary files.
     ///
     /// See the extension module for builder types which will construct certain common extensions.
+    ///
+    /// This function is deprecated, `X509Extension::new_from_der` or the
+    /// types in `x509::extension` should be used in its place.
+    #[deprecated]
```
sfackler, line 848: We should add `note` and `since` sections to these: https://github.com/sfackler/rust-openssl/blob/47abce45f95dbca9c644674abe40d60ede01dc64/openssl/src/pkcs12.rs#L111

openssl/src/x509/mod.rs
```diff
@@ -921,6 +929,40 @@ impl X509Extension {
...
+    ///
+    /// For common extension types, there are Rust APIs provided in
+    /// `openssl::x509::extensions` which are more ergonomic.
+    pub fn new_from_der(
+        oid: &Asn1ObjectRef,
+        critical: bool,
+        der_contents: &[u8],
+    ) -> Result<X509Extension, ErrorStack> {
+        unsafe {
+            ffi::init();
```
sfackler, line 947: No need to do this given we've already constructed an Asn1ObjectRef I think.

openssl/src/x509/mod.rs
```diff
@@ -921,6 +929,40 @@ impl X509Extension {
...
+            let s = Asn1String::from_ptr(cvt_p(ffi::ASN1_STRING_type_new(
+                Asn1Type::OCTET_STRING.as_raw(),
+            ))?);
+            cvt(ffi::ASN1_STRING_set(
+                s.as_ptr(),
+                der_contents.as_ptr().cast(),
+                der_contents.len().try_into().unwrap(),
+            ))?;
```
sfackler, line 955: Should we have this method take an `Asn1OctetStringRef` instead of doing the construction in here?

alex: I don't think it has a public constructor, so we'd have to do that first. This seemed simpler.

sfackler: Should be pretty straightforward to add the constructor.

alex: Done

reaperhulk (approved):

sfackler (approved):

PR #1883, opened 2023-04-15
Fixes #1882 -- added APIs for setting public keys on Dh

### LINKED ISSUE #1882: Expose way to create a new Dh<Public>
Right now it's possible to create a `Dh<Private>` with https://docs.rs/openssl/latest/openssl/dh/struct.Dh.html#method.set_private_key, but there's no `set_public_key` equivalent

alex:


reaperhulk (approved):

openssl/src/dh.rs
```diff
@@ -79,6 +90,17 @@ impl Dh<Params> {
         }
     }
 
+    pub fn set_key(self, pub_key: BigNum, priv_key: BigNum) -> Result<Dh<Private>, ErrorStack> {
```
sfackler, line 93: Needs docs

openssl/src/dh.rs
```diff
@@ -79,6 +90,17 @@ impl Dh<Params> {
...
+        unsafe {
+            let dh_ptr = self.0;
+            cvt(DH_set0_key(dh_ptr, pub_key.as_ptr(), priv_key.as_ptr()))?;
+            mem::forget(pub_key);
+            mem::forget(priv_key);
+            mem::forget(self);
```
sfackler, line 99: This can just be `mem::forget((pub_key, priv_key, self))`.

PR #1885, opened 2023-04-16
Fixes #1884 -- don't leave an error on the stack in public_eq

### LINKED ISSUE #1884: PKey::public_eq can leave errors on the stack
https://docs.rs/openssl/latest/openssl/pkey/struct.PKeyRef.html#method.public_eq

If the keys are different types, an error will be left on the stack.

alex:


sfackler:
Are the errors useful? Should we instead deprecate public_eq in favor of a new method that returns a `Result<bool, ErrorStack>`?

alex:
The only case I'm aware of is "these keys are different types", which is just a weird case of "they are not equal". I think it's fine to ignore them.

sfackler (approved):

PR #1886, opened 2023-04-18
DTLS1 and DTLS1_2 SslVersion for set_min_proto_version()

algesten:
Expose constants to allow limiting the DTLS version.

The relevant openssl constants are taken from here: https://github.com/openssl/openssl/blob/418c6c520764491262018c45481a20ef10cd3bca/include/openssl/prov_ssl.h#L28-L29

The comment about there being no DTLS 1.1 is from the RFC: https://www.rfc-editor.org/rfc/rfc9147.html#name-introduction


sfackler:
Thanks!

PR #1888, opened 2023-04-20
Remove size_t-is-usize argument to bindgen

alex:
It's been on by default for a while: https://github.com/rust-lang/rust-bindgen/commit/cc78b6fdb6e829e5fb8fa1639f2182cb49333569


sfackler:
LGTM but it looks like there are some boringssl ABI changes to deal with as well.

alex:
It looks like it's actually build changes -- I believe the directory that bssl-sys is in changed.

PR #1902, opened 2023-04-24
Add X509 pubkey related types and methods

zh-jq:
Added X509Pubkey:
https://www.openssl.org/docs/man1.1.1/man3/X509_PUBKEY_get.html

Added pubkey_digest method to X509:
https://www.openssl.org/docs/man3.1/man3/X509_pubkey_digest.html


openssl/src/x509/mod.rs
```diff
@@ -491,6 +491,15 @@ impl X509Ref {
         }
     }
 
+    #[corresponds(X509_get_X509_PUBKEY)]
+    #[cfg(ossl110)]
+    pub fn x509_pubkey(&self) -> Result<X509Pubkey, ErrorStack> {
```
sfackler, line 496: > X509_get_X509_PUBKEY() returns an internal pointer to the X509_PUBKEY structure which encodes the certificate of x. The returned value must not be freed up after use.

zh-jq:
@sfackler @alex a review is needed here

Issue #1908, opened 2023-04-26
Unable to build on Ubuntu 18.04


oskirby:
The latest version of OpenSSL available from the `bionic-updates` is currently at version 1.1.1 and this doesn't build cleanly, as this is lacking the `X509_get0_authority_issuer` and `X509_get0_authority_serial` methods where were introduced in OpenSSL 1.1.1d. This results in a linker error something like this:

```
  = note: /home/naomi/rust-openssl/target/debug/deps/openssl-a64f8372bdce9d24.4842uy0p6nd2xqil.rcgu.o: In function `openssl::x509::X509Ref::authority_issuer':
          /home/naomi/rust-openssl/openssl/src/x509/mod.rs:511: undefined reference to `X509_get0_authority_issuer'
          /home/naomi/rust-openssl/target/debug/deps/openssl-a64f8372bdce9d24.4842uy0p6nd2xqil.rcgu.o: In function `openssl::x509::X509Ref::authority_serial':
          /home/naomi/rust-openssl/openssl/src/x509/mod.rs:521: undefined reference to `X509_get0_authority_serial'
          collect2: error: ld returned 1 exit status
          
  = help: some `extern` functions couldn't be found; some native libraries may need to be installed or have their path specified
  = note: use the `-l` flag to specify native libraries to link
  = note: use the `cargo:rustc-link-lib` directive to specify the native libraries to link with Cargo (see https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorustc-link-libkindname)
```

Furthermore, if I hack my way around the linker issues, there is a const-correctness warning in the systest that is due to a change in the function signatures of `SSL_get_num_tickets` and `SSL_CTX_get_num_tickets` methods in OpenSSL 1.1.1b. This results in a handful of warnings like this:

```
naomi@ubuntu-bionic:~/rust-openssl$ cargo build --tests
   Compiling openssl-sys v0.9.87 (/home/naomi/rust-openssl/openssl-sys)
   Compiling garando_syntax v0.1.0
   Compiling openssl v0.10.52 (/home/naomi/rust-openssl/openssl)
   Compiling openssl-errors v0.2.0 (/home/naomi/rust-openssl/openssl-errors)
   Compiling ctest2 v0.4.5
   Compiling openssl-macros v0.1.1 (/home/naomi/rust-openssl/openssl-macros)
   Compiling systest v0.1.0 (/home/naomi/rust-openssl/systest)
The following warnings were emitted during compilation:

[...]

PR #1909, opened 2023-04-26
Fix link errors for X509_get0_authority_xxx methods on Ubuntu/bionic

oskirby:
The `X509_get0_authority_issuer` and `X509_get0_authority_serial` methods were introduced in OpenSSL version 1.1.1d which can cause linker errors when building for Ubuntu/bionic which provides OpenSSL 1.1.1 in the `bionic-updates` package suite.

See issue: #1908


sfackler:
Thanks!

Issue #1907, opened 2023-04-26
Undefined references to EVP_rc4 on distros with OPENSSL_NO_RC4


oskirby:
Over on the [Mozilla VPN](https://github.com/mozilla-mobile/mozilla-vpn-client) project, we have recently introduced some Rust components that make use of the OpenSSL package and this has been causing link issues on some of our users who are trying to build the client on Linux distributions without RC4 support.

Even though we aren't making use of RC4 anywhere in our codebase, we still encounter errors like this:
```
[ 92%] Linking CXX executable mozillavpn
/usr/bin/ld: ../vpnglean/x86_64-unknown-linux-gnu/debug/libvpnglean.a(openssl-473657804fb859a9.openssl.7f8ba334-cgu.5.rcgu.o): in function `openssl::symm::Cipher::rc4':
(REMOVED)/mozilla-vpn-client/build/cargo_home/registry/src/github.com-1ecc6299db9ec823/openssl-0.10.44/src/symm.rs:287: undefined reference to `EVP_rc4'
/usr/bin/ld: ../vpnglean/x86_64-unknown-linux-gnu/debug/libvpnglean.a(openssl-473657804fb859a9.openssl.7f8ba334-cgu.6.rcgu.o): in function `openssl::cipher::Cipher::rc4':
(REMOVED)/mozilla-vpn-client/build/cargo_home/registry/src/github.com-1ecc6299db9ec823/openssl-0.10.44/src/cipher.rs:328: undefined reference to `EVP_rc4'
collect2: error: ld returned 1 exit status
gmake[2]: *** [src/CMakeFiles/mozillavpn.dir/build.make:3977: src/mozillavpn] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:1120: src/CMakeFiles/mozillavpn.dir/all] Error 2
gmake: *** [Makefile:146: all] Error 2
```

I believe that we should check for the `OPENSSL_NO_RC4` configuration when providing methods that use RC4.

This was reported by one of our users here: mozilla-mobile/mozilla-vpn-client#6228

sfackler:
Ah yeah, that should require a patch like https://github.com/sfackler/rust-openssl/commit/62c8dfdedf67d947f2ff5da0b0ae94741ea8671a

PR #1910, opened 2023-04-26
Check for OPENSSL_NO_RC4 when using EVP_rc4

oskirby:
Not all Linux distributions provide a version of OpenSSL with RC4 cipher support. We should make any function that uses it conditional upon `OPENSSL_NO_RC4` being undefined.

Closes: #1907


sfackler:
Thanks!

PR #1915, opened 2023-05-01
add other name support

huettner94:
the issue with other name SANs is that they can contain arbitary data. As we can no longer use the old method for other_name for security reasons the easiest way now is to do individual implementations per datatype.

We start with strings for now, but it should be quite easy to expand upon that.

Closes: https://github.com/sfackler/rust-openssl/issues/1911

### DIFF
--- a/openssl/src/asn1.rs
+++ b/openssl/src/asn1.rs
@@ -655,6 +655,7 @@ impl Asn1OctetStringRef {
 foreign_type_and_impl_send_sync! {
     type CType = ffi::ASN1_OBJECT;
     fn drop = ffi::ASN1_OBJECT_free;
+    fn clone = ffi::OBJ_dup;
 
     /// Object Identifier
     ///
--- a/openssl/src/x509/extension.rs
+++ b/openssl/src/x509/extension.rs
@@ -434,6 +434,7 @@ enum RustGeneralName {
     Uri(String),
     Ip(String),
     Rid(String),
+    OtherName(Asn1Object, Vec<u8>),
 }
 
 /// An extension that allows additional identities to be bound to the subject
@@ -506,12 +507,21 @@ impl SubjectAlternativeName {
 
     /// Sets the `otherName` flag.
     ///
-    /// Not currently actually supported, always panics.
-    #[deprecated = "other_name is deprecated and always panics. Please file a bug if you have a use case for this."]
+    /// Not currently actually supported, always panics. Please use other_name2
+    #[deprecated = "other_name is deprecated and always panics. Please use other_name2."]
     pub fn other_name(&mut self, _other_name: &str) -> &mut SubjectAlternativeName {
-        unimplemented!(
-            "This has not yet been adapted for the new internals. File a bug if you need this."
-        );
+        unimplemented!("This has not yet been adapted for the new internals. Use other_name2.");
+    }
+
+    /// Sets the `otherName` flag.
+    ///
+    /// `content` must be a valid der encoded ASN1_TYPE
+    ///
+    /// If you want to add just a ia5string use `other_name_ia5string`
+    pub fn other_name2(&mut self, oid: Asn1Object, content: &[u8]) -> &mut SubjectAlternativeName {
+        self.items
+            .push(RustGeneralName::OtherName(oid, content.into()));
+        self
     }
 
     /// Return a `SubjectAlternativeName` extension as an `X509Extension`.
@@ -526,6 +536,9 @@ impl SubjectAlternativeName {
                     GeneralName::new_ip(s.parse().map_err(|_| ErrorStack::get())?)?
                 }
                 RustGeneralName::Rid(s) => GeneralName::new_rid(Asn1Object::from_str(s)?)?,
+                RustGeneralName::OtherName(oid, content) => {
+                    GeneralName::new_other_name(oid.clone(), content)?
+                }
             };
             stack.push(gn)?;
         }
--- a/openssl/src/x509/mod.rs
+++ b/openssl/src/x509/mod.rs
@@ -2054,6 +2054,37 @@ impl GeneralName {
             Ok(GeneralName::from_ptr(gn))
         }
     }
+
+    pub(crate) fn new_other_name(
+        oid: Asn1Object,
+        value: &Vec<u8>,
+    ) -> Result<GeneralName, ErrorStack> {
+        unsafe {
+            ffi::init();
+
+            let typ = cvt_p(ffi::d2i_ASN1_TYPE(
+                ptr::null_mut(),
+                &mut value.as_ptr().cast(),
+                value.len().try_into().unwrap(),
+            ))?;
+
+            let gn = cvt_p(ffi::GENERAL_NAME_new())?;
+            (*gn).type_ = ffi::GEN_OTHERNAME;
+
+            if let Err(e) = cvt(ffi::GENERAL_NAME_set0_othername(
+                gn,
+                oid.as_ptr().cast(),
+                typ,
+            )) {
+                ffi::GENERAL_NAME_free(gn);
+                return Err(e);
+            }
+
+            mem::forget(oid);
+
+            Ok(GeneralName::from_ptr(gn))
+        }
+    }
 }
 
 impl GeneralNameRef {
--- a/openssl-sys/src/handwritten/asn1.rs
+++ b/openssl-sys/src/handwritten/asn1.rs
@@ -10,6 +10,7 @@ pub struct ASN1_ENCODING {
 
 extern "C" {
     pub fn ASN1_OBJECT_free(x: *mut ASN1_OBJECT);
+    pub fn OBJ_dup(x: *const ASN1_OBJECT) -> *mut ASN1_OBJECT;
 }
 
 stack!(stack_st_ASN1_OBJECT);
@@ -94,7 +95,14 @@ extern "C" {
     #[cfg(ossl110)]
     pub fn ASN1_ENUMERATED_get_int64(pr: *mut i64, a: *const ASN1_ENUMERATED) -> c_int;
 
+    pub fn ASN1_TYPE_new() -> *mut ASN1_TYPE;
+    pub fn ASN1_TYPE_set(a: *mut ASN1_TYPE, type_: c_int, value: *mut c_void);
     pub fn ASN1_TYPE_free(x: *mut ASN1_TYPE);
+    pub fn d2i_ASN1_TYPE(
+        k: *mut *mut ASN1_TYPE,
+        buf: *mut *const u8,
+        len: c_long,
+    ) -> *mut ASN1_TYPE;
 }
 
 const_ptr_api! {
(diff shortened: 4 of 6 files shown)


huettner94:
note: i did not test with boringssl i hope the ci does that

openssl/src/x509/extension.rs
```diff
@@ -506,14 +507,23 @@ impl SubjectAlternativeName {
...
+    /// Not currently actually supported, always panics. Please use other_name2
+    #[deprecated = "other_name is deprecated and always panics. Please use other_name2."]
     pub fn other_name(&mut self, _other_name: &str) -> &mut SubjectAlternativeName {
         unimplemented!(
             "This has not yet been adapted for the new internals. File a bug if you need this."
         );
     }
 
+    /// Sets the `otherName` flag to a ia5string value.
+    pub fn other_name_string(&mut self, oid: &str, content: &str) -> &mut SubjectAlternativeName {
```
alex, line 519: I think we want to use more precise types here: oid should be an Asn1ObjectIdentifier type, and contents should be `&[u8]` since it's DER.

openssl/src/x509/mod.rs
```diff
@@ -2046,6 +2046,29 @@ impl GeneralName {
...
+        oid: Asn1Object,
+        value: &[u8],
+    ) -> Result<GeneralName, ErrorStack> {
+        unsafe {
+            ffi::init();
+            let gn = cvt_p(ffi::GENERAL_NAME_new())?;
+            (*gn).type_ = ffi::GEN_OTHERNAME;
+
+            let s = cvt_p(ffi::ASN1_STRING_type_new(Asn1Type::IA5STRING.as_raw()))?;
+            ffi::ASN1_STRING_set(s, value.as_ptr().cast(), value.len().try_into().unwrap());
```
alex, line 2060: This isn't right, the value of an OtherName is not necessarily an IA5String, we need to handle arbitrary DER TLVs in the value.

alex (commented):
Thanks for diving into this.

openssl/src/x509/extension.rs
alex: IMO limiting to strings doesn't very much sense, OtherName is a very general purpose specification, and we should have an API that's similarly broad. Why do you think `unsafe` would be required?

huettner94: thats why i would allow both. To get the combination of ease of use and all the features you might need.

I see `unsafe` as required as the data would then be passed to `GENERAL_NAME_set0_othername`. In there the openssl lib expects (if i understood it correctly) certain data formats (e.g. during deserialization). At least the last parameter must be a `ASN1_TYPE` as otherwise you will get a SegmentationFault (which is a clear sign to me that we need unsafe here, or guarantee that we do not need it like in the string case)

alex: If the public API takes `&[u8]`, a sequence of DER bytes, then it will be the responsibility of rust-openssl to have the unsafe code to turn that into an `ASN1_TYPE`, we can do that with `d2i_ASN1_TYPE`.

huettner94: Thanks for the clarification.

I have now added `other_name2` that will take a provided `&[u8]` and try to convert it to an `ASN1_TYPE`.

For ease of use i have adapted the `other_name_ia5string` method, but if you want to get rid of it that is also fine for me.

openssl/src/x509/mod.rs
huettner94: fixed

openssl/src/x509/extension.rs
```diff
@@ -506,14 +511,64 @@ impl SubjectAlternativeName {
...
+            typ = cvt_p(ffi::d2i_ASN1_TYPE(
+                ptr::null_mut(),
+                &mut content.as_ptr().cast(),
+                content.len().try_into().unwrap(),
+            ))?;
+        }
+        self.items
+            .push(RustGeneralName::OtherName(oid.to_string(), typ));
+        Ok(self)
+    }
```
alex, line 545: Would like @sfackler's take on this, specifically whether we should have this return a `Result`, or if it should be consistent with the other ones and just return `&mut SAN` and then do the error handling lazily.

openssl/src/x509/extension.rs
```diff
@@ -506,14 +511,64 @@ impl SubjectAlternativeName {
...
+
+    /// Sets the `otherName` flag for ia5strings.
+    pub fn other_name_ia5string(
+        &mut self,
+        oid: &str,
+        content: &str,
+    ) -> &mut SubjectAlternativeName {
```
alex, line 552: I'd prefer not to expose this -- the right way to do this is a separate, orthogonal API that makes it easy to DER serialize a https://docs.rs/openssl/latest/openssl/asn1/struct.Asn1String.html

alex (commented):
Getting close! Can you please add a test case or two to this.

openssl/src/x509/extension.rs
```diff
@@ -506,14 +511,64 @@ impl SubjectAlternativeName {
...
     }
 
+    /// Sets the `otherName` flag.
+    ///
+    /// `content` must be a valid der encoded ASN1_TYPE
+    ///
+    /// If you want to add just a ia5string use `other_name_ia5string`
+    pub fn other_name2(
+        &mut self,
+        oid: &str,
```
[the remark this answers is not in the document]
huettner94: done

openssl/src/x509/extension.rs
huettner94: done, added that api in a separate commit

openssl/src/x509/mod.rs
```diff
@@ -2046,6 +2047,23 @@ impl GeneralName {
...
+    pub(crate) fn new_other_name(
+        oid: Asn1Object,
+        value: *mut ASN1_TYPE,
+    ) -> Result<GeneralName, ErrorStack> {
+        unsafe {
+            ffi::init();
+            let gn = cvt_p(ffi::GENERAL_NAME_new())?;
+            (*gn).type_ = ffi::GEN_OTHERNAME;
+
+            ffi::GENERAL_NAME_set0_othername(gn, oid.as_ptr().cast(), value);
```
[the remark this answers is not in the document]
huettner94: done

huettner94:
i will add tests soon(tm)

openssl/src/x509/extension.rs
```diff
@@ -506,14 +511,38 @@ impl SubjectAlternativeName {
...
+    /// Sets the `otherName` flag.
+    ///
+    /// `content` must be a valid der encoded ASN1_TYPE
+    ///
+    /// If you want to add just a ia5string use `other_name_ia5string`
+    pub fn other_name2(
+        &mut self,
+        oid: Asn1Object,
+        content: &[u8],
+    ) -> Result<&mut SubjectAlternativeName, ErrorStack> {
```
alex, line 531: @sfackler do you have an opinion on this returning a `Result`, vs. returning `&mut SAN` and doing the fallible ops in `build`?

openssl/src/x509/mod.rs
```diff
@@ -2046,6 +2047,27 @@ impl GeneralName {
...
+        unsafe {
+            ffi::init();
+            let gn = cvt_p(ffi::GENERAL_NAME_new())?;
+            (*gn).type_ = ffi::GEN_OTHERNAME;
+
+            cvt(ffi::GENERAL_NAME_set0_othername(
+                gn,
+                oid.as_ptr().cast(),
+                value,
+            ))?;
```
alex, line 2064: If this fails then `gn` is leaked.

openssl/src/asn1.rs
```diff
@@ -430,6 +430,30 @@ foreign_type_and_impl_send_sync! {
...
+            ffi::ASN1_TYPE_set(typ, ffi::V_ASN1_IA5STRING, self.0.cast());
+
+            let len = cvt(ffi::i2d_ASN1_TYPE(typ, ptr::null_mut()))?;
+            let mut buf = vec![0; len as usize];
+            crate::cvt(ffi::i2d_ASN1_TYPE(typ, &mut buf.as_mut_ptr()))?;
+            Ok(buf)
+        }
+    }
+}
+
```
alex, line 456: Can we do this in a separate PR? It should be orthagonal to the rest of this.

openssl/src/asn1.rs
```diff
@@ -705,6 +729,12 @@ impl Asn1Object {
     }
 }
 
+impl Clone for Asn1Object {
+    fn clone(&self) -> Self {
+        unsafe { Asn1Object::from_ptr(cvt_p(ffi::OBJ_dup(self.as_ptr())).unwrap()) }
```
alex, line 734: Instead of this you can do `clone = OBJ_dup;` in the `foreign_type_and_impl_send_sync` macro.

openssl/src/x509/extension.rs
```diff
@@ -506,14 +511,38 @@ impl SubjectAlternativeName {
 
     /// Sets the `otherName` flag.
     ///
-    /// Not currently actually supported, always panics.
-    #[deprecated = "other_name is deprecated and always panics. Please file a bug if you have a use case for this."]
+    /// Not currently actually supported, always panics. Please use other_name2 or other_name_ia5string
+    #[deprecated = "other_name is deprecated and always panics. Please use other_name2 or other_name_ia5string."]
```
alex, line 515:
```suggestion
    /// Not currently actually supported, always panics. Please use other_name2
    #[deprecated = "other_name is deprecated and always panics. Please use other_name2."]
```

openssl/src/x509/extension.rs
```diff
@@ -506,14 +511,38 @@ impl SubjectAlternativeName {
...
     pub fn other_name(&mut self, _other_name: &str) -> &mut SubjectAlternativeName {
         unimplemented!(
-            "This has not yet been adapted for the new internals. File a bug if you need this."
+            "This has not yet been adapted for the new internals. Use other_name2 or other_name_ia5string."
```
alex, line 518:
```suggestion
            "This has not yet been adapted for the new internals. Use other_name2."
```

alex (commented):
A few small comments, this is very close. Thanks for your work.

openssl/src/x509/extension.rs
huettner94: thanks fixed

openssl/src/asn1.rs
huettner94: thanks, done

openssl/src/x509/mod.rs
huettner94: fixed

openssl/src/x509/extension.rs
```diff
@@ -506,12 +511,34 @@ impl SubjectAlternativeName {
 
     /// Sets the `otherName` flag.
     ///
-    /// Not currently actually supported, always panics.
-    #[deprecated = "other_name is deprecated and always panics. Please file a bug if you have a use case for this."]
+    /// Not currently actually supported, always panics. Please use other_name2 or other_name_ia5string
```
alex, line 514:
```suggestion
    /// Not currently actually supported, always panics. Please use other_name2
```

openssl/src/x509/mod.rs
```diff
@@ -2054,6 +2055,30 @@ impl GeneralName {
...
+        unsafe {
+            ffi::init();
+            let gn = cvt_p(ffi::GENERAL_NAME_new())?;
+            (*gn).type_ = ffi::GEN_OTHERNAME;
+
+            if let Err(e) = cvt(ffi::GENERAL_NAME_set0_othername(
+                gn,
+                oid.as_ptr().cast(),
+                value,
+            )) {
```
alex, line 2072: Unfortunately this isn't sound -- because `build` takes `&self` you can call it multiple times, in which case the `value` ptr will be reused.

openssl/src/x509/extension.rs
sfackler: Given that we're deferring the other fallible work to build currently, I'd have this do the same I think.

alex: Ok, so in that case this should just return the builder, and we can conver to `ASN1_TYPE` in `build` and I think that will fix the other soundness problem

huettner94: ok, done

openssl/src/x509/extension.rs
huettner94: thanks fixed

openssl/src/x509/mod.rs
huettner94: handled as already commented somewhere else

PR #1919, opened 2023-05-06
Link README and Repo for openssl-macros

CosmicHorrorDev:
Just adding some missing metadata that I noticed from the [openssl-macros crates.io page](https://crates.io/crates/openssl-macros/0.1.1). Follows the same technique used for the other crates in this workspace


sfackler:
This crate was left intentionally light on crates.io metadata since it's only intended for use transitively through openssl.

CosmicHorrorDev:
I feel like copying the parent crate's README file is the typical way that I see "internal" crates, although that's admittedly not consistent across the ecosystem

Feel free to handle this however you want of course

PR #1921, opened 2023-05-09
Add bindings to SSL_bytes_to_cipher_list

RoastVeg:
I have a use case where I need to change the SSL context on the server side based on the supported cipher suite provided in the client hello callback. Without `SSL_bytes_to_cipher_list` support in rust-openssl, there really isn't a convenient way to handle the output from `client_hello_ciphers`, as the format is a real mess.

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -72,7 +72,7 @@ use crate::srtp::{SrtpProtectionProfile, SrtpProtectionProfileRef};
 use crate::ssl::bio::BioMethod;
 use crate::ssl::callbacks::*;
 use crate::ssl::error::InnerError;
-use crate::stack::{Stack, StackRef};
+use crate::stack::{Stack, StackRef, Stackable};
 use crate::util::{ForeignTypeExt, ForeignTypeRefExt};
 use crate::x509::store::{X509Store, X509StoreBuilderRef, X509StoreRef};
 #[cfg(any(ossl102, libressl261))]
@@ -1940,6 +1940,10 @@ impl ForeignType for SslCipher {
     }
 }
 
+impl Stackable for SslCipher {
+    type StackType = ffi::stack_st_SSL_CIPHER;
+}
+
 impl Deref for SslCipher {
     type Target = SslCipherRef;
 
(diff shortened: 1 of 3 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -3083,6 +3093,40 @@ impl SslRef {
...
+    ///
+    /// This can only be used inside of the client hello callback. Otherwise, `None` is returned.
+    ///
+    /// Requires OpenSSL 1.1.1 or newer.
+    #[corresponds(SSL_bytes_to_cipher_list)]
+    #[cfg(ossl111)]
+    pub fn client_hello_ciphers_stack(&self) -> Option<(Stack<SslCipher>, Stack<SslCipher>)> {
+        unsafe {
+            let mut ptr = ptr::null();
+            let len = ffi::SSL_client_hello_get0_ciphers(self.as_ptr(), &mut ptr);
```
sfackler, line 3107: I generally prefer to bind to each OpenSSL API as directly as possible, so I think we'd want separate methods for SSL_client_hello_get0_ciphers, SSL_bytes_to_cipher list, and SSL_client_hello_isv2 (though really the answer to that is just `false` as this point!).

RoastVeg: In that case as `SSL_client_hello_get0_ciphers` and `SSL_client_hello_isv2` both already have bindings, would you accept a new method like the following?
```rust
// struct Ssl
#[corresponds(SSL_bytes_to_cipher_list)]
pub fn bytes_to_ciphers_stack(&self, bytes: &[u8], isv2format: bool) -> Option<(Stack<SslCipher>, Stack<SslCipher>)> { // ... }
```

openssl/src/ssl/mod.rs
```diff
@@ -3083,6 +3093,39 @@ impl SslRef {
...
+    /// Returns a tuple of ciphers and signalling cipher-suite values if valid, or `None` otherwise.
+    ///
+    /// Requires OpenSSL 1.1.1 or newer.
+    #[corresponds(SSL_bytes_to_cipher_list)]
+    #[cfg(ossl111)]
+    pub fn bytes_to_ciphers_stack(
+        &self,
+        bytes: &[u8],
+        isv2format: bool,
+    ) -> Option<(Stack<SslCipher>, Stack<SslCipher>)> {
```
sfackler, line 3107: Can we have this return a `struct CipherLists { pub suites: Stack<SslCipher>, pub signaling_suites: Stack<SslCipher> }` so it's a bit more descriptive?

sfackler: Does the function not push errors onto the stack on failure, or should this return a `Result<CipherLists, ErrorStack>` instead?

PR #1922, opened 2023-05-10
fix: BIO_set_retry_write when BIO_CTRL_FLUSH to allow writer returns WouldBlock on flush

ihciah:
This PR fixes a BIO_flush related issue. According to openssl doc([link](https://www.openssl.org/docs/man1.0.2/man3/BIO_flush.html)):
> BIO_flush(), because it can write data may return 0 or -1 indicating that the call should be retried later in a similar manner to BIO_write(). The BIO_should_retry() call should be used and appropriate action taken is the call fails.

We have to call `BIO_set_retry_write` on `retriable_error` like WouldBlock. Otherwise, when the `flush` returns WouldBlock, it will fail.

Since we usually wrap openssl with a `TcpStream`, there is no buffer inside and `flush` always returns `Ok(())`, this issue is not obvious. However, if users use a buffered io, the flush support will be essential.

However, only fixes rust-openssl is not enough. I found a problem in openssl that makes this change not working. The issue can be found at https://github.com/openssl/openssl/pull/20919 and I will try to fix it.


ihciah:
As a memo:
To check the feature, clone the [forked openssl repo](https://github.com/ihciah/openssl), compile it and inject envs:
```
env::set_var("OPENSSL_LIB_DIR", "/some/path/openssl");
env::set_var("OPENSSL_INCLUDE_DIR", "/some/path/openssl/include");
env::set_var("OPENSSL_STATIC", "1");
```

sfackler:
Could you add a test for this change?

ihciah:
> Could you add a test for this change?

Sure. I've added a new test for it. But since openssl still has bug(https://github.com/openssl/openssl/pull/20919) on flush, to make the the test pass, I only tested read and write.

Once the openssl issue solved, change `first_flush` to true(L563) and the flush retriable can be tested.

ihciah:
Hi! Just a remind. @sfackler

sfackler:
I will look at this once the upstream change lands so it can actually be tested.

ihciah:
Hi! @sfackler 
The upstream fix is merged: https://github.com/openssl/openssl/pull/21298

PR #1925, opened 2023-05-15
Update vendored version to openssl 3

### LINKED ISSUE #1645: Update openssl-src to 300?
Currently openssl-sys crate references openssl-src version 111. Is there any plan to update to OpenSSL 3.0?

amousset:
Fixes #1645.

### DIFF
--- a/openssl-sys/Cargo.toml
+++ b/openssl-sys/Cargo.toml
@@ -25,7 +25,7 @@ bssl-sys = { version = "0.1.0", optional = true }
 [build-dependencies]
 bindgen = { version = "0.64.0", optional = true, features = ["experimental"] }
 cc = "1.0.61"
-openssl-src = { version = "111", optional = true }
+openssl-src = { version = "300.1.2", optional = true, features = ["legacy"] }
 pkg-config = "0.3.9"
 vcpkg = "0.2.8"
 
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -115,6 +115,16 @@ fn main() {
         println!("cargo:rustc-link-lib={}={}", kind, lib);
     }
 
+    // https://github.com/openssl/openssl/pull/15086
+    if version == Version::Openssl3xx
+        && kind == "static"
+        && (env::var("CARGO_CFG_TARGET_OS").unwrap() == "linux"
+            || env::var("CARGO_CFG_TARGET_OS").unwrap() == "android")
+        && env::var("CARGO_CFG_TARGET_POINTER_WIDTH").unwrap() == "32"
+    {
+        println!("cargo:rustc-link-lib=dylib=atomic");
+    }
+
     if kind == "static" && target.contains("windows") {
         println!("cargo:rustc-link-lib=dylib=gdi32");
         println!("cargo:rustc-link-lib=dylib=user32");


openssl-sys/Cargo.toml
```diff
@@ -25,7 +25,7 @@ bssl-sys = { version = "0.1.0", optional = true }
 [build-dependencies]
 bindgen = { version = "0.64.0", optional = true, features = ["experimental"] }
 cc = "1.0.61"
-openssl-src = { version = "111", optional = true }
+openssl-src = { version = "300", optional = true, features = ["legacy"] }
```
amousset, line 28: Not sure if we want to enable this in the built openssl (or only use it for tests).

openssl-sys/Cargo.toml
```diff
@@ -25,7 +25,7 @@ bssl-sys = { version = "0.1.0", optional = true }
 [build-dependencies]
 bindgen = { version = "0.64.0", optional = true, features = ["experimental"] }
 cc = "1.0.61"
-openssl-src = { version = "111", optional = true }
+openssl-src = { version = "300.1.2", optional = true, features = ["legacy"] }
```
amousset, line 28: Legacy feature is required for the tests at least.

openssl-sys/build/main.rs
```diff
@@ -115,6 +115,16 @@ fn main() {
...
     }
 
+    // https://github.com/openssl/openssl/pull/15086
+    if version == Version::Openssl3xx
+        && kind == "static"
+        && (env::var("CARGO_CFG_TARGET_OS").unwrap() == "linux"
+            || env::var("CARGO_CFG_TARGET_OS").unwrap() == "android")
+        && env::var("CARGO_CFG_TARGET_POINTER_WIDTH").unwrap() == "32"
+    {
+        println!("cargo:rustc-link-lib=dylib=atomic");
```
amousset, line 125: When upgrading to 3.1, vendored tests fail with `undefined reference to `__atomic_stuff'` link errors, fixed by re-applying the patch removed in 32a303a752732fdd27c407a3a7fd668e546c05db. I did not dig the problem deeper.

amousset:
@sfackler Fixed test issues, ready for review.

openssl-sys/Cargo.toml
sfackler: It does seem like we realistically probably want to enable this - thoughts @alex?

openssl-sys/build/main.rs
sfackler: Hmm, I guess I'll need to poke around a bit to see if the commit that fixed this was reverted or something.

This is also something that may actually be best off moved to openssl-src.

sfackler: Ah, I think we just need to update the logic in openssl-src to include `atomic` as a lib when appropriate: https://github.com/alexcrichton/openssl-src-rs/blob/main/src/lib.rs#L549

openssl-sys/Cargo.toml
alex: We enable it by default in pyca/cryptography, its required for basically any encrypted key serialization (since they all use garbage algorithms).

openssl-sys/build/main.rs
amousset: The libs to link don't seem to come from openssl-src (only the library path is [used](https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/build/find_vendored.rs#L13)) but are [redefined in openssl-sys](https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/build/main.rs#L109). We could try to take lib list from openssl-src but it would still require defining them for non-vendored cases so I'm not sure it would make sense.

amousset: @sfackler We're new ~ 1 week from 1.1 EOL, is the current PR acceptable or do you want additional changes? As stated in the previous comment it does not seem practical to include the atomic lib on the openssl-src side.

BlackDex: I Know this is merged already, but i wanted to provide my feedback.
This now broke armv6 `arm-unknown-linux-gnueabi` again, for which this was removed.
If this really is needed for vendored tests, then maybe only enable this on vendored tests, or, as i do this in my build scripts add it when needed like this `RUSTFLAGS="-Clink-args=-latomic"`.

PR #1928, opened 2023-05-17
simplify converting asn1string

huettner94:
in combination with the previous commit this makes it easier to get an asn1string from a `&str` and convert it to a der encoding with a asn1 type tag.

Note that this PR depends on https://github.com/sfackler/rust-openssl/pull/1915 as they require in some places the same ffi functions.


openssl/src/asn1.rs
```diff
@@ -430,6 +431,34 @@ foreign_type_and_impl_send_sync! {
     pub struct Asn1StringRef;
 }
 
+impl Asn1String {
+    /// Converts the Asn1String to the ASN.1 representation with a IA5String tag
+    pub fn as_asn1type_der(&self) -> Result<Vec<u8>, ErrorStack> {
+        unsafe {
+            let typ = cvt_p(ffi::ASN1_TYPE_new()).unwrap();
+            ffi::ASN1_TYPE_set(typ, ffi::V_ASN1_IA5STRING, self.0.cast());
```
sfackler, line 439: I prefer minimizing the number of operations individual methods are performing. E.g. this should probably be one method to create an asn1type, one to fill it with the string, one to turn that into der.

huettner94: I split it now into `TryFrom<Asn1String>` and a function `as_der`. 

They are defined on the Struct `Asn1TypeWithValue`. I choose that name as there already seems to be a struct `Asn1Type` defined, but it holds the potential asn1 types, and does not wrap the `ASN1_TYPE` of openssl. 

Would you have a recommendation for the new struct name as i honestly do not like that one

PR #1941, opened 2023-05-31
Remove converting PKCS#8 passphrase to CString

alex:
It's not required, there's an explicit length.

### DIFF
--- a/openssl/src/pkey.rs
+++ b/openssl/src/pkey.rs
@@ -57,7 +57,7 @@ use cfg_if::cfg_if;
 use foreign_types::{ForeignType, ForeignTypeRef};
 use libc::{c_int, c_long};
 use openssl_macros::corresponds;
-use std::convert::TryFrom;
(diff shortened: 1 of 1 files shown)


openssl/src/pkey.rs
```diff
@@ -362,14 +358,12 @@ where
...
             let bio = MemBio::new()?;
-            let len = passphrase.len();
-            let passphrase = CString::new(passphrase).unwrap();
             cvt(ffi::i2d_PKCS8PrivateKey_bio(
                 bio.as_ptr(),
                 self.as_ptr(),
                 cipher.as_ptr(),
                 passphrase.as_ptr() as *const _ as *mut _,
-                len as ::libc::c_int,
+                passphrase.len() as ::libc::c_int,
```
sfackler, line 366: We should use try_into here to panic instead of overflowing on comically large passphrases

sfackler (approved):

Issue #1944, opened 2023-06-03
Building with BoringSSL should set ossl110, etc.


davidben:
In C, BoringSSL defines `OPENSSL_VERSION_NUMBER` to match OpenSSL 1.1.1 because, by default, you're expected to use the code you use for OpenSSL 1.1.1. Occasionally things vary from the target version, so we have `OPENSSL_IS_BORINGSSL`, but the default is to match 1.1.1.

rust-openssl's initial BoringSSL port wasn't done right. Rather than following this standard pattern, it defaults to giving BoringSSL the code for the very oldest OpenSSL! This has caused a lot of mess over the short period of time this port has existed:
* 0194e3f9decf0820615ce5b70f26433ac15eaba7
* 0257e2611d01127607b724a043642b01adf41706
* 2fe8b94066f1063ec78b0502052e4558379514a0
* 65a75a818f280ed578e9e68f7d6c1ca203b10e6f (the dh.rs bugfix)
* 3f68c0e5c77bd27ece67eb589ac71fc734fffe5b
* d3e557cf4836e49d97f38dcf2b349b8e7c30d9a8

We're still not done. The code below was deprecated years before the port existed, and will break very soon. Also it means rust-openssl loses [this fix](https://boringssl.googlesource.com/boringssl/+/9939e14cffc66f9b9f3374fb52c97bd8bfb0bfbe).
https://github.com/sfackler/rust-openssl/blob/master/openssl/src/rsa.rs#L584
https://github.com/sfackler/rust-openssl/blob/master/openssl/src/dsa.rs#L317

I'll upload a PR to do a point fix there, but that is merely fixing a symptom of a structural flaw in the port. The true fix is for rust-openssl to follow the supported pattern in C. The default for OpenSSL derivatives should be to target the OpenSSL version they advertise.

That'll simplify a lot of the `cfg(ossl110, boringssl)` lines. Though we'll need to clean up some tech debt in the process:

* Adding `cfg(not(boringssl))` on features we don't support (bindings libraries are, alas, a pain point because they tend to expose every feature that exists, even if not used in practice)
* In some places, we'll probably need to add some compatibility functions to BoringSSL
* There's some mess with `__fixed_rust` suffixes on function callbacks which we may need to replicate in `bssl-sys`. (The old names have been deprecated for a while... is it time to unwind that mess?)

[...]

PR #1945, opened 2023-06-03
Fix some deprecated patterns when using BoringSSL

davidben:
The RSA and DSA changes will be needed to avoid build breakage soon. The others are mostly tidying up. There's another place around BIO that we'd ideally also switch over, but that depends on resolving the __fixed_rust mess first.

This addresses a symptom of #1944, but not the root cause.

(@maurer FYI. Android will need to pick this patch up to avoid causing another problem with BoringSSL updates.)

### DIFF
--- a/openssl/src/asn1.rs
+++ b/openssl/src/asn1.rs
@@ -738,7 +738,7 @@ impl fmt::Debug for Asn1ObjectRef {
 }
 
 cfg_if! {
-    if #[cfg(any(ossl110, libressl273))] {
+    if #[cfg(any(ossl110, libressl273, boringssl))] {
         use ffi::ASN1_STRING_get0_data;
     } else {
         #[allow(bad_style)]
--- a/openssl/src/dsa.rs
+++ b/openssl/src/dsa.rs
@@ -7,6 +7,7 @@
 
 use cfg_if::cfg_if;
 use foreign_types::{ForeignType, ForeignTypeRef};
+#[cfg(not(boringssl))]
 use libc::c_int;
 use std::fmt;
 use std::mem;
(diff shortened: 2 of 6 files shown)


davidben:
CI failures are odd. They are with BoringSSL, but I'm also confused how they could be caused by this PR:

```
error: ambiguous glob re-exports
  --> openssl-sys/src/lib.rs:17:9
   |
17 | pub use libc::*;
   |         ^^^^^^^ the name `FILE` in the type namespace is first re-exported here
...
36 | pub use boringssl::*;
   |         ------------ but the name `FILE` in the type namespace is also re-exported here
   |
   = note: `-D ambiguous-glob-reexports` implied by `-D warnings`

error: ambiguous glob re-exports
  --> openssl-sys/src/lib.rs:17:9
   |
17 | pub use libc::*;
   |         ^^^^^^^ the name `timeval` in the type namespace is first re-exported here
...
36 | pub use boringssl::*;
   |         ------------ but the name `timeval` in the type namespace is also re-exported here

error: ambiguous glob re-exports
  --> openssl-sys/src/lib.rs:17:9
   |
17 | pub use libc::*;
   |         ^^^^^^^ the name `tm` in the type namespace is first re-exported here
...
36 | pub use boringssl::*;
   |         ------------ but the name `tm` in the type namespace is also re-exported here

error: ambiguous glob re-exports
  --> openssl-sys/src/lib.rs:17:9
   |
17 | pub use libc::*;
   |         ^^^^^^^ the name `time_t` in the type namespace is first re-exported here
...
36 | pub use boringssl::*;
   |         ------------ but the name `time_t` in the type namespace is also re-exported here
```

Is there perhaps an issue with bindgen over-binding things?

davidben:
I see a similar failure in #1946, so guessing this is not related to this PR and rather the CI broke.

alex (approved):

PR #1948, opened 2023-06-04
Fix warnings from BoringSSL on Rust 1.70

alex:


openssl-sys/build/run_bindgen.rs
```diff
@@ -110,11 +110,15 @@ pub fn run_boringssl(include_dirs: &[PathBuf]) {
...
         .ctypes_prefix("::libc")
+        .raw_line("use libc::*;")
         .derive_default(false)
         .enable_function_attribute_detection()
         .default_macro_constant_type(MacroTypeVariation::Signed)
         .rustified_enum("point_conversion_form_t")
         .allowlist_file(".*/openssl/[^/]+\\.h")
+        .allowlist_recursively(false)
+        .blocklist_function("BIO_vsnprintf")
+        .blocklist_function("OPENSSL_vasprintf")
```
sfackler, line 121: What was the issue with the varargs functions?

alex: Same as for vanilla (these are skipped there too) -- they generate a `va_list` argument which doesn't have a definition in `libc`.

PR #1953, opened 2023-06-04
Pick up OPENSSL_NO_* with BoringSSL

davidben:
This PR is (I expect) *not* yet ready to merge. It probably breaks the bssl-sys path, which unfortunately diverges even further from upstream BoringSSL's supported path than the bindgen path. (Android's rust-openssl story was, alas, not set up with a clear enough understanding of BoringSSL's needs. CC @maurer) I suspect we'll need to figure out a cleaner story for how cfg options flow from C libraries up through to crates that use them.

In the meantime, I've uploaded this for reference and to confirm that other configurations work in CI the way I expect them to.

---

BoringSSL defines OpenSSL-compatible OPENSSL_NO_* values to aid porting. This removes a ton of boringssl cfg lines that weren't actually necessary. This addresses part of issue #1944, though not the version number half.

### DIFF
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -226,6 +226,11 @@ See rust-openssl documentation for more information:
         }
     }
 
+    for enabled in &enabled {
+        println!("cargo:rustc-cfg=osslconf=\"{}\"", enabled);
+    }
+    println!("cargo:conf={}", enabled.join(","));
+
     if is_boringssl {
         println!("cargo:rustc-cfg=boringssl");
         println!("cargo:boringssl=true");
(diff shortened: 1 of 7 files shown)


alex:
Conceptually looks ok, but there seem to be a few methods not working.

davidben:
Yeah, I think the issue is the bssl-sys path doesn't have any way to ingest BoringSSL's opensslconf.h right now. Will need to ponder that, probably over the week. (That Android doesn't run build.rs and instead just specifies the final cfgs in the build file is going to be a challenge.)

May be that we need to resolve that divergence first, I dunno.

openssl-sys/build/main.rs
```diff
@@ -226,6 +226,11 @@ See rust-openssl documentation for more information:
         }
     }
 
+    for enabled in &enabled {
+        println!("cargo:rustc-cfg=osslconf=\"{}\"", enabled);
+    }
+    println!("cargo:conf={}", enabled.join(","));
+
```
alex, line 233: I think if you move the contents of `check_ssl_kind` to right here, you'll get what you want, functionally.

alex:
Not at the rustc level, AFAIK, all cross-crate coordination is done via
cargo.

alex:
This ended up happening a while ago.

PR #1954, opened 2023-06-04
Fixed type mutability on asn1_flag

alex:


reaperhulk:
(Needs ec.rs:1278 to change variable decl type too)

reaperhulk (approved):

Issue #1965, opened 2023-06-19
Segfault with empty `String` domain


DaanA32:
A segfault occurs in the following (code adapted from [example/google-connect.rs](https://github.com/sfackler/rust-native-tls/blob/36871ea6ab007a55b8d8abd43d5dd4ec655793fd/examples/google-connect.rs)):

It does work when the `String` is not empty.

```rust
extern crate native_tls;

use native_tls::TlsConnector;
use std::io::{Read, Write};
use std::net::TcpStream;

fn main() {
    let connector = TlsConnector::builder()
        .use_sni(false)
        .build()
        .unwrap();
    let domain = String::from("");
    // if domain is &'static str it works
    let stream = TcpStream::connect("google.com:443").unwrap();
    let mut stream = connector.connect(domain.as_str(), stream).unwrap();

    stream.write_all(b"GET / HTTP/1.0\r\n\r\n").unwrap();
    let mut res = vec![];
    stream.read_to_end(&mut res).unwrap();
    println!("{}", String::from_utf8_lossy(&res));
}
```

Output:
```
    Finished dev [unoptimized + debuginfo] target(s) in 0.39s
     Running `target/debug/examples/google-segfault`
[1]    136936 segmentation fault (core dumped)  cargo run --example google-segfault
```

Valgrind log
```
==135192== Memcheck, a memory error detector
==135192== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.
==135192== Using Valgrind-3.21.0 and LibVEX; rerun with -h for copyright info
==135192== Command: target/debug/examples/google-segfault
==135192== 
==135192== Invalid read of size 1
==135192==    at 0x4847EC6: strlen (vg_replace_strmem.c:501)
==135192==    by 0x4B74C85: int_x509_param_set_hosts (x509_vpm.c:46)
==135192==    by 0x12CD5B: openssl::x509::verify::X509VerifyParamRef::set_host (verify.rs:123)
==135192==    by 0x12F25E: openssl::ssl::connector::setup_verify_hostname (connector.rs:400)
==135192==    by 0x12F043: openssl::ssl::connector::ConnectConfiguration::into_ssl (connector.rs:188)
==135192==    by 0x11F610: openssl::ssl::connector::ConnectConfiguration::connect (connector.rs:201)
==135192==    by 0x11A15A: native_tls::imp::TlsConnector::connect (openssl.rs:344)
[...]

PR #1967, opened 2023-06-19
Fix segfault [issue#1965]

DaanA32:
Fix segfault [issue#1965](https://github.com/sfackler/rust-openssl/issues/1965)

### DIFF
--- a/openssl/src/ssl/connector.rs
+++ b/openssl/src/ssl/connector.rs
@@ -395,9 +395,10 @@ cfg_if! {
 
             let param = ssl.param_mut();
             param.set_hostflags(X509CheckFlags::NO_PARTIAL_WILDCARDS);
-            match domain.parse() {
-                Ok(ip) => param.set_ip(ip),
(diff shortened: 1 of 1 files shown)


openssl/src/ssl/connector.rs
```diff
@@ -395,9 +395,10 @@ cfg_if! {
 
             let param = ssl.param_mut();
             param.set_hostflags(X509CheckFlags::NO_PARTIAL_WILDCARDS);
-            match domain.parse() {
-                Ok(ip) => param.set_ip(ip),
-                Err(_) => param.set_host(domain),
+            match (domain.parse(), domain.len()) {
+                (Ok(ip), _) => param.set_ip(ip),
+                (Err(_), 0) => param.set_host(""),
```
sfackler, line 400: The problem is in set_host, not here.

DaanA32: Agreed :) This is a simple solution, but does not fix the core issue.

PR #1972, opened 2023-06-21
Support AES wrap and wrap_pad

mingweishih:
This PR adds the support of AES wrap (RFC 3394, requires OpenSSL 1.0.2) and wrap_pad (RFC 5649 requires OpenSSL 1.1.0) to `Cipher`.

OpenSSL currently requires user to explicitly opt-in the feature by setting the `EVP_CIPHER_CTX_FLAG_WRAP_ALLOW` flag in `EVP_CIPHER_CTX` before `EVP_Encryptinit_ex`/`EVP_Decryptinit_ex`. To support this mechanism, this PR also exposes the `EVP_CTX_set_flags` function that takes a bitflags value as an argument (currently only supports `EVP_CIPHER_CTX_FLAG_WRAP_ALLOW` flag).

### DIFF
--- a/openssl/src/cipher_ctx.rs
+++ b/openssl/src/cipher_ctx.rs
@@ -55,6 +55,8 @@ use crate::error::ErrorStack;
 #[cfg(not(boringssl))]
 use crate::pkey::{HasPrivate, HasPublic, PKey, PKeyRef};
 use crate::{cvt, cvt_p};
+#[cfg(ossl102)]
+use bitflags::bitflags;
 use cfg_if::cfg_if;
 use foreign_types::{ForeignType, ForeignTypeRef};
 use libc::{c_int, c_uchar};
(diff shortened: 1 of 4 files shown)


openssl/src/cipher_ctx.rs
```diff
@@ -509,6 +509,14 @@ impl CipherCtxRef {
         Ok(())
     }
 
+    /// Set ctx flags.
+    #[corresponds(EVP_CIPHER_CTX_set_flags)]
+    pub fn set_flags(&mut self, flags: i32) {
```
sfackler, line 514: If this is going to be publicly exposed, it should take a strongly-typed bitflags value.

openssl/src/symm.rs
```diff
@@ -503,6 +539,11 @@ impl Crypter {
             Mode::Decrypt => CipherCtxRef::decrypt_init,
         };
 
+        if cfg!(ossl102) {
+            // Explicitly opt in the AES key wrap feature
+            ctx.set_flags(ffi::EVP_CIPHER_CTX_FLAG_WRAP_ALLOW);
```
sfackler, line 544: I don't think this should be done automatically.

openssl/src/cipher_ctx.rs
mingweishih: Changed to private, which is sufficient for this PR.

openssl/src/symm.rs
mingweishih: I updated to only set the flag if it's an aes wrap `cipher`. Does this work for you?

sfackler: I don't think this should be done automatically.

openssl/src/cipher_ctx.rs
mingweishih: I made the function public and make the flags as bitflags vlaue.

openssl/src/symm.rs
mingweishih: Reverted the change in symm.

mingweishih:
I addressed the comments. Would the current change work for you? I need this feature for a project. Hopefully it can go in for the upcoming release. Thanks!

sfackler:
Thanks!

PR #1973, opened 2023-06-21
bn: Add mod_sqrt

anpage:
This is a naive implementation of `mod_sqrt` for BigNum. The [docs for OpenSSL](https://www.openssl.org/docs/man1.1.1/man3/BN_mod_sqrt.html) say `BN_mod_sqrt` allocates a new BIGNUM for the return value, but the code suggests this is only the case when `in` is `NULL`. Otherwise, it seems to operate on `in` directly.

I needed this for a project and thought I'd submit what I came up with.

### DIFF
--- a/openssl-sys/src/handwritten/bn.rs
+++ b/openssl-sys/src/handwritten/bn.rs
@@ -73,6 +73,13 @@ extern "C" {
         m: *const BIGNUM,
         ctx: *mut BN_CTX,
     ) -> c_int;
+    #[cfg(ossl110)]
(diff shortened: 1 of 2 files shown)


sfackler:
Can you add a test?

anpage:
I added a simple test with some arbitrary constants (calculated by squaring a prime for `s` and an arbitrary other prime for `p`). It matches the result when calling openssl directly in C.

Is that sufficient?

sfackler:
Yep! Just want something basic to make sure it doesn't trivially segfault and produces the expected output.

PR #1977, opened 2023-06-23
Correctly handle errors being on the stack when EVP_PKEY_verify returns 0

alex:


sfackler:
This seems like a boringssl bug, yeah? Is it filed upstream?

alex:
I believe this is intentional behavior by BoringSSL. /cc @davidben

davidben:
Neither the PR, nor the explanation look correct.

Are you seeing BoringSSL report an error with r == 1 or r == 0? The comment suggests r == 1 ("without EVP_PKRY_verify [sic] returning an error"), which should not happen and would be a bug. If it does so with r == 0, that's natural and to be expected because r == 0 indicates an error. In that case, the comment here is wrong.

Also the code in the PR can't be right because you should *never* condition on the presence of things in the error queue. That's a perfect way to get bugs when other functions leave stuff on the queue. In some APIs, you're forced to, but then you need to be extra careful to clear the error queue first.

Reading between the lines, I gather this is because rust-openssl believes -1 is failing to verify a signature while 0 is "successfully" rejecting a signature.  BoringSSL intentionally rejects this notion for safety reasons. It is not worth the risk of critical security bugs to somehow distinguish between failing to verify a signature and successfully rejecting a signature. BoringSSL simply returns 0 on error and treats all forms of invalid input as an error.

From the discussion on the error queue, I gather rust-openssl believes OpenSSL's behavior supports this three-way split. This is not really true. OpenSSL does have a tri-state negative/zero/one return value, but the error queue behavior is not what rust-openssl thinks. It's a bit more haphazard:

* `EVP_PKEY_verify` for EC keys calls `ECDSA_verify`: https://github.com/openssl/openssl/blob/OpenSSL_1_1_1u/crypto/ec/ec_pmeth.c#L146
  * It's true that `ECDSA_verify` returns -1 with error queue in most error conditions and 0 without error queue on verification failure: https://github.com/openssl/openssl/blob/OpenSSL_1_1_1u/crypto/ec/ecdsa_ossl.c#L317
  * But `ECDSA_verify` also considers an unparsable signature to be failing to verify the signature, not successfully rejecting an invalid signature. So if you were hoping to use this distinction for "bad input from peer" vs "internal error", you can't. There is no reason to ever behavior differently on unparsable signature and parsable signature with bad values. It's all a bag of bytes at the end of the day. https://github.com/openssl/openssl/blob/OpenSSL_1_1_1u/crypto/ec/ecdsa_ossl.c#L304
* But, for RSA keys... https://github.com/openssl/openssl/blob/OpenSSL_1_1_1u/crypto/rsa/rsa_pmeth.c#L251
  * Sometimes it just maps all random errors to zero: https://github.com/openssl/openssl/blob/OpenSSL_1_1_1u/crypto/rsa/rsa_pmeth.c#L268
  * Sometimes it preserves the low-level 0 vs -1 distinction... https://github.com/openssl/openssl/blob/OpenSSL_1_1_1u/crypto/rsa/rsa_pmeth.c#L251
    * But said low-level function actually just always returns zero and pushes to the error queue, even when you "successfully" reject an invalid signature: https://github.com/openssl/openssl/blob/OpenSSL_1_1_1u/crypto/rsa/rsa_sign.c#L220

So what we have here is just a huge mess and trying to distinguish the error cases is futile. In BoringSSL, `EVP_PKEY_verify` will behave uniformly and follow the "failure conditions always do or always don't push to the error queue" rule. OpenSSL doesn't, so you may need to mess with it at the edges.

As for what to do, I guess this depends on whether you want to be consistently returning an error or consistently returning a successful false. If you want the former, change `cvt_n` to `cvt`. If you want the latter, you can keep `cvt_n` I guess, but then you probably need to clear the error queue because, in both BoringSSL and OpenSSL, returning zero can leave stuff on the error queue. BoringSSL is consistent about this because we simply use zero as the error value. OpenSSL is inconsistent, but does this for RSA.

alex:
Errors on the stack are observed with return code == 0 on BoringSSL. As you note, this is because BoringSSL behaves differently than OpenSSL, with respect to ECDSA.

That said, I don't think this is a case of what rust-openssl believes, so much as it having an unfortunate API inherited from OpenSSL. See the return values section: https://www.openssl.org/docs/man3.1/man3/EVP_PKEY_verify.html.

Not for nothing, but I agree with you that this is all a little ridiculous, once you've got the bytes, either they form a valid signature or not. Whether they're valid DER or not _must_ be irrelevant to all sane callers. Alas, legacy.

Based on your observations about RSA keys, I think the existing code is simply wrong, as it will leave errors on the stack. We now must choose between:

- Return `Err(..)` in more cases
- Return `Ok(false)` in more cases

Left to my own devices, I'd probably go with the latter, so that's what I'm going to have this PR do. But I'm happy to take feedback and go the other way.

davidben:
Yeah, OpenSSL's documentation tries to claim this tri-state thing, but it doesn't match reality. :-) There's the RSA mess, but also ECDSA doesn't match the docs. If it cannot parse the ECDSA signature, it returns -1 w/ error queue. By the docs, it should return zero. A byte sequence that cannot be parsed as `SEQUENCE { INTEGER, INTEGER }` is as much an invalid signature as a byte sequence which parses but whose integer values do not form a valid signature. They're both "the other side sent me a byte string, and that byte string isn't a valid signature".

openssl/src/pkey_ctx.rs
```diff
@@ -210,13 +210,22 @@ where
...
-            ))?;
+            );
+            // `EVP_PKEY_verify` is not terribly consistent about how it,
+            // reports errors. It does not clearly distinguish between 0 and
+            // -1, and may put errors on the stack in both cases.
+            // As a practical matter, there's no reasonable need for a caller
+            // to distinguish between an "error" (which often means things
+            // like failing to parse the DER bytes of the signature) and a
+            // signature verification, so treat them all the same.
+            ErrorStack::get();
```
sfackler, line 227: It feels a bit unfortunate to throw away the errors here. We could potentially return the error stack if nonempty and r == 0, and Ok(false) otherwise? Not sure if that'd break existing users though.

alex: as a practical matter, I think its basically impossible to use this function without handling both. I can switch to the impl you suggested easily enough.

PR #1978, opened 2023-06-23
Add SSL_CTX_sess_number and friends

drbrain:
From https://www.openssl.org/docs/manmaster/man3/SSL_CTX_sess_number.html

These allow inspection of statistics for the internal and external session cache and the number of connections started, completed, and renegotiated for servers and clients.

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -1911,6 +1911,80 @@ impl SslContextRef {
     pub fn num_tickets(&self) -> usize {
         unsafe { ffi::SSL_CTX_get_num_tickets(self.as_ptr()) }
     }
+
+    /// Gets the current number of sessions in the internal session cache.
+    #[corresponds(SSL_CTX_sess_number)]
+    pub fn get_session_count(&self) -> i64 {
+        unsafe { ffi::SSL_CTX_sess_number(self.as_ptr()) as i64 }
+    }
+
+    /// Gets the number of started SSL/TLS handshakes in client mode
+    #[corresponds(SSL_CTX_sess_connect)]
+    pub fn get_connect_count(&self) -> i64 {
+        unsafe { ffi::SSL_CTX_sess_connect(self.as_ptr()) as i64 }
+    }
+
+    /// Gets the number of successfully established SSL/TLS handshakes in client mode
+    #[corresponds(SSL_CTX_sess_connect_good)]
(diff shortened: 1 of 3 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -1911,6 +1911,80 @@ impl SslContextRef {
...
+
+    /// Gets the number of started SSL/TLS handshakes in client mode
+    #[corresponds(SSL_CTX_sess_connect)]
+    pub fn get_connect_count(&self) -> i64 {
+        unsafe { ffi::SSL_CTX_sess_connect(self.as_ptr()) as i64 }
+    }
+
+    /// Gets the number of successfully established SSL/TLS handshakes in client mode
+    #[corresponds(SSL_CTX_sess_connect_good)]
+    pub fn get_connect_good_count(&self) -> i64 {
```
sfackler, line 1929: Rust APIs typically exclude the `get_` prefix.

sfackler: I also usually like to mirror the OpenSSL function names in the method names, though in this case `ctx.sess_connect_good()` is a bit of a weird name, so I don't know.

sfackler:
You'll probably just have to disable the clippy warning on each method.

openssl/src/ssl/mod.rs
drbrain: I'll remove the `get_`

I wasn't sure about the names either. I'm fine matching OpenSSL names more-exactly, but I also find them weird when they're for statistics purposes.

drbrain: `get_` removed in 164528c4948e507f18ef7aaff3763a2baa48fbe2

Let me know how you feel about the names

drbrain:
> You'll probably just have to disable the clippy warning on each method.

I needed to remove the `as i64`, fixed in 081de7c2f301e866183c5d4b581a2488dbc25b99

drbrain:
> I needed to remove the `as i64`, fixed in [081de7c](https://github.com/sfackler/rust-openssl/commit/081de7c2f301e866183c5d4b581a2488dbc25b99)

haha, oh, I see now

drbrain:
Clippy disabled in 5da45a2ef29cbb286841f40817d55d7139f5bcf5

drbrain:
I think this is ready for re-review.

drbrain:
Is there anything outstanding I need to do before this PR is ready to merge?

PR #1982, opened 2023-06-27
Allow running vcpkg on any Windows target

micolous:
`vcpkg-rs` doesn't support `-pc-windows-gnu` (mingw) _yet_, but its integration tests will try to build `openssl-sys`, which currently prevents using `vcpkg` for non-`msvc` targets.

This change should have no effect for `openssl-sys` right now, but [unblocks that work](https://github.com/mcgoo/vcpkg-rs/pull/52).

### DIFF
--- a/openssl-sys/Cargo.toml
+++ b/openssl-sys/Cargo.toml
@@ -27,8 +27,6 @@ bindgen = { version = "0.64.0", optional = true, features = ["experimental"] }
 cc = "1.0.61"
 openssl-src = { version = "111", optional = true }
 pkg-config = "0.3.9"
-
-[target.'cfg(target_env = "msvc")'.build-dependencies]
 vcpkg = "0.2.8"
 
 # We don't actually use metadeps for annoying reasons but this is still here for tooling
--- a/openssl-sys/build/find_normal.rs
+++ b/openssl-sys/build/find_normal.rs
@@ -232,8 +232,12 @@ fn try_pkg_config() {
 ///
 /// Note that if this succeeds then the function does not return as vcpkg
 /// should emit all of the cargo metadata that we need.
-#[cfg(target_env = "msvc")]
 fn try_vcpkg() {
+    let target = env::var("TARGET").unwrap();
+    if !target.contains("windows") {
+        return;
+    }
+
     // vcpkg will not emit any metadata if it can not find libraries
     // appropriate for the target triple with the desired linkage.
 
@@ -256,9 +260,6 @@ fn try_vcpkg() {
 
     process::exit(0);
 }
-
-#[cfg(not(target_env = "msvc"))]
-fn try_vcpkg() {}
 
 fn execute_command_and_get_output(cmd: &str, args: &[&str]) -> Option<String> {
     let out = Command::new(cmd).args(args).output();
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -4,7 +4,6 @@ extern crate cc;
 #[cfg(feature = "vendored")]
 extern crate openssl_src;
 extern crate pkg_config;
-#[cfg(target_env = "msvc")]
 extern crate vcpkg;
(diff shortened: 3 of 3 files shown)


openssl-sys/Cargo.toml
```diff
@@ -27,8 +27,6 @@ bindgen = { version = "0.64.0", optional = true, features = ["experimental"] }
 cc = "1.0.61"
 openssl-src = { version = "111", optional = true }
 pkg-config = "0.3.9"
-
-[target.'cfg(target_env = "msvc")'.build-dependencies]
```
sfackler, line 31: We should probably switch this over to `cfg(target_os = "windows")` rather than making it an unconditional dependency.

micolous: Unfortunately, that doesn't do what you expect: [`target_os = "windows"` matches the _host_ platform for `build-dependencies` and `build.rs`, not the target][0]:

> ```
> [target.'cfg(unix)'.build-dependencies]
> cc = "1.0.3"
> ```
>
> In this case, the dependency will only be built when the host platform matches the specified target.

So if you apply that change:

```diff
diff --git a/openssl-sys/Cargo.toml b/openssl-sys/Cargo.toml
index 9b102fa8..81b493d0 100644
--- a/openssl-sys/Cargo.toml
+++ b/openssl-sys/Cargo.toml
@@ -27,6 +27,8 @@ bindgen = { version = "0.64.0", optional = true, features = ["experimental"] }
 cc = "1.0.61"
 openssl-src = { version = "111", optional = true }
 pkg-config = "0.3.9"
+
+[target.'cfg(target_os = "windows")'.build-dependencies]
 vcpkg = "0.2.8"
 
 # We don't actually use metadeps for annoying reasons but this is still here for tooling
diff --git a/openssl-sys/build/main.rs b/openssl-sys/build/main.rs
index 21ccf3d0..8a240dba 100644
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -4,6 +4,7 @@ extern crate cc;
 #[cfg(feature = "vendored")]
 extern crate openssl_src;
 extern crate pkg_config;
+#[cfg(target_os = "windows")]
 extern crate vcpkg;
 
 use std::collections::HashSet;
```

and then try to build from a non-Windows host, you get:

```
% RUST_BACKTRACE=1 cargo build --target x86_64-pc-windows-gnu --manifest-path systest/Cargo.toml  --release
   Compiling openssl-sys v0.9.90 (/Users/michael/dev/rust-openssl/openssl-sys)
   Compiling curl-sys v0.4.63+curl-8.1.2 (/Users/michael/dev/curl-rust/curl-sys)
   Compiling libz-sys v1.1.9 (/Users/michael/dev/libz-sys)
error[E0433]: failed to resolve: use of undeclared crate or module `vcpkg`
   --> /Users/michael/dev/rust-openssl/openssl-sys/build/find_normal.rs:244:21
    |
244 |     let lib = match vcpkg::Config::new()
    |                     ^^^^^ use of undeclared crate or module `vcpkg`
    |
help: consider importing this struct
    |
1   + use pkg_config::Config;
    |
help: if you import `Config`, refer to it directly
    |
244 -     let lib = match vcpkg::Config::new()
244 +     let lib = match Config::new()
    |

For more information about this error, try `rustc --explain E0433`.
error: could not compile `openssl-sys` (build script) due to previous error
```

Instead you'd need [to peek at environment variables][1], which you can't do from that point, so you unfortunately need to unconditionally pull in that dependency. Fortunately, you can build it just fine on non-Windows platforms and for non-Windows targets, this PR will just never run it.

[0]: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#build-dependencies
[1]: https://kazlauskas.me/entries/writing-proper-buildrs-scripts

sfackler: Oh right yeah

PR #1983, opened 2023-06-27
OpenBSD case set openss_dir to /usr

jpcw:
According with @alex This PR is related to https://github.com/pyca/cryptography/issues/9144#issuecomment-1609436924

### DIFF
--- a/openssl-sys/build/find_normal.rs
+++ b/openssl-sys/build/find_normal.rs
@@ -97,7 +97,7 @@ fn find_openssl_dir(target: &str) -> OsString {
         return OsString::from("/usr");
     }
 
-    // OpenBSD ships with OpenSSL but doesn't include a pkg-config file :(
+    // OpenBSD ships with LibreSSL but doesn't include a pkg-config file :(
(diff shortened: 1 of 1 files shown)


openssl-sys/build/find_normal.rs
```diff
@@ -97,6 +97,11 @@ fn find_openssl_dir(target: &str) -> OsString {
         return OsString::from("/usr");
     }
 
+    // OpenBSD ships with LibreSSL but doesn't include a pkg-config file :(
+    if host == target && target.contains("openbsd") {
```
sfackler, line 101: Can we merge this with the freebsd case above?

jpcw:
seems the merged 'openbsd', freebsd', braks Linux with Libressl
first simple precise commit (and then more verbose) commit was good.
I think there is a mistake order wich fail linux with libre ssl, but i don't see the link.
Sorry but it's too hard for me now to purpose anything working, different than my first commit.

Plz revert to my first commit, or take the hand.

alex (approved):

PR #1998, opened 2023-07-19
Expose Poly1305 bindings on libressl and boringssl

facutuesca:
This PR exposes the Poly1305 API from LibreSSL ([source](https://github.com/libressl/openbsd/blob/master/src/lib/libcrypto/poly1305/poly1305.h)) and BoringSSL ([source](https://github.com/google/boringssl/blob/master/include/openssl/poly1305.h))

The API was already exposed for Boring through `bssl-sys` (when building with the `unstable_boringssl` feature), but was not present in normal (`bindgen`) builds. I did not add a handwritten version for BoringSSL, because from what I could see Boring is either built with `bindgen` or `bssl-sys` bindings, not the manually-written ones.

For LibreSSL, both the handwritten and the `bindgen` bindings should work now.

### DIFF
--- a/openssl-sys/CHANGELOG.md
+++ b/openssl-sys/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+* Expose `poly1305_state`, `CRYPTO_poly1305_init`, `CRYPTO_poly1305_update`, and `CRYPTO_poly1305_finish` on BoringSSL and LibreSSL.
+
 ## [v0.9.90] - 2023-06-20
 
 ### Fixed
--- a/openssl-sys/build/run_bindgen.rs
+++ b/openssl-sys/build/run_bindgen.rs
@@ -55,6 +55,10 @@ const INCLUDES: &str = "
 #if OPENSSL_VERSION_NUMBER >= 0x30000000
 #include <openssl/provider.h>
 #endif
+
+#if defined(LIBRESSL_VERSION_NUMBER) || defined(OPENSSL_IS_BORINGSSL)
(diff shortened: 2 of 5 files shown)


sfackler:
The LibreSSL build is red.

I am pretty suspicious of adding support of APIs that OpenSSL doesn't support to the rust-openssl library.

sfackler:
> it's just that you can also use `Poly1305` through the `EVP_Digest` API (which is not the case for Boring and Libre)

Right, so this PR is adding support of APIs that OpenSSL doesn't support.

woodruffw:
We just realized a hiccup: OpenSSL _does_ have these APIs and the symbols aren't hidden, but they're included in a header that isn't part of OpenSSL's normal build-installed headers (i.e.,  `crypto/poly1305.h`).

Given that, there's a sense in which these aren't part of the OpenSSL public API (even if they support them); apologies for the confusion.

@sfackler how would you prefer to move forwards here? We could probably bind directly to OpenSSL's symbols, but I wanted to confirm that that'd be okay with you as a matter of crate/library policy.

sfackler:
Ah that is unfortunate. I definitely wouldn't want to bind to non-public APIs.

Is there a reason you can't work through the EVP interface on all implementations?

sfackler:
smh

facutuesca:
Maybe that's a good reason to only expose the raw bindings in `openssl-sys`? `openssl-sys` already exposes this API when compiled with BoringSSL + `unstable_boring`. So in that sense, this API is already exposed, and this PR only adds support for it with Libre and Boring+`bindgen`

PR #2000, opened 2023-07-20
add EcPointRef::to_hex_str and EcPoint::from_hex_str

zh-jq:
https://www.openssl.org/docs/manmaster/man3/EC_POINT_point2hex.html

### DIFF
--- a/openssl-sys/src/handwritten/ec.rs
+++ b/openssl-sys/src/handwritten/ec.rs
@@ -152,6 +152,20 @@ extern "C" {
         ctx: *mut BN_CTX,
     ) -> c_int;
 
+    pub fn EC_POINT_point2hex(
+        group: *const EC_GROUP,
+        p: *const EC_POINT,
+        form: point_conversion_form_t,
+        ctx: *mut BN_CTX,
+    ) -> *mut c_char;
+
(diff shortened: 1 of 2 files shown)


openssl-sys/src/handwritten/ec.rs
```diff
@@ -152,6 +152,22 @@ extern "C" {
         ctx: *mut BN_CTX,
     ) -> c_int;
 
+    #[cfg(not(boringssl))]
```
sfackler, line 155: boringssl doesn't use these definitions so you don't need the cfgs.

openssl/src/ec.rs
```diff
@@ -631,6 +659,27 @@ impl EcPoint {
         }
         Ok(point)
     }
+
+    /// Creates point from a hexadecimal string representation
+    #[corresponds(EC_POINT_hex2point)]
+    #[cfg(not(boringssl))]
+    pub fn from_hex_str(
```
sfackler, line 666: I think we'd want this to take `&mut self` rather than being a static method taking the group.

zh-jq: The definition is the same as EcPoint::from_bytes. For a &mut self method, it seems the name set_from_hex_str is more suited.

PR #2002, opened 2023-07-24
Gate BIO_new_socket in OPENSSL_NO_SOCK in handwritten bindings

davidben:
OpenSSL conditions the availability of this symbol on `OPENSSL_NO_SOCK`. Match that in the handwritten bindings.

This doesn't really matter as it's only in the handwritten bindings and not used by rust-openssl, but we may as well make it match OpenSSL.


sfackler:
You'll also need to add it here so we can discover the #define properly: https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/build/expando.c

davidben:
> You'll also need to add it here so we can discover the #define properly: https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/build/expando.c

done

alex (approved):

sfackler (approved):

PR #2003, opened 2023-07-25
Add support for CRL extensions and the Authority Information Access e…

AdmiralGT:
We're looking to do CRL validation and as such, want to be able to access the Authority Information Access extension within the CRL to validate the chain of trust. 

I've used a similar model to the extensions as used in the `X509RevokedRef` and then added a AuthorityInformationAccess struct for the extension on the advice of @Skepfyr .

I've updated the `entry_extension.crl` to include a Authority Information Access extension. I've tried to keep the remainder as similar to the existing CRL as possible. I used a python script to create this CRL, not sure if that's worth adding to the PR such that others can easily update these with the required extensions. Do let me know.


openssl/src/x509/mod.rs
```diff
@@ -1915,6 +1926,36 @@ impl X509CrlRef {
     {
         unsafe { cvt_n(ffi::X509_CRL_verify(self.as_ptr(), key.as_ptr())).map(|n| n != 0) }
     }
+
+    /// Get the criticality and value of an extension.
+    ///
+    /// This returns None if the extension is not present or occurs multiple times.
+    #[corresponds(X509_CRL_get_ext_d2i)]
+    pub fn extension<T: ExtensionType>(&self) -> Result<Option<(bool, T::Output)>, ErrorStack> {
```
Skepfyr, line 1934: @sfackler This change looks good to me, I'd like to check that you're happy with carrying on with this approach to extensions?

Skepfyr (approved):

PR #2017, opened 2023-08-20
Add openssl::cipher_ctx::CipherCtx::clone

johntyner:
I have a use case that benefits (speed-wise) from being able to set/expand the key once in the context and then reuse the context via copies for multiple encryptions. This change adds the `clone()` function on the CipherCtx object which calls `EVP_CIPHER_CTX_copy()` behind the scenes.

### DIFF
--- a/openssl/src/cipher_ctx.rs
+++ b/openssl/src/cipher_ctx.rs
@@ -106,7 +106,7 @@ impl CipherCtx {
 
 impl CipherCtxRef {
     #[corresponds(EVP_CIPHER_CTX_copy)]
-    pub fn copy(&mut self, src: &CipherCtx) -> Result<(), ErrorStack> {
+    pub fn copy(&mut self, src: &CipherCtxRef) -> Result<(), ErrorStack> {
         unsafe {
             cvt(ffi::EVP_CIPHER_CTX_copy(self.as_ptr(), src.as_ptr()))?;
             Ok(())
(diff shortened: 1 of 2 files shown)


openssl/src/cipher_ctx.rs
```diff
@@ -102,6 +102,15 @@ impl CipherCtx {
             Ok(CipherCtx::from_ptr(ptr))
         }
     }
+
+    #[corresponds(EVP_CIPHER_CTX_copy)]
+    pub fn clone(&self) -> Result<Self, ErrorStack> {
+        let n = CipherCtx::new()?;
+        unsafe {
+            cvt(ffi::EVP_CIPHER_CTX_copy(n.as_ptr(), self.as_ptr()))?;
```
sfackler, line 110: I think I'd probably make this a `pub fn copy(&mut self, other: &CipherCtxRef) -> Result<(), ErrorStack>` to follow the OpenSSL API.

johntyner: Done.

openssl/src/cipher_ctx.rs
```diff
@@ -105,6 +105,14 @@ impl CipherCtx {
 }
 
 impl CipherCtxRef {
+    #[corresponds(EVP_CIPHER_CTX_copy)]
+    pub fn copy(&mut self, src: &CipherCtx) -> Result<(), ErrorStack> {
```
sfackler, line 109:
```suggestion
    pub fn copy(&mut self, src: &CipherCtxRef) -> Result<(), ErrorStack> {
```

johntyner: Oops. Thanks!

PR #2018, opened 2023-08-22
Add X509VerifyParam::set_email

dhouck:
This is substantially similar to the other X.509 verification parameter options like host.  I tested manually by building and verifying signatures both with and without the correct email address; I did not add unit tests because I am not very familiar with this crate and the nearest similar code did not have tests I could find.


sfackler:
Thanks!

PR #2021, opened 2023-08-26
Re-export ffi crate and foreign-types traits

### LINKED ISSUE #2020: Using openssl with openssl-sys
Thereʼs a function in the C OpenSSL library I might want to use, which doesnʼt have bindings in rust-openssl and where it probably doesnʼt make sense to add bindings (it is deliberately part of the API, but undocumented, and I bet very few people want to use it).

When I try to call `as_ptr` on the various openssl types implementing `ForeignType`, I get an error that I need to `use` the trait first.  I might be missing something, but I think this is not accessible through this crate, so I need to explicitly depend on *the correct version of* `foreign-types-shared` for it?

If thatʼs the case, then either there should be a better way to call those functions, or the trait should be re-exported; I shouldnʼt need to know internal implementation details to use functions documented in the public API.  Alternately, they should not be documented if my use case is not supported.

dhouck:
Closes #2020 

I also re-exported the `ffi` crate, for the same reason; I can remove that part if it isnʼt wanted, but it seems reasonable to me for the same reason I mention there.

At a quick glance I couldnʼt find other things that obviously looked like they should be re-exported.


dhouck:
I just confirmed that this PR does not touch the file Clippy [is complaining about](https://github.com/sfackler/rust-openssl/actions/runs/5983652926/job/16234016399?pr=2021), or indeed anything in `openssl-sys`.  I do not know why this PR fails that check when my previous PR [passed it](https://github.com/sfackler/rust-openssl/actions/runs/5945216540/job/16123903569?pr=2018) and the relevant files have not been changed since that previous PR was made.

The `rustfmt` check is entirely on me, though; fixing that now.

dhouck:
I cannot figure out why `clippy` started complaining now.  GitHub updated the test runner images, but the new one has the same version of all rust tools and of the openssl libraries.  I can see the issue itʼs complaining about has existed for two years.  I cannot get Clippy to complain on my machine, although that might just be that I should have been in bed hours ago.

In the morning, Iʼll make another PR to fix the problem Clippy is complaining about, then hopefully once thatʼs merged I can rebase this one and itʼll pass all checks.

romen:
@sfackler what is the status of this PR? Are there plans to merge it in the foreseeable future or is something holding it back?

PR #2024, opened 2023-08-26
Fix auto-fixable Clippy warnings

dhouck:
(Creating this, as I said I would in #2021)

This does not fix all the new Clippy warnings, assuming the set I see after I ran `rustup` this morning is the same set GitHub sees.  There are four warnings internal to the `bitflags!` macro which Iʼm not sure how to fix.  Clippy doesnʼt seem to have a good option to disable lints inside a macro, and disabling it globally seems excessive.  Bumping the `bitflags` crate to the latest version makes it stop compiling correctly; bumping it to the latest 1.x version compiles and passes all tests but does not get rid of the warning.

Thoughts?  Or should this be merged as-is and the rest handled separately by someone who knows this codebase better?


dhouck:
Never mind; you got there first.

PR #2032, opened 2023-08-30
Add reading X509 certificate extensions

wiktor-k:
Hi :wave: 

This PR adds `X509::get_ext` function for retrieving extensions out of X509 certificates as well as `X509Extension::data` function for reading the actual value.

I've also added test fixture - a certificate with extension to be read.

I've made it into a single function that reads the first extension with the given Nid (OID) but I'm open to suggestions if I should split them into two functions. (current approach allows reading only the first extension with given Nid but it's sufficient for my needs).


openssl/src/x509/mod.rs
```diff
@@ -425,6 +425,64 @@ impl X509Ref {
...
+    /// let extension = cert.get_ext_by_nid(nid).unwrap().unwrap();
+    /// let value = extension.data().as_slice();
+    /// assert_eq!(value, [4, 3, 5, 2, 7]);
+    /// ```
+    #[corresponds(X509_get_ext_by_NID)]
+    pub fn get_ext_by_nid(&self, nid: Nid) -> Result<Option<&X509ExtensionRef>, ErrorStack> {
+        unsafe {
+            let loc = ffi::X509_get_ext_by_NID(self.as_ptr(), nid.as_raw(), -1);
+            Ok(if loc >= 0 {
+                Some(X509ExtensionRef::from_ptr(cvt_p(ffi::X509_get_ext(
```
wiktor-k, line 447: I'm kind of not sure if `cvt_p` is necessary here or whether returned null pointer can just be converted into None (that'd remove the Result from function signature, making it simpler). The docs doesn't seem to be conclusive on this matter: [X509_get_ext ](https://www.openssl.org/docs/man3.1/man3/X509_get_ext.html) :/

wiktor-k:
Ping @alex. I hope you don't mind :)

wiktor-k:
@sfackler is there anything preventing this from being merged? Thanks! :wave:

PR #2042, opened 2023-09-12
[openssl-sys] Add X509_check_{host,email,ip,ip_asc} fns

jgallagher:
I took these definitions from `bindgen` output and double-checked them against https://www.openssl.org/docs/man1.0.2/man3/X509_check_host.html. The constants needed for `flags` are already defined (https://docs.rs/openssl-sys/latest/openssl_sys/?search=check_flag).


sfackler:
Thanks!

PR #2052, opened 2023-10-10
Fix clippy warnings produced by newer Rust

wiktor-k:


sfackler:
I'm assuming the hostent lints are coming from the bindgen builds. We should probably be excluding that from the codegen and pulling from libc explicitly?

wiktor-k:
> I'm assuming the hostent lints are coming from the bindgen builds. We should probably be excluding that from the codegen and pulling from libc explicitly?

Yep. That was a good lead, thank you!

(For the record local `cargo test --features bindgen` showed these warnings.)

alex (approved):

PR #2053, opened 2023-10-12
Add S-BGP extension handling

thillux:
This PR adds support for parsing SBGP ASNs and IP-Ranges. A basic test is also added.


wiktor-k:
FTR I think the job fails because of issues fixed in #2052 (just saying that it's nothing that you've introduced).

thillux:
> FTR I think the job fails because of issues fixed in #2052 (just saying that it's nothing that you've introduced).

Thank you for the hint!

thillux:
This currently depends on #2055. After it is merged, I'll cleanup my temporary commits.

thillux:
> This currently depends on #2055. After it is merged, I'll cleanup my temporary commits.
I'll found another fix for the compilation with Rust 1.56. This PR is ready now from my side.

openssl-sys/src/handwritten/x509_sbgp.rs
```diff
@@ -0,0 +1,222 @@
...
+     */
+    flags: c_int,
+}
+
+/*
+ * Expand the bitstring form of an address into a raw byte array.
+ * At the moment this is coded for simplicity, not speed.
+ */
+#[cfg(ossl110)]
+fn addr_expand(addr: *mut u8, bs: *const ASN1_BIT_STRING, length: isize, fill: u8) -> bool {
```
sfackler, line 150: This doesn't seem like something that should be in openssl-sys.

openssl/src/x509/extension.rs
```diff
@@ -428,6 +428,145 @@ impl AuthorityKeyIdentifier {
...
+        append(&mut value, &mut first, self.critical, "critical");
+        for (asn_min, asn_max) in &self.asn {
+            let asn_format = if asn_min == asn_max {
+                format!("AS:{}", asn_min)
+            } else {
+                format!("AS:{}-{}", asn_min, asn_max)
+            };
+            append(&mut value, &mut first, true, &asn_format);
+        }
+        X509Extension::new_nid(None, Some(ctx), Nid::SBGP_AUTONOMOUSSYSNUM, &value)
```
sfackler, line 486: We need to move away from the string APIs for x509 extensions, and instead directly work with the relevant OpenSSL types.

openssl/src/x509/sbgp.rs
```diff
@@ -0,0 +1,232 @@
...
+impl ASIdentifiers {
+    pub fn inherited(&self) -> bool {
+        unsafe {
+            let asptr = self.0;
+            let asnum = (*asptr).asnum;
+            (*asnum).type_ == ASIdentifierChoice_inherit
+        }
+    }
+
+    pub fn ranges(&self) -> Option<Vec<(u32, u32)>> {
```
sfackler, line 61: Does this correspond to some OpenSSL function?

sfackler: Same question for the other methods in this file.

botovq:
This is a very dangerous API full of traps and bugs. It needs to be wrapped very carefully. What is the use case you want to support by exposing these? Do you have some working code I could look at?

thillux:
@botovq We are currently in the process of reworking this PR. Our use case is certifying address ranges and ASNs in a datacenter fabric setting. Sadly I have no public code that I can share.

@petricor is currently working on this.

openssl-sys/src/handwritten/x509_sbgp.rs
PetrichorIT: This function, and associated functions , has now been replaced with a FFI binding to `X509v3_addr_get_range` and `X509v3_addr_get_afi`.

openssl/src/x509/sbgp.rs
PetrichorIT: These function correspond to `X509v3_addr_get_range`, but executed for the whole extension, since RFC 3779 encodings are rather tedious to work with. Same thing applies to the other functions in this file, they just parse the openssl-internal representation into their Rust-equivalents.

openssl-sys/src/handwritten/x509_sbgp.rs
```diff
@@ -0,0 +1,147 @@
+#[cfg(ossl110)]
+use super::super::*;
+#[cfg(ossl110)]
+use libc::*;
+
+#[repr(C)]
+#[cfg(ossl110)]
```
botovq, line 7: All these should probably condition on `OPENSSL_NO_RFC3779` as well.

botovq (changes requested):
I really think it is a bad idea to expose this API. What stops you from using [rpki-rs][1] for this?

As already mentioned, and you will have no doubt found while developing this, the OpenSSL RFC 3779 API is really bad. And I am not aware of any OpenSSL developer with domain expertise to speak of. So that's not going to change anytime soon.

You don't magically get memory safety by using this via ffi from Rust and then omitting error checks for functions that allocate internally. Very much the opposite of that.

There are omissions of necessary features, missing checks, etc. This all looks taylored to a narrow use case. I do not think this is anywhere close to being ready. People who will try to use this in the future will have a very hard time interoperating with modern RPKI validators.

[1]:  https://github.com/nlnetlabs/rpki-rs/

openssl/src/x509/extension.rs
```diff
@@ -428,6 +440,195 @@ impl AuthorityKeyIdentifier {
...
+        SbgpAsIdentifier::new()
+    }
+}
+
+#[cfg(ossl110)]
+impl SbgpAsIdentifier {
+    /// Construct a new `SbgpAsIdentifier` extension.
+    pub fn new() -> SbgpAsIdentifier {
+        SbgpAsIdentifier {
+            critical: false,
```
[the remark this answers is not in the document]
PetrichorIT: yes, critical=true is now the default

openssl/src/x509/extension.rs
```diff
@@ -428,6 +440,195 @@ impl AuthorityKeyIdentifier {
...
+            asn: Vec::new(),
+        }
+    }
+
+    /// Sets the `critical` flag to `true`. The extension will be critical.
+    pub fn critical(&mut self) -> &mut SbgpAsIdentifier {
```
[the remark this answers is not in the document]
PetrichorIT: In its newer version `fn critical(&mut self, value: bool)` is might be useful to have this configuration option for some legacy systems ? I personally cannot think of a use case, but just in case?

openssl/src/x509/extension.rs
```diff
@@ -428,6 +440,195 @@ impl AuthorityKeyIdentifier {
...
+        unsafe {
+            let asid = ffi::ASIdentifiers_new();
+            for (min, max) in &self.asn {
+                let min = BigNum::from_u32(*min)?.to_asn1_integer()?;
+                let max = BigNum::from_u32(*max)?.to_asn1_integer()?;
+
+                if min == max {
+                    ffi::X509v3_asid_add_id_or_range(asid, 0, min.as_ptr(), std::ptr::null_mut());
+                } else {
+                    ffi::X509v3_asid_add_id_or_range(asid, 0, min.as_ptr(), max.as_ptr());
```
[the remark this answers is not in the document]
PetrichorIT: The newer implementation should prevent double frees + free all nessecary allocations.

thillux:
> I really think it is a bad idea to expose this API. What stops you from using [rpki-rs](https://github.com/nlnetlabs/rpki-rs/) for this?
> 
* We have custom OpenSSL providers for random number generation we'd like to use here.
* We already use OpenSSL with RFC3779 support in a C++-based application, which consumes certificates from our Rust-based CA.

> As already mentioned, and you will have no doubt found while developing this, the OpenSSL RFC 3779 API is really bad. And I am not aware of any OpenSSL developer with domain expertise to speak of. So that's not going to change anytime soon.
> 
* We are aware of this and never use OpenSSL without address sanitizers as the bare minimum during development in our C++ applications using OpenSSL and RFC3779.
> You don't magically get memory safety by using this via ffi from Rust and then omitting error checks for functions that allocate internally. Very much the opposite of that.
> 
* Sure, due to memory safety reasons we chose Rust for our CA application, but it does not completely outweigh missing features in the OpenSSL API subset exposed for our use case.

> There are omissions of necessary features, missing checks, etc. This all looks taylored to a narrow use case. I do not think this is anywhere close to being ready. People who will try to use this in the future will have a very hard time interoperating with modern RPKI validators.
* We currently do not deploy a full RPKI and are a bit taylored yes. But we are eager to make this PR usable in a broader fashion. If somebody steps in and tells us, what is missing in his opinion.

botovq (commented):
Thanks for the explanations. I won't have time for a thorough review in the next days/week, so I only left a few quick comments.

I think we should aim at a good middle ground between the two options you describe: let's make it flexible enough to be generally usable and correct, but let's make it easy enough to use by not exposing things that aren't used in the wild.

So, for example, you currently don't expose the SAFI for IPAddressFamily and you don't expose the RDI variant for ASIdentifiers (which I think is good). In a similar vein, let's not expose criticality as a knob.

openssl/src/x509/extension.rs
```diff
@@ -428,6 +443,258 @@ impl AuthorityKeyIdentifier {
     }
 }
 
+#[cfg(ossl110)]
+pub struct SbgpAsIdentifier {
+    critical: bool,
+    asn: Vec<(u32, u32)>,
+    inherit: bool,
```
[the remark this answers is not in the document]
PetrichorIT: I think this is a good idea. However I think patterns like type-state would be overkill in this scenario, so assertions / panics remain nessecary to prevent silent failures.

PetrichorIT:
So this PR should be ready now.

In regards to the extension builders, criticality will no longer be a configuration option,
and all builder should now make invalid states irrepresentable. To prevent missuse of the current
API , builders will panic if an operation results in a loss of information, so e.g.
```rust
SbgpAsIdentifier::new()
    .add_asn(1234)
    .add_inherit()
    .build()
```
would panic, since the assigned AS number would be lost when the inherit flag is set.

The functions `X509V3_{addr, asid}_{inherits, subset, is_canonical}` are now represented in the API.
`X509V3_{addr, asid}_canonize` is not publicly exposed, since the builder only needs to use it internally.
In general, i think is good to keep the read API and write/builder API seperate, since mutating a raw extension
could go wrong in a number of ways and we dont really want to expose the underlying datastructures to the user.

In regards to `X509V3_asid_subset`, since this API is faulty in 1.1.0, it was moved to 1.1.1 (https://github.com/openssl/openssl/pull/18514).

The APIs for interacting with store contextes and validation paths `X509v3_{asid, addr}_{validate_path, validate_resource_set}`
are not supported, since they seem like a source of unwanted complexity regarding this PR. They are however bound in `openssl-sys`,
for completions sake.

@botovq

botovq:
Thanks. I will try to find time to take a look soon.

botovq:
Apologies for the delay. I haven't forgotten. I won't be able to review this properly before mid-March.

HolyShitMan:
> Apologies for the delay. I haven't forgotten. I won't be able to review this properly before mid-March.

Did you find any time to review the PR?

Greetings

HolyShitMan:
Knowing that my request may be annoying: I would like to kindly remind you that this pull request is still open.

openssl/src/x509/sbgp.rs
```diff
@@ -0,0 +1,252 @@
...
+    /// Extracts the SBGP AS number extension from the `X509` certificate.
+    #[corresponds(X509_get_ext_d2i)]
+    pub fn sbgp_asn(&self) -> Option<ASIdentifiers> {
+        unsafe {
+            let asn = ffi::X509_get_ext_d2i(
+                self.as_ptr(),
+                ffi::NID_sbgp_autonomousSysNum,
+                std::ptr::null_mut(),
+                std::ptr::null_mut(),
+            );
```
botovq, line 234: `X509_get_ext_d2i` is a tricky API to use correctly. As far as I can see, no caller in rust-openssl gets it right.

This can return `NULL` for several reasons: the benign reason is that there is no extension corresponding to the nid in the cert. However, there could also be a library-internal error (malformed extension, memory allocation error, ...), or other errors, like multiple extensions of the same type, which this idiom silently ignores.

The correct way is to pass in a pointer `crit` to a `c_int` as third argument. If `NULL` is returned and `crit` is set to anything but `-1`, an error should be raised.

Not sure if this should be fixed in this PR or if rust-openssl wants to land this as-is and fix the issue in a sweep.

openssl/src/x509/extension.rs
```diff
@@ -428,6 +433,306 @@ impl AuthorityKeyIdentifier {
...
+    /// Construct a new `SbgpAsIdentifier` extension.
+    pub fn new() -> SbgpAsIdentifier {
+        Self(SbgpAsIdentifierOrInherit::List(Vec::new()))
+    }
+
+    /// Sets the `inherit` flag to `true`.
+    pub fn add_inherit(&mut self) -> &mut SbgpAsIdentifier {
+        if let SbgpAsIdentifierOrInherit::List(ref l) = self.0 {
+            if !l.is_empty() {
+                panic!("cannot set extension to inherit, list allready contains elements");
```
botovq, line 468:
```suggestion
                panic!("cannot set extension to inherit, list already contains elements");
```

Similar typos are throughout the panic strings

botovq (commented):
Ignoring the comment on `X509_get_ext_d2i()`, the use of the C API looks about as sane as it can be now.

Regress coverage is decent. As noted, various panic strings need a typo fix (allready -> already). Whether rust-openssl would like panics or different error handling is up to the maintainers of the crate -- as are other API concerns on the rust side of things.

As far as the reservations I had are concerned, this should be ready to proceed.

openssl/src/x509/extension.rs
HolyShitMan: No problem,  I will fix all these

openssl/src/x509/sbgp.rs
HolyShitMan: After considering this further, if there would be a minimally invasive solution available, I'd opt to address it within this PR. However, it seems that we require some new Error construct due to the fact that X509_get_ext_d2i doesn't set the internal OpenSSL error, making the ErrorStack inapplicable.

I'm unsure if this issue extends to other OpenSSL functions as well, where a new Error construct would also help to handle return values properly. After all, I propose addressing this in a separate PR to avoid overloading the current one.

HolyShitMan: In the meantime of the last few months we also got some additional X509 functionality.
Yet, I suggest following the same procedure: let's complete the current task first, and then I'll create an additional PR specifically for the new functions.

HolyShitMan: @botovq

botovq: @HolyShitMan I'm inclined to agree, but it isn't my call. I opened #2226, which, once addressed, should also fix the instance in this PR.

HolyShitMan: @botovq: #2226 and it's solution proposed in #2240 have been open for over a month without any significant progress. Could we find a way to expedite this? I am keen to get this MR resolved.

botovq:
> @botovq: #2226 and it's solution proposed in #2240 have been open for over a month without any significant progress. Could we find a way to expedite this? I am keen to get this MR resolved.

Sorry, I can't help. You need one of the maintainers of this crate to
approve and merge this.

PR #2055, opened 2023-10-12
rustc: bump min version to 1.60

thillux:
The currently tested min version of 1.56 does no longer compile. Due to the transitive dependency log, set the min version to 1.60. The most recent bindgen target <= is 1.59.

$ cargo +1.56.0 build --release
error: package `log v0.4.20` cannot be built because it requires rustc 1.60.0 or newer, while the currently active rustc version is 1.56.0


alex:
There are semver compatible versions of log that support 1.56, so I don't think this is strictly required.

PR #2056, opened 2023-10-12
Use osslconf on BoringSSL

alex:
This reduces a bunch of special casing. Relies on changes from the latest BoringSSL HEAD.


alex:
Hmm, looks like something is needing updating with the CI paths.

alex:
Ok, CI should be green.

sfackler:
Nice!

PR #2058, opened 2023-10-15
Remove manual libatomic reference

### LINKED ISSUE #2043: [REGRESSION] lib atomic is loaded dynamic again on armv6 with openssl-sys v0.9.93
Since #1925 the issue described in #1779 is back again.

Not sure why this is reverted though. If i pin `openssl-sys` to v0.9.92` it works again, but that doesn't seem to be a good long-term-solution.

Is there some way this can be optional, or that it uses `static` instead of `dylib` when some flag is set?
This makes static build binaries break because this forces rustc to dynamically link to atomic.

If you need more information, please let me know.

link2xt:
Reverts revert of 32a303a752732fdd27c407a3a7fd668e546c05db from ddd04786155e7c3855e5c3f4303a7d2b8c908d76

Fixes #2043

### DIFF
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -123,16 +123,6 @@ fn main() {
         println!("cargo:rustc-link-lib={}={}", kind, lib);
     }
 
-    // https://github.com/openssl/openssl/pull/15086
-    if version == Version::Openssl3xx
-        && kind == "static"
-        && (env::var("CARGO_CFG_TARGET_OS").unwrap() == "linux"
-            || env::var("CARGO_CFG_TARGET_OS").unwrap() == "android")
-        && env::var("CARGO_CFG_TARGET_POINTER_WIDTH").unwrap() == "32"
-    {
-        println!("cargo:rustc-link-lib=dylib=atomic");
-    }
-
     if kind == "static" && target.contains("windows") {
         println!("cargo:rustc-link-lib=dylib=gdi32");
         println!("cargo:rustc-link-lib=dylib=user32");


sfackler:
The build is red.

sfackler:
Those are the builds that are failing, yes.

BlackDex:
@sfackler I don't know about Android, but the Linux builds should work just fine with changing `dylib` to `static` i think.
Also, for that exact same system i need to use `RUSTFLAGS: "-Clink-args=-latomic"` else building on that platform fails for anything that uses OpenSSL (v1.1.1 or v3.x.x).
See:
https://github.com/BlackDex/rust-musl/blob/fe53e9361b6f9e0d9efd578f606dcad0dffd9730/.github/workflows/rust-musl.yml#L265C36-L268

And:
https://github.com/dani-garcia/vaultwarden/blob/f863ffb89a0f2a0a682c67110af32c731b5b9fcb/docker/Dockerfile.alpine#L70-L71

What i find a bit strange is that the check is done to see if it is `static`, and then the `atomic` lib is added dynamic.
That will always break builds which try to build statically compiled binaries, since now that is a dynamically linked library.

One compromise would be to either say, if kind is `static` use `static`, and else use `dylib`.
Because of this bug/regression i'm currently forced to pin the crates to a version before this, and people are having issues now trying to build Vaultwarden on FreeBSD for example.

What would be the best way to go to have this solved?

sfackler:
Yeah, I think it makes sense to statically link to libatomic in these contexts. I'm not totally sure why we're dynamically linking currently.

BlackDex:
Ill test what happens if i convert it to `static` and use a newer version and report back.

PR #2060, opened 2023-10-19
Make X509_ALGOR opaque for LibreSSL

botovq:
The struct is still public because that is also the case in OpenSSL, but it should no longer be accessed directly.

I have chosen `libressl350` as conditional since that is the one used for the bindings `X509_ALGOR_get0` . If that is not acceptable for semver reasons, it would be nice if that could be done for 3.8.2 (which is the not yet released stable version of the 3.8 branch).


alex:
Uggghhh, yeah we should be responsible and use 3.8.2 to avoid any compatibility question. (Annoyingly, you'll need to add that in `openssl-sys/build/cfgs.rs`)

botovq:
Yeah, I figured. There, that should be better.

PR #2062, opened 2023-10-20
Clarify 'possible LibreSSL bug'

botovq:
These test fail by default because of lack of PSK support in LibreSSL's TLSv1.3 stack. They do work with SslOptions::NO_TLSV1_3 but it seems preferable to keep ignoring the tests until they are properly supported.


sfackler (approved):

sfackler:
Thanks!

PR #2063, opened 2023-10-20
Enable BN_mod_sqrt() for upcoming LibreSSL 3.8.2

botovq:
This API was inherited from OpenSSL, so it has always been present. Enable it for the upcoming LibreSSL release.

The test as it was written would fail since LibreSSL returns the other square root. Improve the test to work with all possible implementations with a more interesting test case. Also check for error in the simplest situation where no square root exists.

### DIFF
--- a/openssl-sys/src/handwritten/bn.rs
+++ b/openssl-sys/src/handwritten/bn.rs
@@ -75,7 +75,7 @@ extern "C" {
         m: *const BIGNUM,
         ctx: *mut BN_CTX,
     ) -> c_int;
-    #[cfg(ossl110)]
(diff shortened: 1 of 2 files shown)


alex:
If it's always been present, is there a reason to limit it to libressl382, and not support it on all libre versions?

botovq:
Not really. It's been present sincd OpenSSL 0.9.7 which should be old enough to count as always. I also do not know why it was enabled in #1973 for ossl110 only. Let me drop the cfg and see what CI says.

alex:
Thanks for all these PRs!

PR #2068, opened 2023-10-21
Two build script fixes for LibreSSL

botovq:
1. In 04ffe960 and ae74dec45 I added conditionals on libressl382 but missed the build script.

2. The DEP_OPENSSL_LIBRESSL_VERSION_NUMBER handling has two near-duplicate parsers. The new one added in 4ecaf691 was subsequently extended and the old one went stale: libressl250, 310, 370, 380 are missing from it.


alex (approved):
that's perplexing

Issue #1380, opened 2020-11-27
Overriding security level requirements for single hosts - SSL_CTX_set_security_level missing from bindings


Ten0:
The following seems to exist:
https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_security_level.html

But [does not exist in `openssl_sys`](https://docs.rs/openssl-sys/0.9.58/openssl_sys/?search=SSL_CTX_set_).

This prevents from tolerating lower security level (https://askubuntu.com/questions/1233186/ubuntu-20-04-how-to-set-lower-ssl-security-level) for a few hosts instead of globally.

sfackler:
Happy to take a PR adding it!

PR #2074, opened 2023-10-27
add security level bindings

jmayclin:
This PR adds bindings for a limited set of the `set_security_level` methods

```
void SSL_CTX_set_security_level(SSL_CTX *ctx, int level);
void SSL_set_security_level(SSL *s, int level);

int SSL_CTX_get_security_level(const SSL_CTX *ctx);
int SSL_get_security_level(const SSL *s);
```

https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_security_level.html

>  These functions were added in OpenSSL 1.1.0. 

>  SSL_CTX_set_security_level() and SSL_set_security_level() do not return values.
> SSL_CTX_get_security_level() and SSL_get_security_level() return a integer that represents the security level with SSL_CTX or SSL, respectively. 

I chose to expose the security level as an unsigned integer, even through the C api uses a signed integer because the only described security levels are 0 - 5 inclusive. This follows the precedent for the `set_verify_depth` bindings which similarly uses a `u32` in the rust bindings even through the C api uses a signed integer. Let me know if you'd prefer the security level to be represented as an `i32` instead.

This PR addresses #1380

### DIFF
--- a/openssl-sys/src/handwritten/ssl.rs
+++ b/openssl-sys/src/handwritten/ssl.rs
@@ -924,3 +924,17 @@ extern "C" {
     #[cfg(all(ossl111, not(ossl111b)))]
     pub fn SSL_get_num_tickets(s: *mut SSL) -> size_t;
 }
+
+extern "C" {
+    #[cfg(ossl110)]
+    pub fn SSL_CTX_set_security_level(ctx: *mut SSL_CTX, level: c_int);
+
+    #[cfg(ossl110)]
+    pub fn SSL_set_security_level(s: *mut SSL, level: c_int);
+
+    #[cfg(ossl110)]
+    pub fn SSL_CTX_get_security_level(ctx: *const SSL_CTX) -> c_int;
+
+    #[cfg(ossl110)]
+    pub fn SSL_get_security_level(s: *const SSL) -> c_int;
+}
--- a/openssl-sys/src/handwritten/ssl.rs
+++ b/openssl-sys/src/handwritten/ssl.rs
@@ -926,15 +926,15 @@ extern "C" {
 }
 
 extern "C" {
-    #[cfg(ossl110)]
+    #[cfg(any(ossl110, libressl360))]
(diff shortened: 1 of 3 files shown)


openssl-sys/src/handwritten/ssl.rs
```diff
@@ -924,3 +924,17 @@ extern "C" {
     #[cfg(all(ossl111, not(ossl111b)))]
     pub fn SSL_get_num_tickets(s: *mut SSL) -> size_t;
 }
+
+extern "C" {
+    #[cfg(ossl110)]
```
botovq, line 929: Please add `libressl360` throughout:
```diff
-    #[cfg(ossl110)]
+    #[cfg(any(ossl110, libressl360))]
```

openssl/src/ssl/mod.rs
```diff
@@ -1718,6 +1718,16 @@ impl SslContextBuilder {
         unsafe { cvt(ffi::SSL_CTX_set_num_tickets(self.as_ptr(), num_tickets)).map(|_| ()) }
     }
 
+    /// Set the context's security level, which controls the allowed parameters
+    /// and algorithms.
```
botovq, line 1722: worth adding "a number between 0 and 5"?

(The C API allows any integer and clamps to the interval [0, 5] internally, but that's not directly observable.)

openssl/src/ssl/mod.rs
```diff
@@ -3405,6 +3425,26 @@ impl SslRef {
     pub fn num_tickets(&self) -> usize {
         unsafe { ffi::SSL_get_num_tickets(self.as_ptr()) }
     }
+
+    /// Set the connection's security level, which controls the allowed parameters
+    /// and algorithms.
```
botovq, line 3430: ditto

botovq (changes requested):
Setting the security level is [already possible](https://www.openssl.org/docs/man1.1.1/man1/ciphers.html) via  `set_cipher_list` and `set_ciphersuites`  using  the `@SECLEVEL=n` cipher string.

PR #2077, opened 2023-10-31
20231030 2059 pkcs11

### LINKED ISSUE #2059: Is it possible to provide a URI to create a PKey?
I'm working with the pkcs11-provider for OpenSSL 3; one way to use the provider is via a URI; openssl internally will make use of `OSSL_STORE_open_ex` to get the provider key and use it internally. I can't see any function on `PKey` or a store module, so I'm wondering if that's possible with the library

For example, this command works with OpenSSL, but I'm unsure if `-key "pkcs11:type=private;id=%00%01"` will translate into rust-openssl

`openssl req -new -x509 -days 365 -subj '/CN=my key/' -sha256 -provider pkcs11 -key "pkcs11:type=private;id=%00%01" -out cert.pem`

Thank you!

Firstyear:
Fixes #2059 - This allows loading a PKey<Private> from a store by uri.  I have already tested this as working with a pkcs11 tpm provider.

I have some questions about how to get this to a "mergeable"  state. 

* Do you have a preferred method for running tests for a feature like this, especially where there are external requirements like pkcs11?
* Is the approach in the handwritten section correct for ffi?
* I'm not sure how you would like the api structured generally in Store - Stores can be re-used so I could make this use new(uri) instead and then have methods for get_private_key / get_public_key. So some advice on how to shape the api would be good. 

I think after this the next step will be to allow creating keys with propqueries per https://github.com/latchset/pkcs11-provider/blob/main/tests/tgenkey.c#L141


Firstyear:
@WolfDan

Firstyear:
Is it possible to get someone to look at this, as I'd like to be able to contribute this.

ajfouts:
Thank you for contributing this! I was able to use your changes successfully as well. It would be great if this could be looked at sometime & eventually merged.

Firstyear:
Given that it's been nearly 2 years, there is little hope this will be merged. I am also not interested to contribute either.

If you need something for TPM support, look at the work on tss-esapi and hsm-crypto libraries.

PR #2078, opened 2023-11-01
Release openssl v0.10.58 and openssl-sys v0.9.94

alex:


openssl/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "openssl"
-version = "0.10.57"
+version = "0.10.58"
```
sfackler, line 3: We'll want to bump the openssl-sys dep down below.

sfackler (approved):

PR #2079, opened 2023-11-01
Fixed cfg for RSA_PSS

alex:


openssl-sys/src/evp.rs
```diff
@@ -7,7 +7,7 @@ pub const PKCS5_SALT_LEN: c_int = 8;
 pub const PKCS12_DEFAULT_ITER: c_int = 2048;
 
 pub const EVP_PKEY_RSA: c_int = NID_rsaEncryption;
-#[cfg(any(openssl111, boringssl))]
+#[cfg(any(ossl111, boringssl))]
```
botovq, line 10: You could add `libressl310` to both.

alex: Ah great, thanks

botovq (approved):

PR #2081, opened 2023-11-02
Add binding to NID of Chacha20-Poly1305 cipher

Arnavion:
Ref:
- https://github.com/openssl/openssl/blob/OpenSSL_1_1_0l/include/openssl/obj_mac.h#L4325
- https://github.com/openbsd/src/blob/d781822394e40621101778573b197bbb39bc8d5b/lib/libcrypto/objects/obj_mac.num#L967

### DIFF
--- a/openssl-sys/src/obj_mac.rs
+++ b/openssl-sys/src/obj_mac.rs
@@ -996,3 +996,7 @@ pub const NID_sha3_512: c_int = 1034;
 pub const NID_shake128: c_int = 1100;
 #[cfg(ossl111)]
 pub const NID_shake256: c_int = 1101;
+#[cfg(ossl110)]
+pub const NID_chacha20_poly1305: c_int = 1018;
+#[cfg(libressl271)]
+pub const NID_chacha20_poly1305: c_int = 967;
--- a/openssl/src/nid.rs
+++ b/openssl/src/nid.rs
@@ -1090,6 +1090,8 @@ impl Nid {
     pub const SHAKE128: Nid = Nid(ffi::NID_shake128);
     #[cfg(ossl111)]
     pub const SHAKE256: Nid = Nid(ffi::NID_shake256);
+    #[cfg(any(ossl110, libressl271))]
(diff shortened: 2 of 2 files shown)


openssl-sys/src/obj_mac.rs
```diff
@@ -996,3 +996,5 @@ pub const NID_sha3_512: c_int = 1034;
 pub const NID_shake128: c_int = 1100;
 #[cfg(ossl111)]
 pub const NID_shake256: c_int = 1101;
+#[cfg(ossl110)]
+pub const NID_chacha20_poly1305: c_int = 1018;
```
botovq, line 1000: Please add

```
#[cfg(libressl271)]
pub const NID_chacha20_poly1305: c_int = 967;
```

https://github.com/openbsd/src/blob/master/lib/libcrypto/objects/obj_mac.num#L967

openssl/src/nid.rs
```diff
@@ -1090,6 +1090,8 @@ impl Nid {
     pub const SHAKE128: Nid = Nid(ffi::NID_shake128);
     #[cfg(ossl111)]
     pub const SHAKE256: Nid = Nid(ffi::NID_shake256);
+    #[cfg(ossl110)]
```
botovq, line 1093: And change this to `#[cfg(any(ossl110, libressl271))]`.

openssl-sys/src/obj_mac.rs
Arnavion: Should I add boringssl too then?

https://github.com/google/boringssl/blob/f714cb2935906a2f085c3b89d7e206af94627b56/include/openssl/nid.h#L4204

alex:
Boringssl doesn't use these bindings, so it's not needed

openssl-sys/src/obj_mac.rs
Arnavion: Done.

openssl/src/nid.rs
Arnavion: Done.

alex (approved):

Issue #2087, opened 2023-11-04
-include argument is not supported by MSVC's cl.exe


HMaker:
I think the equivalent `-include` arg is [/FI][1], see https://github.com/pyca/cryptography/issues/9746#issuecomment-1773616876.

https://github.com/sfackler/rust-openssl/blob/d009660833267a2f4527a6d00e768b757387679b/openssl-sys/build/run_bindgen.rs#L148-L159

https://github.com/sfackler/rust-openssl/blob/d009660833267a2f4527a6d00e768b757387679b/openssl-sys/build/run_bindgen.rs#L203-L214

[1]: https://learn.microsoft.com/en-us/cpp/build/reference/fi-name-forced-include-file?view=msvc-170

davidben:
Needing the `-include` flag was just due to a flaw in bindgen. I believe the latest version will actually generate correct output, so I'd suggest rust-openssl just update to that.

HMaker:
I tested removing the `-include` flag with bindgen 0.68.1, it worked.

alex:
Please send a PR with that.

Issue #2086, opened 2023-11-04
bindgen's allowlist-file does not match anything for BoringSSL on Windows


HMaker:
I am trying to link Python's cryptography against BoringSSL, it uses rust-openssl. See [related issue][1].

There bindgen is generating empty rust bindings for boringssl, I found the `allowlist-file` doesn't match anything, when removing it the expected rust bindings are generated.
https://github.com/sfackler/rust-openssl/blob/d009660833267a2f4527a6d00e768b757387679b/openssl-sys/build/run_bindgen.rs#L183-L184

[1]: https://github.com/pyca/cryptography/issues/9746#issuecomment-1775911491

davidben:
This is because bindgen rather unhelpfully evaluates the regex with a mix of forward and backslashes on Windows. The regex needs to be fixed to accept both

HMaker:
The following also failed (using backslash only)
```rust
 .arg("--allowlist-file=.*\\openssl\\.*") 
 .arg("--no-recursive-allowlist") 
 ```

PR #2089, opened 2023-11-05
fix bindgen for BoringSSL on Windows

HMaker:
- remove the `-include` flag (works for bindgen 0.68+) (fixes https://github.com/sfackler/rust-openssl/issues/2087)
- fix allowlist-file regex to match Windows' path backslash (fixes https://github.com/sfackler/rust-openssl/issues/2086)

### DIFF
--- a/openssl-sys/build/run_bindgen.rs
+++ b/openssl-sys/build/run_bindgen.rs
@@ -119,7 +119,7 @@ pub fn run_boringssl(include_dirs: &[PathBuf]) {
         .enable_function_attribute_detection()
         .default_macro_constant_type(MacroTypeVariation::Signed)
         .rustified_enum("point_conversion_form_t")
-        .allowlist_file(".*/openssl/[^/]+\\.h")
+        .allowlist_file(".*[/\\\\]openssl/[^/]+\\.h")
         .allowlist_recursively(false)
         .blocklist_function("BIO_vsnprintf")
         .blocklist_function("OPENSSL_vasprintf")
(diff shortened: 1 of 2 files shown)


openssl-sys/build/run_bindgen.rs
```diff
@@ -180,8 +173,8 @@ pub fn run_boringssl(include_dirs: &[PathBuf]) {
         .arg("--enable-function-attribute-detection")
         .arg("--default-macro-constant-type=signed")
         .arg("--rustified-enum=point_conversion_form_t")
-        .arg("--allowlist-file=.*/openssl/[^/]+\\.h")
-        .arg("--no-recursive-allowlist")
+        .arg("--allowlist-file=.*[/\\\\]openssl/[^/]+\\.h") 
+        .arg("--no-recursive-allowlist")
```
alex, line 177: Trailing whitespace

openssl/src/x509/mod.rs
```diff
@@ -761,7 +761,7 @@ impl X509 {
                     ffi::PEM_read_bio_X509(bio.as_ptr(), ptr::null_mut(), None, ptr::null_mut());
                 if r.is_null() {
                     let err = ffi::ERR_peek_last_error();
-                    if ffi::ERR_GET_LIB(err) as X509LenTy == ffi::ERR_LIB_PEM
+                    if ffi::ERR_GET_LIB(err) as X509LenTy == ffi::ERR_LIB_PEM.try_into().unwrap()
```
alex, line 764: What's this change for?

alex:
If this relies on a new version of bindgen, you'll need to change `openssl-sys/Cargo.toml` as well

openssl/src/x509/mod.rs
HMaker: I had a build error and cargo suggested me to make that change, it had a mismatch between `i32` and `u32`. I don't know what caused that, maybe the boringssl version? I built boringssl from commit `5d58c559ace6a24ea6613e412b26bd4c50668ab3`.

alex: It's not needed for either the version we're using in CI, nor for boringssl HEAD as far as I can tell, so it shouldn't be part of this PR

HMaker:
Looks like the linux build of boringssl now fails, probably that `-include` flag is still needed there?

alex:
Some debugging will be required. Testing locally, it seems to work without `-include`.

openssl/src/x509/mod.rs
HMaker: I removed that and now got the same error
```
error[E0308]: mismatched types
   --> openssl\src\x509\mod.rs:764:62
    |
764 |                     if ffi::ERR_GET_LIB(err) as X509LenTy == ffi::ERR_LIB_PEM
    |                        ----------------------------------    ^^^^^^^^^^^^^^^^ expected `u32`, found `i32`
    |                        |
    |                        expected because this is `u32`
    |
help: you can convert an `i32` to a `u32` and panic if the converted value doesn't fit
    |
764 |                     if ffi::ERR_GET_LIB(err) as X509LenTy == ffi::ERR_LIB_PEM.try_into().unwrap()
    |                                                                              ++++++++++++++++++++
```
That BoringSSL version is used by Python's cryptography 0.42

HMaker:
Yes looks like it's the -include flag, I guess the CI is using old version of bindgen?

PR #2093, opened 2023-11-11
Expose ChaCha20 on LibreSSL

alex:


openssl-sys/src/handwritten/evp.rs
```diff
@@ -370,7 +370,7 @@ extern "C" {
     pub fn EVP_aes_256_wrap() -> *const EVP_CIPHER;
     #[cfg(ossl110)]
     pub fn EVP_aes_256_wrap_pad() -> *const EVP_CIPHER;
-    #[cfg(all(ossl110, not(osslconf = "OPENSSL_NO_CHACHA")))]
+    #[cfg(all(any(ossl110, libressl), not(osslconf = "OPENSSL_NO_CHACHA")))]
```
botovq, line 373: This should only be exposed with `libressl310` because of this https://github.com/openbsd/src/commit/51ac9fe0e26a1799a50686e197edcb0aebbb045e

I now think this commit was a mistake. We should have tried to make OpenSSL fix their implementation.

openssl/src/symm.rs
```diff
@@ -1592,7 +1592,7 @@ mod tests {
     }
 
     #[test]
-    #[cfg(ossl110)]
+    #[cfg(any(ossl110, libressl370))]
```
botovq, line 1595: Why is this restricted to 3.7.0 and later?

alex:
thanks

botovq (approved):

PR #2108, opened 2023-11-26
update bindgen and no longer rely on `-include`

### LINKED ISSUE #2086: bindgen's allowlist-file does not match anything for BoringSSL on Windows

### LINKED ISSUE #2087: -include argument is not supported by MSVC's cl.exe

alex:
also make path seperator windows friendly

fixes #2087
fixes #2086
closes #2089

### DIFF
--- a/openssl-sys/build/run_bindgen.rs
+++ b/openssl-sys/build/run_bindgen.rs
@@ -111,6 +111,12 @@ pub fn run(include_dirs: &[PathBuf]) {
 #[cfg(feature = "bindgen")]
 pub fn run_boringssl(include_dirs: &[PathBuf]) {
     let out_dir = PathBuf::from(env::var_os("OUT_DIR").unwrap());
+
+    fs::File::create(out_dir.join("boring_static_wrapper.h"))
+        .expect("Failed to create boring_static_wrapper.h")
+        .write_all(INCLUDES.as_bytes())
+        .expect("Failed to write contents to boring_static_wrapper.h");
+
     let mut builder = bindgen::builder()
         .rust_target(RustTarget::Stable_1_47)
         .ctypes_prefix("::libc")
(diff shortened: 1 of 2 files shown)


openssl-sys/build/run_bindgen.rs
```diff
@@ -119,14 +125,19 @@ pub fn run_boringssl(include_dirs: &[PathBuf]) {
         .enable_function_attribute_detection()
         .default_macro_constant_type(MacroTypeVariation::Signed)
         .rustified_enum("point_conversion_form_t")
-        .allowlist_file(".*/openssl/[^/]+\\.h")
+        .allowlist_file(".*[/\\\\]openssl/[^/]+\\.h")
```
sfackler, line 128: Don't the trailing separators need to be adjusted too?

alex: Great question. This part I just cargo culted from #2089

HMaker: No, I think the trailling separator comes directly from the include directive of the boringssl headers like `#include <openssl/cipher.h>`. The windows path will be like `\path\to\boringssl\src\include\openssl/cipher.h`.

PR #2111, opened 2023-12-01
Add two methods to the PKCS7 API

facutuesca:
This PR adds two new methods:
- `Pkcs7Ref::type_` returns the type of the PKCS#7 object (signed, envelope, etc)

- `Pkcs7Ref::signed` returns the PKCS7_SIGNED member of a signed PKCS#7 object (as a `Pkcs7SignedRef`)

It also adds a `Pkcs7SignedRef::certificates` method, to access the stack of certificates of a signed PKCS7 object.

The motivation behind adding these methods is that `cryptography` [is migrating](https://github.com/pyca/cryptography/issues/8770) its PKCS7 backend implementation from Python to Rust.
[Here](https://github.com/pyca/cryptography/blob/6359dc0e0483b123bb81a329e393314786cb7dff/src/cryptography/hazmat/backends/openssl/backend.py#L1104-L1126) is the Python code that would be replaced by Rust, which manually access the data structures that this PR exposes through the API.

One thing I'm not sure about (in `certificates()`) is my handling of the Stack as a reference, and the ownership status of the certificates inside of it, so any suggestions/fixes are welcome.

### DIFF
--- a/openssl/src/pkcs7.rs
+++ b/openssl/src/pkcs7.rs
@@ -4,15 +4,30 @@ use libc::c_int;
 use std::mem;
 use std::ptr;
 
+use crate::asn1::Asn1ObjectRef;
 use crate::bio::{MemBio, MemBioSlice};
 use crate::error::ErrorStack;
+use crate::nid::Nid;
 use crate::pkey::{HasPrivate, PKeyRef};
-use crate::stack::{Stack, StackRef};
+use crate::stack::{Stack, StackRef, Stackable};
 use crate::symm::Cipher;
+use crate::util::ForeignTypeRefExt;
 use crate::x509::store::X509StoreRef;
 use crate::x509::{X509Ref, X509};
 use crate::{cvt, cvt_p};
 use openssl_macros::corresponds;
+
+foreign_type_and_impl_send_sync! {
+    type CType = ffi::PKCS7_SIGNER_INFO;
+    fn drop = ffi::PKCS7_SIGNER_INFO_free;
+
+    pub struct Pkcs7SignerInfo;
+    pub struct Pkcs7SignerInfoRef;
+}
+
+impl Stackable for Pkcs7SignerInfo {
+    type StackType = ffi::stack_st_PKCS7_SIGNER_INFO;
+}
 
 foreign_type_and_impl_send_sync! {
     type CType = ffi::PKCS7;
@@ -25,6 +40,19 @@ foreign_type_and_impl_send_sync! {
 
     /// Reference to `Pkcs7`
     pub struct Pkcs7Ref;
+}
+
+foreign_type_and_impl_send_sync! {
+    type CType = ffi::PKCS7_SIGNED;
+    fn drop = ffi::PKCS7_SIGNED_free;
+
+    /// A PKCS#7 signed data structure.
+    ///
+    /// Contains signed data.
+    pub struct Pkcs7Signed;
+
+    /// Reference to `Pkcs7Signed`
+    pub struct Pkcs7SignedRef;
 }
 
 bitflags! {
@@ -281,11 +309,43 @@ impl Pkcs7Ref {
             Ok(stack)
         }
     }
+
+    /// Return the type of a PKCS#7 structure as an Asn1Object
+    pub fn type_(&self) -> Option<&Asn1ObjectRef> {
+        unsafe {
+            let ptr = (*self.as_ptr()).type_;
+            Asn1ObjectRef::from_const_ptr_opt(ptr)
+        }
+    }
+
+    /// Get the signed data of a PKCS#7 structure of type PKCS7_SIGNED
+    pub fn signed(&self) -> Option<&Pkcs7SignedRef> {
+        unsafe {
+            if self.type_().map(|x| x.nid()) != Some(Nid::PKCS7_SIGNED) {
+                return None;
+            }
+            let signed_data = (*self.as_ptr()).d.sign;
+            Pkcs7SignedRef::from_const_ptr_opt(signed_data)
+        }
+    }
+}
+
+impl Pkcs7SignedRef {
+    /// Get the stack of certificates from the PKCS7_SIGNED object
+    pub fn certificates(&self) -> Option<&StackRef<X509>> {
+        unsafe {
+            self.as_ptr()
+                .as_ref()
+                .and_then(|x| x.cert.as_mut())
+                .and_then(|x| StackRef::<X509>::from_const_ptr_opt(x))
+        }
+    }
 }
 
 #[cfg(test)]
 mod tests {
     use crate::hash::MessageDigest;
+    use crate::nid::Nid;
     use crate::pkcs7::{Pkcs7, Pkcs7Flags};
     use crate::pkey::PKey;
     use crate::stack::Stack;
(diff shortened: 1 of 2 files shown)


openssl/src/pkcs7.rs
```diff
@@ -281,11 +296,47 @@ impl Pkcs7Ref {
             Ok(stack)
         }
     }
+
+    // Return the type of a PKCS#7 structure as a NID
+    pub fn type_nid(&self) -> Option<Nid> {
```
sfackler, line 301: It seems like this should just return the type, and leave the NID conversion to the caller.

openssl/src/pkcs7.rs
```diff
@@ -281,11 +296,47 @@ impl Pkcs7Ref {
...
+            }
+            // Get the stack of certificates from the PKCS7_SIGNED object
+            let stack_certs = (*self.as_ptr())
+                .d
+                .sign
+                .as_mut()
+                .and_then(|x| x.cert.as_mut())
+                .and_then(|x| StackRef::<X509>::from_const_ptr_opt(x));
+            match stack_certs {
+                None => Ok(new_stack),
```
alex, line 323: I think you should maek this return `Result<Option<T>>` so callers can know if `.d.sign` was null (which the RFC says you're supposed to handle in a specific way)

openssl/src/pkcs7.rs
```diff
@@ -281,11 +296,37 @@ impl Pkcs7Ref {
             Ok(stack)
         }
     }
+
+    // Return the type of a PKCS#7 structure as a NID
+    pub fn type_nid(&self) -> Option<&Asn1ObjectRef> {
```
alex, line 301: Needs to be updated for the new signature.

openssl/src/pkcs7.rs
```diff
@@ -281,11 +296,47 @@ impl Pkcs7Ref {
...
+    // Return the type of a PKCS#7 structure as a NID
+    pub fn type_nid(&self) -> Option<Nid> {
+        unsafe {
+            let ptr = (*self.as_ptr()).type_;
+            Asn1ObjectRef::from_const_ptr_opt(ptr).map(|x| x.nid())
+        }
+    }
+
+    // Retrieve all the certificates from a PKCS#7 structure used for signed data
+    pub fn signed_data_certificates(&self) -> Result<Stack<X509>, ErrorStack> {
```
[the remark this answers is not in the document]
facutuesca: Changed to return a `Result<Option<&StackRef<X509>>, ErrorStack>`, so it matches better the actual type of the `d.sign` member (a pointer to a stack)

openssl/src/pkcs7.rs
facutuesca: Changed to return a `Result<Option<&StackRef<X509>>, ErrorStack>`, so it matches better the actual type of the `.d.sign` member (a pointer to a stack)

openssl/src/pkcs7.rs
```diff
@@ -281,11 +296,37 @@ impl Pkcs7Ref {
             Ok(stack)
         }
     }
+
+    // Return the type of a PKCS#7 structure as an Asn1Object
+    pub fn type_nid(&self) -> Option<&Asn1ObjectRef> {
```
botovq, line 301: `type_nid()` seems not ideal for something returning a (reference to an) object identifier. Perhaps `object()` as in `X509AlgorithmRef`, or `type_oid()` or similar?

facutuesca:
@sfackler @alex Anything else needed for this PR?

openssl/src/pkcs7.rs
```diff
@@ -281,11 +296,37 @@ impl Pkcs7Ref {
             Ok(stack)
         }
     }
+
+    // Return the type of a PKCS#7 structure as an Asn1Object
+    pub fn type_oid(&self) -> Option<&Asn1ObjectRef> {
```
sfackler, line 301: Should this just be called `type`?

openssl/src/pkcs7.rs
```diff
@@ -281,11 +296,37 @@ impl Pkcs7Ref {
...
+        unsafe {
+            let ptr = (*self.as_ptr()).type_;
+            Asn1ObjectRef::from_const_ptr_opt(ptr)
+        }
+    }
+
+    // Retrieve all the certificates from a PKCS#7 structure used for signed data
+    pub fn signed_data_certificates(&self) -> Result<Option<&StackRef<X509>>, ErrorStack> {
```
sfackler, line 309: I think I'd rather go "1 step" here - instead having a `pub fn signed(&self) -> Option<&Pkcs7SignedRef>` method.

openssl/src/pkcs7.rs
botovq: Can't use that here, unfortunately, because `type` is a [strict keyword][1].

[1]: https://doc.rust-lang.org/reference/keywords.html#strict-keywords

woodruffw: It could be `r#type` instead to use the raw identifier syntax, but yeah -- I think bare `type` won't work.

sfackler: You can go with a raw ident, but `type_` is a bit easier to work with.

facutuesca: Fixed (changed to `type_`)

openssl/src/pkcs7.rs
facutuesca: Wouldn't this mean that users of this library would have to use unsafe code to extract the stack of certificates from the `Pkcs7SignedRef`? Since they would have to do the extra step that this function now does:
```rust
                .and_then(|x| x.cert.as_mut())
                .and_then(|x| StackRef::<X509>::from_const_ptr_opt(x));
```
which is unsafe.

alex: I assume @sfackler was suggesting that `Pkcs7SignedRef` should have safe methods for accessing this data.

facutuesca: Got it. Fixed, now the stack of certificates can be accessed in two steps:
```rust
let signed_certificates = pkcs7.signed().and_then(|x| x.certificates());
```

openssl/src/pkcs7.rs
```diff
@@ -281,11 +309,44 @@ impl Pkcs7Ref {
...
+    }
+
+    pub fn signed(&self) -> Option<&Pkcs7SignedRef> {
+        unsafe {
+            if self.type_().map(|x| x.nid()) != Some(Nid::PKCS7_SIGNED) {
+                return None;
+            }
+            // Get the stack of certificates from the PKCS7_SIGNED object
+            let signed_data = (*self.as_ptr()).d.sign;
+            // Pkcs7Signed::from_ptr_opt(signed_data).map(|x| x.as_ref())
```
alex, line 328:
```suggestion
```

openssl/src/pkcs7.rs
```diff
@@ -281,11 +309,44 @@ impl Pkcs7Ref {
...
+
+    // Return the type of a PKCS#7 structure as an Asn1Object
+    pub fn type_(&self) -> Option<&Asn1ObjectRef> {
+        unsafe {
+            let ptr = (*self.as_ptr()).type_;
+            Asn1ObjectRef::from_const_ptr_opt(ptr)
+        }
+    }
+
+    pub fn signed(&self) -> Option<&Pkcs7SignedRef> {
```
alex, line 321: Doc comment please

facutuesca: Fixed

openssl/src/pkcs7.rs
facutuesca: Fixed

sfackler (approved):

sfackler:
Sorry for the delay! LGTM

PR #2115, opened 2023-12-02
Use X509_PURPOSE_get_id instead of struct access

davidben:
The accessor was added at the same version as the struct, so better to just use it. As with X509_PURPOSE_get_by_sname, it was const-corrected later on.

### DIFF
--- a/openssl-sys/src/handwritten/x509.rs
+++ b/openssl-sys/src/handwritten/x509.rs
@@ -687,21 +687,12 @@ extern "C" {
     pub fn X509_REQ_print(bio: *mut BIO, req: *mut X509_REQ) -> c_int;
 }
 
-#[repr(C)]
-pub struct X509_PURPOSE {
-    pub purpose: c_int,
-    pub trust: c_int, // Default trust ID
-    pub flags: c_int,
-    pub check_purpose:
-        Option<unsafe extern "C" fn(*const X509_PURPOSE, *const X509, c_int) -> c_int>,
-    pub name: *mut c_char,
-    pub sname: *mut c_char,
-    pub usr_data: *mut c_void,
-}
+pub enum X509_PURPOSE {}
 
 const_ptr_api! {
     extern "C" {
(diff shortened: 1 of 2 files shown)


alex (approved):
@sfackler are there backwards compataibility issues with making teh struct opaque?

Otherwise the PR looks good to me.

sfackler:
Yeah, we should really keep the struct definition as-is to preserve back compat. If you want, we could make it opaque for future open/boring/libre releases though.

davidben:
We will never, ever support using the handwritten bindings with BoringSSL because of memory safety, so I can just put it back. Though you all really should rethink the stability expectations there.

alex (approved):

davidben:
> X509_PURPOSE_get0() returns a pointer to it (or worse application configured global mutable state if someone called X509_PURPOSE_add()), so X509PurposeRef::from_idx() exposes a pointer to global mutable state to Rust.

Oh, the entire X509PurposeRef API should be removed. It makes no sense in the first place. This is just the usual issue where binding APIs without thinking through use cases leads to problems.

The only thing one can do with `X509_PURPOSE_get_by_sname` is to then chain it with `X509_PURPOSE_get0`, followed by `X509_PURPOSE_get_id`, and then configure it on an `X509_STORE`. Not only that, short of `X509_PURPOSE_add` (which is globally not thread-safe, so it is impossible to expose in Rust safety), the only possible outputs of `X509_PURPOSE_get_by_sname` are the predefined constants anyway.

That means the only APIs here that made sense were the built-in `X509PurposeId` constants and `set_purpose`. Everything else was just a mistake and should be removed.

davidben:
> I plan on changing that in libre

BoringSSL will be doing that too. Note that, similar to https://github.com/sfackler/rust-openssl/pull/2113, it may run into a design flaw in foreign-types. From talking with @alex, it sounds like that crate uses exclusively `mut` pointers and doesn't honor `const` in C? This is, of course, a pretty serious safety issue.

I think I'll also just remove the APIs that `X509PurposeRef` relies on, since it's now clear rust-openssl's use is invalid.

PR #2121, opened 2023-12-07
Fix building with latest BoringSSL

alex:


alex:
Huh. What do you know. Updated to use that and two other places we should have used it recently.

PR #2122, opened 2023-12-09
Configure OpenSSL data dir on vendored builds.

DanielSidhion:
Currently, any build using Nix fails if it's attempting to build a static binary (musl target) that uses this crate with the `vendored` flag. This happened because the `openssl-src` crate looked for the openssl dir using a hardcoded path, and was recently fixed in https://github.com/alexcrichton/openssl-src-rs/pull/220.

This PR bumps the version of `openssl-src` to get the fix, and propagates the value of the `OPENSSL_DIR` environment variable to `openssl-src` if it's set. This is all that's needed to unblock static musl builds on Nix.

### DIFF
--- a/openssl-sys/Cargo.toml
+++ b/openssl-sys/Cargo.toml
@@ -25,7 +25,7 @@ bssl-sys = { version = "0.1.0", optional = true }
 [build-dependencies]
 bindgen = { version = "0.65.0", optional = true, features = ["experimental"] }
 cc = "1.0.61"
-openssl-src = { version = "300.1.2", optional = true, features = ["legacy"] }
+openssl-src = { version = "300.2.0", optional = true, features = ["legacy"] }
 pkg-config = "0.3.9"
 vcpkg = "0.2.8"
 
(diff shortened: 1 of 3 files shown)


sfackler:
The `OPENSSL_DIR` environment variable is already used for an entirely different purpose in openssl-sys.

DanielSidhion:
My apologies for the initial changes in this PR - I just dug more into this matter and understand better why not use `OPENSSL_DIR`, and also what was the actual build failure I was having. The build failure turned out to be a missing `perl` during the build with Nix (the error message made it look like a completely different issue).

As for not using `OPENSSL_DIR` - I'm leaning towards creating and documenting a new environment variable for this. This variable would only be used in a `vendored` build to configure the directory that OpenSSL uses by default to look for configuration file and certificates. This crate's current documentation makes it seem like OpenSSL will not look in any place on a `vendored` build, but it does. This is especially important on `vendored` builds with Nix, since `/usr/local/ssl` (the default directory as configured by `openssl-src`) may contain files for a completely different version of OpenSSL, and/or not be controlled by whoever runs the binary with the `vendored` build.

Skepfyr:
This seems neat, is clearly needed, and is documented clearly, @sfackler shout of you disagree but I'm hitting merge. Thanks for this!

PR #2123, opened 2023-12-09
Fix tests on macOS

alex:
It seems that `OSSL_PROVIDER_try_load` can leave errors on the stack.

### DIFF
--- a/openssl/src/provider.rs
+++ b/openssl/src/provider.rs
@@ -55,6 +55,10 @@ impl Provider {
                 retain_fallbacks as _,
             ))?;
 
+            // OSSL_PROVIDER_try_load seems to leave errors on the stack, even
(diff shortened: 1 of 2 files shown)


openssl/src/provider.rs
```diff
@@ -55,6 +55,10 @@ impl Provider {
                 retain_fallbacks as _,
             ))?;
 
+            // OSSL_PROVIDER_try_load seems to leave errors on the stack, even
+            // when it succeeds.
+            let _ = ErrorStack::get();
```
sfackler, line 60: I assume this is a regression in 3.2? We should probably file a bug upstream if so.

alex: I'm not sure what changed

PR #2126, opened 2023-12-13
Add `rand_priv_bytes`

overvenus:
`rand_priv_bytes` has the same semantics as `rand_bytes`, and it is intended to be used for generating values that should remain private.

See https://www.openssl.org/docs/manmaster/man3/RAND_priv_bytes.html


sfackler:
The concept of RAND_priv_bytes seems truly bizarre to me, but no reason not to expose it I suppose.

PR #2129, opened 2023-12-19
Add nid constant for curve brainpoolP320r1

nicklaswj:
For the same reasons and approach taking in PR #1714 but I've added the NID constant for `brainpoolP320r1` too.
There are more variants of `brainpoolP*` left, but I didn't want to add them if nobody is using them (but I can if you want me to).


sfackler:
Thanks!

PR #2130, opened 2023-12-20
add openssl async error code

zh-jq:
See https://www.openssl.org/docs/man3.2/man3/SSL_get_error.html.
We need them when working with async engines.


sfackler:
You'll need to flag these off of unsupported versions.

PR #2131, opened 2023-12-20
add X509Name::digest

zh-jq:


openssl/src/x509/mod.rs
```diff
@@ -1261,6 +1261,27 @@ impl X509NameRef {
         to_der,
         ffi::i2d_X509_NAME
     }
+
+    /// Returns a digest of the DER representation of this 'X509Name'.
+    #[corresponds(X509_NAME_digest)]
+    pub fn digest(&self, hash_type: MessageDigest) -> Result<DigestBytes, ErrorStack> {
```
zh-jq, line 1267: A macro may be needed to generate all digest method as described here https://www.openssl.org/docs/man1.1.1/man3/X509_digest.html.

zh-jq:
closed, see https://github.com/sfackler/rust-openssl/pull/2134

PR #2134, opened 2023-12-24
add various X509 digest methods via macro

zh-jq:
Add a digest! macro and add all digest methods here https://www.openssl.org/docs/manmaster/man3/X509_digest.html


zh-jq:
> Out of curiosity: what do you need all these for?

For X509Name::digest, just like this https://github.com/nginx/nginx/blob/f255815f5d161fab0dd310fe826d4f7572e141f2/src/event/ngx_event_openssl.c#L3867.
For X509::pubkey_digest, we need to calc the SKI of the public key.

PR #2135, opened 2023-12-25
Allow passing a passphrase callback when loading a public key

alex:


alex:
Unfortuantely, in OpenSSL 3, if you have a password protected PEM and you load it as a public key, it'll prompt on the CLI for a passphrase: https://github.com/pyca/cryptography/issues/7129

sfackler:
Do you need to know the passphrase to load the public key?

alex:
I'll be honest, I don't really care about loading the keys, I care about being able to use the callback to error out, rather than to try to read the key from stdin.

sfackler (approved):

PR #2137, opened 2023-12-26
add ASYNC mode and more mode methods

zh-jq:
And also fix the comment about the return value of SslCtx::set_mode


openssl/src/ssl/mod.rs
```diff
@@ -847,11 +858,31 @@ impl SslContextBuilder {
...
     #[corresponds(SSL_CTX_set_mode)]
     pub fn set_mode(&mut self, mode: SslMode) -> SslMode {
         unsafe {
-            let bits = ffi::SSL_CTX_set_mode(self.as_ptr(), mode.bits() as MtuTy) as SslBitType;
+            let bits = ffi::SSL_CTX_set_mode(self.as_ptr(), mode.bits() as ModeTy) as SslBitType;
+            SslMode::from_bits_retain(bits)
+        }
+    }
+
+    /// Clear the mode used by the context, returning the new mode bit mask.
```
sfackler, line 872: This only clears the set bits of the `mode` argument, right?

zh-jq: yes, updated

taikulawo:
@sfackler r? we need it too

PR #2141, opened 2024-01-03
brew: add openssl@3.0 (for 3.0.x LTS releases)

chenrui333:
Also, it might be consider deprecating/removing openssl@1.1 (as it is already EOL)

image


sfackler:
Thanks!

PR #2142, opened 2024-01-04
Add PKey::from_dhx

alex:
DHX keys have a different serialization from DH keys, but otherwise behave similarly.


alex:
One note: This re-introduces `EVP_PKEY_assign`, which @davidben previously removed for type-safety. Unfortunately no `EVP_PKEY_assign_DHX` exists.

sfackler (approved):

PR #2145, opened 2024-01-11
Make X509_PURPOSE opaque for LibreSSL >= 3.9.0

botovq:
As discussed in https://github.com/sfackler/rust-openssl/pull/2115


alex (approved):

alex:
As an aside, do you have any sense of when 3.9.0 will be out?

botovq:
There are no concrete plans. I do not expect it to happen earlier second half of February. But it might be quite a later. Is there anything you're blocking on?

alex:
Not particularly, so no rush.

Sp1l:
FWIW... LibreSSL 3.9.1 is now release, together with OpenBSD 7.5

PR #2148, opened 2024-01-11
PEM parsing: check last error instead of first

### LINKED ISSUE #2146: Error while loading PEM generated on the fly
We are getting error while reading pem file using `X509::stack_from_pem`. What's happening is an error is left on the error stack and it is pulled later when pem file is read to certificate stack.

This is seen with 0.10.62 and not with 0.10.61.

Note this error cannot be reproduced in macos and windows but can be easily reproduced on linux.

**PEM generation code**:
```
    impl TestResources {
        pub fn init() -> Self {
			let rsa_key = Rsa::<Private>::generate(2048).unwrap();
			let key_pair = PKey::<Private>::from_rsa(rsa_key).unwrap();	

			let mut x509_name = openssl::x509::X509NameBuilder::new().unwrap();
			x509_name.append_entry_by_text("C", "US").unwrap();
			x509_name.append_entry_by_text("ST", "CA").unwrap();
			x509_name.append_entry_by_text("O", "Some organization").unwrap();
			x509_name.append_entry_by_text("CN", "www.example.com").unwrap();
			let x509_name = x509_name.build();
			
			let mut x509 = openssl::x509::X509::builder().unwrap();
			x509.set_subject_name(&x509_name).unwrap();
			x509.set_issuer_name(&x509_name).unwrap();
			x509.set_pubkey(key_pair.as_ref()).unwrap();
			x509.set_not_before(asn1::Asn1Time::days_from_now(0).unwrap().as_ref()).unwrap();
			x509.set_not_after(asn1::Asn1Time::days_from_now(1).unwrap().as_ref()).unwrap();
			x509.set_version(2).unwrap();
			x509.set_serial_number(asn1::Asn1Integer::from_bn(BigNum::from_u32(0).unwrap().as_ref()).unwrap().as_ref()).unwrap();
			x509.sign(key_pair.as_ref(), MessageDigest::sha256()).unwrap();

			let x509 = x509.build();

			let server_certificate_filename = Path::new(&env::temp_dir()).join("server_certificate.pem").to_str().unwrap().to_string();
			let server_key_filename = Path::new(&env::temp_dir()).join("server_key.pem").to_str().unwrap().to_string();
			let server_key_pass_filename = Path::new(&env::temp_dir()).join("server_key_pass.pem").to_str().unwrap().to_string();
			let server_certificate_content = x509.to_pem().unwrap();
			let server_key_content_pkcs1 = key_pair.rsa().as_ref().unwrap().as_ref().private_key_to_pem().unwrap();
			let server_key_content_pkcs8 = key_pair.private_key_to_pem_pkcs8().unwrap();
			let server_key_pass_content_pkcs8 = key_pair.private_key_to_pem_pkcs8_passphrase(Cipher::aes_256_ecb(), server_key_pass_filename.as_bytes()).unwrap();

			fs::write(&server_certificate_filename, &server_certificate_content).unwrap();		
			fs::write(&server_key_filename, &server_key_content_pkcs8).unwrap();	
			fs::write(&server_key_pass_filename, &server_key_pass_content_pkcs8).unwrap();

			Self {
				server_certificate_filename,
				server_key_filename,
				server_key_pass_filename,				
				server_certificate_content,
				server_key_content_pkcs1,
				server_key_content_pkcs8,
				server_key_pass_content_pkcs8,
			}
        }
    }
```

**Panic from line below:**
`let root_certificates = X509::stack_from_pem(&pem).context(format!("failed to load certificates from {}", filename))?;`

**Error:**
```
Caused by:
    error:060CC07A:digital envelope routines:EVP_CIPHER_asn1_to_param:cipher parameter error:../crypto/evp/evp_lib.c:79:, error:0909006C:PEM routines:get_name:no start line:../crypto/pem/pem_lib.c:745:
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

botovq:
Regression introduced in #2120. Ideally, tests would not leave other errors behind.

Fixes #2146

### DIFF
--- a/openssl/src/x509/mod.rs
+++ b/openssl/src/x509/mod.rs
@@ -756,12 +756,13 @@ impl X509 {
                     ffi::PEM_read_bio_X509(bio.as_ptr(), ptr::null_mut(), None, ptr::null_mut());
                 if r.is_null() {
                     let e = ErrorStack::get();
-                    let errors = e.errors();
-                    if !errors.is_empty()
-                        && errors[0].library_code() == ffi::ERR_LIB_PEM as libc::c_int
-                        && errors[0].reason_code() == ffi::PEM_R_NO_START_LINE as libc::c_int
-                    {
-                        break;
+
+                    if let Some(err) = e.errors().last() {
+                        if err.library_code() == ffi::ERR_LIB_PEM as libc::c_int
+                            && err.reason_code() == ffi::PEM_R_NO_START_LINE as libc::c_int
+                        {
+                            break;
+                        }
                     }
 
                     return Err(e);


alex:
Hmm, my impression when I wrote that code is that `[0]` is the item most recently added to the stack. Is that not right?

Using last seems to presume other errors will be added there _after_ the no start line error.

botovq:
It uses [`Err_get_error_line_data()`][1] which is the oldest error. The code you replaced used `ERR_peek_last_error()` which is the latest error. Running this in a loop  has been stable for more than an hour while it would previously crash after a few minutes..

https://github.com/sfackler/rust-openssl/blob/06143eb363f4f59bc4af32da89f0e8d046ebe33c/openssl/src/error.rs#L367-L377

[1]: https://man.openbsd.org/ERR_get_error

alex:
😱. We probably should make that unambigiously clear in the docs. But for now: thank you!

alex (approved):

PR #2155, opened 2024-01-25
enable x509 verify and groups list for boringssl

zh-jq:


openssl-sys/src/ssl.rs
```diff
@@ -411,7 +411,7 @@ cfg_if! {
                 s as *const c_void as *mut c_void,
             )
         }
-    } else if #[cfg(libressl251)] {
+    } else if #[cfg(any(boringssl, libressl251))] {
```
alex, line 414: boringssl does not use `openssl-sys` bindings, so there shouldn't be any conditionals for it here.

alex (approved):

PR #2157, opened 2024-01-30
rebuild openssl-sys if the underlying openssl has changed

reaperhulk:


openssl-sys/build/main.rs
```diff
@@ -79,6 +79,9 @@ fn main() {
     let target = env::var("TARGET").unwrap();
 
     let (lib_dirs, include_dir) = find_openssl(&target);
+    if let Some(printable_include) = include_dir.to_str() {
+        println!("cargo:rerun-if-changed={}", printable_include);
+    }
```
sfackler, line 84: I've considered doing something like this before, but I think it's a bit dangerous. With this specific setup, `include_dir` will be e.g. `/usr/include` on a system-wide install so you'd rebuild any time you updated any system header.

alex:
`include_dir.join("openssl")`?

openssl-sys/build/main.rs
reaperhulk: I've updated the PR to use `<include>/openssl`

openssl-sys/build/main.rs
```diff
@@ -79,6 +79,11 @@ fn main() {
     let target = env::var("TARGET").unwrap();
 
     let (lib_dirs, include_dir) = find_openssl(&target);
+    let mut ossl_include_dir = include_dir.clone();
+    ossl_include_dir.push("openssl");
```
alex, line 83: This can just be `let ossl_include_dir = include_dir.join("openssl");` as far as I can tell

alex (approved):
(Not merging until @sfackler reviews)

sfackler (approved):

PR #2164, opened 2024-02-09
add signature type retrieval functions

jmayclin:
This PR adds bindings for signature retrieval apis. I also added a unit test to make sure that basic functionality is working.

https://www.openssl.org/docs/man1.1.1/man3/SSL_get_peer_signature_type_nid.html

### History

I dug through the git branch to find the first branch that the various APIs appeared in.

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -3484,6 +3484,60 @@ impl SslRef {
             }
         }
     }
+
+    /// this returns the hash, e.g. SHA1, SHA512
+    #[corresponds(SSL_get_signature_nid)]
+    #[cfg(ossl111)]
+    pub fn signature_nid(&self) -> Result<Nid, ErrorStack> {
+        unsafe {
+            let mut pnid: c_int = 0;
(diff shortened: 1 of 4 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -3484,6 +3484,63 @@ impl SslRef {
             }
         }
     }
+
+    /// this returns the hash, e.g. SHA1, SHA512
```
sfackler, line 3488: What hash?

jmayclin: I just updated the comment to more explicitly refer to the signature scheme. I find the `signature_nid` and `signature_type_nid` naming to be really confusing because it's not totally obvious which one returns the digest and which returns the signature algorithm, so I though it would be useful to include an example.

jmayclin:
Hi @sfackler! Could I have another review on this?

jmayclin:
Hi, @sfackler ,sorry for the ping but I was wondering if you have had a chance to look at this? Thanks!

jmayclin:
Hi @sfackler ! I apologize for another ping, but I was wondering if you might be able to take a look at this? I'd love to get this merged in, and I'm very happy to address any feedback that you might have.

PR #2167, opened 2024-02-12
Expose alias on X509 structs

alex:


openssl/src/x509/mod.rs
```diff
@@ -649,6 +649,21 @@ impl X509Ref {
         }
     }
 
+    /// Returns this certificate's "alias". This field is populated by
+    /// OpenSSL in some contexts, such as PKCS12 parsing.
```
botovq, line 653: Since OpenSSL doesn't document this API, it might be worth being a bit more specific: it contains the `friendlyName` attribute if it was present on the PKCS12.

As far as I know, this is the only way the library will set it. The x509 "app" has `-setalias` and the pcks12 app also sets it in some circumstances.

alex: Good idea.

botovq (approved):

sfackler (approved):

openssl/src/x509/mod.rs
```diff
@@ -649,6 +649,22 @@ impl X509Ref {
         }
     }
 
+    /// Returns this certificate's "alias". This field is populated by
+    /// OpenSSL in some situations -- specifically OpenSSL will store a
+    /// PKCS#12 `friendlyName` in this field.
```
davidben, line 654: I might suggest adding a sentence that this is not part of the certificate itself, just some auxiliary info that OpenSSL staples onto it. In particular, it doesn't get serialized. It's just a way to get the friendly name out of you used pkcs12 APIs.

alex: Yeah, I should make that explicit rather than implicit.

PR #2187, opened 2024-02-22
don't emit rerun-if-changed unless the path exists and is readable

reaperhulk:


openssl-sys/build/main.rs
```diff
@@ -84,7 +84,9 @@ fn main() {
     // every time when vendoring so we disable it.
     #[cfg(not(feature = "vendored"))]
     if let Some(printable_include) = include_dir.join("openssl").to_str() {
-        println!("cargo:rerun-if-changed={}", printable_include);
+        if include_dir.join("openssl").exists() {
```
alex, line 87: This seems silly and duplicative. No reason to do the join multiple times

reaperhulk: I agree, I put it up so someone could tell me whether it's a safe assumption that since `find_openssl` returns a PathBuf I should just be able to do
```rust
let potential_path = include_dir.join("openssl")
if potential_path.exists() { 
  // emit
}
```
😄

alex: I don't understand the question, why wouldn't you be able to?

openssl-sys/build/main.rs
```diff
@@ -83,8 +83,11 @@ fn main() {
     // dir has changed since the last build. However, this causes a rebuild
     // every time when vendoring so we disable it.
     #[cfg(not(feature = "vendored"))]
```
alex, line 85: Need to drop this and add `&& !cfg(feature = "vendored")` to the branch below

alex (approved):

PR #2193, opened 2024-02-28
Add copy to MdCtxRef

haroldbruintjes:
Adds the copy method to MdCtx so shared state can be used. I did not implement Clone as copy is fallible, nor did I use dup since that is only available since openssl 3.1.0.


haroldbruintjes:
Whoops, dup of #1679

PR #2199, opened 2024-03-05
Updating openssl-sys build error to inform the user about version incompatibilities.

jmracek:
I lost a couple of hours trying to figure out why openssl-sys wouldn't compile, even though I had a local system installation of OpenSSL.  Turns out I was using 3.0.7, which is unsupported.  When I downgrade to 3.0.0 and update my pkg-config path for the new installation, everything compiles correctly.  

The current error message is too implicit, since the reader is left to assume that by simply installing OpenSSL that the error should be resolved.  The new error message makes it explicit that only specific OpenSSL versions are supported.

### DIFF
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -190,9 +190,12 @@ fn validate_headers(include_dirs: &[PathBuf]) -> Version {
 Header expansion error:
 {:?}
 
-Failed to find OpenSSL development headers.
+Failed to find OpenSSL development headers.  Note: openssl-sys is compatible
+only with OpenSSL versions: 1.0.1, 1.0.2, 1.1.0, 1.1.1, and 3.0.0.  Please
+ensure that the OpenSSL installation being linked is one of the supported
+versions.
 
-You can try fixing this setting the `OPENSSL_DIR` environment variable
+You can try fixing this by setting the `OPENSSL_DIR` environment variable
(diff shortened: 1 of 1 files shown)


sfackler:
Why do you think 3.0.7 isn't supported?

jmracek:
The comment immediately above the error message I'm modifying indicates which minor versions should be supported, and does not include 3.0.7.  Given that I wasn't able to compile openssl-sys with 3.0.7 and a downgrade to 3.0.0 solved the problem, I assumed this to be the root cause and didn't look further into it.  It's entirely possible I have some other error in my build setup is causing the issue.  I wanted to open the PR anyway and let other more knowledgeable folks guide the decision on what to do.

jmracek:
[This](https://github.com/sfackler/rust-openssl/pull/2199/files#diff-d325ad8c35f84eefe9ae678462fec0a3a77fee9b850b5c9ccb46fac3d1d91b6fL170-R181) comment, I meant

sfackler:
That comment is out of date. All versions of OpenSSL from 1.0.1 and newer are supported.

jmracek:
Feel free to close the PR then, if it's not accurate.  Unfortunately my error logs are lost to my bash history after building the downgraded OpenSSL version so I don't think I can reproduce the error anymore or give more details.

Thanks for your attention on this and sorry for the false positive!

PR #2201, opened 2024-03-07
Add support for SSL_group_to_name and SSL_get_negotiated_group

romen:
This PR adds support for OpenSSL 3.0 functions to retrieve information about the effective group negotiated in the TLS handshake from an `SSL*` object.

Marked as draft as I need more guidance from maintainers:
- [ ] `cargo test` fails with
  ```
  error: returning ‘const char * (*)(SSL *, int)’ {aka ‘const char * (*)(struct ssl_st *, int)’} from a function with incompatible 
         return type ‘const char * (*)(const SSL *, int)’ {aka ‘const char * (*)(const struct ssl_st *, int)’}
         [-Werror=incompatible-pointer-types]
  ```
  so I must be doing something wrong with the declaration, but I am struggling to see it

- [ ] in `openssl/src/ssl/mod.rs:3523` I currently have an `.expect()` which is probably undesirable, but I'd like guidance on the proper way to handle that error to be consistent with the rest of the crate.

### DIFF
--- a/openssl-sys/src/handwritten/ssl.rs
+++ b/openssl-sys/src/handwritten/ssl.rs
@@ -951,3 +951,8 @@ extern "C" {
     #[cfg(any(ossl110, libressl360))]
     pub fn SSL_get_security_level(s: *const SSL) -> c_int;
 }
+
+extern "C" {
+    #[cfg(ossl300)]
+    pub fn SSL_group_to_name(ssl: *const SSL, id: c_int) -> *const c_char;
+}
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -82,7 +82,7 @@ use crate::x509::store::{X509Store, X509StoreBuilderRef, X509StoreRef};
 #[cfg(any(ossl102, boringssl, libressl261))]
 use crate::x509::verify::X509VerifyParamRef;
 use crate::x509::{X509Name, X509Ref, X509StoreContextRef, X509VerifyResult, X509};
-use crate::{cvt, cvt_n, cvt_p, init};
+use crate::{cvt, cvt_n, cvt_p, cvt_p_const, init};
 use bitflags::bitflags;
 use cfg_if::cfg_if;
 use foreign_types::{ForeignType, ForeignTypeRef, Opaque};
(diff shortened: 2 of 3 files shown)


openssl-sys/src/handwritten/ssl.rs
```diff
@@ -951,3 +951,8 @@ extern "C" {
     #[cfg(any(ossl110, libressl360))]
     pub fn SSL_get_security_level(s: *const SSL) -> c_int;
 }
+
+extern "C" {
+    #[cfg(ossl300)]
+    pub fn SSL_group_to_name(ssl: *const SSL, id: c_int) -> *const c_char;
```
botovq, line 957: This matches the prototype in the documentation, but not the prototype in ssl.h.

botovq: To be explicit, I think this should fix your compilation issue. Not sure if you need to fix more as a consequence of that:

```suggestion
    pub fn SSL_group_to_name(ssl: *mut SSL, id: c_int) -> *const c_char;
```

romen: Addressed with a988bdf1c5376c943a31fe5000174a8ca3fba8e5

openssl/src/ssl/mod.rs
```diff
@@ -3484,6 +3484,47 @@ impl SslRef {
...
+    /// Note that the return value is valid only during the lifetime of the
+    /// SSL object ssl.
+    #[corresponds(SSL_group_to_name)]
+    #[cfg(ossl300)]
+    pub fn group_to_name(&self, id: c_int) -> Result<&str, ErrorStack> {
+        unsafe {
+            match cvt_p_const(ffi::SSL_group_to_name(self.as_ptr(), id)) {
+                Ok(constp) => Ok(CStr::from_ptr(constp)
+                    .to_str()
+                    .expect("Invalid UTF8 in input")),
```
romen, line 3523: Having `.expect()` here seems wrong to me, but I am not sure what is the best way to address this, consistently with best practices in `rust-openssl`.

openssl/src/ssl/mod.rs
```diff
@@ -3484,6 +3486,48 @@ impl SslRef {
...
+    /// For earlier TLS versions, when a session has been resumed, it always
+    /// reflects the group used for key exchange during the initial handshake
+    /// (otherwise it is from the current, non-resumption, connection).
+    /// This can be called by either client or server.
+    ///
+    /// If the NID for the shared group is unknown then the value is set to the
+    /// bitwise OR of TLSEXT_nid_unknown (0x1000000) and the id of the group.
+    #[corresponds(SSL_get_negotiated_group)]
+    #[cfg(ossl300)]
+    pub fn negotiated_group(&self) -> Result<c_int, ErrorStack> {
```
sfackler, line 3504: I don't think we shouldn't be exposing this as a raw c_int. Since their doing the weird unknown thing, I think we'd need to make a `NegotiatedGroup` type that provides a `pub fn nid(self) -> Option<Nid>` and `pub fn unknown_group_id(self) -> Option<i32>`.

openssl/src/ssl/mod.rs
sfackler: I think this is fine - the set of names that can be returned should be a static set from within OpenSSL rather than being provided by the peer.

openssl/src/ssl/mod.rs
romen: I can do this change.

Then again in `SSL::group_to_name` which calls `SSL_group_to_name()`, we need the raw unaltered `c_int`.

Should the `NegotiatedGroup` type implement `From/To` trait to convert back to the raw `c_int`?
Suggestions on how to implement that?

openssl/src/ssl/mod.rs
romen: The string could be coming from a 3rd party provider, but in the OpenSSL threat model, one should trust the providers they load.

Alternatively is there a way to fold this Rust error within the returned `ErrorStack` type?

openssl/src/ssl/mod.rs
sfackler: Yep, these wrapper types have an `as_raw` method to get the c_int out: https://docs.rs/openssl/latest/openssl/nid/struct.Nid.html#method.as_raw

romen: I tried addressing this with 893ce6a796a0c16b05578b996289e49d8f008f6d

PR #2211, opened 2024-03-30
Add repository field to openssl-macros crate

paolobarbolini:
While scraping our dependencies from crates.io I've found `openssl-macros` to be missing the `repository` field. This PR fixed it by adding it.


sfackler:
Thanks!

PR #2224, opened 2024-04-24
Small additions to X509 names/extensions and ASN1 time

pamaury:
This PR adds a few useful functions that currently require the ffi:
- retrieve the set index of a `X509NameRef`: this is useful to deal with multi-valued RDNs
- retrieve extensions from an `X509` and for each `X509ExtensionRef`, get the criticality, type and data: this is usefult to deal with non-standard extensions
- add a couple of methods to make `Asn1GeneralizedTime` more useful

### DIFF
--- a/openssl/src/x509/mod.rs
+++ b/openssl/src/x509/mod.rs
@@ -667,6 +667,29 @@ impl X509Ref {
         }
     }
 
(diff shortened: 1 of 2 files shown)


pamaury:
@sfackler Would you consider merging this pull request?

PR #2238, opened 2024-05-13
Add `Display` implementation to `X509NameRef`

overhacked:
Add a convenient way of displaying the OIDs of a certificate without writing downstream code to iterate over and convert them individually. Also makes for easier interop with other OpenSSL-based projects (including `openssl` CLI), because the output should match what those tools output if the library version is the same (or similar).

Adapted & simplified from #1688, based on @davidben's https://github.com/sfackler/rust-openssl/pull/1688#issuecomment-1945078714. Removes access to all the wonky flag possibilities in `X509_NAME_print_ex(3)`. Internally, still calls `X509_NAME_print_ex(3)` with RFC2253 flags and disable UTF-8 escaping so the result is a UTF-8-encoded string.

Test included ensures that:

* Printing the name of a certificate with Unicode characters, emoji in this case, functions.
* Formatting a name with ASCII-only OID values results in the expected RFC2253 output.


overhacked:
Failure in CI, saying that `const`s from `bssl_sys` are the wrong type, even though they are correctly typed in `bssl_sys/src/lib.rs`. This builds successfully with `--features bindgen,unstable_boringssl` on my machine with the same commit hash for boringssl that the CI is using (2db0eb3f).  I'm happy to troubleshoot, but I'm not sure where to start figuring out how the CI environment is different.

PR #2240, opened 2024-05-17
New x509d2i error to improve error checking for X509_ext_get_d2i()

### LINKED ISSUE #2226: Improve error checking for `X509_ext_get_d2i()`
Quoting [the BoringSSL documentation][1], which is very explicit about it:

```c++
// WARNING: This function is difficult to use correctly. Callers should pass a
// non-NULL |out_critical| and check both the return value and |*out_critical|
// to handle errors. If the return value is NULL and |*out_critical| is not -1,
// there was an error. Otherwise, the function succeeded and but may return NULL
// for a missing extension. Callers should pass NULL to |out_idx| so that
// duplicate extensions are handled correctly.
```

The internal callers of `X509_CRL_get_ext_d2i()`  and `X509_REVOKED_get_ext_d2i()` handle this closer to correctly, although you can't rely on the library setting an error code if `NULL` is returned and `critical != -1`, and I think `-2` (multiple extensions of type `nid` found) should be an error.

https://github.com/sfackler/rust-openssl/blob/83b3186f5f0c82c28339c2b235b3bb4e5d324b7b/openssl/src/x509/mod.rs#L1950-L1974

[1]: https://github.com/google/boringssl/blob/54821d806d574dd8f2869a8c7f5725b65a67af42/include/openssl/x509.h#L1909C1-L1914C47

HolyShitMan:
First try to fix #2226. If you got other ideas or expectations let me know.

### DIFF
--- a/openssl/src/error.rs
+++ b/openssl/src/error.rs
@@ -23,6 +23,7 @@ use std::convert::TryInto;
 use std::error;
 use std::ffi::CStr;
 use std::fmt;
+use std::fmt::Formatter;
 use std::io;
 use std::ptr;
 use std::str;
@@ -396,6 +397,49 @@ cfg_if! {
                 }
             }
         }
+    }
+}
+
+pub enum X509D2iError {
+    InternalOpenSSLError(ErrorStack),
+    ExtensionNotFoundError,
+    ExtensionUnambiguousError,
+}
+
+impl X509D2iError {
+    pub fn internal_openssl_error(error: ErrorStack) -> Self {
+        Self::InternalOpenSSLError(error)
+    }
+
+    pub fn extension_not_found_error() -> Self {
+        Self::ExtensionNotFoundError
+    }
+
+    pub fn extension_unambiguous_error() -> Self {
+        Self::ExtensionUnambiguousError
+    }
+
+    fn format(&self, fmt: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::InternalOpenSSLError(stack) =>
+                write!(fmt, "Error: Could not get X509 extension; {}", stack),
+            Self::ExtensionNotFoundError =>
+                write!(fmt, "Error: Could not get X509 extension; Reason: Could not find any matching extension."),
+            Self::ExtensionUnambiguousError =>
+                write!(fmt, "Error: Could not get X509 extension; Reason: Tried to read on extension, but found multiple."),
+        }
+    }
+}
+
+impl fmt::Debug for X509D2iError {
+    fn fmt(&self, fmt: &mut Formatter<'_>) -> fmt::Result {
+        self.format(fmt)
+    }
+}
+
+impl fmt::Display for X509D2iError {
+    fn fmt(&self, fmt: &mut Formatter<'_>) -> fmt::Result {
+        self.format(fmt)
     }
 }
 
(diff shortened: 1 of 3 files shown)


openssl/src/error.rs
```diff
@@ -399,6 +400,49 @@ cfg_if! {
     }
 }
 
+pub enum X509D2iError {
+    InternalOpenSSLError(ErrorStack),
+    ExtensionNotFoundError,
+    ExtensionUnambiguousError,
```
botovq, line 406:
```suggestion
    ExtensionAmbiguousError,
```

openssl/src/error.rs
```diff
@@ -399,6 +400,49 @@ cfg_if! {
...
+impl X509D2iError {
+    pub fn internal_openssl_error(error: ErrorStack) -> Self {
+        Self::InternalOpenSSLError(error)
+    }
+
+    pub fn extension_not_found_error() -> Self {
+        Self::ExtensionNotFoundError
+    }
+
+    pub fn extension_unambiguous_error() -> Self {
```
botovq, line 418:
```suggestion
    pub fn extension_ambiguous_error() -> Self {
```

openssl/src/error.rs
```diff
@@ -399,6 +400,49 @@ cfg_if! {
...
+    }
+
+    fn format(&self, fmt: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::InternalOpenSSLError(stack) =>
+                write!(fmt, "Error: Could not get X509 extension; {}", stack),
+            Self::ExtensionNotFoundError =>
+                write!(fmt, "Error: Could not get X509 extension; Reason: Could not find any matching extension."),
+            Self::ExtensionUnambiguousError =>
+                write!(fmt, "Error: Could not get X509 extension; Reason: Tried to read on extension, but found multiple."),
```
botovq, line 429:
```suggestion
                write!(fmt, "Error: Could not get X509 extension; Reason: Tried to read an extension, but found multiple."),
```

botovq (commented):
The approach looks sensible to me. I do not know when and how this crate is willing to do breaking changes to its public API. Maybe the less brutal way to do it would be to deprecate the existing API and point at new variants that return `Result<_, X509D2iError>`.

That's @sfackler and @alex's call, though

HolyShitMan:
> The approach looks sensible to me. [...]

 @sfackler or @alex: Could you please share your thoughts on this? This MR has been open for a month without significant updates, and #2053, which is related, is open since October. We aim to close both so we can resume using the upstream repository in our dependencies.

HolyShitMan:
> @sfackler or @alex: Could you please share your thoughts on this? This MR has been open for a month without significant updates, and #2053, which is related, is open since October. We aim to close both so we can resume using the upstream repository in our dependencies.

Hi @sfackler and @alex, I hope you’re both doing well. I wanted to follow up on the message above. Both issues have been open for a considerable time now, and we're really eager to close them out so we can get back to using the upstream repository. Could you please share your thoughts or provide an update on the current status? If there's anything we can do to help move things forward, we’re more than happy provide them.

HolyShitMan:
Hey @sfackler & @alex,
just a quick reminder that this PR is still uncomment by you guys and that there is another PR that depends on this one being resolved.

Best regards

Issue #2244, opened 2024-05-27
Encrypt AES 256 CCM with 12 byte IV silently uses 7 byte IV


lwestlund:
I have been trying to use an IV of length 12 for encryption with AES 256 CCM but the result is as if I had used an IV of length 7. I suspect that this might be an issue for other ciphers that use an IV as well but I haven't tested.

This is observed using https://docs.rs/openssl/0.10.64/openssl/index.html (latest at time of writing) and OpenSSL 3.3.0.

The problem appears to be that the AES 256 CCM cipher defines the default IV length as 12, but OpenSSL still [defaults to 7](https://wiki.openssl.org/index.php/EVP_Authenticated_Encryption_and_Decryption), causing the last 5 bytes in a 12 byte IV to be ignored.

Minimal reproducible example to show that 7 and 12 byte IV gives the same result:
<details><summary>Details</summary>
<p>

```rust
fn main() {
    let cipher = openssl::symm::Cipher::aes_256_ccm();

    let shared_key = {
        let peer_public_key = openssl::pkey::PKey::public_key_from_raw_bytes(
            &[
                96, 12, 76, 143, 122, 170, 174, 93, 156, 154, 227, 185, 173, 46, 41, 35, 235, 2,
                22, 165, 46, 125, 79, 218, 86, 184, 14, 121, 206, 138, 227, 84,
            ],
            openssl::pkey::Id::X25519,
        )
        .unwrap();

        let private = openssl::pkey::PKey::private_key_from_raw_bytes(
            &[
                19, 163, 196, 101, 143, 222, 15, 123, 7, 149, 252, 132, 136, 99, 29, 231, 1, 16,
                107, 167, 59, 16, 49, 186, 3, 145, 119, 78, 56, 25, 6, 60,
            ],
            openssl::pkey::Id::X25519,
        )
        .unwrap();
        let mut deriver = openssl::derive::Deriver::new(&private).unwrap();

        deriver.set_peer(&peer_public_key).unwrap();

        deriver.derive_to_vec().unwrap()
    };

    let plain_text = &[
        3, 2, 9, 16, 19, 193, 87, 25, 215, 46, 127, 232, 223, 212, 18, 211,
    ];

    let iv_7 = [155, 127, 74, 111, 232, 50, 255];
    let iv_12 = [155, 127, 74, 111, 232, 50, 255, 59, 59, 70, 43, 169];
    assert_eq!(iv_7, iv_12[..7]);
[...]

PR #2245, opened 2024-05-27
Always set IV length for ciphers that use an IV

lwestlund:
This fixes an issue where the IV length would not be set if the length
was equal to the recommended length. The issue shows up at least when an
IV of length 12 (which is returned by `t.iv_len()`) is used with the
AES256 CCM cipher, as OpenSSL defaults the IV length to 7 bytes [^1] and it
would not be correctly set to 12.

[^1]: https://wiki.openssl.org/index.php/EVP_Authenticated_Encryption_and_Decryption

Closes https://github.com/sfackler/rust-openssl/issues/2244.


lwestlund:
> Yeah it unfortunately looks like we can't just unconditionally do this :(

No it looks like it ain't that easy :disappointed:

Based on the [OpenSSL wiki](https://wiki.openssl.org/index.php/EVP_Authenticated_Encryption_and_Decryption) on AES with CCM, I still think that a change is warranted to make 12 byte IVs work so I'm gonna self review another proposal!

openssl/src/symm.rs
```diff
@@ -626,10 +626,8 @@ impl Crypter {
 
         ctx.set_key_length(key.len())?;
 
-        if let (Some(iv), Some(iv_len)) = (iv, t.iv_len()) {
-            if iv.len() != iv_len {
-                ctx.set_iv_length(iv.len())?;
-            }
+        if let (Some(iv), Some(_iv_len)) = (iv, t.iv_len()) {
+            ctx.set_iv_length(iv.len())?;
```
lwestlund, line 630: What do we think about this instead? I reintroduced the length check and added specific handling for the AES CCM ciphers, which are the ones with the IV length issue.

```suggestion
        if let (Some(iv), Some(iv_len)) = (iv, t.iv_len()) {
            if iv.len() != iv_len
                || matches!(
                    t.nid(),
                    Nid::AES_128_CCM | Nid::AES_192_CCM | Nid::AES_256_CCM
                )
            {
                ctx.set_iv_length(iv.len())?;
            }
```

sfackler: Looks like boringssl's not happy with it though.

lwestlund: Also, agree that it's not very pretty but it should (hopefully!) get the job done!

lwestlund:
@sfackler please have another look at this!

PR #2260, opened 2024-07-07
openssl-sys: Provide information about the default certificate paths

hvenev:
The goal of these functions is to contain the logic implemented by the `openssl-probe` crate. The `openssl-sys` crate knows whether we are linked against the system OpenSSL library, and if so, it can use it to get the correct certificate paths.

If we are using the vendored version, we keep the old insecure behavior where we randomly guess paths and hope one works.

### DIFF
--- a/openssl-sys/Cargo.toml
+++ b/openssl-sys/Cargo.toml
@@ -17,6 +17,7 @@ edition = "2018"
 [features]
 vendored = ['openssl-src']
 unstable_boringssl = ['bssl-sys']
+probe = []
 
 [dependencies]
 libc = "0.2"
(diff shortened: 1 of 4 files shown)


sfackler:
> The `openssl-sys` crate knows whether we are linked against the system OpenSSL library, and if so, it can use it to get the correct certificate paths.

It does not. It only knows if we are linked to the vendored copy or not.

sfackler:
I don't see how this change would affect the behavior of native-tls.

sfackler:
Fundamentally, I am not interested in maintaining the big-list-of-paths in openssl-sys.

Again, `not(vendored)` does not guarantee "the system OpenSSL library".

sfackler:
That is a thing that you can do with dynamic linking, but I wouldn't say that it's the "point" of it. You can absolutely ship a dynamically linked libssl with your executable if you want.

PR #2262, opened 2024-07-10
Switch Pkey::from_ to use set1 functions

alex:
They're more type safe, because they aren't macros, and they slightly simplify reference counting.

### DIFF
--- a/openssl/src/pkey.rs
+++ b/openssl/src/pkey.rs
@@ -60,6 +60,7 @@ use openssl_macros::corresponds;
 use std::convert::{TryFrom, TryInto};
 use std::ffi::CString;
 use std::fmt;
+#[cfg(all(not(boringssl), ossl110))]
 use std::mem;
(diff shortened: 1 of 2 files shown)


openssl/src/pkey.rs
```diff
@@ -407,38 +408,35 @@ impl<T> Clone for PKey<T> {
 
 impl<T> PKey<T> {
     /// Creates a new `PKey` containing an RSA key.
-    #[corresponds(EVP_PKEY_assign_RSA)]
+    #[corresponds(EVP_PKEY_set1_RSA)]
     pub fn from_rsa(rsa: Rsa<T>) -> Result<PKey<T>, ErrorStack> {
```
sfackler, line 412: We should add some todo's to switch these to take `&RsaRef<T>` in the future if we're going to use these setters.

alex: Done

PR #2268, opened 2024-07-21
Added a utility function to ensure we never have an issue with 0-length slices from pointers again

alex:
Added a clippy lint to ensure we use it.


openssl/src/bio.rs
```diff
@@ -66,7 +66,7 @@ impl MemBio {
             if len == 0 {
```
sfackler, line 66: You can remove this check now.

sfackler (approved):

PR #2274, opened 2024-08-04
Support OCSP nonce

sgilson:
Will close https://github.com/sfackler/rust-openssl/issues/1045.

Based on the work in https://github.com/sfackler/rust-openssl/pull/1046, but with more tests and a simpler Rust interface to match the existing conventions of the crate.

Reference documentation:


openssl-sys/src/handwritten/ocsp.rs
```diff
@@ -1,5 +1,6 @@
 use super::super::*;
 use libc::*;
+use std::os::raw::c_int;
```
zh-jq, line 3: This `c_int` type has been imported by `use libc::*`.

sgilson:
Hi @sfackler, may I get your input on this PR?

PR #2275, opened 2024-08-08
Expose EVP_DigestSqueeze from Hasher

initsecret:
This completes the streaming API.


openssl/src/hash.rs
```diff
@@ -290,6 +292,21 @@ impl Hasher {
         Ok(())
     }
 
+    /// Squeezes buf out of the hasher. Can be called multiple times, unlike `finish_xof`.
+    /// The output will be as long as the buf.
+    #[cfg(ossl330)]
+    pub fn squeeze_xof(&mut self, buf: &mut [u8]) -> Result<(), ErrorStack> {
```
alex, line 300: Does this need to check that state isn't finalized? Is it permissible to update after a squeeze?

sgmenda: Re finalized: good point, it indeed fails if already finalized: https://github.com/openssl/openssl/blob/master/crypto/sha/sha3.c#L144-L145

sgmenda: Re update: that is also not allowed: https://github.com/openssl/openssl/blob/master/crypto/sha/sha3.c#L55-L57

sgmenda: Thanks for the catch, lemme fix the PR.

alex:
Hmm, so it looks like if you do a `squeeze()` followed by a `finalize()`, the `finalize()` will do an `init()`. It seems like that should be an allowable sequence, that's basically equivalent to calling `squeeze()` 2x. WDYT?

reaperhulk:
I am having trouble following the state machine transitions here. If I call `squeeze_xof`, followed by `update`, followed by `squeeze_xof` again it will initialize the underlying hasher 3 separate times since that ordering is not allowable.

Existing behavior (prior to this PR) appears to allow calling `finish` multiple times where each subsequent invocation re-initializes and then immediately finalizes the hasher. Instead we should really just error (as the underlying OpenSSL hasher would!) on disallowed behavior like this.

The particular case Alex mentioned should be fixed, but we should really fix the state machine here in general since it allows nonsense.

alex:
Unfortunately we can't change the existing behavior, but I guess there's no reason we can't just make it an error to call other methods after squeeze!

sgmenda:
Yes, I tried to preserve prior behavior of silently re-initializing instead of error-ing.

Nope, it's not allowable to call finalize after squeeze. We could allow it, but that would further deviate from the underlying openssl.

reaperhulk:
@initsecret are you still planning to revise this to make it an error to call other methods after squeeze while retaining the rest of the behavior of this API? I think, despite the deviation from the rest of the API, that's what I'd prefer, but @alex may disagree.

initsecret:
I updated the PR. But `hash::tests::test_squeeze_then_update` seems to fail locally on MacOS and I am not sure why. I feel like it is a bug in OpenSSL because this snippet implies that the intended behavior is that it should fail: https://github.com/openssl/openssl/blob/master/crypto/sha/sha3.c#L52-L63

initsecret:
Interesting, all OSes seem to allow Updates after Squeeze. The documentation seems to only promise that Finalize will fail if called after Squeeze:

> Similar to EVP_DigestFinalXOF() but allows multiple calls to be made to squeeze variable length output data. EVP_DigestFinalXOF() should not be called after this.
https://docs.openssl.org/master/man3/EVP_DigestInit/

And indeed, the Finalized check is [done at the EVP level](https://github.com/openssl/openssl/blob/master/crypto/evp/digest.c#L385-L393) while the Squeeze check is only done in the [above cited implementation code](https://github.com/openssl/openssl/blob/master/crypto/sha/sha3.c#L52-L63).

When I get a chance, I will dig deeper because afaict this---allowing updates after squeezing---is inconsistent with Section 4 of [FIPS 202](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.202.pdf).

reaperhulk:
@initsecret Are you still planning to look at this? I'm triaging issues for our next pyca/cryptography release and this is in our milestone right now 😄

initsecret:
@reaperhulk thanks for the ping. Sorry this dropped off my list, I made a note to look deeper tonight and update the PR.

openssl/src/hash.rs
```diff
@@ -274,6 +278,18 @@ impl Hasher {
...
+            // But, [FIPS 202], as shown in Figure 7, has a distinguished absorbing phase followed by a squeezing phase.
+            // Indeed, the [`sha3.c`] implmentation disallows Updates after Squeezes.
+            // For consistency, we always return an error when Update is called after Squeeze.
+            //
+            // [`EVP_DigestUpdate`]: https://github.com/openssl/openssl/blob/b3bb214720f20f3b126ae4b9c330e9a48b835415/crypto/evp/digest.c#L385-L393
+            // [FIPS 202]: https://dx.doi.org/10.6028/NIST.FIPS.202
+            // [`sha3.c`]: https://github.com/openssl/openssl/blob/b3bb214720f20f3b126ae4b9c330e9a48b835415/crypto/sha/sha3.c#L52-L63
+            let errors = ErrorStack::get();
+            return Err(errors);
+        }
```
initsecret, line 292: Doing this check in rust-openssl is not ideal, but it is the best I could come up with.

The alternative is to update individual implementations in OpenSSL, because they likely will not accept this change at the `EVP_DigestUpdate` level, since they may want to use it for cryptographic objects that allow updates after squeezes.

Here, I think we can be more opinionated and say that if we have objects that support updates after squeezes, they should define a new API, rather than further overload this API.

reaperhulk: What impls allow update after squeeze? I'm fine with being opinionated, I just don't understand when this occurs or if it's a hypothetical for the future 😄

initsecret: Seemingly all of them allow updates after squeezes, see this comment and the preceding CI run: https://github.com/sfackler/rust-openssl/pull/2275#issuecomment-2325393699

I double-checked again that this behavior not expected. Also, I noticed that openssl does not have a test for this (disallowing updates after squeezes): https://github.com/openssl/openssl/blob/fbd34c03e3ca94d3805e97a01defdf8b6037f61c/test/evp_xof_test.c

reaperhulk (approved):

alex (approved):

alex:
thanks!

PR #2277, opened 2024-08-17
Bump hex dev-dependency version

alex:


sfackler:
I should set up dependabot...

PR #2278, opened 2024-08-17
Raise bindgen version

alex:


openssl-sys/Cargo.toml
```diff
@@ -23,7 +24,7 @@ libc = "0.2"
 bssl-sys = { version = "0.1.0", optional = true }
 
 [build-dependencies]
-bindgen = { version = "0.65.0", optional = true, features = ["experimental"] }
+bindgen = { version = "0.69.0", optional = true, features = ["experimental"] }
```
sfackler, line 27: Do we still need the experimental feature?

sfackler:
Do you know if there are any codegen breaks we need to worry about for back compat?

openssl-sys/Cargo.toml
alex: Yes, not static-fns are not yet non-experimental. (I'll file a ticket about that if there's not one).

alex:
Not that I know of.

PR #2279, opened 2024-08-17
Ensure Rsa::check_key doesn't leave errors on the stack

alex:


sfackler (approved):

sfackler:
It looks like this is flaky on LibreSSL: https://github.com/sfackler/rust-openssl/actions/runs/10544624544/job/29213875028?pr=2282

botovq:
As long as you don't ensure that the error stack is clean after every single test, I think you need to make sure you clean the stack *before* running a test that cares about it. Otherwise the assert isn't asserting what you want it to assert.

In this case, I'm pretty sure it's the `md_ctx::test::verify_fail` that left an error behind that the assert then picked up.

PR #2283, opened 2024-08-25
Don't leave errors on the stack in `MdCtxRef::digest_verify_final`

alex:


alex:
@sfackler let me know if you think this should just ignore teh errors, I'm not sure if `Ok(false)` ever actually occurs.

botovq:
`Ok(false)` can happen: 1 signature is valid, 0 signature is invalid (typically no error on the stack), < 0 a library error occurred (typically error on the stack).

sfackler (approved):

sfackler:
This seems right to me. Must have been another example of a function where the return value docs used to be more specific.

PR #2285, opened 2024-08-28
Explicit rustfmt config

kornelski:
I have my custom rustfmt configuration, which automatically "breaks" formatting in projects that assume the default formatting. This change explicitly selects the formatting that is required.

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,0 +1,1 @@
+# this project uses the default rustfmt settings


alex:
Hmm, is this typical? It seems to me that part of the point of using defaults it not needing to configure anything :-)

kornelski:
Rustfmt supports system-wide configuration. 

I need to use non-default config, which clashes with projects that demand rustfmt-formatted code assuming defaults.

alex (approved):
I'm skeptical that system-wide config was a good idea for rustfmt, but given that it supports it, this seems like an inevitable need.

PR #2286, opened 2024-08-29
Timestamp Protocol Wrapper (RFC 3161)

JM4ier:
This is an updated (and a bit cleaned up) version of this stale PR: https://github.com/sfackler/rust-openssl/pull/1336


openssl/src/ts.rs
```diff
@@ -0,0 +1,501 @@
...
+        let hash = h.finish()?;
+        Self::from_prehash_with_algo(&hash, md)
+    }
+
+    /// Creates a ready-to-use message imprint from the hash of a message and a specified hash algorithm.
+    ///
+    /// `hash` must have originated from the hash function specified by `md`.
+    pub fn from_prehash_with_algo(hash: &[u8], md: MessageDigest) -> Result<Self, ErrorStack> {
+        let mut algo = X509Algorithm::new()?;
+        algo.set_md(md);
```
botovq, line 95: By using `set_md()` that wraps `X509_ALGOR_set_md()`, you will have a different encoding of the `AlgorithmIdentifier`'s parameters than OpenSSL uses in their [`ts.c`][1]. This PR's code will omit the parameters for the SHA-1/SHA-2 families of hashes, while OpenSSL encodes them as an explicit ASN.1 NULL for all `md`. While your choice is likely the correct one -- the backstory is an [incorrect translation][2] from older to newer ASN.1 syntax -- it poses an interop concern with OpenSSL.

[1]: https://github.com/openssl/openssl/blob/d1c2c054a4b585eed8c883367d80e2a972c4846f/apps/ts.c#L464-L470
[2]: https://www.rfc-editor.org/rfc/rfc5754#page-4

JM4ier: what's the best way forward here?

should I special case OpenSSL's X509_ALGOR to be a struct exposing the underlying fields such that I can exactly replicate the behaviour in ts.c?

botovq: I think you can simply expose and use `X509_ALGOR_set0`.

openssl/src/ts.rs
```diff
@@ -91,8 +92,13 @@ impl TsMsgImprint {
...
     /// `hash` must have originated from the hash function specified by `md`.
     pub fn from_prehash_with_algo(hash: &[u8], md: MessageDigest) -> Result<Self, ErrorStack> {
-        let mut algo = X509Algorithm::new()?;
-        algo.set_md(md);
+        let algo = X509Algorithm::new()?;
+
+        let aobj = unsafe { cvt_p(OBJ_nid2obj(EVP_MD_get_type(md.as_ptr())))? };
+        let res = unsafe { X509_ALGOR_set0(algo.as_ptr(), aobj, ffi::V_ASN1_NULL, ptr::null_mut()) };
+        cvt(res).inspect_err(|_| unsafe {
+            ASN1_OBJECT_free(aobj);
```
botovq: Thanks, this part looks good now.

This free isn't really needed, but doesn't hurt (OBJ_nid2obj() returns objects from an internal hash, the result never needs to be freed)..

JM4ier:
@sfackler could you look at this PR and decide if there's further work needed or if this feature can be merged like this?

PR #2290, opened 2024-08-31
add argon2id support for ossl 3.2+

reaperhulk:
In argon2 there is a concept of parallelism (called lanes). Generally you'd want to have at least as many threads as you have lanes since otherwise you are trading advantage to an attacker, but OpenSSL maintains a global thread limit (as well as allowing building with `no-threads`). So to use threads in this you must both set the global thread limit to greater than 1 and then also pass a thread count to the underlying `EVP_KDF_derive` function.

For this PR we simply always set threads to 1, and follow-ups will expose the global thread limit APIs and then make use of potential parallelism.


openssl/src/argon2.rs
```diff
@@ -0,0 +1,147 @@
...
+///
+/// Requires OpenSSL 3.2.0 or newer.
+#[allow(clippy::too_many_arguments)]
+pub fn argon2id(
+    pass: &[u8],
+    salt: &[u8],
+    ad: Option<&[u8]>,
+    secret: Option<&[u8]>,
+    mut iter: u32,
+    mut threads: u32,
```
alex, line 18: Didn't we agree that it didn't make sense to expose this? lanes-only?

openssl/src/argon2.rs
```diff
@@ -0,0 +1,147 @@
+use libc::c_void;
```
alex, line 1: You can get this from std https://doc.rust-lang.org/stable/std/ffi/enum.c_void.html

openssl/src/argon2.rs
```diff
@@ -0,0 +1,147 @@
...
+    assert!(threads == 1);
+    let pass_field = CStr::from_bytes_with_nul(b"pass\0").unwrap();
+    let salt_field = CStr::from_bytes_with_nul(b"salt\0").unwrap();
+    let ad_field = CStr::from_bytes_with_nul(b"ad\0").unwrap();
+    let secret_field = CStr::from_bytes_with_nul(b"secret\0").unwrap();
+    let iter_field = CStr::from_bytes_with_nul(b"iter\0").unwrap();
+    let size_field = CStr::from_bytes_with_nul(b"size\0").unwrap();
+    let threads_field = CStr::from_bytes_with_nul(b"threads\0").unwrap();
+    let lanes_field = CStr::from_bytes_with_nul(b"lanes\0").unwrap();
+    let memcost_field = CStr::from_bytes_with_nul(b"memcost\0").unwrap();
```
alex, line 34: I don't think there's a huge advantage to creating all these `CStr`, you can just do `b"iter\0".as_ptr()` below

openssl/src/argon2.rs
```diff
@@ -0,0 +1,147 @@
...
+    unsafe {
+        ffi::init();
+        let mut params = vec![];
```
alex, line 37: No need to be a `vec`, just use a stack-allocated array.

openssl/src/argon2.rs
```diff
@@ -0,0 +1,147 @@
...
+        params.push(param_size);
+        let param_end = ffi::OSSL_PARAM_construct_end();
+        params.push(param_end);
+
+        let argon2id_field = CStr::from_bytes_with_nul(b"ARGON2ID\0").unwrap();
+        let argon2 = cvt_p(ffi::EVP_KDF_fetch(
+            ptr::null_mut(),
+            argon2id_field.as_ptr(),
+            ptr::null(),
+        ))?; // This needs to be freed
```
alex, line 85: ok, well, you should do that

openssl/src/argon2.rs
```diff
@@ -0,0 +1,147 @@
...
+        let ctx = cvt_p(ffi::EVP_KDF_CTX_new(argon2))?; // this also needs to be freed
```
alex, line 86: this too

openssl/src/lib.rs
```diff
@@ -147,6 +147,8 @@ mod bio;
 #[macro_use]
 mod util;
 pub mod aes;
+#[cfg(ossl320)]
+pub mod argon2;
```
alex, line 151: Should this just be a `kdf` module?

reaperhulk: I don't have a strong opinion on this. The other KDFs all live in `pkcs5` right now since that's where their headers were declared. We could do this as `kdf`, although then we might want it to be visible for 3.0+ just with the argon2id function missing (since technically EVP_KDF and EVP_KDF_CTX were added in 3.0).

openssl/src/kdf.rs
```diff
@@ -0,0 +1,162 @@
...
+                    idx += 1;
+                }
+                params[idx] = ffi::OSSL_PARAM_construct_end();
+
+                let argon2_p = cvt_p(ffi::EVP_KDF_fetch(
+                    ptr::null_mut(),
+                    b"ARGON2ID\0".as_ptr() as *const c_char,
+                    ptr::null(),
+                ))?;
+                let argon2 = EvpKdf(argon2_p);
```
alex, line 107:
```suggestion
                let argon2 = EvpKdf(cvt_p(ffi::EVP_KDF_fetch(
                    ptr::null_mut(),
                    b"ARGON2ID\0".as_ptr() as *const c_char,
                    ptr::null(),
                ))?);
```

openssl/src/kdf.rs
```diff
@@ -0,0 +1,162 @@
...
+                let ctx_p = cvt_p(ffi::EVP_KDF_CTX_new(argon2.0))?;
+                let ctx = EvpKdfCtx(ctx_p);
```
alex, line 109:
```suggestion
                let ctx = EvpKdfCtx(cvt_p(ffi::EVP_KDF_CTX_new(argon2.0))?);
```

openssl/src/kdf.rs
```diff
@@ -0,0 +1,162 @@
...
+        )
+        .unwrap();
+        assert_eq!(hex::encode(&actual[..]), expected);
+    }
+
+    #[test]
+    #[cfg(all(ossl320, not(osslconf = "OPENSSL_NO_ARGON2")))]
+    fn argon2id_no_ad_secret() {
+        // Test vector from OpenSSL
+        let pass = "";
```
alex, line 154:
```suggestion
        let pass = b"";
```

openssl/src/kdf.rs
```diff
@@ -0,0 +1,162 @@
...
+        let salt = hex::decode("02020202020202020202020202020202").unwrap();
+        let expected = "0a34f1abde67086c82e785eaf17c68382259a264f4e61b91cd2763cb75ac189a";
+
+        let mut actual = [0u8; 32];
+        super::argon2id(pass.as_bytes(), &salt, None, None, 3, 4, 32, &mut actual).unwrap();
```
alex, line 159:
```suggestion
        super::argon2id(pass, &salt, None, None, 3, 4, 32, &mut actual).unwrap();
```

alex (approved):

PR #2292, opened 2024-09-01
allow conversion from MessageDigest to Md

reaperhulk:
This eases porting from legacy const EVP_MD to dynamic EVP_MD, especially in codebases that need to support boringssl/libressl in addition to OpenSSL 3.

### DIFF
--- a/openssl/src/md.rs
+++ b/openssl/src/md.rs
@@ -76,6 +76,12 @@ pub struct Md(Inner);
 unsafe impl Sync for Md {}
 unsafe impl Send for Md {}
 
+impl From<crate::hash::MessageDigest> for Md {
+    fn from(value: crate::hash::MessageDigest) -> Self {
+        unsafe { Md::from_ptr(value.as_ptr() as *mut _) }
+    }
+}
+
 impl Md {
     /// Returns the `Md` corresponding to an [`Nid`].
     #[corresponds(EVP_get_digestbynid)]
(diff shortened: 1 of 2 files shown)


openssl/src/md.rs
```diff
@@ -31,32 +30,6 @@ cfg_if! {
...
-
-            #[inline]
-            fn deref(&self) -> &Self::Target {
-                unsafe {
-                    MdRef::from_ptr(self.as_ptr())
-                }
-            }
-        }
-
         impl DerefMut for Md {
```
alex, line 33: Is there a reason this stays in the conditional?

reaperhulk:
@sfackler Alex and I aren't sure of the motivation behind the design of Md vs MessageDigest so we wanted to get your input here 😄

sfackler:
The MessageDigest/Hasher API is one of the oldest parts of the crate, and tries to provide a "cleaner" interface which doesn't IIRC actually cover all of the weird edge cases. Md/MdCtx is a newer iteration of the same thing that tries to provide a more direct binding. The same thing is true for the asymmetric encryption APIs, which I think had even more issues.

We can hopefully remove MessageDigest/Hasher in the upcoming breaking release, but in the meantime I think this conversion should be fine.

openssl/src/md.rs
```diff
@@ -30,52 +27,52 @@ cfg_if! {
...
-        impl DerefMut for Md {
-            #[inline]
-            fn deref_mut(&mut self) -> &mut Self::Target {
-                unsafe {
-                    MdRef::from_ptr_mut(self.as_ptr())
-                }
-            }
-        }
-    } else {
-        enum Inner {}
```
sfackler, line 69: I'm a bit suspicious that this is a safe change. I vaguely remember that the new OpenSSL 3 dynamic lookup APIs were the reason I had to make the new types in the first place.

PR #2302, opened 2024-09-08
Test against 3.4.0-alpha1

sfackler:


alex (approved):

alex:
oh, they changed the URLs that tarballs are downloaded from

sfackler:
Yeah looks like it's github releases now

PR #2314, opened 2024-10-10
Bump to LibreSSL 4.0.0

bilaliscarioth:


alex:
Is LibreSSL 4.0 portable released? I don't see it on https://www.libressl.org/releases.html

botovq:
On Thu, Oct 10, 2024 at 11:30:34AM -0700, Alex Gaynor wrote:
> Is LibreSSL 4.0 portable released? I don't see it on https://www.libressl.org/releases.html

No, it's not yet released. @busterb is working on it, but it will take a
bit longer still.

alex:
Ok, there's no action for us to take until that's released.

Once it's out, before we can merge this:

a) The 4.0.x branch shouldn't be there until the 4.0 channel is declared stable by LibreSSL (i.e., that line should be removed)
b) We'll need to add 4.0.0 to ci.yml

bilaliscarioth:
The main issue is in OpenBSD current & release branches use `LibreSSL 4.0.0`.. Without it we neither be able to compile via cargo without change it in the `$HOME/.cargo/` and remove Cargo.lock.

However, can we made a pre-processor conditions to made it available only for OpenBSD ?
Or we had to wait the release of LibreSSL-portable ?

botovq:
This can be closed now.

PR #2315, opened 2024-10-15
LibreSSL 4.0.0

botovq:
LibreSSL 4.0.0 was released. It's stable.

Alternate PR to #2314

### DIFF
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -380,6 +380,8 @@ See rust-openssl documentation for more information:
             (3, 8, _) => ('3', '8', 'x'),
             (3, 9, 0) => ('3', '9', '0'),
             (3, 9, _) => ('3', '9', 'x'),
+            (4, 0, 0) => ('4', '0', '0'),
+            (4, 0, _) => ('4', '0', 'x'),
             _ => version_error(),
         };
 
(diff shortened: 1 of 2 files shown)


openssl-sys/build/main.rs
```diff
@@ -380,6 +380,8 @@ See rust-openssl documentation for more information:
             (3, 8, _) => ('3', '8', 'x'),
             (3, 9, 0) => ('3', '9', '0'),
             (3, 9, _) => ('3', '9', 'x'),
+            (4, 0, 0) => ('4', '0', '0'),
+            (4, 0, _) => ('4', '0', 'x'),
```
alex, line 384: Are y'all committing that there will be no ABI breaking changes and that 4.0.1 will be stable?

botovq:
> Are y'all committing that there will be no ABI breaking changes and that 4.0.1 will be stable?

yes.

openssl-sys/build/main.rs
botovq: Yes, we do. As the [release announcement][1] says: 

> We have released LibreSSL 4.0.0, which will be arriving in the LibreSSL directory of your local OpenBSD mirror soon. This is the first stable release for the 4.0.x branch, also available with OpenBSD 7.6


[1]: https://marc.info/?l=openbsd-announce&m=172897399729957&w=2

alex: Ah, in that case I think we don't need the first line, and we only need the 40x line (and to update the existing checks in the code)?

alex: Actually, maybe I misunderstood the intent of the "x" lines, I think this is fine as is.

alex (approved):

openssl-sys/build/main.rs
botovq: I was also unsure, but I think it makes sense to stick with what we used to do previously. Usually we release one or two test releases, but this didn't happen this time for various reasons.

PR #2327, opened 2024-11-13
Enable set_alpn_select_callback for BoringSSL

ViktoriiaKovalova:


sfackler:
There are some ALPN tests that should also be enabled for boringssl I think.

sfackler:
Thanks! (failing tests are due to an unrelated issue)

ViktoriiaKovalova:
Thanks for the quick review!

PR #2345, opened 2025-01-06
Expose operations for RSA flags

notgull:
This commit adds the functions needed to manipulate the flags on the RSA object. These flags are not used in normal SSL as far as I know, but are used by custom providers/engines to change some functionality. The functions I've added are as follows:

- RSA_test_flags
- RSA_set_flags
- RSA_clear_flags

Since these operations are not available on OpenSSL 1.0.2 or earlier, I've also added shims that allow one to directly manipulate the "flags" variable on these older versions.

This patch is made on behalf of Marvell Technology Inc.


notgull:
@alex @sfackler Any chance this can be reviewed?

alex:
Sorry for the delay.

I'm not sure this is sound: Rsas are reference counted objects, which means multiple can exist with the same pointer. As a result, I don't think mutation operations are sound on them?

PR #2357, opened 2025-01-31
Attempt to fix CI by pinning to the Ubuntu 22.04 image

alex:


botovq:
Thanks.

While this will most likely fix CI, some of the failures I saw suggest that rust-openssl has actual problems on arm platforms on newer ubuntu:

This here is one of them related to time_t:

https://github.com/sfackler/rust-openssl/actions/runs/13067958829/job/36463478280

and this one is downright scary since it suggests that on very old OpenSSL releases, rust-openssl and C disagree about the struct layout of X509_VERIFY_PARAM:

https://github.com/sfackler/rust-openssl/actions/runs/13067958829/job/36463480631

Now it is unlikely that many people will use such an OpenSSL on ubuntu 20.04, but you never know.

alex:
Yes, there's a real substantive problem here: Ubuntu 24.04 for armhf has a 64-bit `time_t`, and the `libc` crate does not know about this: https://github.com/rust-lang/libc/issues/3223

I don't think it's something we're in a position to mitigate ourselves, frustratingly.

sfackler:
Sounds like it's time to jettison support for 1.0.x?

alex:
That'd ameliorate a lot of the problem! https://endoflife.date/openssl indicates you can pay OpenSSL to still support it, but no one is paying me!

botovq:
Yes, that makes sense and seems the right move given these problems. FWIW, ruby/openssl dropped support for 1.0.x and 1.1.0 (and all unmaintained libressl versions) a few weeks ago: https://github.com/ruby/openssl/issues/835

sfackler:
We could add a `sizeof(time_t)` to expando.c to try to pick the right size for that field, but that assumes the build and runtime hosts made the same choice here.

PR #2361, opened 2025-02-05
Expose rc2 ciphers on symm::Cipher

alex:
They're terrible, but they're still used in PKCS#8 and PKCS#12


sfackler (approved):
boo, hiss, etc

PR #2365, opened 2025-02-13
tag functions not relevant for BoringSSL

frncs-rss:
it would simplify some of the downstream patches in Android and other places, e.g., https://android.googlesource.com/platform/external/rust/crates/openssl/+/refs/heads/main/patches/

### DIFF
--- a/openssl/src/cipher.rs
+++ b/openssl/src/cipher.rs
@@ -238,6 +238,7 @@ impl Cipher {
         unsafe { CipherRef::from_ptr(ffi::EVP_aes_192_cfb1() as *mut _) }
     }
 
+    #[cfg(not(boringssl))]
     pub fn aes_192_cfb128() -> &'static CipherRef {
         unsafe { CipherRef::from_ptr(ffi::EVP_aes_192_cfb128() as *mut _) }
     }
@@ -295,6 +296,7 @@ impl Cipher {
         unsafe { CipherRef::from_ptr(ffi::EVP_aes_256_cfb1() as *mut _) }
     }
 
+    #[cfg(not(boringssl))]
     pub fn aes_256_cfb128() -> &'static CipherRef {
         unsafe { CipherRef::from_ptr(ffi::EVP_aes_256_cfb128() as *mut _) }
     }
@@ -345,12 +347,12 @@ impl Cipher {
         unsafe { CipherRef::from_ptr(ffi::EVP_bf_ecb() as *mut _) }
     }
 
-    #[cfg(not(osslconf = "OPENSSL_NO_BF"))]
+    #[cfg(not(any(osslconf = "OPENSSL_NO_BF", boringssl)))]
     pub fn bf_cfb64() -> &'static CipherRef {
         unsafe { CipherRef::from_ptr(ffi::EVP_bf_cfb64() as *mut _) }
     }
 
-    #[cfg(not(osslconf = "OPENSSL_NO_BF"))]
+    #[cfg(not(any(osslconf = "OPENSSL_NO_BF", boringssl)))]
     pub fn bf_ofb() -> &'static CipherRef {
         unsafe { CipherRef::from_ptr(ffi::EVP_bf_ofb() as *mut _) }
     }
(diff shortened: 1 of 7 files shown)


alex:
Sorry, can you explain the motivation for the PR? This code works today, and other callers may rely on these functions being available.

openssl/src/ec.rs
```diff
@@ -1009,6 +1009,20 @@ impl EcKey<Private> {
...
+    pub fn private_key_from_der_for_group(
+        der: &[u8],
+        group: &EcGroupRef,
+    ) -> Result<EcKey<Private>, ErrorStack> {
+        unsafe {
+            let mut cbs = ffi::CBS { data: der.as_ptr(), len: der.len() };
+            cvt_p(ffi::EC_KEY_parse_private_key(&mut cbs as *mut ffi::CBS, group.as_ptr()))
+                .map(|p| EcKey::from_ptr(p))
+        }
+    }
```
alex, line 1025: I don't think this is related to the subject of the PR?

openssl/src/encrypt.rs
```diff
@@ -148,7 +148,7 @@ impl<'a> Encrypter<'a> {
     /// This corresponds to [`EVP_PKEY_CTX_set_rsa_oaep_md`].
     ///
     /// [`EVP_PKEY_CTX_set_rsa_oaep_md`]: https://www.openssl.org/docs/manmaster/man3/EVP_PKEY_CTX_set_rsa_oaep_md.html
-    #[cfg(any(ossl102, libressl310))]
+    #[cfg(any(ossl102, libressl310, boringssl))]
```
alex, line 151: Adding these functions looks fine. There may be test functions you also want to allow though.

openssl/src/cipher.rs
```diff
@@ -238,6 +238,7 @@ impl Cipher {
         unsafe { CipherRef::from_ptr(ffi::EVP_aes_192_cfb1() as *mut _) }
     }
 
+    #[cfg(not(boringssl))]
```
alex, line 241: What's the motivation here, boring has this function.

openssl/src/cipher.rs
```diff
@@ -345,12 +347,12 @@ impl Cipher {
         unsafe { CipherRef::from_ptr(ffi::EVP_bf_ecb() as *mut _) }
     }
 
-    #[cfg(not(osslconf = "OPENSSL_NO_BF"))]
+    #[cfg(not(any(osslconf = "OPENSSL_NO_BF", boringssl)))]
```
alex, line 350: Boring already sets `OPENSSL_NO_BF`, why is that not sufficient?

openssl/src/cipher.rs
frncs-rss: Indeed.

It looks like I got that one and the one below from https://android.googlesource.com/platform/external/rust/crates/openssl/+/refs/heads/main/patches/0001-cfgs.diff but indeed in the bssl-sys bindgen I can see them. Removing.

openssl/src/cipher.rs
frncs-rss: I see it in https://boringssl.googlesource.com/boringssl/+/refs/heads/main/rust/bssl-sys/build.rs#22, but when using the crate, I only found folks [passing](https://android.googlesource.com/platform/external/rust/crates/openssl/+/refs/heads/main/Android.bp#25) "boringssl" as the cfg and then maintaining that [patch](https://android.googlesource.com/platform/external/rust/crates/openssl/+/refs/heads/main/patches/0007-cfgs.diff) on all these functions. Perhaps I am holding it wrong in Rust, but where would the "osslconf" come from in the BoringSSL Rust bindings?

alex: I'm not sure I understand your question, when built by `cargo`, the value you identified is exported here: https://boringssl.googlesource.com/boringssl/+/refs/heads/main/rust/bssl-sys/build.rs#137

frncs-rss:
> Sorry, can you explain the motivation for the PR? This code works today, and other callers may rely on these functions being available.

I am trying to build rust-openssl with bssl-sys from BoringSSL and I kept hitting missing symbol issues until I found a set of [patches](https://android.googlesource.com/platform/external/rust/crates/openssl/+/refs/heads/main/patches/) in Android that matched the issues I was having so I felt like maybe we should just upstream these patches instead of carrying them in two places.

It seems like I am missing something, potentially due to different build systems (cargo, soong in Android and GN/ninja for my case). I'll take a deeper look tomorrow. Thanks for the very fast review today!

alex:
I don't know much about GN/ninja, but it seems quite likely however it's building things isn't reproducing what upstream expects.

openssl/src/cipher.rs
frncs-rss: I now understand that it's with a non-cargo build system that the problem arises and we basically need to do the equivalent with the other build system.

openssl/src/ec.rs
frncs-rss: Yeah sorry. Let me start a different PR for the new symbols I am trying to have defined and likely abandon this PR as I fix the build invocation for bssl-sys to define all these osslconf symbols.

openssl/src/encrypt.rs
frncs-rss: Let me make a separate PR for this. The broader context is trying to build [Android KeyMint](https://android.googlesource.com/platform/system/keymint/+/refs/heads/main/boringssl/src/rsa.rs#164) and there is an [Android patch](https://android.googlesource.com/platform/external/rust/crates/openssl/+/refs/heads/main/patches/0001-cfgs.diff#52) to rust-openssl to make these functions visible.

frncs-rss:
Closing the PR as it turns out I was holding it wrong with the non-cargo build system. Thanks for the quick feedback!

PR #2372, opened 2025-02-19
make set_rsa_oaep_md visible to boringssl config

frncs-rss:
Upstreaming https://android-review.git.corp.google.com/c/platform/external/rust/crates/openssl/+/2177444

Android's KeyMint is [using](https://android.googlesource.com/platform/system/keymint/+/refs/heads/main/boringssl/src/rsa.rs#164) this function and it uses rust-openssl with boringssl, but via a downstream patch


alex (approved):

alex:
Looks like there's an import that needs to be included for boringssl as well.

frncs-rss:
> Looks like there's an import that needs to be included for boringssl as well.

Good catch. I missed the test section had its own imports. Amended.

alex (approved):

PR #2373, opened 2025-02-19
Add X509_NAME_hash_ex to X509NameRef

oscarh:
Add the ability to call X509_NAME_hash_ex so that we can also get hashes from issuer names, and not just X509 Certificate subject names.


oscarh:
Sorry about pinging people, but is it possible to get feedback on pull requests here?

I assume that @sfackler is the maintainer, but I've also seen @alex participate in discussions.

Thanks

elwerene:
I also need this functionality to provide crls for warp-openssl

PR #2386, opened 2025-03-18
Enable additional capabilities for AWS-LC

skmcgrail:
This was the follow-up work to the recently merged support for AWS-LC in https://github.com/sfackler/rust-openssl/pull/1805.

This enables some additional API support in various parts of the openssl crate. This includes some specific AES cipher support likes CFB, CCM, and XTS. Also includes support for SHA3 digests and SHAKE128 and SHAKE256.

### DIFF
--- a/openssl/src/cipher_ctx.rs
+++ b/openssl/src/cipher_ctx.rs
@@ -809,8 +809,173 @@ mod test {
         aes_128_cbc(cipher);
     }
 
+    #[cfg(not(boringssl))]
+    #[test]
+    fn default_aes_128_ccm() {
+        // from https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Algorithm-Validation-Program/documents/mac/ccmtestvectors.zip
+        let cipher = Cipher::aes_128_ccm();
+        aes_ccm(
+            cipher,
+            "26511fb51fcfa75cb4b44da75a6e5a0e",
+            "ea98ec44f5a86715014783172e",
+            "4da40b80579c1d9a5309f7efecb7c059a2f914511ca5fc10",
+            "e4692b9f06b666c7451b146c8aeb07a6e30c629d28065c3dde5940325b14b810",
+            "1bf0ba0ebb20d8edba59f29a9371750c9c714078f73c335d",
+            "2f1322ac69b848b001476323aed84c47",
+        );
+    }
+
+    #[cfg(not(boringssl))]
+    #[test]
+    fn default_aes_192_ccm() {
+        // from https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Algorithm-Validation-Program/documents/mac/ccmtestvectors.zip
+        let cipher = Cipher::aes_192_ccm();
+        aes_ccm(
+            cipher,
+            "26511fb51fcfa75cb4b44da75a6e5a0eb8d9c8f3b906f886",
+            "ea98ec44f5a86715014783172e",
+            "4da40b80579c1d9a5309f7efecb7c059a2f914511ca5fc10",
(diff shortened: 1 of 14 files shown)


openssl/src/cipher_ctx.rs
```diff
@@ -809,8 +809,173 @@ mod test {
...
+        ctx.set_tag(&tag).unwrap();
+        ctx.decrypt_init(None, Some(&key), Some(&iv)).unwrap();
+        ctx.set_data_len(pt.len()).unwrap();
+
+        let mut buf = vec![];
+        ctx.cipher_update(&aad, None).unwrap();
+        ctx.cipher_update_vec(&ct, &mut buf).unwrap();
+        // Some older libraries don't support calling EVP_CipherFinal/EVP_DecryptFinal for CCM
+        // https://wiki.openssl.org/index.php/EVP_Authenticated_Encryption_and_Decryption#Authenticated_Decryption_using_CCM_mode
+        #[cfg(any(libressl410, ossl111, awslc, boringssl))]
```
botovq, line 906: Please skip this call also for LibreSSL 4.1.0. This isn't fixed yet. We're close to cutting our release and I'm not sure I want to change this behavior at this point in the release cycle, everything EVP is super hairy.

```suggestion
        #[cfg(any(ossl111, awslc, boringssl))]
```

skmcgrail:
It looks like everything passed except for one item: https://github.com/sfackler/rust-openssl/actions/runs/13935540844/job/39002532120?pr=2386

This failed with an transient error caused by rustup. Can someone poke the button to retry this failed action? Thanks!

openssl/src/hash.rs
```diff
@@ -796,14 +796,17 @@ mod tests {
         }
 
         assert_eq!(MessageDigest::shake_128().block_size(), 168);
+        #[cfg(ossl111)]
         assert_eq!(MessageDigest::shake_128().size(), 16);
+        #[cfg(awslc)]
+        assert_eq!(MessageDigest::shake_128().size(), 0);
```
alex, line 802: This seems wrong?

openssl/src/hash.rs
```diff
@@ -818,7 +821,10 @@ mod tests {
         }
 
         assert_eq!(MessageDigest::shake_256().block_size(), 136);
+        #[cfg(ossl111)]
         assert_eq!(MessageDigest::shake_256().size(), 32);
+        #[cfg(awslc)]
+        assert_eq!(MessageDigest::shake_128().size(), 0);
```
alex, line 827: This seems both wrong, and a copy-paste error 😂

openssl/src/pkcs12.rs
```diff
@@ -269,7 +271,10 @@ impl Pkcs12Builder {
                 cvt(ffi::PKCS12_set_mac(
                     pkcs12.as_ptr(),
                     pass,
-                    -1,
+                    // -1 could be passed in for the password length in OpenSSL,
+                    // but AWS-LC does not support that at this time.
+                    // We know the length so explicitly provide it.
```
alex, line 276: Personally I don't think we have to justify why we're passing the length. (It's useful to this diff reviewer, but not really to future readers of the code.)

openssl/src/hash.rs
skmcgrail: Ah this one is legitimate mistake to swap the algorithm.

openssl/src/hash.rs
skmcgrail: The SHAKE algorithms are extendable output algorithms that don't have an explicit digest output size as it's determined based on the requested output size. So we opt to return `0` in this instance if `EVP_MD_size()` is called on such an XOF algorithm like SHAKE-128 and SHAKE-256.

alex (approved):

PR #2405, opened 2025-05-12
RFC: Add bindings for ML-KEM and ML-DSA.

### LINKED ISSUE #2393: feature request: PQC algoritm bindings to openssl 3.5.0
Hi

Openssl 3.5.0 added support for PQC algorithms (ML-KEM, ML-DSA and SLH-DSA).

Could these be added?

teythoon:
This patch series adds support for the post-quantum ML-KEM and ML-DSA algorithms.  I intend to also add support for SLH-DSA, but didn't get around to that yet, and want to get the conversation going.  Once SLH-DSA is added, this will close #2393.

I only added the interfaces required to implement ML-KEM and ML-DSA in Sequoia PGP.  And, I'm not too familiar with the OpenSSL API, so I'm happy to add more functions, but I need guidance there.  Similarly, I'm not too familiar with how the Rust bindings abstract over the differences between OpenSSL versions and variants, so any advice in that direction is appreciated (and please be concrete ;)).

It builds upon parts of Jakub's patch series #2380 modulo the RSA and ECDSA parts.


openssl/src/pkey_ml_dsa.rs
```diff
@@ -0,0 +1,279 @@
...
+    }
+}
+
+pub struct PKeyMlDsaParams<T> {
+    params: OsslParam,
+    _m: ::std::marker::PhantomData<T>,
+}
+
+impl<T> PKeyMlDsaParams<T> {
+    /// Creates a new `PKeyMlDsaParams` from existing ECDSA PKey. Internal.
```
Jakuje, line 154:
```suggestion
    /// Creates a new `PKeyMlDsaParams` from existing ML-DSA PKey. Internal.
```

openssl/src/pkey_ml_dsa.rs
```diff
@@ -0,0 +1,279 @@
...
+            Ok(PKeyMlDsaParams::<T> {
+                params: OsslParam::from_ptr(params),
+                _m: ::std::marker::PhantomData,
+            })
+        }
+    }
+}
+
+impl PKeyMlDsaParams<Public> {
+    /// Creates a new `PKeyMlDsaParams` from existing Public ECDSA PKey.
```
Jakuje, line 172:
```suggestion
    /// Creates a new `PKeyMlDsaParams` from existing Public ML-DSA PKey.
```

openssl/src/pkey_ml_dsa.rs
```diff
@@ -0,0 +1,279 @@
...
+    /// Returns a reference to the public key.
+    pub fn public_key(&self) -> Result<&[u8], ErrorStack> {
+        self.params
+            .locate(OSSL_PKEY_PARAM_PUB_KEY).unwrap()
+            .get_octet_string()
+    }
+}
+
+impl PKeyMlDsaParams<Private> {
+    /// Creates a new `PKeyMlDsaParams` from existing Private ECDSA PKey.
```
Jakuje, line 188:
```suggestion
    /// Creates a new `PKeyMlDsaParams` from existing Private ML-DSA PKey.
```

openssl/src/pkey_ml_dsa.rs
teythoon: Good catch, fixed!

Jakuje:
@alex any thoughts regarding to adding the new PQC algorithms?

alex:
I have not had a chance to review. For similar reasons to what I noted in https://github.com/sfackler/rust-openssl/pull/2380, I have incredibly low motivation to do anything involving OpenSSL's new (bad) APIs.

Jakuje:
In #2393 you wrote you would be happy for the PR. Does it mean if somebody else would review, this, it would be acceptable?

openssl/src/pkey_ml_dsa.rs
```diff
@@ -0,0 +1,276 @@
...
+}
+
+impl PKeyMlDsaBuilder<Private> {
+    /// Returns the Private ML-DSA PKey from the provided parameters.
+    #[corresponds(EVP_PKEY_fromdata)]
+    pub fn build(self) -> Result<PKey<Private>, ErrorStack> {
+        self.build_internal(ffi::EVP_PKEY_KEYPAIR)
+    }
+
+    /// Creates a new `PKeyRsaBuilder` to generate a new ML-DSA key
```
ArthurHeymans, line 117: PKeyMlDsaBuilder

teythoon: Good catch, fixed!

openssl/src/lib.rs
```diff
@@ -177,18 +177,28 @@ pub mod memcmp;
...
+#[cfg(ossl350)]
+pub mod pkey_slh_dsa;
 #[cfg(ossl300)]
 pub mod provider;
 pub mod rand;
 pub mod rsa;
 pub mod sha;
 pub mod sign;
+#[cfg(ossl300)]
+pub mod signature;
```
Jakuje, line 201: Should we use `#[cfg(ossl350)]`  here to avoid the need to add the `#[allow(unused_imports)]` inside? Even though the functions might be available in older versions, they are for no good use as far as I know.

swenson:
One thing I noticed was missing is support for ML-DSA deterministic signatures. I believe we need an additional FFI function exported for this, [`EVP_PKEY_sign_init_ex2`](https://docs.openssl.org/3.5/man3/EVP_PKEY_sign/).

https://docs.openssl.org/3.5/man7/EVP_SIGNATURE-ML-DSA/#ml-dsa-signature-parameters

teythoon:
Before doing more polishing, and adding more features, it would be good to get some kind of confirmation that this work can be merged, and what changes are necessary before that can happen.

alex:
I apologize if the review was confusing, I wasn't objecting substantively to this work, but the size of the PR makes it difficult to review.

Several of the components here can be done independently of the top line ML-KEM APIs:

- the changse to openssl-sys are all self-contained and verye asy to review, so they can split out
- similarly, the introduction of the new `OsslParamBuilder` and refactors of existing code are independent and can be split out.
- And then I think it'd be easier if the KEM and DSA pieces were in their own PRs, since they'r independent APIs.

Thanks for your work on this. If anyone is interested in carrying this forward, hopefully this gives a roadmap for next steps.

openssl/src/lib.rs
mouse07410: My preference **in this case** would be `#[allow(unused_imports)]`.

Issue #2406, opened 2025-05-12
`cargo test --package openssl test_ec_key_accessor` crashes with 0xc0000005 on arm64 Windows


saschanaz:
Previous discussion: https://github.com/pyca/cryptography/issues/12867

When I run the test command:

```
> cargo test --package openssl test_ec_key_accessor
warning: openssl-sys@0.9.108: expando.c
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.09s
     Running unittests src\lib.rs (target\debug\deps\openssl-82bd68d2fab20804.exe)

running 1 test
error: test failed, to rerun pass `-p openssl --lib`

Caused by:
  process didn't exit successfully: `D:\rust-openssl\target\debug\deps\openssl-82bd68d2fab20804.exe test_ec_key_accessor` (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
note: test exited abnormally; to see the full output pass --nocapture to the harness.
```

Attaching windbg, it shows a recursive loop on `evp_pkey_get_legacy`:

```
(9360.308): Access violation - code c0000005 (first chance)
First chance exceptions are reported before any exception handling.
This exception may be expected and handled.
openssl_82bd68d2fab20804!evp_pkey_get_legacy+0x10:
00007ff7`5d3e1238 f90003ff str         xzr,[sp]
```

```
[0x0]   openssl_82bd68d2fab20804!evp_pkey_get_legacy+0x10   0xdea2100000   0x7ff75d3e1234   
[0x1]   openssl_82bd68d2fab20804!evp_pkey_get_legacy+0xc   0xdea2100030   0x7ff75d3e1234   
[0x2]   openssl_82bd68d2fab20804!evp_pkey_get_legacy+0xc   0xdea2100050   0x7ff75d3e1234   
[0x3]   openssl_82bd68d2fab20804!evp_pkey_get_legacy+0xc   0xdea2100070   0x7ff75d3e1234   
[0x4]   openssl_82bd68d2fab20804!evp_pkey_get_legacy+0xc   0xdea2100090   0x7ff75d3e1234   
[0x5]   openssl_82bd68d2fab20804!evp_pkey_get_legacy+0xc   0xdea21000b0   0x7ff75d3e1234   
[0x6]   openssl_82bd68d2fab20804!evp_pkey_get_legacy+0xc   0xdea21000d0   0x7ff75d3e1234   
[...]

PR #2407, opened 2025-05-12
Run tests on windows-11-arm

saschanaz:
(To investigate #2406)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,7 +93,7 @@ jobs:
           - runner: windows-latest
             arch: x64
           - runner: windows-11-arm
-            arch: aarch64
+            arch: arm64
     runs-on: ${{ matrix.os.runner }}
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -87,14 +87,23 @@ jobs:
...
+        os:
+          - runner: windows-latest
+            arch: x64
+          - runner: windows-11-arm
+            arch: arm64
+    runs-on: ${{ matrix.os.runner }}
     steps:
       - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        if: matrix.os.runner == 'windows-11-arm'
```
alex: this shouldn't be required anymore

alex (approved):

.github/workflows/ci.yml
```diff
@@ -87,14 +87,21 @@ jobs:
 
   windows-vcpkg:
     name: windows-vcpkg
```
alex: oh, can you include the arch in the job name

saschanaz: I think it's already visible as `windows-vcpkg (windows-11-arm, arm64)` but sure.

PR #2417, opened 2025-06-01
Add missing Fedora build dependency

pka:
Tested on Fedora 42

### DIFF
--- a/openssl/src/lib.rs
+++ b/openssl/src/lib.rs
@@ -45,7 +45,7 @@
 //! $ sudo apt-get install pkg-config libssl-dev
 //!
 //! # Fedora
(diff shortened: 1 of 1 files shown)


openssl/src/lib.rs
```diff
@@ -45,7 +45,7 @@
 //! $ sudo apt-get install pkg-config libssl-dev
 //!
 //! # Fedora
-//! $ sudo dnf install pkgconf perl-FindBin perl-IPC-Cmd openssl-devel
+//! $ sudo dnf install pkgconf perl-FindBin perl-File-Compare perl-IPC-Cmd openssl-devel
```
alex, line 48: Looks like `perl-File-Copy` is also required, can you add that in?

PR #2422, opened 2025-06-07
Added X509_pubkey_digest

scarsi42:


openssl/src/x509/mod.rs
```diff
@@ -554,6 +554,27 @@ impl X509Ref {
         }
     }
 
+    /// Returns a digest of the DER representation of the public key in the certificate.
```
botovq, line 557: I would add that this is usually combined with SHA-1 to compute the SKI or AKI of a cert.

botovq (commented):
A test exercising this would be nice.

openssl/src/x509/mod.rs
scarsi42: Yes I agree, but for compatibility reasons I need to call the openssl exported X509_pubkey_digest .
This is computing a generic digest on the ASN1 bit string.

PR #2423, opened 2025-06-09
Improve support for OPENSSL_NO_COMP and OPENSSL_NO_SRTP

justsmth:
Improve support for OPENSSL_NO_COMP and OPENSSL_NO_SRTP. 

This PR addresses two potential failures:

* During bindgen -- the [INCLUDES](https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/build/run_bindgen.rs#L11-L75) tries to `#include <openssl/comp.h>` even when [comp.h isn't available](https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/crypto/comp/comp.h#L7).

* During linking -- when built with `OPENSSL_NO_SRTP`, the SRTP functions [are not available](https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/ssl/d1_srtp.c#L122).


alex:
the libressl test failures look relevant here, if you can take a look

justsmth:
Is anything else needed for this PR?

alex (approved):
looks good, thanks for your contribution!

PR #2424, opened 2025-06-10
Add aws-lc-fips feature to allow linking the aws-lc-fips-sys crate

skmcgrail:
Adds the ability for users to enable the `aws-lc-fips` crate feature, which when enabled will utilize the `aws-lc-fips-sys` crate for the AWS-LC bindings. This version of AWS-LC provides bindings to [AWS-LC-FIPS 3.x](https://github.com/aws/aws-lc/tree/fips-2024-09-27), which has completed FIPS validation testing by an accredited lab and has been submitted to NIST for certification.

Due to the nature of having two crate features for AWS-LC bidnings (`aws-lc`, and `aws-lc-fips`), this feature will take precedence if both features are specified due to feature unification by cargo.

### DIFF
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -74,11 +74,9 @@ fn check_ssl_kind() {
     }
 
     let is_aws_lc = cfg!(feature = "aws-lc");
+    let is_aws_lc_fips = cfg!(feature = "aws-lc-fips");
 
-    if is_aws_lc {
-        println!("cargo:rustc-cfg=awslc");
-        println!("cargo:awslc=true");
-
+    if is_aws_lc || is_aws_lc_fips {
         // The aws-lc-sys crate uses a link name that embeds
         // the version number of crate. Examples (crate-name => links name):
         //   * aws-lc-sys => aws_lc_0_26_0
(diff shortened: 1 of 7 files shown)


openssl-sys/build/main.rs
```diff
@@ -73,12 +73,10 @@ fn check_ssl_kind() {
         std::process::exit(0);
     }
 
-    let is_aws_lc = cfg!(feature = "aws-lc");
-
-    if is_aws_lc {
-        println!("cargo:rustc-cfg=awslc");
-        println!("cargo:awslc=true");
+    let is_aws_lc = cfg!(all(feature = "aws-lc", not(feature = "aws-lc-fips")));
```
alex: I think it'd be easier to read if this was just `is_aws_lc = cfg!(feature = "aws-lc")` and then below for the env var prefix just for fips first.

openssl-sys/src/lib.rs
```diff
@@ -38,7 +38,10 @@ mod aws_lc {
     #[cfg(feature = "aws-lc")]
     pub use aws_lc_sys::*;
 
-    #[cfg(not(feature = "aws-lc"))]
+    #[cfg(feature = "aws-lc-fips")]
+    pub use aws_lc_fips_sys::*;
+
```
alex: This will import both, which I think is a compilation error. My impression is we wanted to support feature-unifying aws-lc with aws-lc-fips into => aws-lc-fips only, right?

alex (approved):

PR #2426, opened 2025-07-02
Some API adjustments for LibreSSL 4.2.0

botovq:


alex:
Pretty sure that's a flakey test, that we ought to debug eventually, but not today.

alex (approved):

PR #2427, opened 2025-07-02
variety of fixes for warnings in new rust

alex:


alex:
I think this locking code may not be strictly sound (creates multiple `&mut` references to the `Vec`). But also I want to spend 0 time on it, and if I was going to, I'd want to just delete it all :-)

sfackler:
Yeah the locking callbacks are only used in EOL versions of OpenSSL so when we drop support for those it should all go away.

openssl-sys/build/main.rs
```diff
@@ -1,3 +1,8 @@
+#![allow(
+    // This can be removed once our MSRV is raised to 1.66.
```
sfackler, line 2: I'd be fine doing that now imo, but doesn't matter too much either way.

alex: I'm also fine bumping the MSRV (cryptography's is now 1.74 without incident), but I'm trying to only do one thing 😂

PR #2450, opened 2025-08-29
ML-KEM/ML-DSA part 1: openssl-sys changes

swenson:
Splitting up https://github.com/sfackler/rust-openssl/pull/2405 into a few parts as suggest by @alex.

This adds the non-param-builder openssl-sys changes.


openssl/build.rs
```diff
@@ -175,5 +177,11 @@ fn main() {
         if version >= 0x3_03_00_00_0 {
             println!("cargo:rustc-cfg=ossl330");
         }
+        if version >= 0x3_04_00_00_0 {
+            println!("cargo:rustc-cfg=ossl340");
+        }
+        if version >= 0x3_05_00_00_0 {
+            println!("cargo:rustc-cfg=ossl350");
+        }
```
alex, line 185: You also need these in openssl-sys/build I believe

swenson: Looks like 340 is already there; adding 350

alex (approved):

alex:
thanks for looking at this

PR #2453, opened 2025-08-29
sys: add symbols to construct an EVP_PKEY from a param builder

huwcbjones:
- **sys/params: add OSSL_PARAM_BLD symbols**
- **sys/params: add OSSL_PARAM dup, free, merge, locate, locate_const**
- **sys/evp: add selection consts**

### DIFF
--- a/openssl-sys/src/evp.rs
+++ b/openssl-sys/src/evp.rs
@@ -38,14 +38,14 @@ pub const EVP_CTRL_GCM_SET_IVLEN: c_int = 0x9;
 pub const EVP_CTRL_GCM_GET_TAG: c_int = 0x10;
 pub const EVP_CTRL_GCM_SET_TAG: c_int = 0x11;
 
-#[cfg(ossl300)]
-pub const EVP_PKEY_KEY_PARAMETERS: c_int = OSSL_KEYMGMT_SELECT_ALL_PARAMETERS;
-#[cfg(ossl300)]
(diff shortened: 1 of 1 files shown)


openssl-sys/src/handwritten/params.rs
```diff
@@ -1,8 +1,14 @@
 use super::super::*;
 use libc::*;
 
+#[cfg(ossl300)]
```
alex: I don't think this is required, the entire module has this cfg.

openssl-sys/src/handwritten/types.rs
```diff
@@ -1134,12 +1134,15 @@ pub enum OSSL_LIB_CTX {}
 #[repr(C)]
 pub struct OSSL_PARAM {
     key: *const c_char,
-    data_type: c_uchar,
+    data_type: c_uint,
```
alex, line 1137: how did it ever work!

openssl-sys/src/handwritten/params.rs
huwcbjones: Yep, removed - rebase conflict!

openssl-sys/src/handwritten/types.rs
huwcbjones: I think because the size is the same, but caught this whilst I was debugging segfaults 😅

alex (approved):

alex (approved):

PR #2454, opened 2025-08-29
sys: add encoder & decoder symbols

huwcbjones:
- **sys: add decoder symbols**
- **sys: add encoder symbols**


huwcbjones:
@alex please could I also get some eyes on this openssl-sys change?

PR #2455, opened 2025-08-29
ec-point: add set_affine_coordinates

huwcbjones:
`EC_POINT_set_affine_coordinates_GFp` is deprecated, `EC_POINT_set_affine_coordinates` is the replacement for the deprecated symbol.

- **sys/ec: add EC_POINT_set_affine_coordinates**
- **EcPointRef: add set_affine_coordinates**

### DIFF
--- a/openssl/src/ec.rs
+++ b/openssl/src/ec.rs
@@ -561,6 +561,29 @@ impl EcPointRef {
         }
     }
 
+    /// Sets affine coordinates of a point on an elliptic curve using the provided
+    /// `x` and `y` `BigNum`s
+    #[corresponds(EC_POINT_set_affine_coordinates)]
+    #[cfg(any(ossl111, boringssl, libressl350, awslc))]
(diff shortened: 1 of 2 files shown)


openssl/src/ec.rs
```diff
@@ -561,6 +561,29 @@ impl EcPointRef {
         }
     }
 
+    /// Sets affine coordinates of a curve over a prime field using the provided
```
botovq:
```suggestion
    /// Sets affine coordinates of a point on an elliptic curve using the provided
```

It sets the coordinates of a *point*. it works for both, prime and binary curves.

huwcbjones: Thanks for the sound feedback!
I just copied the docstring from `EC_POINT_set_affine_coordinates_GFp`, so not too surprised it's not that accurate 🙈

alex (commented):
Please add a test.

alex (approved):
Please add a test.

PR #2459, opened 2025-09-09
 ML-KEM/ML-DSA part 4: ML-DSA

swenson:
Splitting up #2405 into a few parts as suggest by @alex.

Adapting ML-DSA changes from #2405 to use the OsslParamBuilder changes over the last few PRs and to juggle lifetimes appropriately

### DIFF
--- a/openssl/src/pkey_ctx.rs
+++ b/openssl/src/pkey_ctx.rs
@@ -952,9 +952,8 @@ impl<T> PkeyCtxRef<T> {
         Ok(())
     }
 
-    /// Generates a new public/private keypair.
-    ///
-    /// New OpenSSL 3.0 function that should do the same thing as keygen()
+    /// Performs the generation operation and returns the resulting
+    /// key parameters or key.
     #[corresponds(EVP_PKEY_generate)]
     #[cfg(ossl300)]
     #[inline]


openssl/src/pkey_ml_dsa.rs
```diff
@@ -0,0 +1,332 @@
...
+    pub(crate) fn as_cstr(&self) -> &'static CStr {
+        match self {
+            Variant::MlDsa44 => MLDSA44_CSTR,
+            Variant::MlDsa65 => MLDSA65_CSTR,
+            Variant::MlDsa87 => MLDSA87_CSTR,
+        }
+    }
+}
+
+pub struct PKeyMlDsaBuilder<'a, 'opb, T>
```
alex: Why do we need a builder? There's exactly 2 APIs of consequnce: generate a new key, and load one from a seed.

openssl/src/pkey_ctx.rs
```diff
@@ -913,6 +990,14 @@ impl<T> PkeyCtxRef<T> {
         }
         Ok(NonceType(nonce_type))
     }
+
+    /// Initializes a conversion from `OsslParam` to `PKey` on given `PkeyCtx`.
```
alex, line 993: This seems like a really wacky API, and I don't even understand the theory of it. Why is `fromdata` an operation you'd do with an `EVPK_PKEY_CTX`?

openssl/src/pkey_ctx.rs
```diff
@@ -889,6 +952,20 @@ impl<T> PkeyCtxRef<T> {
         Ok(())
     }
 
+    /// Generates a new public/private keypair.
+    ///
+    /// New OpenSSL 3.0 function that should do the same thing as keygen()
```
alex: wat.

if they're behaviorally identical, a) why did OpenSSL expose it, b) can we just change what API keygen() uses on OpenSSL 3+?

openssl/src/pkey_ctx.rs
swenson: From the docs:

> EVP_PKEY_fromdata_init() initializes a public key algorithm context for creating a key or key parameters from user data

So if we're going to pass a param builder in, then we have to call this first it looks like.

openssl/src/pkey_ctx.rs
swenson: It looks like this comment is not accurate. This performs either paramgen or keygen depending on context. I'll update the comment. Either is an option.

> EVP_PKEY_paramgen() and EVP_PKEY_keygen() do exactly the same thing as EVP_PKEY_generate(), after checking that the corresponding EVP_PKEY_paramgen_init() or EVP_PKEY_keygen_init() was used to initialize ctx. These are older functions that are kept for backward compatibility. It is safe to use EVP_PKEY_generate() instead

openssl/src/pkey_ctx.rs
alex: I feel almost stupid for asking this, but why is an EVP_PKEY_CTX even involved in creating an EVP_PKEY? This is the complete opposite of how EVP_PKEY normally are (where you have a key and then you uuse the CTX to do operations with it)?

openssl/src/pkey_ml_dsa.rs
```diff
@@ -0,0 +1,254 @@
...
+        pkey: &PKey<S>,
+        selection: c_int,
+    ) -> Result<PKeyMlDsaParams<T>, ErrorStack> {
+        unsafe {
+            let mut params: *mut ffi::OSSL_PARAM = ptr::null_mut();
+            cvt(ffi::EVP_PKEY_todata(pkey.as_ptr(), selection, &mut params))?;
+            Ok(PKeyMlDsaParams::<T> {
+                params: OsslParamArray::from_ptr(params),
+                _m: PhantomData,
+            })
```
alex, line 137: What, if any, are the lifetime constraints an `OSSL_PARAM` from `EVP_PKEY_todata`?

openssl/src/signature.rs
```diff
@@ -0,0 +1,93 @@
...
+impl Clone for Signature {
+    fn clone(&self) -> Signature {
+        SignatureRef::to_owned(self)
+    }
+}
+
+impl Signature {
+    /// Creates a new `Signature` for use with ML-DSA.
+    #[cfg(ossl350)]
+    pub fn for_ml_dsa(variant: crate::pkey_ml_dsa::Variant) -> Result<Signature, ErrorStack> {
```
alex, line 70: Err, what's the use case for this being public?

openssl/src/pkey_ctx.rs
beldmit: You can use `EVP_PKEY_Q_keygen` for quick key generation, but normally EVP_PKEY_CTX is a structure keeping the operational context so it is involved.

openssl/src/pkey_ml_dsa.rs
beldmit: They are allocated on export and should be explicitly freed.

openssl/src/pkey_ml_dsa.rs
```diff
@@ -0,0 +1,254 @@
...
+    let params = OsslParamBuilder::new()?.to_param()?;
+    unsafe {
+        cvt(ffi::EVP_PKEY_CTX_set_params(ctx.as_ptr(), params.as_ptr()))?;
+    }
+    ctx.generate()
+}
+
+/// Returns the Private ML-DSA PKey from the provided parameters.
+/// Creates a new `PKeyMlDsaBuilder` to generate a new ML-DSA key
+/// pair.
```
alex, line 75: Comment is out of date

openssl/src/pkey_ml_dsa.rs
```diff
@@ -0,0 +1,254 @@
...
+        match self {
+            Variant::MlDsa44 => MLDSA44_CSTR,
+            Variant::MlDsa65 => MLDSA65_CSTR,
+            Variant::MlDsa87 => MLDSA87_CSTR,
+        }
+    }
+}
+
+/// Generate an ML-DSA PKey.
+#[corresponds(EVP_PKEY_fromdata)]
```
alex, line 62: I don't think this corresponds is particularly useful, it's too generic. We can drop it. (Same for all others like this)

openssl/src/pkey_ml_dsa.rs
```diff
@@ -0,0 +1,254 @@
...
+                _m: PhantomData,
+            })
+        }
+    }
+}
+
+impl PKeyMlDsaParams<Public> {
+    /// Creates a new `PKeyMlDsaParams` from existing Public ML-DSA PKey.
+    #[corresponds(EVP_PKEY_todata)]
+    pub fn from_pkey<S>(pkey: &PKey<S>) -> Result<PKeyMlDsaParams<Public>, ErrorStack> {
```
alex, line 145: None of the pkey using functions seem to actually verify the type of pkey?

Also, this inverts the way these APIs usually work. Usually there's a PKey method to get the type-specific reprsentation.

FinnRG: @alex Hi, I would like to move ahead with this work and create a separate PR which addresses your review comments. I just want to clarify regarding the pkey type verification:

As far as I can tell the cleanest way would be to use the `EVP_PKEY_is_a` method, since the ML-DSA keys return an id of -1. We currently don't seem to have `EVP_PKEY_is_a` in our ffi bindings, should I add it?

FinnRG:
I continued this PR in #2519.

swenson:
Thanks @FinnRG. Closing for now.

PR #2462, opened 2025-09-11
Add bindings for SSL_CIPHER_get_protocol_id

jedenastka:
[`SSL_CIPHER_get_protocol_id`](https://docs.openssl.org/master/man3/SSL_CIPHER_get_name/)

This includes both bindings for `openssl-sys` and high-level bindings for `openssl`. I wasn't sure whether to use `[u8; 2]` or `u16` for the high-level return type, decided on `[u8; 2]` for now as it feels more natural, I can change this if `u16` (the same as in C) is better.

### DIFF
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -2143,6 +2143,15 @@ impl SslCipherRef {
             Some(Nid::from_raw(n))
         }
     }
+
+    /// Returns the two-byte ID of the cipher
+    #[corresponds(SSL_CIPHER_get_protocol_id)]
+    pub fn protocol_id(&self) -> [u8; 2] {
+        unsafe {
+            let id = ffi::SSL_CIPHER_get_protocol_id(self.as_ptr());
+            id.to_be_bytes()
+        }
+    }
 }
 
 impl fmt::Debug for SslCipherRef {
--- a/openssl/src/ssl/mod.rs
+++ b/openssl/src/ssl/mod.rs
@@ -2146,6 +2146,7 @@ impl SslCipherRef {
 
     /// Returns the two-byte ID of the cipher
     #[corresponds(SSL_CIPHER_get_protocol_id)]
+    #[cfg(ossl111)]
     pub fn protocol_id(&self) -> [u8; 2] {
         unsafe {
             let id = ffi::SSL_CIPHER_get_protocol_id(self.as_ptr());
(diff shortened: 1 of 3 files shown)


openssl/src/ssl/mod.rs
```diff
@@ -2143,6 +2143,16 @@ impl SslCipherRef {
             Some(Nid::from_raw(n))
         }
     }
+
+    /// Returns the two-byte ID of the cipher
+    #[corresponds(SSL_CIPHER_get_protocol_id)]
+    #[cfg(ossl111)]
+    pub fn protocol_id(&self) -> [u8; 2] {
```
alex, line 2152: I'm almost positive the TLS spec defines these as u16s

jedenastka: I was almost positive it is two bytes, so I checked, and it is indeed two bytes: https://www.rfc-editor.org/rfc/rfc8446#appendix-B.3.1

alex: Huh, good call, thanks for double checking me!

openssl/src/ssl/mod.rs
```diff
@@ -2143,6 +2143,18 @@ impl SslCipherRef {
...
+
+    /// Returns the two-byte ID of the cipher
+    ///
+    /// Requires OpenSSL 1.1.1 or newer.
+    #[corresponds(SSL_CIPHER_get_protocol_id)]
+    #[cfg(ossl111)]
+    pub fn protocol_id(&self) -> [u8; 2] {
+        unsafe {
+            let id = ffi::SSL_CIPHER_get_protocol_id(self.as_ptr());
+            id.to_be_bytes()
```
alex, line 2155: Is big endian always right?

jedenastka: It definitely should be. Those are two byte constants (they are not defined as u16), moreover TLS is big-endian in general, even if they weren't: https://www.rfc-editor.org/rfc/rfc8446#section-3.1.

OpenSSL *could* have a bug with this though, I think it's unlikely, but I don't know. Two bytes are received, then they are turned into an integer, which is native-endian - in other words, on little-endian architectures a byte swap is necessary. It seems like on all architectures you test it did properly byte swap as needed, since tests passed, but I don't see a test on PowerPC so who knows. I don't have a big-endian machine to test myself.

Simply said - on big-endian architectures, there should be no byte swap, and on little-endian there should be two byte swaps, which cancel each other out.

botovq: Building the branch on a sparc64 machine with OpenSSL 3.5 gives:

```
test ssl::test::cipher_id ... ok
```

alex (approved):
great, thanks

PR #2472, opened 2025-10-08
drop support for libressl < 3.0

reaperhulk:


openssl-sys/src/bio.rs
```diff
@@ -34,37 +34,37 @@ pub unsafe fn BIO_get_mem_data(b: *mut BIO, pp: *mut *mut c_char) -> c_long {
 
 extern "C" {
     #[deprecated(note = "use BIO_meth_set_write__fixed_rust instead")]
-    #[cfg(any(ossl110, libressl273))]
+    #[cfg(ossl110)]
```
botovq, line 37: Is there someting enabling this for libressl via ossl110? Otherwise this doesn't look right.

reaperhulk: Yeah this is going to be wrong since it doesn't take both a libre and openssl path when emitting cfgs. It's also still way too much code to review effectively so any removal probably needs to be more incremental (this was already an attempt at incremental).

PR #2485, opened 2025-10-13
clippy fixes + don't build locking code on libressl

alex:


openssl-sys/src/lib.rs
```diff
@@ -158,7 +158,10 @@ mod openssl {
         })
     }
 
-    #[cfg(not(ossl110))]
+    #[cfg(libressl)]
+    pub fn init() {}
+
+    #[cfg(all(not(ossl110), not(libressl)))]
```
botovq: I would have done this, but if you prefer your way, I'm not fussed:

```suggestion
    #[cfg(not(any(ossl110, libressl)))]
```

PR #2498, opened 2025-10-14
Add support for LibreSSL 4.2 (stable release)

botovq:


openssl-sys/build/main.rs
```diff
@@ -435,6 +435,8 @@ See rust-openssl documentation for more information:
             (4, 0, _) => ('4', '0', 'x'),
             (4, 1, 0) => ('4', '1', '0'),
             (4, 1, _) => ('4', '1', 'x'),
+            (4, 2, 0) => ('4', '2', '0'),
+            (4, 2, _) => ('4', '2', 'x'),
```
alex: If this is immediately stable, do we actually need this distinction?

botovq: I guess we don't. It matches what we had since 3.9.0.

PR #2499, opened 2025-10-14
Remove support for LibreSSL < 3.5

botovq:


openssl-sys/src/handwritten/x509.rs
```diff
@@ -205,12 +205,12 @@ extern "C" {
 
 const_ptr_api! {
     extern "C" {
-        #[cfg(any(ossl102, libressl350))]
+        #[cfg(any(ossl102, libressl))]
```
alex: this is now a no-op, right?

botovq: I suppose so but I'm unsure: I haven't looked closely at how boringssl and aws-lc are hooked in, but if I understand correctly they both use bindgen, so don't run the handwritten code?

alex: Yeah, that's right.

PR #2507, opened 2025-10-23
SSL{,_CTX}_get_ex_new_index: expose CRYPTO-based definition to LibreSSL

vishwin:
Not sure how this got missed until some linker errors I experienced now, but this has apparently been the case for time. Present in all currently supported LibreSSL versions.

### DIFF
--- a/openssl-sys/src/handwritten/ssl.rs
+++ b/openssl-sys/src/handwritten/ssl.rs
@@ -735,7 +735,7 @@ extern "C" {
 }
 
 extern "C" {
-    #[cfg(not(ossl110))]
+    #[cfg(not(any(ossl110, libressl)))]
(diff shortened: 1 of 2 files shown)


botovq:
I would like to understand the motivation for this change better. What is that linker error you run into?

I don't particularly like replacing the binding to an existing public symbol with a hand-written translation to rust just because for some reason OpenSSL decided to replace said symbol with a preprocessor macro that rust doesn't understand.

vishwin:
This is no longer necessary, turns out there was some symbol clashing between this and rustls (ie embedded AWS-LC) with one of the dependent crates that forcing rustls there fixed the issue.

botovq:
Thanks for the update.

PR #2508, opened 2025-10-28
Implement set_rsa_oaep_label for AWS-LC/BoringSSL

goffrie:
This function works on boringssl derivatives too, just with different integer and pointer types.


alex:
Sigh, MSRV CI appears busto. Need to fix before we can land stuff.

PR #2512, opened 2025-10-30
openssl: add wrappers around EVP_MAC APIs

huwcbjones:
Adds `Mac` and `MacRef`

### DIFF
--- a/openssl/src/lib.rs
+++ b/openssl/src/lib.rs
@@ -171,6 +171,8 @@ pub mod hash;
 pub mod kdf;
 #[cfg(ossl300)]
 pub mod lib_ctx;
+#[cfg(ossl300)]
+pub mod mac;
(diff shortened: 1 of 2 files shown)


openssl/src/lib.rs
```diff
@@ -171,6 +171,8 @@ pub mod hash;
 pub mod kdf;
 #[cfg(ossl300)]
 pub mod lib_ctx;
+#[cfg(ossl300)]
+pub mod mac;
```
huwcbjones, line 175: Public mod because I want it to replace the deprecated use of `HMAC` and `CMAC` in both:
* https://github.com/pyca/cryptography/blob/main/src/rust/src/backend/cmac.rs
* https://github.com/pyca/cryptography/blob/main/src/rust/src/backend/hmac.rs

PR #2517, opened 2025-11-06
Fix unsound OCSP find_status handling of optional next_update field

alex:
The OCSP find_status function was unsound because it treated next_update as always present, even though it's optional per RFC 6960. When absent, the null pointer from FFI was passed to from_ptr which doesn't check for null, causing undefined behavior.

Fixed by using from_const_ptr_opt to safely handle the null case. Added next_update() method returning Option<&Asn1GeneralizedTimeRef> and deprecated the existing field which now contains a sentinel max time value (99991231235959Z) for backwards compatibility.

Fixes https://github.com/rust-openssl/rust-openssl/issues/2516


alex:
I don't understand the warning in the MSRV builder. We have an `allow()`, does that not work on older Rust?

openssl/src/ocsp.rs
```diff
@@ -14,6 +15,27 @@ use crate::x509::{X509Ref, X509};
...
 
+// Sentinel value used when next_update is not present in OCSP response
+// This represents the maximum possible time (9999-12-31 23:59:59 UTC)
+static SENTINEL_MAX_TIME: OnceLock<Asn1Time> = OnceLock::new();
+
+fn get_sentinel_max_time() -> &'static Asn1GeneralizedTimeRef {
+    // ASN1_TIME and ASN1_GENERALIZEDTIME have compatible representations
+    // and can be safely cast. ASN1_TIME is the generic type that can hold
+    // both UTCTIME and GENERALIZEDTIME. For dates outside 1950-2049,
+    // it automatically uses GENERALIZEDTIME format.
```
botovq: This is somewhat confusingly and misleadingly describing the RFC 5280 semantics.

In OpenSSL the `ASN1_TIME_set_string()` function (underlying `Asn1Time::from_str()`) does *not* use these semantics, that's what `ASN1_TIME_set_string_X509()` is for. It decides between UtcTime and GeneralizedTime solely on basis of the length of the string. This does not matter here, but it will happily accept `20251106080730Z` as a GeneralizedTime, for example.

Regarding the "safely cast" bit, that's a bit sloppily phrased. It is safe to "cast" from `ASN1_GENERALIZEDTIME` to `ASN1_TIME` (they have the same underlying C type), but not the other way around.

openssl/src/ocsp.rs
```diff
@@ -14,6 +15,27 @@ use crate::x509::{X509Ref, X509};
...
+    let time = SENTINEL_MAX_TIME.get_or_init(|| {
+        // Create an Asn1Time for 99991231235959Z (max time)
```
botovq: Instead of stating the obvious and repeating what we already know from prior comments, I would suggest:

```suggestion
        // XXX - use `from_str_x509()` once the oldest supported OpenSSL is 1.1.1
```

botovq (commented):
The code itself reads fine modulo one thing that could be better if we didn't support ancient library versions.

(I really do not like Claude's overly repetitive and verbose commenting style.)

openssl/src/ocsp.rs
alex: The better fix is that really I should just add a function that calls `ASN1_GENERALIZEDTIME_set_string`.

botovq: The downside of doing that is that it has the same issue of not ensuring that the result is what X.509/RFC 5280 want.

alex: `nextUpdate` is _always_ a GeneralizedTime -- it's not a Time (CHOICE) like many other X.509 things.

botovq: Ah. Then it makes sense. I agree.

botovq: OCSP is so weird

openssl/src/asn1.rs
```diff
@@ -84,6 +84,29 @@ impl fmt::Display for Asn1GeneralizedTimeRef {
...
+            let time_str = CString::new(s).unwrap();
+            let ptr = cvt_p(ffi::ASN1_GENERALIZEDTIME_new())?;
+            let time = Asn1GeneralizedTime::from_ptr(ptr);
+
+            cvt(ffi::ASN1_GENERALIZEDTIME_set_string(
+                time.as_ptr(),
+                time_str.as_ptr(),
+            ))?;
+
+            Ok(time)
```
botovq, line 105: Probably just a matter of taste, but I'd have written this as
```suggestion
            cvt(ffi::ASN1_GENERALIZEDTIME_set_string(ptr, time_str.as_ptr()))?;
            Ok(Asn1GeneralizedTime::from_ptr(ptr))
```

alex: Then you have a leak if `ASN1_GENERALIZEDTIME_set_string` fails

botovq: Oh right. It's been too long... Thanks.

PR #2519, opened 2025-11-13
Add bindings for ML-DSA

FinnRG:
Based on #2459. I addressed all review comments and refactored the key generation API.
For example `pkey_ml_dsa::new_generate(variant)` is now a method on PKey: `PKey::generate_mldsa(variant)`. This is more in line with the other key types.

I had to extend `openssl-sys` with various new methods, because `ML-DSA` doesn't have a NID and needs to be handled with new methods like `EVP_PKEY_new_raw_private_key_ex`, `EVP_PKEY_new_raw_public_key_ex` and `EVP_PKEY_CTX_new_from_name`.

I've also added a few new tests based on boringssl test vectors.

### DIFF
--- a/openssl/src/pkey.rs
+++ b/openssl/src/pkey.rs
@@ -706,7 +706,7 @@ impl PKey<Private> {
 
     /// Generates a new ML-DSA key with the provided variant.
     ///
-    /// Requires OpenSSL 3.0.0 or newer.
+    /// Requires OpenSSL 3.5.0 or newer.
     #[cfg(ossl350)]
     pub fn generate_ml_dsa(variant: pkey_ml_dsa::Variant) -> Result<PKey<Private>, ErrorStack> {
         Self::generate_key_from_name(variant.as_str())
--- a/openssl/src/pkey_ml_dsa.rs
+++ b/openssl/src/pkey_ml_dsa.rs
@@ -14,7 +14,7 @@ use std::ffi::CStr;
 use std::marker::PhantomData;
 
 // Safety: these all have null terminators.
-// We cen remove these CStr::from_bytes_with_nul_unchecked calls
+// We can remove these CStr::from_bytes_with_nul_unchecked calls
 // when we upgrade to Rust 1.77+ with literal c"" syntax.
 const OSSL_PKEY_PARAM_SEED: &CStr = unsafe { CStr::from_bytes_with_nul_unchecked(b"seed\0") };
 const OSSL_PKEY_PARAM_PUB_KEY: &CStr = unsafe { CStr::from_bytes_with_nul_unchecked(b"pub\0") };


openssl-sys/src/handwritten/evp.rs
```diff
@@ -742,6 +737,30 @@ cfg_if! {
         }
     }
 }
+cfg_if! {
```
alex, line 746: These can be split out into a separate PR -- merging new bindings is always easy (and keeps the size of this PR down).

openssl/src/pkey.rs
```diff
@@ -207,6 +230,15 @@ impl<T> PKeyRef<T> {
     pub fn size(&self) -> usize {
         unsafe { ffi::EVP_PKEY_size(self.as_ptr()) as usize }
     }
+
+    /// Returns true if the key matches the string key_type.
+    #[corresponds(EVP_PKEY_is_a)]
+    #[cfg(ossl300)]
+    pub fn is_key_type(&self, key_type: &str) -> Result<bool, ErrorStack> {
```
alex: Hmm, let's not make this public for now.

openssl/src/pkey.rs
```diff
@@ -779,6 +832,27 @@ impl PKey<Private> {
             .map(|p| PKey::from_ptr(p))
         }
     }
+
+    /// Creates a private key from its raw byte representation using a string keytype
+    #[corresponds(EVP_PKEY_new_raw_private_key)]
+    #[cfg(ossl300)]
+    pub fn private_key_from_raw_bytes_ex(
```
alex, line 839: is there anyways to consolidate this with the existing private key from bytes?

alex (commented):
(I only started reviewing, didn't finish, but wanted to leave these since I won't get back to this for a bit)

openssl/src/pkey.rs
FinnRG: We could refactor Id to have a string internally and then change `private_key_from_raw_bytes` to use the `_ex` method for all key types. I'm not sure if this will have larger implications though (and it would be a larger refactoring effort).

As I understand it the supported key types from `private_key_from_raw_bytes_ex` are a superset of `private_key_from_raw_bytes`.

FinnRG: Also `EVP_PKEY_new_raw_public_key_ex` was only introduced in ossl300, while `EVP_PKEY_new_raw_public_key` exists since ossl110.

openssl-sys/src/handwritten/evp.rs
FinnRG: I've created a separate PR #2521. Once that is merged I'll rebase this PR onto master.

FinnRG: I've removed the openssl-sys changes from this branch and synced it with master.

openssl/src/pkey.rs
```diff
@@ -207,6 +230,15 @@ impl<T> PKeyRef<T> {
     pub fn size(&self) -> usize {
         unsafe { ffi::EVP_PKEY_size(self.as_ptr()) as usize }
     }
+
+    /// Returns true if the key matches the string key_type.
+    #[corresponds(EVP_PKEY_is_a)]
+    #[cfg(ossl350)] // Available since ossl300, but currently only used in a function gated behind ossl350
+    fn is_key_type(&self, key_type: &str) -> Result<bool, ErrorStack> {
+        let key_type = CString::new(key_type).unwrap();
```
DemiMarie, line 238: Any reason to not take `&CStr` and avoid an allocation?

frasertweedale:
Hi @FinnRG and all,

Based on this PR and @abbra's experimental branch, I have pushed this work a bit further including adding BoringSSL support.  I am in the process of cleaning up the code and commit history and will prepare a new PR(s) soon.

Fixup commits, and in some cases more substantive commits, for the sake of clarity and ease of reviewing, will be squashed down into the relevant preceding commits.  In cases where commit authorship differs, I acknowledge the contributions via `Co-Authored-By: ` lines in the commit message.  Primary authorship of smaller commits may be ceded in some cases to the author of the larger commit.  I want to be transparent about this ahead of time.  If you have concerns (now, or after I submit the PRs) please do not hesitate to speak up.

In line with expressed priorities of python-cryptography upstream, I also intend to **reorder** the work somewhat to ensure BoringSSL support comes before OpenSSL.  And certainly ML-DSA before ML-KEM.

openssl/src/pkey.rs
frasertweedale: To do this neatly *and consistently* is a bigger change than it seems at first.  I will have a go at it and see if the juice is worth the squeeze.

PR #2523, opened 2025-11-20
Fix invalid value parsing of OCSP revocation reason

danpashin:
Seems like [920ab0d](https://github.com/rust-openssl/rust-openssl/commit/920ab0d6fb60c17077f43d7f08ad3ff391201689) introduced a copy-pasted behaviour in OCSP module https://github.com/rust-openssl/rust-openssl/blob/920ab0d6fb60c17077f43d7f08ad3ff391201689/openssl/src/ocsp.rs#L179-L180 which makes `reason` field to be always equal to `status` field and drop the original value. I believe there must be another behavior like https://github.com/rust-openssl/rust-openssl/blob/f2a461c1f4f76f2cfea18080ff759217da53146b/openssl/src/ocsp.rs#L239-L240 which makes more sense in terms of reason field


alex:
@botovq any reason you didn't merge?

botovq:
@alex no. only status == distracted.

Issue #2527, opened 2025-11-25
Reintroduce support for SubjectAlternativeName::dir_name


mqqz:
### Background
`SubjectAlternativeName::dir_name` has been deprecated since v0.10.48 by [this commit](https://github.com/rust-openssl/rust-openssl/commit/482575bff434f58b80ffea34a9610d0ff265ac1f#diff-3105f7585b1063b001e5ff3f482774ddf31242e7390eb0f9b61fbd145754f223R544-R550) to patch an [arbitrary read file vulnerability](https://rustsec.org/advisories/RUSTSEC-2023-0023.html).

> Fixed injection vulnerabilities where OpenSSL's configuration mini-language could be used via `x509::extension::SubjectAlternativeName` and `x509::extension::ExtendedKeyUsage`. The mini-language can read arbitrary files amongst other things.
> As part of fixing this `SubjectAlternativeName::dir_name` and `SubjectAlternativeName::other_name` are deprecated and their implementations always panic!. If you have a use case for these, **please file an issue**.

### Use Case
... and unfortunately, I find myself in a place where I need to use `dir_name`. Namely, I need to generate CSRs that match a government spec which requires a `SubjectAlternativeName` of type `DirectoryName` (see [section 2.2.2 if you care](https://zatca.gov.sa/ar/E-Invoicing/SystemsDevelopers/Documents/20230519_ZATCA_Electronic_Invoice_Security_Features_Implementation_Standards_vF.pdf#page=16)).

### Fix 
Support for `other_name` has been since reintroduced with issue #1911 and PR #1915 (by adding new `other_name2`), so we could do something similar to make `dir_name` compatible with the new internals i.e. 
- Add `SubjectAlternativeName::dir_name2(X509Name)` in `openssl/src/x509/extension.rs`
  - Internally constructs a `GENERAL_NAME` with type `GEN_DIRNAME`  via  `GeneralName::new_dir_name()` 
  
Hopefully, by using a pre-constructed `X509Name` argument, no config language and no arbitrary file reads.

_Optionally_, we could add support for [GENERAL_NAME_set1_X509_NAME](https://docs.openssl.org/master/man3/GENERAL_NAME) for OpenSSL 3.4+.

-------

I **will open a PR soon** that implements this fix, or at minimum provides a working proof-of-concept to guide the proper implementation.

PR #2528, opened 2025-11-25
reintroduce dir_name support for subject_alt_names

mqqz:
This reintroduces support for `GENERAL_NAME` values of type `GEN_DIRNAME` without relying on OpenSSL’s configuration mini-language. The deprecated `dir_name(&str)` constructor remains unchanged and continues to panic.

A new method, dir_name2(X509Name), is added to construct `directoryName` SAN entries using an owned X509Name. Internally, this uses safe duplication (`X509_NAME_dup` or `.to_owned()`) and manually builds a `GENERAL_NAME` with G`EN_DIRNAME`, following the same pattern used for `other_name2` in #1915.

This avoids the vulnerability described in RUSTSEC-2023-0023 while restoring needed functionality for consumers that rely on DirectoryName SANs (e.g., government CSR specifications). New tests verify correctness, pointer ownership, and OpenSSL/BoringSSL/AWS-LC union handling.

Addresses #2527

### DIFF
--- a/openssl/src/x509/mod.rs
+++ b/openssl/src/x509/mod.rs
@@ -2087,6 +2087,35 @@ impl GeneralName {
             Ok(GeneralName::from_ptr(gn))
         }
     }
+
+    pub(crate) fn new_dir_name(name: &X509NameRef) -> Result<GeneralName, ErrorStack> {
+        unsafe {
+            ffi::init();
+            let gn = cvt_p(ffi::GENERAL_NAME_new())?;
+            (*gn).type_ = ffi::GEN_DIRNAME;
+
+            let dup = match name.to_owned() {
+                Ok(dup) => dup,
+                Err(e) => {
+                    ffi::GENERAL_NAME_free(gn);
+                    return Err(e);
+                }
+            };
+
+            #[cfg(any(boringssl, awslc))]
+            {
+                (*gn).d.directoryName = dup.as_ptr();
+            }
+            #[cfg(not(any(boringssl, awslc)))]
+            {
+                (*gn).d = dup.as_ptr().cast();
+            }
+
+            std::mem::forget(dup);
+
+            Ok(GeneralName::from_ptr(gn))
+        }
+    }
 }
 
 impl GeneralNameRef {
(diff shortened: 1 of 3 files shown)


openssl/src/x509/mod.rs
```diff
@@ -2087,6 +2087,50 @@ impl GeneralName {
...
+                Ok(dup) => dup,
+                Err(e) => {
+                    ffi::GENERAL_NAME_free(gn);
+                    return Err(e);
+                }
+            };
+
+            #[cfg(not(any(boringssl, ossl110, libressl, awslc)))]
+            let dup = {
+                let ptr = match cvt_p(ffi::X509_NAME_dup(name.as_ptr())) {
```
alex: if this exists, is there a reason `to_owned` can't just exist on `name`?

mqqz: silly silly me. 
I tried to fix it by doing a workaround for OpenSSL 1.0.x by converting to der and back
```rust
            #[cfg(not(any(boringssl, ossl110, libressl, awslc)))]
            let dup = {
                // OpenSSL 1.0.x doesn't have X509_NAME_dup, use DER roundtrip
                let der = match name.to_der() {
                    Ok(der) => der,
                    Err(e) => {
                        ffi::GENERAL_NAME_free(gn);
                        return Err(e);
                    }
                };
                match X509Name::from_der(&der) {
                    Ok(name) => name,
                    Err(e) => {
                        ffi::GENERAL_NAME_free(gn);
                        return Err(e);
                    }
                }
            };
 ```
 do you think this is ok?

Comment (alex) [reply openssl/src/x509/mod.rs]: https://github.com/rust-openssl/rust-openssl/pull/2529 should expose to_owned unconditionally.

Comment (mqqz) [reply openssl/src/x509/mod.rs]: @alex nice I'll edit to keep `to_owned` only

Comment (mqqz) [reply openssl/src/x509/mod.rs]: done

File: openssl/src/x509/tests.rs [hunk 2]
```diff
@@ -1143,6 +1145,83 @@ fn other_name_as_subject_alternative_name() {
...
+    );
+
+    // Ensure no pointer aliasing between original and SAN copy
+    assert_ne!(name.as_ptr(), dirname.as_ptr());
+
+    // Check underlying GENERAL_NAME type and union layout
+    let first = alt_names.iter().next().unwrap();
+    unsafe {
+        assert_eq!((*first.as_ptr()).type_, ffi::GEN_DIRNAME);
+    }
```
alex: We shouldn't need to make assertions against the internals

mqqz: fixed, I removed it

PR #2529, opened 2025-11-25
Expose X509_NAME_dup on all versions of OpenSSL

alex:
Remove the version restriction (#[cfg(any(ossl110, libressl))]) from X509_NAME_dup in openssl-sys and the corresponding X509NameRef::to_owned() method in openssl. This function has been available since early versions of OpenSSL, so there's no need to restrict it to 1.1.0+.

Also removes the version restriction from the test_name_to_owned test.


botovq (approved):
You could also remove the cfg for `X509_dup`, `X509_CRL_dup` and `X509_REQ_dup` which were added at the same time. They have all been available since the last millenium.

alex:
Making a note to do that tomorrow, for now going to bed :-)

PR #2531, opened 2025-12-01
Fix cipher comparison with NID instead of pointers 

lwestlund:
The implementation of `Cipher::is_ccm` and `Cipher::is_ocb` currently relies on that OpenSSL returned pointers to static structs when this is in fact not true. Depending on the state of the library you can get different pointers, see examples.

The assumption that a cipher can be identified based on pointer comparison has consequences for at least `openssl::symm::{encrypt,decrypt}_aead()` (what failed causing me to find the problem) since those function will do the wrong thing when failing to identify the cipher correctly.

This PR changes the implementation to use NID instead of pointer to compare identity of the cipher algorithms `aes_<bits>_{ccm,ocb}`.

### Examples

These showcase that the pointers are not static and depends on the library state. Both "work" with CCM and OCB algorithms.

C, compile and run with `gcc main.c -lssl -lcrypto && ./a.out`.

```c
#include <openssl/evp.h>
#include <openssl/types.h>
#include <stdio.h>

int main(void) {
    const EVP_CIPHER *p1 = EVP_aes_256_ccm();
    printf("p1: %p\n", p1);

    // Change the library state.
    const EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_id(EVP_PKEY_X25519, NULL);

    const EVP_CIPHER *p2 = EVP_aes_256_ccm();
    printf("p2: %p\n", p2);

    return 0;
}
```

Rust, build with `openssl` as dependency.

```rust
fn main() {
    let c1 = openssl::symm::Cipher::aes_256_ccm();
    println!("p1: {:p}", c1.as_ptr());

    // Change the library state.
    let _ctx = openssl::pkey_ctx::PkeyCtx::new_id(openssl::pkey::Id::X25519).unwrap();

    let c2 = openssl::symm::Cipher::aes_256_ccm();
    println!("p2: {:p}", c2.as_ptr());
}
```

Both are consistent in giving me output like

```
p1: 0x7f8581d02380
p2: 0x7f8581d02240
```

### DIFF
--- a/openssl/src/symm.rs
+++ b/openssl/src/symm.rs
@@ -514,8 +514,10 @@ impl Cipher {
     /// Determines whether the cipher is using CCM mode
     #[cfg(not(any(boringssl, awslc)))]
     fn is_ccm(self) -> bool {
-        // NOTE: OpenSSL returns pointers to static structs, which makes this work as expected
-        self == Cipher::aes_128_ccm() || self == Cipher::aes_256_ccm()
+        matches!(
+            self.nid(),
+            Nid::AES_128_CCM | Nid::AES_192_CCM | Nid::AES_256_CCM
+        )
     }
 
     #[cfg(any(boringssl, awslc))]
@@ -526,9 +528,10 @@ impl Cipher {
     /// Determines whether the cipher is using OCB mode
     #[cfg(all(ossl110, not(osslconf = "OPENSSL_NO_OCB")))]
     fn is_ocb(self) -> bool {
-        self == Cipher::aes_128_ocb()
-            || self == Cipher::aes_192_ocb()
-            || self == Cipher::aes_256_ocb()
+        matches!(
+            self.nid(),
+            Nid::AES_128_OCB | Nid::AES_192_OCB | Nid::AES_256_OCB
+        )
     }
 
     #[cfg(any(not(ossl110), osslconf = "OPENSSL_NO_OCB"))]


openssl-sys/src/obj_mac.rs
```diff
@@ -631,6 +631,9 @@ pub const NID_aes_192_ctr: c_int = 905;
 pub const NID_aes_256_ctr: c_int = 906;
 pub const NID_aes_128_xts: c_int = 913;
 pub const NID_aes_256_xts: c_int = 914;
+pub const NID_aes_128_ocb: c_int = 958;
+pub const NID_aes_192_ocb: c_int = 959;
+pub const NID_aes_256_ocb: c_int = 960;
```
botovq, line 639: These were added in OpenSSL 1.1.0, after BoringSSL and LibreSSL forked, which never added OCB, so

```suggestion
#[cfg(ossl110)]
pub const NID_aes_128_ocb: c_int = 958;
#[cfg(ossl110)]
pub const NID_aes_192_ocb: c_int = 959;
#[cfg(ossl110)]
pub const NID_aes_256_ocb: c_int = 960;
```

openssl/src/nid.rs
```diff
@@ -746,6 +746,9 @@ impl Nid {
     pub const AES_256_CTR: Nid = Nid(ffi::NID_aes_256_ctr);
     pub const AES_128_XTS: Nid = Nid(ffi::NID_aes_128_xts);
     pub const AES_256_XTS: Nid = Nid(ffi::NID_aes_256_xts);
+    pub const AES_128_OCB: Nid = Nid(ffi::NID_aes_128_ocb);
+    pub const AES_192_OCB: Nid = Nid(ffi::NID_aes_192_ocb);
+    pub const AES_256_OCB: Nid = Nid(ffi::NID_aes_256_ocb);
```
botovq, line 754: Same here:
```suggestion
    #[cfg(ossl110)]
    pub const AES_128_OCB: Nid = Nid(ffi::NID_aes_128_ocb);
    #[cfg(ossl110)]
    pub const AES_192_OCB: Nid = Nid(ffi::NID_aes_192_ocb);
    #[cfg(ossl110)]
    pub const AES_256_OCB: Nid = Nid(ffi::NID_aes_256_ocb);
```

openssl-sys/src/obj_mac.rs
lwestlund: Thanks for the quick feedback! I pushed a fixup with this applied in 04deedd971bf505709a48f9604d701fcbc79e877

PR #2536, opened 2025-12-07
X509ReqBuilder: add attribute

viktorTarasov:
Add attribute to X509Req.
Check if the attribute of a given type is present in X509Req.

X509 attribute (to be added) is defined by its Asn1Object and Asn1Type of value.

Needed for the IoT enrollment procedures:
-- 'challengePassword' for SCEP, EST, BRSKI,
-- 'changeSubjectName', 'masa-URL'  - EST, BRSKI
...

### DIFF
--- a/openssl-sys/src/handwritten/x509.rs
+++ b/openssl-sys/src/handwritten/x509.rs
@@ -400,6 +400,15 @@ extern "C" {
         len: c_int,
     ) -> c_int;
 }
+const_ptr_api! {
+    extern "C" {
+        pub fn X509_REQ_get_attr_by_OBJ(
+            req: *const X509_REQ,
+            obj: #[const_ptr_if(any(ossl110, libressl))] ASN1_OBJECT,
+            lastpos: c_int,
+        ) -> c_int;
+    }
+}
 extern "C" {
     pub fn X509_set_pubkey(x: *mut X509, pkey: *mut EVP_PKEY) -> c_int;
     pub fn X509_REQ_verify(req: *mut X509_REQ, pkey: *mut EVP_PKEY) -> c_int;
--- a/openssl/src/x509/mod.rs
+++ b/openssl/src/x509/mod.rs
@@ -1431,6 +1431,25 @@ impl X509ReqBuilder {
         }
     }
 
+    /// Adds attribute defined by Asn1Object and type of value to certificate request
+    pub fn add_attribute(
+        &mut self,
+        attrname: &Asn1Object,
+        attrtype: Asn1Type,
+        value: &[u8],
+    ) -> Result<(), ErrorStack> {
+        unsafe {
+            cvt(ffi::X509_REQ_add1_attr_by_OBJ(
+                self.0.as_ptr(),
+                attrname.as_ptr(),
+                attrtype.as_raw(),
+                value.as_ptr(),
+                value.len().try_into().expect("value's len outside range"),
+            ))
+            .map(|_| ())
+        }
+    }
+
     /// Sign the request using a private key.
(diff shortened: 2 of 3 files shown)


openssl/src/x509/mod.rs
```diff
@@ -1431,6 +1431,26 @@ impl X509ReqBuilder {
...
+    /// Adds attribute defined by name and type to certificate request
+    pub fn add_attribute(
+        &mut self,
+        attrname: &str,
+        attrtype: Asn1Type,
+        value: &[u8],
+    ) -> Result<(), ErrorStack> {
+        unsafe {
+            let attrname = CString::new(attrname).unwrap();
+            cvt(ffi::X509_REQ_add1_attr_by_txt(
```
alex: I think it'd be preferrable to have our API built around `X509_REQ_add1_attr_by_OBJ` (and take an OID as an argument)

openssl/src/x509/mod.rs
```diff
@@ -1431,6 +1431,26 @@ impl X509ReqBuilder {
...
+                self.0.as_ptr(),
+                attrname.as_ptr(),
+                attrtype.as_raw(),
+                value.as_ptr(),
+                value.len() as c_int,
```
alex: Please use `value.len().try_into().expect("value's len outside range")` so that there's no silent truncation.

openssl/src/x509/tests.rs
```diff
@@ -450,6 +451,10 @@ fn x509_req_builder() {
     extensions.push(subject_alternative_name).unwrap();
     builder.add_extensions(&extensions).unwrap();
 
+    builder
+        .add_attribute("challengePassword", Asn1Type::PRINTABLESTRING, b"password")
+        .unwrap();
+
```
alex, line 468: Please add some sort of assertion that this worked to the test.

alex (commented):
Thanks for contributing!

openssl/src/x509/mod.rs
viktorTarasov: Done

openssl/src/x509/tests.rs
viktorTarasov: Two 'is-ok()' assertions, and one 'is-err()' have been added

openssl/src/x509/mod.rs
viktorTarasov: Ok.
Discovered that Asn1Object can be initiated from numeric OID string, as well as from the 'name', short or long.

openssl-sys/src/handwritten/x509.rs
```diff
@@ -376,6 +376,11 @@ const_ptr_api! {
 extern "C" {
     pub fn X509_REQ_get_attr_count(req: *const X509_REQ) -> c_int;
     pub fn X509_REQ_get_attr_by_NID(req: *const X509_REQ, nid: c_int, lastpos: c_int) -> c_int;
+    pub fn X509_REQ_get_attr_by_OBJ(
+        req: *const X509_REQ,
+        obj: *const ASN1_OBJECT,
+        lastpos: c_int,
+    ) -> c_int;
```
botovq: As a first step towards fixing the breakage in 1.0.2, you'll want to move this into a `const_ptr_api!{}` block and do this:

```suggestion
    pub fn X509_REQ_get_attr_by_OBJ(
        req: *const X509_REQ,
        obj: #[const_ptr_if(any(ossl110, libressl))] ASN1_OBJECT,
        lastpos: c_int,
    ) -> c_int;
```

botovq:
Feel free to squash and rebase, but we haven't done a full review yet and we will need another rebase once the CI is fixed.

PR #2538, opened 2025-12-11
x509: add minimal support for CRL building

DavidFontaineOcd:
**Summary**

This pull request introduces initial support for Certificate Revocation List (CRL) functionality. The current implementation of the library exposes many X.509 certificate features, but the APIs related to CRL creation and manipulation are still missing. This PR fills part of that gap by adding the essential types and extensions required to build and manage CRLs programmatically.

**Motivation**

The OpenSSL C API provides full support for generating, parsing, and extending CRLs. However, rust-openssl currently lacks high-level Rust bindings for these components. As a result, users cannot create CRLs, add revoked certificate entries, or include the standard CRL-related extensions expected in real-world PKI workflows. This PR aims to improve feature parity with OpenSSL and enable broader CRL use cases within Rust applications.

**What’s Added**

This contribution introduces wrappers and bindings for several key X.509 structures used in CRL creation:

- `X509RevokedBuilder`
- `AccessDescriptionBuilder`
- `DistributionPointBuilder`
- `DistributionPointNameBuilder`

Additionally, the PR adds support for CRL extensions and CRL entry extensions, which are necessary for producing standards-compliant CRLs (e.g., CRL Number, Authority Key Identifier, CRL Reason Codes, Issuing Distribution Point).

**Benefits**

Enables programmatic creation and customization of CRLs in Rust
Improves parity with the OpenSSL CRL API
Helps developers build complete PKI workflows without relying on external tools

### DIFF
--- a/openssl/src/x509/extension.rs
+++ b/openssl/src/x509/extension.rs
@@ -18,7 +18,9 @@
 //! ```
 use std::fmt::Write;
 
-use crate::asn1::Asn1Object;
+use crate::asn1::{Asn1Integer, Asn1Object};
+use crate::bn::BigNum;
+use crate::cvt_p;
 use crate::error::ErrorStack;
 use crate::nid::Nid;
 use crate::x509::{GeneralName, Stack, X509Extension, X509Name, X509v3Context};
@@ -551,6 +553,38 @@ impl SubjectAlternativeName {
 
         unsafe {
             X509Extension::new_internal(Nid::SUBJECT_ALT_NAME, self.critical, stack.as_ptr().cast())
+        }
+    }
+}
+
+/// An extension that provides a means of versionning the CRL.
+pub struct CrlNumber(Asn1Integer);
+
+impl CrlNumber {
+    /// Construct a new `CrlNumber` extension.
+    pub fn new(number: BigNum) -> Result<Self, ErrorStack> {
+        let mut max = BigNum::new()?;
+        max.lshift(BigNum::from_u32(1)?.as_ref(), 159)?;
+
+        assert!(
+            !number.is_negative() && number < max,
+            "CrlNumber must be an ASN.1 integer greater than or equal to 0 and less than 2^159"
+        );
+
+        Ok(Self(Asn1Integer::from_bn(&number)?))
+    }
+
+    /// Return a `CrlNumber` extension as an `X509Extension`.
+    pub fn build(self) -> Result<X509Extension, ErrorStack> {
+        unsafe {
+            ffi::init();
+
+            cvt_p(ffi::X509V3_EXT_i2d(
+                Nid::CRL_NUMBER.as_raw(),
+                0,
+                self.0.as_ptr().cast(),
+            ))
+            .map(X509Extension)
         }
     }
 }
--- a/openssl/src/x509/mod.rs
+++ b/openssl/src/x509/mod.rs
@@ -22,7 +22,7 @@ use std::ptr;
 use std::str;
 
 use crate::asn1::{
-    Asn1BitStringRef, Asn1Enumerated, Asn1IntegerRef, Asn1Object, Asn1ObjectRef,
+    Asn1BitStringRef, Asn1Enumerated, Asn1Integer, Asn1IntegerRef, Asn1Object, Asn1ObjectRef,
     Asn1OctetStringRef, Asn1StringRef, Asn1TimeRef, Asn1Type,
 };
 use crate::bio::MemBioSlice;
@@ -38,6 +38,8 @@ use crate::string::OpensslString;
 use crate::util::{self, ForeignTypeExt, ForeignTypeRefExt};
 use crate::{cvt, cvt_n, cvt_p, cvt_p_const};
 use openssl_macros::corresponds;
+
+pub use crate::x509::extension::CrlNumber;
 
 pub mod verify;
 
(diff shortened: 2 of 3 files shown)


ocdlroux:
Hi again,

We're reworking the `CrlBuilder` from what you stated, and:
- We were wondering, if the two required extensions aren't present, should we `assert!` with an explanation message (and with adequate documentation) to let the developer know that this shouldn't be used without those extensions _or_ do we just return an `Err` ?
- Since the builder is going to contain only rust members until the `build` should it take the `pkey` & `digest` to construct the `Crl` and sign it in one go ?

Thanks.

openssl/src/x509/extension.rs
```diff
@@ -555,6 +557,38 @@ impl SubjectAlternativeName {
...
+
+impl CrlNumber {
+    /// Construct a new `CrlNumber` extension.
+    pub fn new(number: BigNum) -> Result<Self, ErrorStack> {
+        let mut max = BigNum::new()?;
+        max.lshift(BigNum::from_u32(1)?.as_ref(), 159)?;
+
+        assert!(
+            number >= BigNum::from_u32(0)? && number < max,
+            "CrlNumber must have n ASN.1 integer greater than or equal to 0 and less than 2^159"
```
botovq:
```suggestion
            !number.is_negative() && number < max,
            "CrlNumber must be an ASN.1 integer greater than or equal to 0 and less than 2^159"
```

openssl/src/x509/mod.rs
```diff
@@ -1721,6 +1766,152 @@ unsafe impl ExtensionType for AuthorityInformationAccess {
...
+        unsafe {
+            cvt(ffi::X509_CRL_set_issuer_name(
+                self.0.as_ptr(),
+                issuer_name.as_ptr(),
+            ))
+            .map(|_| ())
+        }
+    }
+
+    /// Set the lastUpdate time indicating when the CRL was issued.
```
botovq: lastUpdate is an OpenSSL-ism. The standard name is thisUpdate

```suggestion
    /// Set the lastUpdate (thisUpdate) time indicating when the CRL was issued.
```

openssl/src/x509/mod.rs
```diff
@@ -1721,6 +1766,152 @@ unsafe impl ExtensionType for AuthorityInformationAccess {
...
+        unsafe { cvt(ffi::X509_CRL_set1_lastUpdate(self.0.as_ptr(), t.as_ptr())).map(|_| ()) }
+    }
+
+    /// Set the nextUpdate timestamp indicating when a newer CRL is expected.
+    #[corresponds(X509_CRL_set1_nextUpdate)]
+    pub fn set_next_update(&mut self, t: &Asn1TimeRef) -> Result<(), ErrorStack> {
+        unsafe { cvt(ffi::X509_CRL_set1_nextUpdate(self.0.as_ptr(), t.as_ptr())).map(|_| ()) }
+    }
+
+    /// Adds an X509 extension value to the CRL.
```
botovq:
```suggestion
    /// Add an X509 extension value to the CRL.
```

openssl/src/x509/mod.rs
```diff
@@ -1721,6 +1766,152 @@ unsafe impl ExtensionType for AuthorityInformationAccess {
...
+    ///
+    /// This works just as `append_extension` except it takes ownership of the `X509Extension`.
+    pub fn append_extension(&mut self, extension: X509Extension) -> Result<(), ErrorStack> {
+        self.append_extension2(&extension)
+    }
+
+    /// Adds an X509 extension value to the CRL.
```
botovq:
```suggestion
    /// Add an X509 extension value to the CRL.
```

openssl/src/x509/mod.rs
```diff
@@ -1597,6 +1599,49 @@ impl CrlReason {
...
+            cvt(ffi::X509_REVOKED_set_serialNumber(
+                self.0.as_ptr(),
+                serial.as_ptr(),
+            ))
+            .map(|_| ())
+        }
+    }
+
+    /// Consumes the builder, returning the `X509Revoked`.
+    pub fn build(self) -> X509Revoked {
```
botovq, line 1655: I wonder if this should not refuse building an empty list of revoked certificates per RFC 5280, 5.1.2.6, "When there are no revoked certificates, the revoked certificates list MUST be absent."

botovq (commented):
I think we're getting quite close. I few small nits and one suggestion regarding empty list of revoked.

openssl/src/x509/mod.rs
```diff
@@ -1721,6 +1766,157 @@ unsafe impl ExtensionType for AuthorityInformationAccess {
...
+            );
+
+            assert!(
+                !X509_CRL_get0_nextUpdate(self.0.as_ptr()).is_null(),
+                "CRL must have nextUpdate time set"
+            );
+            let revoked = self.0.get_revoked();
+            assert!(
+                revoked.is_some_and(|r| !r.is_empty()),
+                "Revoked should not be empty or null"
```
botovq:
```suggestion
                // XXX - switch to is_none_or() once MSRV is 1.82.
                revoked.is_none() || revoked.is_some_and(|r| !r.is_empty()),
                "Revoked must be absent or non-empty"
```

If we had MSRV 1.82, we could write this as `revoked.is_none_or(|r| !r.is_empty())`, but since we still have 1.70, this needs to be spelled out explicitly.

The reason is that the `revokedCertificates` element is `OPTIONAL`:
```
 revokedCertificates     SEQUENCE OF SEQUENCE  {
             userCertificate         CertificateSerialNumber,
             revocationDate          Time,
             crlEntryExtensions      Extensions OPTIONAL
                                      -- if present, version MUST be v2
                                  }  OPTIONAL,
```
per [RFC 5280, section 5.1.2.6](https://www.rfc-editor.org/rfc/rfc5280#section-5.1.2.6), which states: "When there are no revoked certificates, the revoked certificates list MUST be absent.", so the None case is valid.

botovq (commented):
One last thing for you to fix. Then I believe we're done as far as code is concerned.

Before we can land this we will have to fix the min-version test either by pinning some crate (random?) or bumping the MSRV. I'll ask you to rebase once more once that's done.

botovq (approved):
Thanks for reviving this - I could have sworn I had merged this months ago... I reread the code and I am happy to merge once the CI is green.

If you wish to replace the asserts with the new `ErrorStack::internal_error()` you could do that as a follow-up.

PR #2539, opened 2025-12-11
check extent of rust 1.92 breakage

botovq:


sfackler:
Time to drop 1.0.2? :D

PR #2543, opened 2025-12-19
Add HKDF support

Zenkibou:
OpenSSL 3.0 added generic [KDF](https://docs.openssl.org/3.0/man3/EVP_KDF/) support as well as the [HKDF](https://docs.openssl.org/3.0/man7/EVP_KDF-HKDF/) algorithm ([RFC 5869](https://datatracker.ietf.org/doc/html/rfc5869)).

rust-openssl currently only supports the Argon2 algorithm added in OpenSSL 3.2.

This PR adds HKDF support that can be used with OpenSSL 3.0 and later.
It includes tests for the HKDF support against the test vectors from the RFC.

### DIFF
--- a/openssl/src/kdf.rs
+++ b/openssl/src/kdf.rs
@@ -22,6 +22,14 @@ impl Drop for EvpKdfCtx {
     }
 }
 
+#[cfg(ossl300)]
+#[derive(Copy, Clone)]
+pub enum HkdfMode {
+    ExtractAndExpand,
+    ExtractOnly,
+    ExpandOnly,
+}
+
 cfg_if::cfg_if! {
     if #[cfg(ossl300)] {
         use std::ffi::CStr;
@@ -32,6 +40,13 @@ cfg_if::cfg_if! {
         use crate::lib_ctx::LibCtxRef;
         use crate::error::ErrorStack;
         use crate::ossl_param::{OsslParamBuilder, OsslParamArray};
+        use crate::md::MdRef;
+
+        const OSSL_KDF_PARAM_DIGEST: &CStr = unsafe { CStr::from_bytes_with_nul_unchecked(b"digest\0") };
+        const OSSL_KDF_PARAM_KEY: &CStr = unsafe { CStr::from_bytes_with_nul_unchecked(b"key\0") };
+        const OSSL_KDF_PARAM_SALT: &CStr = unsafe { CStr::from_bytes_with_nul_unchecked(b"salt\0") };
+        const OSSL_KDF_PARAM_INFO: &CStr = unsafe { CStr::from_bytes_with_nul_unchecked(b"info\0") };
+        const OSSL_KDF_PARAM_MODE: &CStr = unsafe { CStr::from_bytes_with_nul_unchecked(b"mode\0") };
 
         // Safety: these all have null terminators.
         // We cen remove these CStr::from_bytes_with_nul_unchecked calls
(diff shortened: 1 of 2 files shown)


openssl/src/kdf.rs
```diff
@@ -22,24 +22,101 @@ impl Drop for EvpKdfCtx {
...
+        const OSSL_KDF_PARAM_INFO: &CStr = unsafe { CStr::from_bytes_with_nul_unchecked(b"info\0") };
+        const OSSL_KDF_PARAM_MODE: &CStr = unsafe { CStr::from_bytes_with_nul_unchecked(b"mode\0") };
 
         // Safety: these all have null terminators.
         // We cen remove these CStr::from_bytes_with_nul_unchecked calls
         // when we upgrade to Rust 1.77+ with literal c"" syntax.
 
+        /// Derives a key using a KDF.
+        #[allow(clippy::too_many_arguments)]
+        fn kdf_helper(
```
alex: helper isn't a good name. it should be kdf_derive or something

openssl/src/kdf.rs
```diff
@@ -22,24 +22,101 @@ impl Drop for EvpKdfCtx {
...
+            }
+        }
+
+        pub const OSSL_HKDF_DIGEST_SN_SHA1: &[u8] = b"SHA1";
+        pub const OSSL_HKDF_DIGEST_SN_SHA256: &[u8] = b"SHA256";
+        pub const OSSL_HKDF_DIGEST_SN_SHA384: &[u8] = b"SHA384";
+        pub const OSSL_HKDF_DIGEST_SN_SHA512: &[u8] = b"SHA512";
+
+        pub fn hkdf(
+            digest: &[u8],
```
alex: I'm not wild about passing the digest as a string. (I'm not wild about OpenSSL's APIs here either...)

Is taking one of our digest types practical?

openssl/src/kdf.rs
```diff
@@ -137,31 +214,257 @@ cfg_if::cfg_if! {
             if let Some(ad) = ad {
                 bld.add_octet_string(OSSL_KDF_PARAM_ARGON2_AD, ad)?;
             }
+            let mut bld = OsslParamBuilder::new()?;
```
alex: Did this get duplicated?

Zenkibou: Indeed, my mistake, I will fix this.

openssl/src/kdf.rs
Zenkibou: I was indeed wondering if an enum would be better. Openssl is using SN_* defines from obj_mac.h.
I'm not totally familiar with the codebase, we could use MessageDigest objects, get the Nid type from there and convert to the SN string with a match ?
Or expose the SN defines as an enum directly ?

openssl/src/kdf.rs
```diff
@@ -22,24 +22,101 @@ impl Drop for EvpKdfCtx {
...
+        const OSSL_KDF_PARAM_SALT: &CStr = unsafe { CStr::from_bytes_with_nul_unchecked(b"salt\0") };
+        const OSSL_KDF_PARAM_INFO: &CStr = unsafe { CStr::from_bytes_with_nul_unchecked(b"info\0") };
+        const OSSL_KDF_PARAM_MODE: &CStr = unsafe { CStr::from_bytes_with_nul_unchecked(b"mode\0") };
 
         // Safety: these all have null terminators.
         // We cen remove these CStr::from_bytes_with_nul_unchecked calls
         // when we upgrade to Rust 1.77+ with literal c"" syntax.
 
+        /// Derives a key using a KDF.
+        #[allow(clippy::too_many_arguments)]
```
alex: Is this allow doing anything?

openssl/src/kdf.rs
alex: We should be able to take an MdRef -> Nid -> short_name() on the nid

openssl/src/ossl_param.rs
```diff
@@ -106,6 +106,25 @@ impl<'a> OsslParamBuilder<'a> {
         }
     }
 
+    /// Adds an utf8 string to `OsslParamBuilder`.
+    #[corresponds(OSSL_PARAM_BLD_push_utf8_string)]
+    #[cfg_attr(any(ossl300), allow(dead_code))]
```
alex: I'm not sure I understand what this cfg_attr is for.

openssl/src/kdf.rs
Zenkibou: No indeed

openssl/src/ossl_param.rs
Zenkibou: I think I mis-understood what it was doing for Argon2, I will remove it.

PR #2545, opened 2025-12-20
drop openssl 1.0.2

alex:
this raises the minimum to 1.1.0


openssl-sys/src/ssl.rs
```diff
@@ -189,40 +189,28 @@ cfg_if! {
...
+    } else {
+        pub const SSL_OP_MICROSOFT_SESS_ID_BUG: ssl_op_type!() = 0x00000000;
+        pub const SSL_OP_NETSCAPE_CHALLENGE_BUG: ssl_op_type!() = 0x00000000;
+        pub const SSL_OP_NETSCAPE_REUSE_CIPHER_CHANGE_BUG: ssl_op_type!() = 0x00000000;
+        pub const SSL_OP_MICROSOFT_BIG_SSLV3_BUFFER: ssl_op_type!() = 0x00000000;
+        pub const SSL_OP_SSLEAY_080_CLIENT_DH_BUG: ssl_op_type!() = 0x00000000;
+        pub const SSL_OP_TLS_D5_BUG: ssl_op_type!() = 0x00000000;
+        pub const SSL_OP_TLS_BLOCK_PADDING_BUG: ssl_op_type!() = 0x00000000;
+        pub const SSL_OP_SINGLE_ECDH_USE: ssl_op_type!() = 0x00000000;
+        pub const SSL_OP_SINGLE_DH_USE: ssl_op_type!() = 0x00000000;
```
botovq, line 204: The only real difference between the two branches is `SSL_OP_SINGLE_DH_USE` (which has been a noop in libressl, since forever). All those extra 0 are strange.

alex: nice catch

PR #2548, opened 2025-12-21
Add Debug implementation for EcGroup{,Ref}

buytenh:
This allows deriving Debug on user-defined structs that contain
instances of EcGroup.

For standard groups, the debug output will look something like this:

```
    EcGroup { curve_name: "secp521r1" }
```

For groups not based on a standard curve, the Debug impl will try to
extract and print the curve's components, such that:

```
    let mut p = BigNum::new().unwrap();
    let mut a = BigNum::new().unwrap();
    let mut b = BigNum::new().unwrap();
    let mut ctx = BigNumContext::new().unwrap();

    EcGroup::from_curve_name(Nid::SECP224R1)
        .unwrap()
        .components_gfp(&mut p, &mut a, &mut b, &mut ctx)
        .unwrap();

    // reconstruct the group from its components
    let group = EcGroup::from_components(p, a, b, &mut ctx).unwrap();

    println!("{:#?}", group);
```

will print:

```
    EcGroup {
        p: "FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF000000000000000000000001",
        a: "FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFFFFFFFFFFFFFFFFFFFE",
        b: "B4050A850C04B3ABF54132565044B0B7D7BFD8BA270B39432355FFB4",
    }
```

If EcGroup::components_gfp() fails, or if any other error occurs, we'll
instead print only:

```
    EcGroup
```

Calling EcGroup::components_gfp() on a curve over GF(2^m) unexpectedly
works, and there does not seem to be an explicit way to distinguish
custom curves over GF(2^m) from custom curves over GF(p), and so, custom
curves over GF(2^m) will be Debug'd by the same logic.  (rust-openssl
does not allow constructing custom curves over GF(2^m), so we can't add
test cases for this.)

### DIFF
--- a/openssl/src/ec.rs
+++ b/openssl/src/ec.rs
@@ -21,7 +21,7 @@ use libc::c_int;
 use std::fmt;
 use std::ptr;
 
-use crate::bn::{BigNum, BigNumContextRef, BigNumRef};
+use crate::bn::{BigNum, BigNumContext, BigNumContextRef, BigNumRef};
 use crate::error::ErrorStack;
 use crate::nid::Nid;
 use crate::pkey::{HasParams, HasPrivate, HasPublic, Params, Private, Public};
@@ -162,6 +162,12 @@ impl EcGroup {
             ))
             .map(EcGroup)
         }
+    }
+}
+
+impl fmt::Debug for EcGroup {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        self.as_ref().fmt(f)
     }
 }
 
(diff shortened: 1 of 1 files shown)


openssl/src/ec.rs
```diff
@@ -320,6 +326,45 @@ impl EcGroupRef {
...
+                f.debug_struct("EcGroup")
+                    .field("curve", &curve_name)
+                    .finish()
+            }
+        } else {
+            // let chains are only allowed in Rust 2024 or later
+            if let Ok(mut p) = BigNum::new() {
+                if let Ok(mut a) = BigNum::new() {
+                    if let Ok(mut b) = BigNum::new() {
+                        if let Ok(mut ctx) = BigNumContext::new() {
```
botovq, line 346: I think you can pass in `ptr::null_mut()` and the library will allocate a `ctx` for you internally (that may not work in 1.1.0, but let's try).

openssl/src/ec.rs
```diff
@@ -320,6 +326,45 @@ impl EcGroupRef {
...
+                            if self
+                                .components_gfp(&mut p, &mut a, &mut b, &mut ctx)
+                                .is_ok()
+                            {
+                                return f
+                                    .debug_struct("EcGroup")
+                                    .field("p", &p)
+                                    .field("a", &a)
+                                    .field("b", &b)
```
botovq: Would it be very annoying to print the components in hex? I've never seen them printed in decimal.

openssl/src/ec.rs
botovq: Never mind. That isn't exposed in the rust interface.

openssl/src/ec.rs
buytenh: That's a good idea.  If we could consider https://github.com/rust-openssl/rust-openssl/pull/2550 first, then this would be slightly easier to do.

buytenh: The components are now printed in hex, relying on impl UpperHex for BigNum (#2550) to do so.  I'll rebase this PR onto master if/when #2550 is merged, if that works for you.

PR #2550, opened 2025-12-21
Add UpperHex implementation for BigNum{,Ref}

buytenh:
This is mainly to help satisfy https://github.com/rust-openssl/rust-openssl/pull/2548#discussion_r2637999963 .

### DIFF
--- a/openssl/src/bn.rs
+++ b/openssl/src/bn.rs
@@ -1186,6 +1186,32 @@ impl fmt::Display for BigNum {
     }
 }
 
+impl fmt::UpperHex for BigNumRef {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self.to_hex_str() {
+            Ok(s) => {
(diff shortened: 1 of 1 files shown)


openssl/src/bn.rs
```diff
@@ -1186,6 +1186,36 @@ impl fmt::Display for BigNum {
     }
 }
 
+impl fmt::UpperHex for BigNumRef {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self.to_hex_str() {
+            Ok(s) => {
+                // BoringSSL's BN_bn2hex() returns lower-case hexadecimal.
+                #[cfg(boringssl)]
+                let s = &s.to_uppercase();
```
botovq: I don't feel strongly here, but I'd just do this unconditionally. Perhaps LibreSSL 4.3 or OpenSSL 4 will decide to change the output to lowercase, and this implementation will keep working then. It's just a debug, after all.

buytenh: Works for me!  `.to_uppercase()` is now done unconditionally.

PR #2560, opened 2026-01-02
Still more OpenSSL 1.0.2 complications

botovq:
This is kind of endless because all the grep I can come up with are very noisy. Let's see how this one goes. There's still more to clean up.

### DIFF
--- a/openssl/src/ssl/callbacks.rs
+++ b/openssl/src/ssl/callbacks.rs
@@ -1,4 +1,3 @@
-use cfg_if::cfg_if;
 use foreign_types::ForeignType;
 use foreign_types::ForeignTypeRef;
 #[cfg(any(ossl111, not(osslconf = "OPENSSL_NO_PSK")))]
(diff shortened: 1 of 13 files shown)


openssl/src/ssl/callbacks.rs
```diff
@@ -330,13 +329,7 @@ pub unsafe extern "C" fn raw_remove_session<F>(
...
 }
 
-cfg_if! {
-    if #[cfg(any(ossl110, libressl, boringssl, awslc))] {
-        type DataPtr = *const c_uchar;
-    } else {
-        type DataPtr = *mut c_uchar;
-    }
-}
+type DataPtr = *const c_uchar;
```
alex: nit: I wonder if we should just inline this now that it's always the same, not sure a type alias makes sense

openssl/src/ssl/callbacks.rs
```diff
@@ -466,13 +459,7 @@ where
...
 
 #[cfg(not(any(boringssl, awslc)))]
-cfg_if! {
-    if #[cfg(any(ossl110, libressl))] {
-        type CookiePtr = *const c_uchar;
-    } else {
-        type CookiePtr = *mut c_uchar;
-    }
-}
+type CookiePtr = *const c_uchar;
```
alex: Same

alex (approved):
two nits, but on a facial read this looks right to me

alex (approved):

PR #2562, opened 2026-01-02
Let AWS-LC use the BoringSSL path for BIO_METHOD

botovq:
I'm not entirely sure about this one, since I don't know how this managed to work for AWS-LC so far. In any case, I think we want to get rid of the "else" branch of this cfg_if.


botovq:
Thanks. I think you're right. How's this?

Not sure if this is semver compatible and how to make it so, though.

PR #2567, opened 2026-01-27
add more brainpool curve NID constants

butteronarchbtw:
This PR adds more NID constants for additional brainpool curves defined in [RFC5639](https://datatracker.ietf.org/doc/html/rfc5639), which are available in openssl, but not rust-openssl.

These curves included are:
- brainpoolP256t1
- brainpoolP320t1
- brainpoolP384t1
- brainpoolP512t1

### DIFF
--- a/openssl-sys/src/obj_mac.rs
+++ b/openssl-sys/src/obj_mac.rs
@@ -117,15 +117,23 @@ pub const NID_brainpoolP512r1: c_int = 934;
 
 #[cfg(ossl110)]
 pub const NID_brainpoolP256t1: c_int = 928;
+#[cfg(libressl)]
+pub const NID_brainpoolP256t1: c_int = 929;
 
 #[cfg(ossl110)]
 pub const NID_brainpoolP320t1: c_int = 930;
+#[cfg(libressl)]
+pub const NID_brainpoolP320t1: c_int = 931;
 
 #[cfg(ossl110)]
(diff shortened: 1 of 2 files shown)


openssl-sys/src/obj_mac.rs
```diff
@@ -115,6 +115,18 @@ pub const NID_brainpoolP512r1: c_int = 933;
 #[cfg(libressl)]
 pub const NID_brainpoolP512r1: c_int = 934;
 
+#[cfg(ossl110)]
+pub const NID_brainpoolP256t1: c_int = 928;
```
botovq, line 119: Please add the corresponding nids 929,...,935 for libressl:

```suggestion
#[cfg(ossl110)]
pub const NID_brainpoolP256t1: c_int = 928;
#[cfg(libressl)]
pub const NID_brainpoolP256t1: c_int = 929;
```

openssl/src/nid.rs
```diff
@@ -222,6 +222,14 @@ impl Nid {
     pub const BRAINPOOL_P384R1: Nid = Nid(ffi::NID_brainpoolP384r1);
     #[cfg(any(ossl110, libressl))]
     pub const BRAINPOOL_P512R1: Nid = Nid(ffi::NID_brainpoolP512r1);
+    #[cfg(ossl110)]
```
botovq: And for all these do:

```suggestion
    #[cfg(any(ossl110, libressl))]
```

PR #2568, opened 2026-02-02
Update ctest to 0.5.0

botovq:


botovq:
Alright. Looks like this is more work than I have time for in the coming days...

PR #2572, opened 2026-02-15
Fix use-after-free of error strings on BoringSSL/aws-lc

### LINKED ISSUE #2470: Incorrect lifetime of error strings when awslc or boringssl features are enabled
rust-openssl assumes that borrowed error data is `'static` and not `free()`'d by the underlying library:

https://github.com/sfackler/rust-openssl/blob/01166fa74c5d398feb5fbf1914a9188869ddabc2/openssl/src/error.rs#L169

However, this isn't the case in boringssl and aws-lc. They have a workaround that explicitly calls out rust-openssl as having a wrong assumption here:

https://github.com/google/boringssl/blob/480148ccc9444f5323b84389a6660e33ae38d4cb/crypto/err/err.cc#L184-L187

https://github.com/aws/aws-lc/blob/f4f0be9fc1a894121c56a2a3b06b043f15465249/crypto/err/err.c#L293-L296

Unfortunately, the workaround has been broken by additions in c3b714e879eac0b7fd6a1ad216671bba50f51156

and now `malloc`ed strings from boringssl and aws-lc are treated again as `'static`, which is incorrect:

https://github.com/sfackler/rust-openssl/blob/01166fa74c5d398feb5fbf1914a9188869ddabc2/openssl/src/error.rs#L136-L138

alex:
The boringssl/awslc cfg gate unconditionally used Cow::Borrowed for error data, treating it as 'static. However, BoringSSL and aws-lc dynamically allocate error strings and may free them on the next error queue operation. Both libraries set ERR_TXT_MALLOCED as a workaround to signal that the string should be copied, but the cfg gate bypassed that check entirely.

Remove the cfg gates and use the same ERR_TXT_MALLOCED check for all backends.

Fixes #2470

### DIFF
--- a/openssl/src/error.rs
+++ b/openssl/src/error.rs
@@ -127,14 +127,11 @@ impl Error {
                     let data = if flags & ffi::ERR_TXT_STRING != 0 {
                         let bytes = CStr::from_ptr(data as *const _).to_bytes();
                         let data = str::from_utf8(bytes).unwrap();
-                        #[cfg(not(any(boringssl, awslc)))]
                         let data = if flags & ffi::ERR_TXT_MALLOCED != 0 {
                             Cow::Owned(data.to_string())
                         } else {
                             Cow::Borrowed(data)
                         };
-                        #[cfg(any(boringssl, awslc))]
-                        let data = Cow::Borrowed(data);
                         Some(data)
                     } else {
                         None


botovq (approved):
Be all that as it may, this seems correct to me.

PR #2573, opened 2026-02-15
fix min-version CI

alex:


alex:
We may want to think about raising MSRV, but I'm too lazy tonight

PR #2583, opened 2026-03-15
Support openssl 4.0.0

theoparis:
OpenSSL 4.0.0 beta 1 is out, and 4.0.0 is planned to be released by April 14th

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -173,6 +173,8 @@ jobs:
             - name: openssl
               version: vendored
             - name: openssl
+              version: 4.0.0-beta1
+            - name: openssl
               version: 3.6.0
             - name: openssl
               version: 3.5.0
--- a/openssl-sys/build/main.rs
+++ b/openssl-sys/build/main.rs
@@ -19,6 +19,7 @@ mod run_bindgen;
 
 #[derive(PartialEq)]
 enum Version {
+    Openssl4xx,
     Openssl3xx,
     Openssl11x,
     Libressl,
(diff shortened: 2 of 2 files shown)


theoparis:
cargo run -p systest fails on aarch64 darwin with a bunch of macro expansion errors, but cargo build -p openssl works fine 🤔  
```
in this scope\n   --> /Volumes/Dev/rust-openssl/openssl-sys/src/./x509v3.rs:105:43\n    |\n105 | pub const CRL_REASON_PRIVILEGE_WITHDRAWN: c_int = 9;\n    |                                           ^^^^^ not found in this scope\n    |\nhelp: consider importing one of these type aliases\n    |\n  1 + use std::ffi::c_int;\n    |\n  1 + use std::os::raw::c_int;\n    |\n  1 + use core::ffi::c_int;\n    |\n\nerror[E0425]: cannot find type `c_int` in this scope\n   --> /Volumes/Dev/rust-openssl/openssl-sys/src/./x509v3.rs:106:37\n    |\n106 | pub const CRL_REASON_AA_COMPROMISE: c_int = 10;\n    |                                     ^^^^^ not found in this scope\n    |\nhelp: consider importing one of these type aliases\n    |\n  1 + use std::ffi::c_int;\n    |\n  1 + use std::os::raw::c_int;\n    |\n  1 + use core::ffi::c_int;\n    |\n\nerror[E0425]: cannot find type `c_char` in this scope\n   --> /Volumes/Dev/rust-openssl/openssl-sys/src/lib.rs:167:19\n    |\n167 |         buf: *mut c_char,\n    |                   ^^^^^^\n    |\nhelp: a builtin type with a similar name exists\n    |\n167 -         buf: *mut c_char,\n167 +   
```

alex:
Thanks for taking a look at this -- it looks like the builds are failing on several platforms.

Also, I think we want to limit ourselves to only allowing 4.0.0-beta1 (or other tested pre-releases) until they reach the point where they consider the ABI stable.

alex:
We landed a different PR for this.

PR #2589, opened 2026-04-12
md_ctx: enable sign/verify/reset on BoringSSL, LibreSSL, and AWS-LC

alex:
EVP_DigestSign, EVP_DigestVerify, and EVP_MD_CTX_reset are all available on BoringSSL, AWS-LC, and LibreSSL (since 3.4.0 for the former two, and 2.7.0 for reset, well under the 3.5.0 minimum).

Also add libressl to the openssl-sys binding for EVP_MD_CTX_reset, which was previously only exposed for ossl111.

EVP_DigestFinalXOF is intentionally left out for BoringSSL: it is declared only as a stub that always fails, since BoringSSL does not support any XOF digests.

https://claude.ai/code/session_01KXL6oJAqpTNfMUmgS36yFZ

### DIFF
--- a/openssl-sys/src/handwritten/evp.rs
+++ b/openssl-sys/src/handwritten/evp.rs
@@ -246,7 +246,7 @@ cfg_if! {
     }
 }
 cfg_if! {
-    if #[cfg(any(ossl111, libressl370))] {
+    if #[cfg(any(ossl111, libressl))] {
         extern "C" {
             pub fn EVP_DigestSign(
                 ctx: *mut EVP_MD_CTX,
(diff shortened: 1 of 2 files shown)


openssl-sys/src/handwritten/evp.rs
```diff
@@ -246,7 +246,7 @@ cfg_if! {
     }
 }
 cfg_if! {
-    if #[cfg(any(ossl111, libressl370))] {
+    if #[cfg(any(ossl111, libressl))] {
```
botovq, line 249: The situation is confusing: we added the `DigestSign/Verify` API because people were keen on merging `Update` and `Final` if possible while the real reason for the existence of this API is EdDSA for which we did the plumbing only in in 3.7.0.

In particular, I'm not sure how many of the `libressl370` cfg in `sign.rs` can and should be changed accordingly. The `_oneshot_` API probably can. The `eddsa()` test won't work and I'm pretty sure `len_internal()` should also stay as it is since we fixed that for EdDSA.

botovq (approved):

openssl-sys/src/handwritten/evp.rs
alex: tbh, I wanted this on boringssl for mldsa, and just figured I'd do everything else for completeness

PR #2591, opened 2026-04-12
openssl 4 support

reaperhulk:


openssl-sys/build/main.rs
```diff
@@ -436,7 +437,8 @@ See rust-openssl documentation for more information:
         let openssl_version = openssl_version.unwrap();
         println!("cargo:version_number={openssl_version:x}");
 
-        if openssl_version >= 0x4_00_00_00_0 {
+        // Only allow 4.0.0 pre-releases for now
+        if openssl_version >= 0x4_00_00_00_f {
```
botovq: This doesn't look right. My OpenSSL 4.0.0-beta1 have `OPENSSL_VERSION_NUMBER` defined to `0x4_00_00_00_0`.

botovq (commented):
There's also a version number to adjust here:

https://github.com/rust-openssl/rust-openssl/blob/78efb65c7866c46a18298384efe06eb05f07c9b8/openssl/src/lib.rs#L3-L4

openssl-sys/build/main.rs
reaperhulk: I originally did this so that it would only be compatible with the pre-release 4.0.0 series since @alex mentioned that in the other PR, but as of beta1 they've effectively committed to their API so it should be safe to simply change it (which I now have).

openssl/src/ssl/test/mod.rs
```diff
@@ -675,6 +675,8 @@ fn default_verify_paths() {
     let mut ctx = SslContext::builder(SslMethod::tls()).unwrap();
     ctx.set_default_verify_paths().unwrap();
     ctx.set_verify(SslVerifyMode::PEER);
+    #[cfg(ossl400)]
+    ctx.set_options(super::SslOptions::IGNORE_UNEXPECTED_EOF);
```
reaperhulk, line 679: This test fails in OpenSSL 4.0 because it doesn't receive a close_notify. Adding this flag resolves it, but I'm unclear why this doesn't fail in 3.0 so calling it out for discussion.

reaperhulk: I have gone down a long rabbit hole here...

Prior to OpenSSL 4.0, when a server closed a TLS connection without sending `close_notify` (basically the entire web at this point) custom BIOs using the `BIO_meth_set_read` callback would return 0 to signal EOF. The record layer would then call `BIO_eof()` — which was a macro expanding to `BIO_ctrl(b, BIO_CTRL_EOF, 0, NULL)` — to determine if this was a true end-of-file. Since custom BIOs typically don't handle BIO_CTRL_EOF (their ctrl callback returns 0 for unknown commands), BIO_eof() returned false. The record layer then fell into a `OSSL_RECORD_RETURN_FATAL` path, but because no alert code was set, no error was pushed to OpenSSL's error stack. `SSL_get_error` saw the empty stack and returned `SSL_ERROR_SYSCALL` with errno=0, which rust-openssl's `SslStream::read` treats as a clean EOF (Ok(0)).

https://github.com/openssl/openssl/pull/29290 changed `BIO_eof()` from a macro to a function and added a `BIO_FLAGS_AUTO_EOF` flag. Now when the legacy `bread_conv` shim sees a 0 return from the old-style callback and the BIO doesn't handle `BIO_CTRL_EOF`, it sets `BIO_FLAGS_AUTO_EOF`, and the new `BIO_eof()` function checks this flag first — correctly returning true. The record layer now takes the `OSSL_RECORD_RETURN_EOF` path, which raises `SSL_R_UNEXPECTED_EOF_WHILE_READING` (a fatal error unless `SSL_OP_IGNORE_UNEXPECTED_EOF` is set). `SSL_get_error` sees the error on the stack and returns `SSL_ERROR_SSL`, which rust-openssl propagates as an `io::Error`.

This PR resolves this issue by setting `SSL_OP_IGNORE_UNEXPECTED_EOF` in the test, but arguably we could update the default flags to set it at all times when available. Python does this (https://github.com/python/cpython/commit/6f37ebc61e9e0d13bcb1a2ddb7fc9723c04b6372).

openssl-sys/build/main.rs
botovq: There was one ABI break between beta1 and the final release: https://github.com/openssl/openssl/pull/30572. Fortunately it doesn't matter for rust-openssl since there are no bindings for `X509V3_EXT_print_fp()`.

The version number still doesn't end in f, so the trick would not have worked.
I think they abandoned that in the lead-up to 3 where this define was temporarily deprecated. Not sure how to realiably recognize a prerelease programmatically. Perhaps check for a non-empty `OPENSSL_VERSION_PRE_RELEASE`?

openssl/src/ssl/test/mod.rs
botovq: Wow. I think working around this in the test ultimately puts the burden on the downstream consumers, which will likely be bitten by this scary change. While it is very unfortunate that such a step is necessary, I believe the better approach is to follow Python and set the flag by default when available, as you suggest in the last paragraph.

.github/workflows/ci.yml
```diff
@@ -182,6 +182,8 @@ jobs:
               version: 338f44af3c92ef665bf740a8127a2d69c872b52a
             - name: openssl
               version: vendored
+            - name: openssl
+              version: 4.0.0-beta1
```
botovq: 4.0.0 is now available, so let's test with that.

botovq (commented):
This looks reasonable overall. As mentioned as a reaction to your sleuthing, I think setting the flag by default is safer, but I do not object to your approach of tweaking the test.

openssl/src/ssl/test/mod.rs
reaperhulk: I think setting it is definitely the least invasive to the consumer and Python's use of it for 5+ years suggests it's fine. @alex @sfackler do you two have opinions?

.github/workflows/ci.yml
reaperhulk:
```suggestion
              version: 4.0.0
```

openssl/src/ssl/test/mod.rs
alex: I think that's _probably_ correct, but I'm going to land this as-is and we can make that decision in a follow up.

alex (approved):

PR #2595, opened 2026-04-15
Set timeout for package installation step

alex:
Sometimes the network is cursed and it takes forever. In these cases it's better to fail quickly and retry it.


botovq (approved):
Not at all a GHA expert so I can't tell if there's a better way, but this makes sense to me.

PR #2596, opened 2026-04-15
Panic in Crypter::new when IV is required but not provided

alex:
Previously, passing `iv: None` to a cipher that requires an IV (CBC, GCM, CTR, etc.) silently passed a null pointer to OpenSSL, which used an all-zero IV. This matched neither the documented contract ("Panics if an IV is required by the cipher but not provided") nor safe cryptographic practice.

Add an assertion that enforces the documented behavior, and add tests for both CBC and GCM to verify the panic.

### DIFF
--- a/openssl/src/symm.rs
+++ b/openssl/src/symm.rs
@@ -620,6 +620,10 @@ impl Crypter {
         key: &[u8],
         iv: Option<&[u8]>,
     ) -> Result<Crypter, ErrorStack> {
+        assert!(
+            iv.is_some() || t.iv_len().is_none(),
+            "an IV is required for this cipher"
+        );
         let mut ctx = CipherCtx::new()?;
 
         let f = match mode {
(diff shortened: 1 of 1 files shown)


botovq (approved):
Looks good.

There's one thing: OpenSSL [docs][1] say:

> EVP_CIPHER_get_iv_length() and EVP_CIPHER_CTX_get_iv_length() return the IV length, zero if the cipher does not use an IV and a negative value on error.

That might simply be a carelessness in the docs ("legacy" behavior was that only `EVP_CIPHER_CTX_iv_length()` could fail), but I'm not sure if fetched ciphers can have `cipher->iv_len < 0`.

https://github.com/rust-openssl/rust-openssl/blob/12b8e79803b797e805adef375792afbdb7da5c13/openssl/src/symm.rs#L492-L498

Maybe we should return `None` for `len <= 0` and only then do `Some(len as usize)`

[1]: https://docs.openssl.org/master/man3/EVP_EncryptInit/#return-values

PR #2597, opened 2026-04-15
add ML-DSA and ML-KEM NIDs for OpenSSL 3.5+

reaperhulk:


reaperhulk:
This is actually unhelpful because with modern OpenSSL `EVP_PKEY_is_a` is effectively required.

PR #2599, opened 2026-04-16
Fix dangling stack pointer in custom extension add callback

alex:
The out pointer given to OpenSSL was captured before the buffer was moved into ex_data. For inline-storage types like [u8; N], as_ptr() points into the value itself, so the move invalidated the pointer. Capture it from the final heap location instead.

### DIFF
--- a/openssl/src/ssl/callbacks.rs
+++ b/openssl/src/ssl/callbacks.rs
@@ -515,9 +515,6 @@ where
         match (*callback)(ssl, ectx, cert) {
             Ok(None) => 0,
             Ok(Some(buf)) => {
-                *outlen = buf.as_ref().len();
-                *out = buf.as_ref().as_ptr();
-
                 let idx = Ssl::cached_ex_index::<CustomExtAddState<T>>();
                 let mut buf = Some(buf);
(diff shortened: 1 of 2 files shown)


openssl/src/ssl/callbacks.rs
```diff
@@ -530,6 +527,15 @@ where
                 if new {
                     ssl.set_ex_data(idx, CustomExtAddState(buf));
                 }
+
+                // We must captures the out pointer AFTER buf has been moved into
+                // ex_data -- otherwise moving a type like [u8; N] would
+                // invalidate the captured pointer.
```
botovq: Maybe we can shorten this a bit:
```suggestion
                // Capture the out pointer AFTER buf has been moved into ex_data.
                // The move invalidates any previous pointer into buf.
```
At least fix the "captures" typo, please.

botovq (approved):
This feels like very dubious and likely dangerous functionality to support...

alex:
Yeah, I'm not sure this is a _good_ feature to support. But this fix lets us kick the backwards compatibility questions down the road 🙃

PR #2601, opened 2026-04-16
Add CMS partial signing, certificate, and signer retrieval APIs

justsmth:
This adds several CMS functions that enable partial/incremental signing workflows and certificate introspection:

- `CMS_add1_signer` — add signers to a `CMS_PARTIAL` structure
- `CMS_final` — finalize a partially-built CMS structure
- `CMS_add1_cert` / `CMS_get1_certs` — add and retrieve certificates
- `CMS_get0_signers` — retrieve signer certificates after verification
- `CMS_CADES` flag (OpenSSL 3.0+)

### Call-outs
`CmsSignerInfoRef` is introduced as a opaque borrowed reference type. No methods are defined on it yet, but it's returned from `add1_signer` to support future work (e.g., setting signed attributes for CAdES).

`get0_signers` uses the same `mem::forget(cert.to_owned())` refcount-bump pattern as `Pkcs7Ref::signers`, since `CMS_get0_signers` returns a caller-owned stack containing borrowed `X509` pointers.

`finalize` accepts `dcont` as `Option<&[u8]>`, matching the `&[u8]` → `MemBioSlice` pattern used by `sign` and `verify` for their BIO parameters.

Co-authored with @burelc-amzn.

### DIFF
--- a/openssl-sys/src/cms.rs
+++ b/openssl-sys/src/cms.rs
@@ -22,3 +22,5 @@ pub const CMS_USE_KEYID: c_uint = 0x10000;
 pub const CMS_DEBUG_DECRYPT: c_uint = 0x20000;
 pub const CMS_KEY_PARAM: c_uint = 0x40000;
 pub const CMS_ASCIICRLF: c_uint = 0x80000;
+#[cfg(ossl300)]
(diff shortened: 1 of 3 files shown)


justsmth:
The CI failures seem unrelated to this change.

PR #2602, opened 2026-04-17
bump for release

reaperhulk:


botovq:
So is the decision to keep the workaround for the EOF thing in the test for now and see if things break?

botovq:
Also, can this release perhaps wait half a day or so? I'm just about to sign LibreSSL 4.3.0 tarballs (dev-release but stable ABI/API) it would be really nice if the allowlisting of (4,3,_) could ride this.

reaperhulk:
Seems reasonable to wait 😄

And yeah I think we see if EOF is an issue in practice and if so we hoist it up into default flags.

openssl/CHANGELOG.md
```diff
@@ -2,6 +2,21 @@
...
+## [v0.10.78] - 2026-04-17
+
+### Added
+
+* Added support for OpenSSL 4.x.
+
+### Fixed
+
+* Fixed dangling stack pointer in custom extension add callback.
+* Fixed panic for overlong OIDs by clamping to maximum buffer size.
```
alex, line 14: nit: fixed panic _for what_

openssl/CHANGELOG.md
```diff
@@ -2,6 +2,21 @@
...
+* `Crypter::new` now panics when an IV is required but not provided, matching documented behavior.
+* Fixed zero-length `SSL_read_ex`/`SSL_write_ex` calling into OpenSSL unnecessarily.
+* Fixed Suite B flag assignments in `verify.rs`.
+* Use `cvt_p` for `OPENSSL_malloc` error handling.
```
alex, line 18: use it where?

alex:
going to take this one over

PR #2606, opened 2026-04-19
Check derive output buffer length on OpenSSL 1.1.x

alex:


botovq (approved):

botovq (commented):
Hmm. I'm not sure I buy the fix after the hkdf failure.

I think this workaround is ok for HKDF, as the comment explains.

However, I'm not seeing where the input key length is being honored for X25519/X448.

EVP_PKEY_derive() is pretty simple. The `M_check_autoargs` macro doesn't do anything since the pmeth don't have `EVP_PKEY_FLAG_AUTOARGLEN` (which is the source of the problem).

https://github.com/openssl/openssl/blob/b372b1f76450acdfed1e2301a39810146e28b02c/crypto/evp/pmeth_fn.c#L284-L296

The derive handler for `X25519()` is `pkey_ecx_derive25519()`:

https://github.com/openssl/openssl/blob/b372b1f76450acdfed1e2301a39810146e28b02c/crypto/ec/ecx_meth.c#L657-L695

If a `key` is given it writes to it and then sets `keylen` to the length it has assumed available.

alex:
This code implicitly assumes that supports NULL for outbuf and "only writes up to *buflen bytes" are mutually exclusive. Is there any case this isn't true?

(This whole thing is insane and I hate it)

botovq:
Ah. Got it. Yes, this should work: `X25519` correctly probes, so we can compare the result, so it is fine that it doesn't check itself. `HKDF` doesn't correctly probe but it does honor the key size.

botovq (approved):

botovq:
And yes, this is completely nuts.

botovq:
Did some more reading. There's one disturbing comment in OpenSSL 1.1.1's ec_pmeth.c's `pkey_ec_derive()`. It indicates that it deliberately allows truncation (`*outlen` is clearly intended to be `*keylen`):

```C
    /*
     * NB: unlike PKCS#3 DH, if *outlen is less than maximum size this is not
     * an error, the result is truncated.
     */

    outlen = *keylen;
```
https://github.com/openssl/openssl/blob/b372b1f76450acdfed1e2301a39810146e28b02c/crypto/ec/ec_pmeth.c#L174-L179

This is reachable for a default `EVP_PKEY_EC` unless you explicitly set the KDF type, e.g., via `EVP_PKEY_CTX_set_ecdh_kdf_type()` (which rust-openssl doesn't do itself).

So unless I'm confusing myself again, I think this fix will break this behavior. I don't know of examples relying on this but I would assume they exist.

alex:
... so this is a security issue on DH, but intended behavior on EC keys? I hate this sooooo much.

I'm no longer certain what the right fix is.

botovq:
A gross hack would be to skip the probing if `EVP_PKEY_base_id(...) == EVP_PKEY_EC`. We don't have bindings for `EVP_PKEY_base_id()`, but this shouldn't be hard.

alex:
Alternative idea: instead of erroring in this case, do a temporary alloc and then copy from that vec into the outbuf. Consistent behavior in all cases, fairly gross.

botovq:
Yes. Your suggestion seems better.

alex:
If I'm being honest with credit, it was Claude's idea.

alex:
@botovq hopefully this looks reasonable. (I tested locally on OpenSSL 1.1.1, OpenSSL 3.x, and LibreSSL...)

openssl/src/derive.rs
```diff
@@ -195,6 +215,26 @@ mod test {
...
+        // buffers before any write happens, so the call errors out.
+        let pkey = PKey::generate_x25519().unwrap();
+        let pkey2 = PKey::generate_x25519().unwrap();
+        let mut deriver = Deriver::new(&pkey).unwrap();
+        deriver.set_peer(&pkey2).unwrap();
+        let mut buf = [0u8; 4];
+        let result = deriver.derive(&mut buf);
+        #[cfg(any(all(ossl110, not(ossl300)), libressl))]
+        assert_eq!(result.unwrap(), 4);
+        #[cfg(all(ossl300, not(libressl)))]
```
botovq, line 234: Let's not kick off another lengthy round of CI runs, but I think this should be simplified to 
```suggestion
        #[cfg(ossl300)]
```
after the release.

botovq (approved):

PR #2614, opened 2026-04-21
Try to fix OpenSSL 1.1.0l download

botovq:
From [recent CI failures][1] it appears that the "/old/1.1.0" dl-path workaround is no longer working. Alternatively, we could switch to using the github release with the variant in capitalization which this currently redirects to.

[1]: https://github.com/rust-openssl/rust-openssl/actions/runs/24720642198/job/72308300577#step:10:121

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -200,7 +200,6 @@ jobs:
             - name: openssl
               version: 1.1.0l
               old: true
-              dl-path: /old/1.1.0
           include:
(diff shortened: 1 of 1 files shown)


botovq:
So this works. The alternative I mentioned would be to keep using the dl-path and point it at https://github.com/openssl/openssl/releases/download/OpenSSL_1_1_0l/openssl-1.1.0l.tar.gz.

alex (approved):

PR #2615, opened 2026-04-22
Require &mut BigNumContextRef for EcPointRef mul/invert

alex:
EC_POINT_mul and EC_POINT_invert mutate the BN_CTX's internal stack of temporaries via BN_CTX_start/BN_CTX_get/BN_CTX_end, so accepting a shared &BigNumContextRef was unsound: BigNumContext is Send+Sync, which let safe code race two threads on the same context.

Deprecate EcPointRef::mul, mul_generator, and invert, and add mul2, mul_generator2, and invert2 taking &mut BigNumContextRef.

### DIFF
--- a/openssl/src/ec.rs
+++ b/openssl/src/ec.rs
@@ -398,13 +398,21 @@ impl EcPointRef {
     }
 
     /// Computes `q * m`, storing the result in `self`.
+    ///
+    /// This method is deprecated because it takes a shared reference to the
+    /// `BigNumContext`, which is unsound: `EC_POINT_mul` mutates the context's
+    /// internal stack of temporaries, so sharing it across threads can race.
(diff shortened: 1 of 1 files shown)


botovq (approved):
In the C API `BN_CTX` is actually intended to be optional for elliptic curve operations. The `EC_POINT_mul() implementation [does this][1]. I can't find where that's documented in OpenSSL - it probably isn't.

I doubt it is worth the effort to audit all the libs to see if `Option<&mut BigNumContextRef>` is safe to use and to change this in the public rust-openssl API. It is certain that some older library versions get this wrong (I know, I fixed some of that or at least reviewed some of the fixes.)

[1]: https://github.com/openssl/openssl/blob/9ac29bc8579dd9632bd50900675ff7900f5f7ef1/crypto/ec/ec_lib.c#L1182-L1183

PR #2619, opened 2026-04-26
Add PKeyRef::is_a and KeyType for name-based key identification

reaperhulk:
EVP_PKEY_id returns -1 for provider-supplied keys in OpenSSL 3.0+ (notably ML-DSA, ML-KEM, and any third-party provider key), so the existing id() method cannot identify them. EVP_PKEY_is_a does a name-based check that works for all keys.

KeyType is a newtype around &'static CStr with constants for common algorithms.

### DIFF
--- a/openssl/src/pkey.rs
+++ b/openssl/src/pkey.rs
@@ -132,33 +132,38 @@ impl Id {
 pub struct KeyType(&'static CStr);
 
 impl KeyType {
-    pub const RSA: KeyType = KeyType::from_bytes_with_nul(b"RSA\0");
-    pub const RSA_PSS: KeyType = KeyType::from_bytes_with_nul(b"RSA-PSS\0");
-    pub const DSA: KeyType = KeyType::from_bytes_with_nul(b"DSA\0");
-    pub const DH: KeyType = KeyType::from_bytes_with_nul(b"DH\0");
-    pub const EC: KeyType = KeyType::from_bytes_with_nul(b"EC\0");
-    pub const ED25519: KeyType = KeyType::from_bytes_with_nul(b"ED25519\0");
-    pub const ED448: KeyType = KeyType::from_bytes_with_nul(b"ED448\0");
-    pub const X25519: KeyType = KeyType::from_bytes_with_nul(b"X25519\0");
-    pub const X448: KeyType = KeyType::from_bytes_with_nul(b"X448\0");
-    pub const HMAC: KeyType = KeyType::from_bytes_with_nul(b"HMAC\0");
-    pub const CMAC: KeyType = KeyType::from_bytes_with_nul(b"CMAC\0");
-    pub const ML_DSA_44: KeyType = KeyType::from_bytes_with_nul(b"ML-DSA-44\0");
-    pub const ML_DSA_65: KeyType = KeyType::from_bytes_with_nul(b"ML-DSA-65\0");
-    pub const ML_DSA_87: KeyType = KeyType::from_bytes_with_nul(b"ML-DSA-87\0");
-    pub const ML_KEM_512: KeyType = KeyType::from_bytes_with_nul(b"ML-KEM-512\0");
-    pub const ML_KEM_768: KeyType = KeyType::from_bytes_with_nul(b"ML-KEM-768\0");
-    pub const ML_KEM_1024: KeyType = KeyType::from_bytes_with_nul(b"ML-KEM-1024\0");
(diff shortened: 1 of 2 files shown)


openssl/src/pkey.rs
```diff
@@ -118,6 +118,64 @@ impl Id {
...
+    pub const HMAC: KeyType = KeyType::from_bytes_with_nul(b"HMAC\0");
+    pub const CMAC: KeyType = KeyType::from_bytes_with_nul(b"CMAC\0");
+    pub const ML_DSA_44: KeyType = KeyType::from_bytes_with_nul(b"ML-DSA-44\0");
+    pub const ML_DSA_65: KeyType = KeyType::from_bytes_with_nul(b"ML-DSA-65\0");
+    pub const ML_DSA_87: KeyType = KeyType::from_bytes_with_nul(b"ML-DSA-87\0");
+    pub const ML_KEM_512: KeyType = KeyType::from_bytes_with_nul(b"ML-KEM-512\0");
+    pub const ML_KEM_768: KeyType = KeyType::from_bytes_with_nul(b"ML-KEM-768\0");
+    pub const ML_KEM_1024: KeyType = KeyType::from_bytes_with_nul(b"ML-KEM-1024\0");
+
+    const fn from_bytes_with_nul(bytes: &'static [u8]) -> KeyType {
```
alex: this has to be unsafe, otherwise a safe caller can violate the invariant

openssl/src/pkey.rs
```diff
@@ -196,11 +254,27 @@ impl<T> PKeyRef<T> {
...
 
+    /// Returns `true` if this key is an instance of the algorithm named
+    /// by `key_type`.
+    ///
+    /// This is the OpenSSL 3.0+ name-based equivalent of [`PKeyRef::id`].
+    /// It is the only reliable way to identify provider-supplied keys,
+    /// which return `-1` from `EVP_PKEY_id`.
+    #[corresponds(EVP_PKEY_is_a)]
+    #[cfg(ossl300)]
+    pub fn is_a(&self, key_type: KeyType) -> bool {
```
alex, line 267: this API is so stupid, it makes me angry knowing it exists

openssl/src/pkey.rs
```diff
@@ -118,6 +118,69 @@ impl Id {
...
+
+    /// Constructs a `KeyType` from a NUL-terminated static C string.
+    ///
+    /// This is the escape hatch for naming algorithms that do not have a
+    /// pre-defined constant on `KeyType` — for example, algorithms exposed
+    /// by a third-party OpenSSL provider.
+    pub const fn from_static(name: &'static CStr) -> KeyType {
+        KeyType(name)
+    }
+
```
alex: is there actually a current use case for a public API? if not, let's not add it until there is

openssl/src/pkey.rs
```diff
@@ -118,6 +118,69 @@ impl Id {
...
+    pub const ML_KEM_768: KeyType = unsafe { KeyType::from_bytes_with_nul(b"ML-KEM-768\0") };
+    pub const ML_KEM_1024: KeyType = unsafe { KeyType::from_bytes_with_nul(b"ML-KEM-1024\0") };
+
+    /// # Safety
+    ///
+    /// `bytes` must be NUL-terminated and contain no interior NULs.
+    ///
+    /// Once MSRV is >= 1.72 this whole helper can become a safe `const fn`
+    /// using `CStr::from_bytes_with_nul(bytes).unwrap()`; once MSRV is >= 1.77
+    /// the helper goes away entirely in favor of `c"..."` literals.
```
alex: nit: this properly shouldn't be a `///` comment -- only the safety bit should be. this shoudl be a `//` comment. the difference is if the audience is the caller or the reader.

openssl/src/pkey.rs
```diff
@@ -118,6 +118,69 @@ impl Id {
...
+    ///
+    /// This is the escape hatch for naming algorithms that do not have a
+    /// pre-defined constant on `KeyType` — for example, algorithms exposed
+    /// by a third-party OpenSSL provider.
+    pub const fn from_static(name: &'static CStr) -> KeyType {
+        KeyType(name)
+    }
+
+    /// Returns the algorithm name as a C string.
+    pub fn as_cstr(&self) -> &'static CStr {
```
alex: hell, does this need to be a public API?

alex (approved):
what a dumb api

PR #2620, opened 2026-04-27
Add PKey::{public,private}_key_from_raw_bytes_ex

reaperhulk:
Wraps EVP_PKEY_new_raw_{public,private}_key_ex. Identifies the algorithm by KeyType and accepts an optional library context and property query string. Required for provider-supplied algorithms that have no associated Id, such as ML-DSA.

### DIFF
--- a/openssl/src/pkey.rs
+++ b/openssl/src/pkey.rs
@@ -859,16 +859,15 @@ impl PKey<Private> {
     pub fn private_key_from_raw_bytes_ex(
         ctx: Option<&LibCtxRef>,
         key_type: KeyType,
-        properties: Option<&str>,
+        properties: Option<&CStr>,
         bytes: &[u8],
     ) -> Result<PKey<Private>, ErrorStack> {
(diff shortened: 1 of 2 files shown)


openssl/src/pkey.rs
```diff
@@ -845,6 +847,34 @@ impl PKey<Private> {
...
+    /// property query string.
+    ///
+    /// This is required for algorithms that are only available via OpenSSL
+    /// 3.0+ providers and have no associated `Id`, such as ML-DSA.
+    #[corresponds(EVP_PKEY_new_raw_private_key_ex)]
+    #[cfg(ossl300)]
+    pub fn private_key_from_raw_bytes_ex(
+        ctx: Option<&LibCtxRef>,
+        key_type: KeyType,
+        properties: Option<&str>,
```
alex: is there a reason not to do `Option<&CStr>`? Avoids an allocation and also a panic condition.

alex (approved):

PR #2621, opened 2026-04-27
Add PKey::private_key_from_seed for ML-DSA/ML-KEM key import

reaperhulk:
Wraps EVP_PKEY_CTX_new_from_name + EVP_PKEY_fromdata with the "seed" OSSL_PARAM, the documented path for ML-DSA and ML-KEM in OpenSSL 3.5+ and uses OsslParamBuilder internally to avoid exposing OSSL_PARAM construction publicly.

This function can be refactored to support boring and aws-lc as well in the future, but is currently openssl only.

### DIFF
--- a/openssl/src/pkey.rs
+++ b/openssl/src/pkey.rs
@@ -858,6 +858,48 @@ impl PKey<Private> {
             .map(|p| PKey::from_ptr(p))
         }
     }
+
+    /// Constructs a private keypair from an algorithm-defined `seed`.
+    ///
+    /// Only ML-DSA (32-byte seed) and ML-KEM (64-byte seed) are supported at
+    /// this time, and require OpenSSL 3.5 or newer at runtime. The seed
+    /// length is algorithm-defined and a wrong-sized seed is rejected by the
+    /// provider.
+    ///
+    /// Internally this calls `EVP_PKEY_CTX_new_from_name` followed by
+    /// `EVP_PKEY_fromdata` with a single `seed` octet-string `OSSL_PARAM` and
+    /// `EVP_PKEY_KEYPAIR` selection.
+    #[corresponds(EVP_PKEY_fromdata)]
+    #[cfg(ossl350)]
+    pub fn private_key_from_seed(
+        ctx: Option<&LibCtxRef>,
+        key_type: KeyType,
+        properties: Option<&CStr>,
+        seed: &[u8],
+    ) -> Result<PKey<Private>, ErrorStack> {
+        let mut builder = crate::ossl_param::OsslParamBuilder::new()?;
+        builder.add_octet_string(c"seed", seed)?;
+        let params = builder.to_param()?;
+        unsafe {
+            ffi::init();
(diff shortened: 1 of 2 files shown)


openssl/src/pkey.rs
```diff
@@ -874,6 +874,49 @@ impl PKey<Private> {
...
+    /// `EVP_PKEY_KEYPAIR` selection.
+    #[corresponds(EVP_PKEY_fromdata)]
+    #[cfg(ossl350)]
+    pub fn private_key_from_seed(
+        ctx: Option<&LibCtxRef>,
+        key_type: KeyType,
+        properties: Option<&CStr>,
+        seed: &[u8],
+    ) -> Result<PKey<Private>, ErrorStack> {
+        let seed_key = CStr::from_bytes_with_nul(b"seed\0").unwrap();
```
alex: shoudl we raise the msrv to get c""? it makes things way nicer

openssl/src/pkey.rs
```diff
@@ -1358,6 +1401,49 @@ mod tests {
         assert!(key.ec_key().is_ok());
     }
 
+    #[cfg(ossl350)]
+    #[test]
+    fn test_private_key_from_seed_mldsa() {
```
alex, line 1389: can we get a "vector" style test. something that loads a known seed and asserts the correct raw public key?

openssl/src/pkey.rs
```diff
@@ -858,6 +858,49 @@ impl PKey<Private> {
...
+    #[cfg(ossl350)]
+    pub fn private_key_from_seed(
+        ctx: Option<&LibCtxRef>,
+        key_type: KeyType,
+        properties: Option<&CStr>,
+        seed: &[u8],
+    ) -> Result<PKey<Private>, ErrorStack> {
+        let seed_key = c"seed";
+        let mut builder = crate::ossl_param::OsslParamBuilder::new()?;
+        builder.add_octet_string(seed_key, seed)?;
```
alex: nit: inline c"seed" here

openssl/src/pkey.rs
```diff
@@ -1342,6 +1385,75 @@ mod tests {
...
+        assert_eq!(
+            hex::encode(&pk[..32]),
+            "e50d03fff3b3a70961abbb92a390008dec1283f603f50cdbaaa3d00bd659bc76",
+        );
+        // SHA-256 of the full pk, computed once from the KAT vector.
+        let pk_sha = crate::sha::sha256(&pk);
+        assert_eq!(
+            hex::encode(pk_sha),
+            "ab0f27cfe4f84d2a9fa74d48467439df9e5dd8401f8715f82ed70e43a371400a",
+        );
```
alex: This is such a weird way to do the assert. Can we just assert on the full thing?

reaperhulk: it will be 3904 bytes of hex, which is why the hash was chosen (damn huge keys). We can do the hex if you want though!

alex: honestly seems much simpler than this

alex (approved):

PR #2626, opened 2026-04-28
Add PKeyRef::seed_into for ML-DSA/ML-KEM seed extraction

reaperhulk:
Wraps EVP_PKEY_get_params with the "seed" OSSL_PARAM. Returns the 32-byte ML-DSA seed or 64-byte ML-KEM seed; inverse of PKey::private_key_from_seed. Errors for non-ML-DSA/ML-KEM keys, or for keys imported from encoded forms that retain only the expanded private key (no seed).

### DIFF
--- a/openssl/src/pkey.rs
+++ b/openssl/src/pkey.rs
@@ -388,6 +388,42 @@ where
         }
     }
 
+    /// Writes the algorithm-defined seed for an ML-DSA or ML-KEM private key
+    /// into `buf`, returning the number of bytes written.
+    ///
+    /// `buf` must be at least 32 bytes long for ML-DSA and 64 bytes long for
+    /// ML-KEM. The inverse of [`PKey::private_key_from_seed`].
+    ///
+    /// Errors when called on a key whose algorithm has no `"seed"`
+    /// `OSSL_PARAM` (e.g. RSA, EC, Ed25519), when `buf` is too small for
+    /// the algorithm's seed, or when the underlying key was imported from
+    /// an encoded form that retains only the expanded private key with no
+    /// seed.
+    #[corresponds(EVP_PKEY_get_params)]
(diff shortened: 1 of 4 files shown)


openssl/src/pkey.rs
```diff
@@ -388,6 +388,46 @@ where
...
+    ///
+    /// 32 bytes for ML-DSA, 64 bytes for ML-KEM. The inverse of
+    /// [`PKey::private_key_from_seed`].
+    ///
+    /// Errors when called on a key whose algorithm is not ML-DSA or ML-KEM,
+    /// or when the underlying key has no `"seed"` `OSSL_PARAM` available
+    /// (most commonly because the key was imported from an encoded form
+    /// that carries only the expanded private key, with no seed retained).
+    #[cfg(ossl350)]
+    pub fn seed(&self) -> Result<Vec<u8>, ErrorStack> {
```
alex: is there really no way to get a borrowed slice from the key? alternatively should callers be able to pass a buffer that we write the seed into?

reaperhulk: The current APIs like `raw_private_key` return a `Vec<u8>`, but we could do `seed_into` here and take a caller supplied buffer.

openssl/src/pkey.rs
```diff
@@ -388,6 +388,51 @@ where
...
+        {
+            64
+        } else {
+            return Err(ErrorStack::get());
+        };
+
+        if buf.len() < seed_len {
+            return Err(ErrorStack::get());
+        }
+
```
alex: do we need to manually do lengths? can we not just pass the buffer to openssl and let it fill in/reject?

reaperhulk: It's not in my nature to trust OpenSSL but let's find out.

alex (approved):

PR #2627, opened 2026-04-28
TLS 1.3 Pre shared keys support

rcollinskets:
This patch adds some callbacks and functions to enable pre-shared keys to work with TLS 1.3.

The need for early key material means that openssl requires some new callbacks for certain session types, namely `SSL_set_psk_find_session_callback` for the server side and `SSL_CTX_use_psk_identity_hint` on the client side. No existing behaviour should have been modified, this is simply new methods.

These features are important for emerging non-public key encryption mechanisms.

There is a change I'm not sure about, [forgetting the session returned by the callback](https://github.com/rust-openssl/rust-openssl/commit/e5be188593d73718e854803425708746ba4655a1) seems fine but I would like to know from if this is the right approach.

This change adds mappings for:
- `SSL_CTX_set_psk_use_session_callback`
- `SSL_CTX_set_psk_find_session_callback`
- `SSL_CIPHER_find`
- `SSL_SESSION_set_protocol_version`
- `SSL_SESSION_get0_cipher`
- `SSL_SESSION_set_cipher`
- `SSL_SESSION_new`
- `SSL_SESSION_set1_master_key`


dasobral:
Hi @rcollinskets, I opened #2687 for the same seven symbols before I noticed this PR, sorry for the duplicate effort. I listed the differences I found in a comment there so the maintainers can decide between the two or combine them. Two things you may want to check on this branch regardless: OpenSSL copies the client identity only after the callback returns, so the `CString::into_raw` pointer is never freed, and since #2624 the PSK trampolines look the closure up through the original `SSL_CTX` in the `SSL` ex data rather than `ssl.ssl_context()`, to survive an SNI context swap.

rcollinskets:
Thanks @dasobral I'll look into that.

PR #2629, opened 2026-05-02
Add PkeyCtxRef::set_context_string for ML-DSA

reaperhulk:
Sets the "context-string" OSSL_PARAM via EVP_PKEY_CTX_set_params, which binds a context string to an ML-DSA signing or verification operation per FIPS 204 §5. Gated on OpenSSL 3.5.0+.


openssl/src/pkey_ctx.rs
```diff
@@ -1368,4 +1384,23 @@ mxJ7imIrEg9nIQ==
         assert_eq!(output, expected_output);
         assert!(ErrorStack::get().errors().is_empty());
     }
+
+    #[test]
+    #[cfg(ossl350)]
+    fn set_context_string_mldsa() {
+        use crate::md_ctx::MdCtx;
+        use crate::pkey::KeyType;
```
alex: imports at top of module

openssl/src/pkey_ctx.rs
```diff
@@ -1368,4 +1386,20 @@ mxJ7imIrEg9nIQ==
         assert_eq!(output, expected_output);
         assert!(ErrorStack::get().errors().is_empty());
     }
+
+    #[test]
+    #[cfg(ossl350)]
+    fn set_context_string_mldsa() {
```
alex, line 1394: can we get a precise vector test?

alex (approved):

PR #2631, opened 2026-05-03
Reject non-UTF-8 OCSP responder URLs in X509Ref::ocsp_responders

alex:
X509_get1_ocsp returns the raw bytes of the IA5String accessLocation from the AIA extension, and OpenSSL's DER parser does not enforce that IA5String content is ASCII. OpensslStringRef's Deref<Target=str> wraps those bytes with str::from_utf8_unchecked, so a malicious cert could cause safe code to construct a &str containing invalid UTF-8 — language-level UB.

Validate each entry against str::from_utf8 (via the raw C pointer, not the unsound Deref) and return an ErrorStack if any URL is non-UTF-8.


botovq (approved):

ia0:
> construct a &str containing invalid UTF-8 — language-level UB

A `&str` that is not UTF-8 is not language-level UB, it is library-level UB. Eventually this may or may not lead to language-level UB (which in turn may or may not lead to a security issue). But on its own, it is not clear how bad this vulnerability actually is. I hope the severity of "high" in https://github.com/rust-openssl/rust-openssl/security/advisories/GHSA-xp3w-r5p5-63rr was not based on this confusion.

alex:
Sure, you're right that the UB only occurs if you ever _use_ the `str` -- any method is allowed to assume the utf-8 invariant. I think this is a distinction without a difference in practice.

ia0:
> UB only occurs if you ever _use_ the `str`

That's still wrong. There's no language UB when you use the `str`.

From the language perspective, there's no difference between `[u8]` and `str`. The difference is only at the library level.

Some libraries may rely on `str` being UTF-8 to do operations that would have language UB if that `str` would not be `UTF-8`. For example they may assume that the first byte is not 0xff and unsafely index into an array of 255 items only. The UB is only when (and due to) indexing, not when the `str` was created or provided.

alex:
The str docs state that any function may assume the str is well formed
utf8, which necessarily includes all methods on the str.

All that is necessary for evil to succeed is for good people to do nothing.

PR #2633, opened 2026-05-10
Prefer Homebrew openssl@4 and stop looking for openssl@1.1

alex:
Updates the macOS Homebrew auto-detection to try `openssl@4` first, falling back to `openssl@3` / `openssl@3.0`. The pre-3.0 `openssl@1.1` formula is no longer probed.


alex:
`openssl@1.1` no longer exists in brew

botovq (approved):

PR #2643, opened 2026-05-27
Add CMAC module for BoringSSL and AWS-LC backends

gshaurya18:
## Summary

Add a standalone cmac module that wraps the CMAC_CTX API directly, providing CMAC support for BoringSSL and AWS-LC backends which do not support CMAC through the EVP_PKEY interface.

## Motivation

The existing `PKey::cmac() function` is gated with `#[cfg(all(not(any(boringssl, awslc)), ossl110))]` because AWS-LC and BoringSSL don't support CMAC key generation through `EVP_PKEY*` API https://github.com/rust-openssl/rust-openssl/blob/master/openssl/src/pkey.rs#L608. 
This leaves users of these backends without CMAC support. The low-level `CMAC_CTX` API is available on AWS-LC and BoringSSL backends, so this module wraps it directly.

## API

- `CmacCtx::new(cipher, key)` — initialize with cipher and key
- `CmacCtxRef::update(data)` — feed data incrementally
- `CmacCtxRef::len()` — query expected output length
- `CmacCtxRef::finalize(buf)` — write MAC into caller-provided buffer
- `CmacCtxRef::finalize_to_vec()` — return MAC as `Vec<u8>`
- `impl Write for CmacCtxRef` — streaming via `std::io::Write`

Gated with `#[cfg(any(boringssl, awslc))]`.

## Testing

Tested on both AWS-LC 1.73.0 and BoringSSL:
- RFC 4493 test vectors (AES-128)
- NIST SP 800-38B test vectors (AES-192, AES-256)
- Non-block-aligned data (40 bytes)
- Invalid key lengths (negative test)
- Invalid cipher modes: CTR, GCM, OFB (negative test)
- len() and finalize() into caller buffer
- Write trait

### DIFF
--- a/openssl/src/cmac.rs
+++ b/openssl/src/cmac.rs
@@ -1,0 +1,250 @@
+//! CMAC (Cipher-based Message Authentication Code).
+//!
+//! CMAC is a MAC algorithm based on AES-CBC, defined in RFC 4493.
+//! https://tools.ietf.org/html/rfc4493#section-2.3.
+//!
+//! # Examples
+//!
+//! ```
+//! use openssl::cmac::CmacCtx;
+//! use openssl::symm::Cipher;
+//!
+//! let key = hex::decode("2b7e151628aed2a6abf7158809cf4f3c").unwrap();
+//! let mut ctx = CmacCtx::new(&Cipher::aes_128_cbc(), &key).unwrap();
+//! ctx.update(b"Hi There").unwrap();
+//! let mac = ctx.finalize_to_vec().unwrap();
+//! assert_eq!(mac.len(), 16);
+//! ```
+
+use std::io;
+use std::io::Write;
+use std::ptr;
+
+use crate::error::ErrorStack;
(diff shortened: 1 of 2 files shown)


openssl/src/cmac.rs
```diff
@@ -0,0 +1,250 @@
...
+//! CMAC is a MAC algorithm based on AES-CBC, defined in RFC 4493.
+//! https://tools.ietf.org/html/rfc4493#section-2.3.
+//!
+//! # Examples
+//!
+//! ```
+//! use openssl::cmac::CmacCtx;
+//! use openssl::symm::Cipher;
+//!
+//! let key = hex::decode("2b7e151628aed2a6abf7158809cf4f3c").unwrap();
```
kyawp5440-code, line 12: တော်​တေ်မိုက်တယွ

kyawp5440-code (changes requested):

kyawp5440-code (approved):

PR #2665, opened 2026-07-26
Remove unnecessary casts

FranciscoTGouveia:
I believe these casts can be safely removed, as the source type is already the target type.

For reference, these changes were found by an optimizer developed as a research project.

CC @nunoplopes


alex:
Assuming CI passes, seems fine. Needs a `cargo fmt` though.

alex (approved):

PR #2673, opened 2026-08-26
Fix null pointer write when feeding AAD to a non-AEAD cipher

### LINKED ISSUE #2670: Safe `aad_update` / `encrypt_aead` / `decrypt_aead` make OpenSSL write through a NULL pointer for non-AEAD ciphers
> [!NOTE]
> This finding was identified by [Crustify](https://github.com/crustify-rs/crustify-audit), an experimental agentic UB/soundness auditor developed at UC Berkeley and running on Claude Opus 5, then reviewed and independently reproduced by hand.

## The Issue

`CipherCtxRef::cipher_update(input, None)` passes `ptr::null_mut()` as `EVP_CipherUpdate`'s output pointer:

https://github.com/rust-openssl/rust-openssl/blob/10bbf71be84742fad3f9f782f282e48aa3ba389a/openssl/src/cipher_ctx.rs#L581-L596

For an **AEAD** cipher that is OpenSSL's documented ["this input is AAD"](https://docs.openssl.org/3.0/man3/EVP_EncryptInit/#aead-interface) convention. For any other cipher there is no such convention: `EVP_CipherUpdate` encrypts `input` into the pointer it was handed, i.e. it writes to address 0.

Nothing on the safe path checks that the cipher is an AEAD. The cipher is an ordinary runtime value, so the choice of cipher and the choice to supply AAD are made independently, by safe code, with no `unsafe` opt-in anywhere. Three public entry points reach it:

- https://github.com/rust-openssl/rust-openssl/blob/10bbf71be84742fad3f9f782f282e48aa3ba389a/openssl/src/symm.rs#L690-L693 — `Crypter::aad_update`
- https://github.com/rust-openssl/rust-openssl/blob/10bbf71be84742fad3f9f782f282e48aa3ba389a/openssl/src/symm.rs#L882 — `encrypt_aead`
- https://github.com/rust-openssl/rust-openssl/blob/10bbf71be84742fad3f9f782f282e48aa3ba389a/openssl/src/symm.rs#L913 — `decrypt_aead`

This is a soundness report, not a vulnerability report. I demonstrate a reachable NULL-pointer write; I have not shown, and do not claim, memory corruption, information disclosure, or exploitability.

## Reproduction

No `unsafe` anywhere in this file:

```rust
use openssl::symm::{Cipher, Crypter, Mode};

fn main() {
    let cipher = Cipher::aes_128_cbc(); // not an AEAD
    let key = [0u8; 16];
    let iv = [0u8; 16];

    let mut c = Crypter::new(cipher, Mode::Encrypt, &key, Some(&iv)).unwrap();
    println!("about to call aad_update on a non-AEAD cipher");
    c.aad_update(&[0u8; 64]).unwrap();
}
```

```text
$ cargo run
about to call aad_update on a non-AEAD cipher
Segmentation fault (core dumped)
```

Under AddressSanitizer, with the direction of the access made explicit:

```text
==1537==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000
==1537==The signal is caused by a WRITE memory access.
==1537==HINT: address points to the zero page.
    #0 (libcrypto.so.3+0x244cee)
    ...
    #4 aad_null_write::main
```

It crashes with and without a sanitizer, in debug and release.

## Environment

`openssl` 0.10.81 at `10bbf71be84742fad3f9f782f282e48aa3ba389a`, linked against system OpenSSL 3.0.20 (Debian 12), `x86_64-unknown-linux-gnu`, rustc 1.98.0 stable and nightly 1.100 for the sanitizer build.

Full reproduction and sanitizer logs available on request.

## Proposed fix

The cheapest correct option is to reject the combination rather than pass NULL: have `cipher_update(input, None)` return an error when the context's cipher is not an AEAD, which `EVP_CIPHER_CTX_get0_cipher` plus a flag check can determine at the call site. `Crypter::aad_update` and the two `*_aead` helpers then inherit the check for free.

If you would rather not add a runtime check on that path, the alternative is to document `cipher_update(.., None)` as requiring an AEAD context and make it `unsafe` — but that changes a public signature, so I would rather you choose.

Would you welcome a PR for either?

ChrisJr404:
Fixes #2670.

`cipher_update` with no output buffer tells OpenSSL to treat the input as AAD, which is only a valid convention for AEAD ciphers. With any other cipher `EVP_CipherUpdate` encrypts the input into the output pointer it was given, which is null here, so it writes to address 0. That is reachable from safe code via `aad_update`, `encrypt_aead`, and `decrypt_aead`, e.g. calling `aad_update` on an AES-128-CBC context segfaults.

This adds an `is_aead` check (mirroring the existing `is_wrap_mode` helper) and returns an error on the no-output path when the cipher is not an AEAD, so the AAD-only call can no longer reach the null write. The AEAD path is unchanged. On boringssl/awslc `is_aead` returns true to preserve existing behavior since `EVP_CIPHER_flags` isn't a reliable AEAD signal there.

You mentioned an AEAD check was probably best, so I went with that. Added tests for both the rejected non-AEAD case and the still-working AEAD case.


openssl/src/cipher_ctx.rs
```diff
@@ -570,6 +588,8 @@ impl CipherCtxRef {
     /// Writes data into the context.
     ///
     /// Providing no output buffer will cause the input to be considered additional authenticated data (AAD).
+    /// That is only meaningful for AEAD ciphers, so passing no output buffer with a non-AEAD cipher returns
+    /// an error rather than handing OpenSSL a null output pointer to encrypt into.
```
alex:
```suggestion
    /// an error.
```

PR #2677, opened 2026-08-28
x509: support the Reason extension in X509RevokedBuilder

DavidFontaineOcd:
Adds Asn1Enumerated::from_bn (via a new BN_to_ASN1_ENUMERATED
binding, mirroring BN_to_ASN1_INTEGER) since there was previously no way
to construct an Asn1Enumerated.

Add X509RevokedBuilder::append_extension/append_extension2 (mirroring
X509CrlBuilder), and a ReasonCode extension builder producing the CRL
entry reasonCode extension, following the same pattern as CrlNumber.

ReasonCode changes from an uninhabited ExtensionType marker into a
concrete builder wrapping Asn1Enumerated, matching how CrlNumber wraps
Asn1Integer.


openssl/src/x509/mod.rs
```diff
@@ -1652,6 +1653,26 @@ impl X509RevokedBuilder {
...
+    /// Add an X509 extension value to the `X509Revoked`.
+    ///
+    /// This works just as `append_extension` except it takes ownership of the `X509Extension`.
+    pub fn append_extension(&mut self, extension: X509Extension) -> Result<(), ErrorStack> {
+        self.append_extension2(&extension)
+    }
+
+    /// Add an X509 extension value to the `X509Revoked`.
+    #[corresponds(X509_REVOKED_add_ext)]
+    pub fn append_extension2(&mut self, extension: &X509ExtensionRef) -> Result<(), ErrorStack> {
```
alex, line 1665: is there a reason to have both of these?

DavidFontaineOcd: It was meant to be as close as possible to the `X509Builder` API, which exposes both of them.

PR #2687, opened 2026-09-17
Add TLS 1.3 external PSK session callbacks and SSL_SESSION setters

### LINKED ISSUE #1351: Support TLSv1.3 PSK functions
OpenSSL 1.1.1 introduces two new functions for TLS-PSK on TLSv1.3:
* https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_psk_use_session_callback.html
* https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_psk_find_session_callback.html

More information: https://wiki.openssl.org/index.php/TLS1.3#PSKs

I can't find these in the `openssl` crate yet.
The old-style TLSv1.2 callbacks seem to work but they default to SHA-256 message digest.

dasobral:
Closes #1351.

## Problem

`set_psk_client_callback` and `set_psk_server_callback` bind the TLS 1.2 PSK API. Under TLS 1.3 OpenSSL runs them with the PSK hash fixed to SHA-256: it builds the implied session with cipher `0x1301` (`ssl/statem/extensions_clnt.c`, `tls_construct_ctos_early_data`). A context restricted to `TLS_AES_256_GCM_SHA384` therefore cannot complete an external PSK handshake through the safe API. OpenSSL 1.1.1 added `SSL_CTX_set_psk_use_session_callback` and `SSL_CTX_set_psk_find_session_callback` for this, plus the `SSL_SESSION` setters needed to build the session they return. None of the seven symbols was bound.

Use case: a TLS 1.3 link between two embedded peers with an external PSK where the security profile requires AES-256-GCM.

## Added

`openssl-sys` (`handwritten/ssl.rs`):

| Symbol | Gate |
|---|---|
| `SSL_CTX_set_psk_use_session_callback` | `ossl111` |
| `SSL_CTX_set_psk_find_session_callback` | `ossl111` |
| `SSL_SESSION_set1_master_key` | `ossl111` |
| `SSL_SESSION_set_cipher` | `ossl111` |
| `SSL_SESSION_set_protocol_version` | `ossl111` |
| `SSL_SESSION_new` | none (present in 1.1.0 and LibreSSL) |
| `SSL_CIPHER_find` | none (present in 1.1.0 and LibreSSL) |

The two callbacks sit outside the `OPENSSL_NO_PSK` guard in `ssl.h`, so they are not gated on it. The three setters do not exist in 1.1.0, hence `ossl111` rather than `ossl110`.

`openssl`, all `#[cfg(ossl111)]`:

- `SslSession::new`, `SslSessionRef::set_master_key`, `set_cipher`, `set_protocol_version`
- `SslRef::cipher_by_id([u8; 2]) -> Option<&SslCipherRef>`
- `SslContextBuilder::set_psk_use_session_callback` and `set_psk_find_session_callback`, in the shape of the existing PSK wrappers. `Ok(None)` offers or finds no PSK, `Err` puts the error on the stack and aborts the handshake.

The trampolines look the closure up through the original `SSL_CTX` stashed in the `SSL` ex data (same reasoning as #2624: OpenSSL copies `psk_use_session_cb` and `psk_find_session_cb` into the `SSL` at `SSL_new` and `SSL_set_SSL_CTX` does not update them). The client identity is stored in the `SSL` ex data because OpenSSL copies it with `OPENSSL_memdup` only after the callback has returned.

## Tests

Four loopback tests in `ssl/test/mod.rs`, all `#[cfg(ossl111)]`:

- `psk_tls13_session_aes256`: both callbacks run, `TLS_AES_256_GCM_SHA384` negotiated, `session_reused()` true.
- `psk_tls13_legacy_callback_skips_psk_with_sha384_only`: the old callbacks under the same pins, OpenSSL drops the SHA-256 bound PSK and falls back to the certificate. Documents the motivation.
- `psk_tls13_find_session_none` and `psk_tls13_use_session_none`: the `Ok(None)` paths.

Verified against the `openssl` 3.0.13 command line tool in both directions with a 32 byte external PSK bound to `TLS_AES_256_GCM_SHA384`, no certificates on either side:

- Rust client (`set_psk_use_session_callback`) to `openssl s_server -nocert -psk_session`: `CIPHER is TLS_AES_256_GCM_SHA384`, `Reused session-id`, application data delivered.
- `openssl s_client -psk_session` to Rust server (`set_psk_find_session_callback`): `Reused, TLSv1.3, Cipher is TLS_AES_256_GCM_SHA384`, application data delivered.
- The session file for the tool was produced by `SslSession::new` plus the three setters and parses with `openssl sess_id`.
- Control: `openssl s_client -psk` (SHA-256 bound, the only thing the legacy callbacks can express) against the same SHA-384 only server fails with `tlsv1 alert decode error`.

Local matrix, all passing `cargo run -p systest` and `cargo test -p openssl`: OpenSSL 1.1.0l, 1.1.1w, 3.5.0, 4.0.0, LibreSSL 4.3.1, and Rust 1.80.0 with OpenSSL 3.5.0 (`cargo check -p openssl`). Not run locally: BoringSSL, AWS-LC, `no-deprecated`, i686, arm, macOS, Windows. The new items are gated off on the first two.

## Question

`SslSession::new` and `cipher_by_id` are gated `ossl111` although both C symbols exist in 1.1.0 and LibreSSL, because a session built through the safe API is only useful with the 1.1.1 setters. Happy to relax the gates if you prefer.


dasobral:
After opening this I found #2627 by @rcollinskets, which binds almost the same symbol set. Apologies for not linking it earlier. For the maintainers, the concrete differences as I read the two diffs, so you can pick either or ask for a merge:

- **Client identity lifetime.** OpenSSL copies the identity with `OPENSSL_memdup` only after the callback returns (`tls_construct_ctos_early_data`). #2627 hands OpenSSL a `CString::into_raw` pointer that is never freed and that panics on an interior NUL. This PR keeps the `Vec<u8>` in the `SSL` ex data, replaced in place on the next call and freed with the connection.
- **Closure lookup.** #2627 recovers the closure through `ssl.ssl_context()`. Since #2624 the PSK trampolines look it up through the original `SSL_CTX` stashed in the `SSL` ex data, because OpenSSL copies these callback pointers at `SSL_new` and `SSL_set_SSL_CTX` does not update them. This PR follows #2624.
- **Error handling.** `SslSession::new` returns `Result` here (`SSL_SESSION_new` can fail); `set_master_key` returns `Result<(), ErrorStack>` rather than a raw `i32`.
- **Gates.** `ossl111` only. The two callbacks sit outside the `OPENSSL_NO_PSK` guard in `ssl.h`, and `SSL_CIPHER_find` and `SSL_SESSION_new` exist in 1.1.0 and LibreSSL so they are ungated in the sys crate. Prototypes are matched to the header (`SSL_CIPHER_find` takes `SSL *`, not `const SSL *`).
- **Changelog entries** for both crates are included.
- **Callback shape.** The client closure returns `Option<(Vec<u8>, SslSession)>`; #2627 passes `&mut Option<SslSession>` and returns the identity. Either works, this one mirrors `raw_client_psk` more closely.

Happy to converge on whichever shape you prefer, or to fold anything from #2627 that you would rather keep.

